PR #28, opened 2013-07-31
Add support for fetching CA from PKCS11 token.

tc-anssi:
Dear Kerberos Team,

When using PKCS#11 token, the current PKINIT code is not able to fetch a CA from the token. 
The following patch adds a PKCS11 target in the in the pkinit_anchors configuration parameter. 
It is possible to use cacert_label/cacert_id to refine the search. 
When cacert_label is not specified, the code will search fetch all certificates and look for a self-signed one with CA=true and reconstruct the certification path.

I'll be waiting for your feedback.

Thomas Calderon.


tc-anssi:
Hi,

I finally had the chance to work on tidying up the code a little bit.

PR #132, opened 2014-05-28
KRB-49 Password history in LDAP KDB plugin

tkuthan:
This is yet another older Solaris fix, we would like to contribute upstream. It enables using password history in policies with LDAP KDB plugin.

The fix is equivalent to patch submitted in Ticket #5889, modulo a modification for new krb5_update_tl_kadm_data() API and explicit parenthesis for mask operations.

### DIFF
--- a/src/lib/kadm5/admin.h
+++ b/src/lib/kadm5/admin.h
@@ -110,7 +110,7 @@ typedef long            kadm5_ret_t;
 #define KADM5_RANDKEY_USED      0x100000
 #endif
 #define KADM5_LOAD              0x200000
-#define KADM5_NOKEY             0x400000
+#define KADM5_KEY_HIST          0x400000
 
 /* all but KEY_DATA, TL_DATA, LOAD */
 #define KADM5_PRINCIPAL_NORMAL_MASK 0x41ffff
--- a/src/lib/kadm5/srv/svr_principal.c
+++ b/src/lib/kadm5/srv/svr_principal.c
@@ -1543,6 +1543,9 @@ kadm5_chpass_principal_3(void *server_handle,
         KADM5_FAIL_AUTH_COUNT;
     /* | KADM5_CPW_FUNCTION */
 
+    if (hist_added == 1)
+        kdb->mask |= KADM5_KEY_HIST;
+
     ret = k5_kadm5_hook_chpass(handle->context, handle->hook_handles,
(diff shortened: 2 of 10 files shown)


fr3ddie:
Hello all,
I knew this patch was foreseen for krb v1.14 but, as far as I can see, it is still in the "need revision" stage on git.
When is this patch supposed to be merged into the main tree?

Moreover, is it sane to think to manually apply the patch on the 1.14 branch and recompile to achieve this feature?

Thank you very much

SaharahSarah:
OK, a bit confused on what the decisions were on the work that needs to be done. I see at least the one thing of sorting the keys when they are retrieved in the DIL, but not exactly sure on what the next work would be. tkuthan had mentioned these 2 things:

> 2) don't duplicate old keys in history (fix create_history_entry)
> 3) password history in LDAP plugin rewritten
> - rely on original kvno for aging passwords out of history

Is that still applicable? If so, can I get some clarity on these requirements?

greghudson:
Take a look at:

http://mailman.mit.edu/pipermail/krbdev/2014-July/012084.html
http://mailman.mit.edu/pipermail/krbdev/2014-August/012156.html

and see if that helps to clarify.

SaharahSarah:
OK I have:

* Keys are sorted when retrieved from an entry.
* Only the most recent keys are stored in a history entry.

Is there anything else that needs to be done?

greghudson:
Yes, the hard part remains to be done.  Because of a miscommunication, Tomas wrote his patch to the LDAP back end assuming that he would need to store multiple key versions in a single osa_princ_ent_rec, so he overloaded the kvno field of the KrbKeySet value to mean something different in password history entries.  With the change to create_history_entry(), this overloading can be removed and the code can be simplified.  Tomas's patch needs to be examined, carefully understood, and rewritten without this extra complexity.  See my comment on August 14, 2014 for more details.

tlyu:
Superseded by #399.

ulysse31:
Hi,

I've got an old MIT kerberos (kerberos 1.6), and seems that your patch was made for another version.
Which MIT kerberos version did you patched here ?
Thanks a lot.

greghudson:
All MIT krb5 work takes place on the master branch, which is more current than the most recent release.  (Bugfixes are then backported to recent release branches, currently 1.14 and 1.13.)  The 1.6 release is almost ten years old, and is not supported.  I would not expect LDAP password history support to be a very easy backport to 1.6, although it's possible in theory.

PR #214, opened 2014-10-05
KDC authdata refactoring

greghudson:
This patch series improves the state of kdc_authdata.c in preparation for CAMMAC and authentication indicator work.  There are no associated RT tickets, because there aren't any public behavior changes, but the KDC side of the not-yet-public authdata pluggable interface is replaced with a new-style interface.

I expect to squash the first three commits before pushing; this is reflected in the commit messages.


greghudson:
I noticed that handle_signedpath() and its helpers weren't using their krbtgt parameters, so I modified the second commit to remove those parameters.

PR #225, opened 2014-10-21
Add support for directories in profile paths

greghudson:
This branch currently contains an unaltered patch from Roland Mainz to add support for directories in profile paths.


greghudson:
Since we are coming up on the 1.14 branch date and have no clear design to solve the timestamp issue, I am going to merge this.  I will open a ticket about profile timestamps, include/includedir, and g_shared_trees.

PR #226, opened 2014-10-24
Add missing salt from enctype in t_kdb.py test

pkis:
kadmin requires enctypes in form "key:salt"
If it is not given in that form, it reports error and the test fails.

The issue might remain hidden because the test is executed only if ldap is installed. If missing, the test silently passes (only the verbose output shows more detail).


tlyu:
Thanks; pushed as 62a4509028ae7c7cf00af38cde879428f03fe5e6.

PR #231, opened 2014-10-31
Add ASN.1 encoder and decoder for CAMMAC

greghudson:
Add internal type declarations for krb5_verifier_mac and krb5_cammac.
Add ASN.1 encoder and decoder functions and an internal free function
for krb5_cammac.  Add ASN.1 tests for krb5_cammac as well as asn1c
test vectors for Verifier and AD-CAMMAC.


tlyu:
Looks OK to me.

PR #238, opened 2014-12-16
Fix bugs in previous cc_file.c changes

greghudson:
In fcc_destroy and krb5int_fcc_new_unique, call set_errmsg_filename
before deleting the cache handle, or else the reference to
data->filename is a use after free.

In set_errmsg_filename, do nothing if the code is 0, as we don't have
an error to annotate.


kaduk:
This looks okay.
It's not clear that having set_errmsg_filename() override 'ret' was the right thing to do in the first place, anyway.

PR #241, opened 2015-01-12
Only return a single key in etype-info and kdcpreauth plugins

npmccallum:
The KDC already knows what etypes the client supports/prefers
and what keys exist in the kdb. By making this choice in the KDC
we gain several benefits. First, less data is transmitted over
the wire. Second, kdcpreauth plugins can know which key/etype
will be used by the exchange. This means that preauth mechs
don't have to make a separate (possibly wrong) decision.

### DIFF
--- a/src/kdc/kdc_preauth.c
+++ b/src/kdc/kdc_preauth.c
@@ -463,36 +463,57 @@ max_time_skew(krb5_context context, krb5_kdcpreauth_rock rock)
 }
 
 static krb5_error_code
+client_key(krb5_context context, const krb5_kdc_req *request,
+           krb5_db_entry *client, krb5_key_data **entry_key,
+           krb5_keyblock *key)
+{
+    int i;
+
+    for (i = 0; i < request->nktypes; i++) {
+        krb5_key_data *kd = NULL;
+        krb5_keyblock kb;
+
+        if (!krb5_is_permitted_enctype(context, request->ktype[i]))
+            continue;
+
+        if (krb5_dbe_find_enctype(context, client, request->ktype[i],
+                                  -1, 0, &kd) != 0)
+            continue;
+
+        if (krb5_dbe_decrypt_key_data(context, NULL, kd, &kb, NULL) != 0)
+            continue;
+
+        if (entry_key != NULL)
+            *entry_key = kd;
+
+        if (key != NULL)
+            *key = kb;
+
+        return 0;
+    }
+
+    return ENOENT;
+}
+
+static krb5_error_code
 client_keys(krb5_context context, krb5_kdcpreauth_rock rock,
             krb5_keyblock **keys_out)
 {
-    krb5_kdc_req *request = rock->request;
-    krb5_db_entry *client = rock->client;
-    krb5_keyblock *keys, key;
-    krb5_key_data *entry_key;
-    int i, k;
+    krb5_error_code retval;
+    krb5_keyblock *keys;
 
-    keys = calloc(request->nktypes + 1, sizeof(krb5_keyblock));
+    keys = calloc(2, sizeof(krb5_keyblock));
     if (keys == NULL)
         return ENOMEM;
 
-    k = 0;
-    for (i = 0; i < request->nktypes; i++) {
-        entry_key = NULL;
-        if (krb5_dbe_find_enctype(context, client, request->ktype[i],
-                                  -1, 0, &entry_key) != 0)
-            continue;
-        if (krb5_dbe_decrypt_key_data(context, NULL, entry_key,
-                                      &key, NULL) != 0)
-            continue;
-        keys[k++] = key;
-    }
-    if (k == 0) {
+    retval = client_key(context, rock->request, rock->client, NULL, &keys[0]);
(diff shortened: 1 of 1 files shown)


npmccallum:
WARNING: I have not tested this. I just want to make sure this is sane.

greghudson:
I think we would want a number of automated tests for this.  I'm not immediately sure what the harness would look like.

Concerns:

* Optimistic encrypted challenge should work with any of an entry's keys (of highest kvno).  This change would break that.  Possibly encrypted challenge should not be using the client_keys() callback.

* We (probably I) need to think really hard about the DES special cases in the old code.  I originally thought it was covered by the similarity check in krb5_dbe_def_search_enctype(), but now that I look closer, it looks like that might not be enough.

* For DES keys, there's a nonstandard "afs3" salt type which the client may or may not implement.  (In particular Java doesn't, or at least didn't used to.)  We could try to somehow prefer non-afs3-salted DES keys over afs3-salted ones; I need to think about whether that breaks ancient AFS clients.

Quibbles (partial list):

* The permitted_enctype check in client_key shouldn't be needed; krb5_dbe_find_enctype() checks that.

* I'm not sure if there's enough commonality between what client_keys() wants and what etype_info_helper() wants to productively share code.  Right now client_key() decrypts and leaks the key data when called from etype_info_helper().

greghudson:
More on how preauth modules should and do interact with client keys:

* A module always has the option of calling client_entry() and looking into the key data via krb5_dbe_search_enctype() or similar.  Doing so couples the module implementation to the officially-unstable KDB API, but is likely to allow greater efficiency in key decryption operations.

* The existing logic for client_keys() dates back to 1.6 and may not have been carefully thought out.  At the time it was used only for an example module.

* One current use is by the SAM-2 module, which only uses the first result.  SAM-2 has more or less the same requirement as a KDC-speaks-first PAKE module; it wants to pick a single key which the client should have possession of based on its enctype list.  The existing client_keys() is wasteful for this purpose but should give the correct result.

* The other current use is in the encrypted challenge module.  Since encrypted challenge should allow any key (of permitted enctype), the existing client_keys() logic is not entirely correct--we don't really want to restrict the choices to the request's enctype list, and we don't want to look at only the first key of each request enctype.

My current design preference is:

* Change client_keys() to fetch and decrypt all of the entry's keys of permitted enctypes, ignoring the request.  This callback can be used for modules which don't want to use the KDB API but need to be able to verify preauthentication against any key (like encrypted timestamp/challenge).

* Add a new preferred_client_key() callback which fetches and decrypts the first entry key it can find going by the order of request enctypes.  This should also be the key we send etype-info2 for, although for etype-info2 we don't need to decrypt the key.

greghudson:
After taking a closer look, I believe that:

* To preserve existing behavior for DES enctypes, all we need to do is filter out des-cbc-md4 from the request enctype list when choosing the preferred key.  (Even that much is probably paranoia.)

* We shouldn't need to make any special allowances for the afs3 salt type.  Anything which breaks after this change is very likely already broken before.

greghudson:
https://github.com/greghudson/krb5/commits/single_etype (top two commits) has my candidate changes.  I still need to add automated tests, which will require a bit of work to create the necessary harnesses.

I elected not to change the behavior of client_keys() at this time, or to change SAM-2 or encrypted challenge not to use it; both of those seem like low-priority changes.

After looking over several incarnations of the single-DES enctype special cases, I decided that they were probably just awkward and buggy attempts to deal with the fact that any DES key data entry can be used with any DES enctype.  I handled that by remembering the request enctype used to match the key data entry, and putting that enctype in the etype-info2 entry or decrypted keyblock.  (client_keys() could use a similar workaround, but again, low priority.)  I couldn't find evidence of a deliberate intent to effectively filter out des-cbc-md4 from the request enctype list, so I didn't preserve that behavior.

greghudson:
While writing a test harness for KDC etype-info2 behavior, I noticed that we do the same key calculation in finish_process_as_req() to set state->client_keyblock and state->rock.client_key, the latter of which is already used in etype_info_as_rep_helper() for the etype-info2 information in the AS-REP.  If we just did that earlier in the AS req processing, we could avoid duplicate code, save a bit of work, and eliminate one of etype_info_as_rep_helper() and etype_info_helper().  So I will have a go at that.

I will probably not add automated tests for the new callback at this time, because we have a test preauth module in the auth-indicator pull request.  But I will make sure to test it by hand, and it will probably be dead simple (just pulling the previously computed client keyblock out of the rock).

greghudson:
Okay, I think the top four commits here are ready: https://github.com/greghudson/krb5/commits/single_etype2

Nathaniel, please review these and see if they look reasonable.

tlyu:
The commits on single_etype2 seem reasonable, though I haven't analyzed them extensively.

greghudson:
Pushed to master as 385cd2d07983a89892dad1606e1a41a78066c6ec 5cf4a7e220141f10f51995ceae9b9e74232a31b7 7b12eb4757f8dd05b79c9b49d4289f0caf1f6eec be20a5f5cee8d6c4072d1b81712520dbf9f6eefd

PR #242, opened 2015-01-14
Check for null *iter_p in profile_iterator()

greghudson:
In profile_iterator(), return PROF_MAGIC_ITERATOR if *iter_p is NULL,
instead of dereferencing a null pointer, as we did prior to 1.10.
Correct calling code will not trigger this case, but incorrect code
has been reported in the field.


kaduk:
Seems reasonable.

PR #243, opened 2015-01-20
Do not loop on principal unknown errors

greghudson:
Simo's commit, not reviewed or edited.

### DIFF
--- a/src/lib/krb5/krb/get_in_tkt.c
+++ b/src/lib/krb5/krb/get_in_tkt.c
@@ -1379,33 +1379,23 @@ note_req_timestamp(krb5_context context, krb5_init_creds_context ctx,
         AUTH_OFFSET : UNAUTH_OFFSET;
 }
 
-/* Determine whether the client realm in a KRB-ERROR is empty. */
-static krb5_boolean
-is_empty_crealm(krb5_error *err)
-{
-
-    return (err->client == NULL || err->client->realm.length == 0);
-}
-
 /*
- * Determine whether a KRB-ERROR is a referral to another realm.
+ * Determine whether err is a client referral to another realm, given the
+ * previously requested client principal name.
  *
- * RFC 6806 Section 7 requires that KDCs return the referral realm in
- * an error type WRONG_REALM, but Microsoft Windows Server 2003 (and
- * possibly others) return the realm in a PRINCIPAL_UNKNOWN message.
- * Detect this case by looking for a non-empty client.realm field in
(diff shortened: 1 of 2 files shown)


greghudson:
I pushed a proposed rework which simplifies the code, although it's not quite as minimal of a change.  Still needs an automated test and RT headers at a minimum.  Tom was concerned that we would be changing the behavior on WRONG_REALM with an empty crealm if retry is true, so we need to examine that path more closely.  I don't think there's any need to single out that particular error case and disable retries, but I'm also not certain what the intended purpose that code path is.

greghudson:
Okay, more information.  The origin of the "retry" intent can be found in 1.6, back when the AS client code was synchronous and we didn't support client referrals.  In this version of the loop, any error other than PREAUTH_REQUIRED, coupled with e_data, is intended to result in another attempt with krb5_do_preauth_tryagain() used to generate potentially better preauthentication information.  For instance, PKINIT can try again after decoding the KDC's list of trusted CAs.

Release 1.7 added the actual "retry" variable, set by the FAST logic based on the presence of a cookie for a FAST error, and the presence of e-data for a non-FAST error.  Release 1.7 also added client referral support.  In this version of the loop, a WRONG_REALM with no crealm field bypasses the "retry" logic--but that's really an unimportant detail.  A WRONG_REALM reply with no crealm and with e_data (or FAST error padata and a cookie) is not expected in practice, and there's nothing dangerous about passing such an error through the krb5_do_preauth_tryagain() step like we would have done in 1.6, and like we do for any random unhandled error code.

So, I'm comfortable with the behavior of my current proposed patch.  I will add a test case.

greghudson:
I narrowed the is_referral() contract so that it doesn't take the krb5_init_creds_context "god object" as input, added RT headers and a comment about the retry behavior change, and added a test case.  I don't like adding to t_general.py since it really shouldn't exist, but that's where we currently have kinit tests.  This should be ready for review by the rest of the team.

tlyu:
I agree with Greg's analysis of the retry case, though I haven't thoroughly analyzed what happens in the retry with ERR_WRONG_REALM case. It's probably not dangerous.

The patch seems OK to me, but I think we should avoid long multi-line boolean expressions. Could we separate the err->error checks into an if-statement that returns early from is_referral?

greghudson:
Okay, I changed is_referral() to break up the boolean expression.

tlyu:
Thanks; that looks OK to me.

PR #244, opened 2015-01-20
Export function gss_add_cred_with_password

frozencemetery:
This function is already present in gssapi_ext.h, but without exporting it, a link error will be produced every time it is used.

An example can be seen [here](https://gist.github.com/frozencemetery/7c8a016db7081f806baa).

This problem is present in master and at least as far back as 1.12 and likely earlier as well.


greghudson:
Pushed to master as 266cce14ee39f6d11b186ee988cffd0c2a119f3d

PR #245, opened 2015-01-25
Add support for KDC_ERR_MORE_PREAUTH_DATA_REQUIRED (RFC 6113)

npmccallum:
This entails some simple changes to the kdc and libkrb5.

On the KDC side, we define the appropriate error code. Additionally,
we allow kdcpreauth plugins to return this error code to clients.

On the libkrb5 side, we treat this error code in the same way as
KDC_ERR_PREAUTH_REQUIRED except that we additionally reset the counter
of tried preauth types, permitting preauth types to continue the converation.


kaduk:
Greg says that coverity will complain about a fallthrough in a switch statement without a comment prior to it (any comment at all, apparently).  But probably that logic could be reworked without using switch anyway.

The kdc_preauth.c chunk might want a /* rfc 6113 */ comment akin to the rfc 4556 comment right there in the context.

npmccallum:
I fixed this in the latest commit.

greghudson:
While I appreciate using a helper function to avoid adding complexity to the already-large init_creds_step_reply(), I'm not fond of having a side-effect in a predicate function.  Also, this is one of three cases where we try again (the other two being client referrals and the somewhat tortured retry-with-edata code path), so the helper is confusingly named.

Taking a step back, I don't think the preauth already-seen-types list was designed correctly back when it was introduced in 1.6.  It should really only apply when trying to continue after a failed preauth attempt, something we briefly tried doing in 1.7 but had to revert pending deeper changes to the preauth framework.  We already introduced a hack in f20a77e879d203cdcb1bdbf9dc8e604a5187c88f to make the list not apply to SAM-2.

So, while it's not as conservative, I think the right thing for this pull request is:

* Treat KDC_ERR_PREAUTH_REQUIRED and KDC_ERR_MORE_PREAUTH_DATA_REQUIRED identically.

* Call k5_reset_preauth_types_tried() for both types of KDC errors.  This will mostly defeat the purpose of the list, but the internal logic will remain for when we do continuation after preauth failures.

* Remove the SAM-2 hack since it won't be required any more.  That should be a separate commit.

Since I'm engaging in scope creep, I will put together candidate commits.

greghudson:
Please have a look at the two commits in https://github.com/greghudson/krb5/commits/more which implement what I discussed.

npmccallum:
Those both look good to me.

greghudson:
I amended the first commit to update a comment and mention MORE_PREAUTH_DATA_REQUIRED errors in the docs.

I also corrected a mistake in the commit message (the part I wrote): it turns out there is no error parameter to the clpreauth process method, and we don't even keep the error around.  There also isn't any way for a process method to distinguish between padata received with an error and padata received along with the issued ticket.  That's a pre-existing issue and not necessarily a serious one; an attacker can't accomplish anything by confusing a clpreauth module if it's properly written.

greghudson:
Pushed to master as 95c3cab051aa1b8b4f7eb309bf135e8f51665baa 922f7d1230fe647821d9767fafef3774c5cfd2fc

PR #247, opened 2015-01-27
Remove starttime hack in EncTicketPart decoder

greghudson:
The EncTicketPart decoder sets starttime to authtime if it wasn't
included in the ASN.1 value.  This is problematic for upcoming CAMMAC
work, as we will need to re-encode a received EncTicketPart to check
the KDC verifier.  Remove that behavior and just use opt_kerberos_time
for the starttime field.  Adjust krb5_decode_test.c to match the new
decoder behavior.

Similarly, remove the process_tgs_req() code which sets starttime in
the header ticket if it isn't set.  Add a comment explaining the
unrelated code adjacent to it.

check_tgs_times() used the ticket starttime without checking if it was
present.  Add a fallback to times->authtime, and narrow the function
contract to make the implementation more concise.

There is a similar hack in the EncKDCRepPart decoder; leave that alone
for now.

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -383,12 +383,7 @@ process_tgs_req(struct server_handle *handle, krb5_data *pkt,
     if (isflagset(server->attributes, KRB5_KDB_OK_AS_DELEGATE))
         setflag(enc_tkt_reply.flags, TKT_FLG_OK_AS_DELEGATE);
 
-    /*
-     * Fix header_ticket's starttime; if it's zero, fill in the
-     * authtime's value.
-     */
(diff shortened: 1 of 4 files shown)


kaduk:
Perhaps the comment "/* Depends only on request, ticket, and time. */" in the last chunk of tgs_policy.c should be updated to match the narrowed contract of check_tgs_times().
Otherwise, it looks okay.

tsitkov:
Consider changing copyrights to 2015 - at least for tgs_policy.c and asn1_k_encode.c

greghudson:
I updated the comment.  I don't think any of the changes are significant enough to be copyrightable, so I didn't update the copyright dates.

PR #248, opened 2015-02-01
kadmin script mode

greghudson:
Here is my implementation of RT #7991, which resulted from design ideas discussed in PR #121.

### DIFF
--- a/doc/admin/admin_commands/kadmin_local.rst
+++ b/doc/admin/admin_commands/kadmin_local.rst
@@ -16,6 +16,7 @@ SYNOPSIS
 [[**-c** *cache_name*]\|[**-k** [**-t** *keytab*]]\|\ **-n**]
 [**-w** *password*]
 [**-s** *admin_server*\ [:*port*]]
+[command args...]
 
 **kadmin.local**
 [**-r** *realm*]
@@ -25,6 +26,7 @@ SYNOPSIS
 [**-e** *enc*:*salt* ...]
 [**-m**]
 [**-x** *db_args*]
+[command args...]
 
 .. _kadmin_synopsis_end:
 
@@ -112,8 +114,7 @@ OPTIONS
     via the process list.
 
 **-q** *query*
-    Perform the specified query and then exit.  This can be useful for
-    writing scripts.
+    Perform the specified query and then exit.
 
 **-d** *dbname*
     Specifies the name of the KDC database.  This option does not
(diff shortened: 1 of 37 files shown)


kaduk:
> The exit status will be 1 if the query fails.

Do we want to lock ourselves into this specific value (as opposed to just "a nonzero value")?

greghudson:
I edited it to say "non-zero" instead of 1.

kaduk:
Comments on 613b5c496:

Do we need prototypes for info() and error() if the prototypes immediately precede the function definitions?

There is a typo "exit_statu" in the comment in the RHS's line 142.

The first commit makes set_com_err_hook() always called, not just for kadmin.local, but does not mention this in the commit message.  I guess this is just one of the many existing bugs in surrounding code that are fixed as part of this commit and not called out explicitly.

Some places were changed to call error() but then have an exit(1) right afterwards.  One could argue that those should become exit(exit_status), but I could see an argument for making that a separate change.

"-q is exclusive with command-line query" has potential for confusion, but I have no better suggestion.

The use of error() within cpw_usage(), kadmin_addprinc_usage(), kadmin_modprinc_usage(), and kadmin_addmodpol_usage() could in principle overwrite an error_status that was set prior to the call to *_usage().  There do not appear to be any such calls at present, so we should proceed to not care.

The patch seems to be inconsistent about whether error() is used in the !script_mode case (compare RHS lines 632 and 1255, for example).  I guess the intent is to always use error() since it is more concise?


I did not review the portions of kadmin.c not visible in the diff for any calls which should be converted to info() or error().

kaduk:
Comments on a8cb9fd648:

I only skimmed some parts of this, but I did run the tests on my FreeBSD machine.
I don't run the OTP tests there, though -- does your machine run them?

In general, it looks fine.

It also occurs to me that with (almost) all the tests converted to the command-line form of kadmin, we have minimal test coverage of the -q form of executing queries.  Perhaps we should explicitly add some.

There are lots of places where we go from checking for (failure) text in the output to just relying on the exit status, which leaves me with some (probably unjustified) worry that there are now new classes of potential bugs which are not covered by the test suite.

t_salt.py retains a -force flag to delprinc that could be removed.  (Possibly others?  I did not do an exhaustive check, since that's easier to do in a working copy than by viewing a diff.)

greghudson:
The prototypes for info() and error() are, to the best my knowledge, the only way to apply gcc printf format attributes to them.

I will fix the "exit_statu" typo.

I think it was deliberate that the extended com_err hook was previously installed for kadmin.local only, because at the time extended errors were generated chiefly by the LDAP KDB module, and wouldn't be communicated over the kadmin protocol.  Today we might concievably get extended errors set from libkadm5, and at any rate I need the extended hook to be called for this patch so that error() is called and exit_status is set.  I will mention that in the commit message.

For the moment I am comfortable assuming that exit_status is always 0 or 1.  If we want to add differentiating error exit statuses (doubtful), we can revisit the design.

I should not have used error() inside kadmin_addprinc() because those messages are notices or warnings and are not indicative of a failure to perform the operation.  I will revert that part of the change.

I did make sure to run all the OTP tests (and in fact caught a mistake as a result).

There are still a couple of kadmin -q operations in the Python tests.  I think the eventual solution is to have a t_kadmin.py, perhaps subsuming t_kadmin_acl.py, which explicitly tests kadmin -q and the interactive mode for very basic operations.  I don't think that's too important for the moment.

I am comfortable relying on the kadmin exit status to detect failure rather than the output.

I will remove the unneeded -force from t_salt.py.  Some grepping suggests that it's the only kadmin -force remaining in the Python tests.

PR #250, opened 2015-02-04
Avoid uninitialized data in t_prf.c

tlyu:
In t_prf.c, make sure that the partially initialized, faked-up
structures gss_union_ctx_id_desc and krb5_gss_ctx_id_rec are zeroed.
This avoids uninitialized reads in gss_pseudo_random(), which can
cause intermittent test failures on some platforms.


greghudson:
This seems reasonable.  (The particular mistake I made was failing to set kgctx.terminated.)

PR #251, opened 2015-02-05
Add configure checks for portability assumptions

tlyu:
Check a few portability assumptions: integers are two's complement,
unsigned-to-signed conversions preserve representations, and bytes are
8 bits.

### DIFF
--- a/src/configure.in
+++ b/src/configure.in
@@ -22,6 +22,40 @@ AC_SUBST(KRB5_VERSION)
 
 AC_REQUIRE_CPP
 
+AC_CACHE_CHECK([whether integers are two's complement],
+  [krb5_cv_ints_twos_compl],
+  [AC_COMPILE_IFELSE(
+    [AC_LANG_BOOL_COMPILE_TRY(
+[#include <limits.h>
+],
+	[/* Basic two's complement check */
+	  ~(-1) == 0 && ~(-1L) == 0L &&
+	  /* Check that values with sign bit 1 and value bits 0 are valid */
+	  -(INT_MIN + 1) == INT_MAX && -(LONG_MIN + 1) == LONG_MAX &&
+	  /* Check that unsigned-to-signed conversions preserve bit patterns */
+	  (int)((unsigned int)INT_MAX + 1) == INT_MIN &&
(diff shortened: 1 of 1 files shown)


greghudson:
Okay, following along:

* C requires that signed integers be represented in two's complement, one's complement, or sign and magnitude.  It also allows padding bits whose values are unspecified and may include trap representations.

* ~(-1) == 0 and ~(-1L) == 0L eliminate one's complement and sign and magnitude.  For one's complement, ~(-1) == 1, and for sign and magnitude, ~(-1) is a large positive value.

* The second pair of tests is, I think, trying to eliminate implementations using padding bits.  I don't see how they do that, though.  It seems like an implementation with padding bits would still be required to pass those conditions.

* The third pair of tests deals with converting values to signed integer types which cannot be represented within the range of the signed type.  C99 6.3.1.3 says "either the
result is implementation-defined or an implementation-defined signal is raised."  "Implementation-defined" means the result must be consistent; unfortunately, the standard doesn't say that a non-signal result must correspond to what you would expect from a bit-for-bit conversion within the chosen integer representation (e.g. two's complement).  These tests are a reasonable spot check, at least.

tlyu:
The second pair of tests is to check that the value with sign bit one and value bits zero is a valid representation, which is only true for two's complement when -INT_MAX != INT_MIN, etc.  C99 allows -INT_MAX = INT_MIN even for two's complement.  Testing for padding bits seems like it would be considerably more involved and might not be possible with a compile-time test.

greghudson:
Ah, I see the relevant C99 clause.  This patch looks okay, then.

kaduk:
Yeah, this seems okay.  Thanks for writing up the analysis, Greg.

tlyu:
Should I update the comments to make them more verbosely describe what they're testing?

greghudson:
It might be cool to reference the relevant sections of C99, but being concise is also good.  (Elaborating a bit in the commit message might make sense; people can find that with git blame if they're really curious.)

tlyu:
Tweaked comments and made a more verbose commit message.

greghudson:
Seems okay to me.

PR #252, opened 2015-02-05
Implement secure cookies

npmccallum:
RFC 6113 specifies the ability for the KDC to send a PA-FX-COOKIE
to the client which will be returned on the next AS-REQ. This permits
the KDC to retain state for the subsequent request. In the majority of
cases, this cookie must be encrypted.

This patch enables the transparent encryption/decryption of such
cookies. After the request is decoded, any cookies found are decrypted.
Any cookies set, either internally or by kdcpreauth plugins, will be
automatically encrypted before the reply is encoded.

Additionally, a timestamp is inserted into the cookie before encryption.
This timestamp is validated for freshness during decryption.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -1469,6 +1469,14 @@ encode_krb5_pa_otp_enc_req(const krb5_data *, krb5_data **);
 krb5_error_code
 encode_krb5_kkdcp_message(const krb5_kkdcp_message *, krb5_data **);
 
+typedef struct _krb5_secure_cookie {
+    krb5_timestamp time;
+    krb5_pa_data data;
+} krb5_secure_cookie;
+
+krb5_error_code
+encode_krb5_secure_cookie(const krb5_secure_cookie *, krb5_data **);
+
 /*************************************************************************
  * End of prototypes for krb5_encode.c
  *************************************************************************/
@@ -1641,6 +1649,9 @@ decode_krb5_pa_otp_enc_req(const krb5_data *, krb5_data **);
 
 krb5_error_code
 decode_krb5_kkdcp_message(const krb5_data *, krb5_kkdcp_message **);
+
+krb5_error_code
+decode_krb5_secure_cookie(const krb5_data *, krb5_secure_cookie **);
 
 struct _krb5_key_data;          /* kdb.h */
 
--- a/src/kdc/Makefile.in
+++ b/src/kdc/Makefile.in
@@ -9,6 +9,7 @@ all:: krb5kdc rtest
 LOCALINCLUDES = -I.
 SRCS= \
 	kdc5_err.c \
+	$(srcdir)/cookies.c \
 	$(srcdir)/dispatch.c \
 	$(srcdir)/do_as_req.c \
 	$(srcdir)/do_tgs_req.c \
@@ -29,6 +30,7 @@ SRCS= \
 
 OBJS= \
 	kdc5_err.o \
+	cookies.o \
 	dispatch.o \
 	do_as_req.o \
 	do_tgs_req.o \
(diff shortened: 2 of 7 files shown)


greghudson:
From a brief scan:

* The ASN.1 encoder/decoder needs associated tests, and should be in a separate commit.

* RFC 6113 only specifies a single cookie from the KDC, and our client implementation only reflects a single token.  Although there wasn't necessarily a good reason for this, there also isn't a compelling reason to change it.

* I also don't think it's a particularly clean design for modules to generate cookies as generic padata which is then post-processed into encrypted cookies.  That design certainly doesn't work well for the hint list, as the get_edata() method doesn't allow more than one padata output.  I would prefer to add a new callback for generating a cookie; kdc_preauth.c can assemble the cookie values and marshal them into a PA-FX-COOKIE padata value.  Similarly, kdc_preauth.c can decrypt any incoming cookie and provide a second new callback to read the cookie value for a specific padata type.

* I think we should document our cookie format, probably in a new top-level section for non-standardized formats (such as the FILE ccache and keytab formats, which are currently only documented as part of GNU Shishi).  If you write the .rst file, I can take care of putting it in the right place.

npmccallum:
I don't think a cookie is needed in get_edata() in pretty much any case. It certainly isn't needed for PAKE.

npmccallum:
I also disagree that there isn't a compelling reason to change it. Having multiple preauth mechs be able to set cookies independently is far less code than having to manage a single monolithic cookie. I'm happy to write a patch to support multiple cookies on the client.

npmccallum:
Why do we need to document the cookie format when it is intentionally opaque?

greghudson:
People have been known to mix KDC implementations within a realm, although it's not common.  A documented cookie format can allow interoperability in that circumstance.  It's why we use the same AD-SIGNTICKET as Heimdal, for instance.

I'll see if I can get Tom and/or Ben to voice opinions on our design disagreement over multiple cookies and postprocessing vs. callbacks and a single cookie.

On IRC you suggested that cookies could be used for PKINIT freshness (although we don't know if that idea will pan out).  That's definitely a circumstance where a cookie would be needed in the hint list.  I'm not sure where you're at with PAKE, but if we use client hints to allow the KDC to supply its public value in the hint list, I would expect the KDC to want to supply a cookie as well.

(Of course, cookies aren't strictly required for any new preauth mech.  As I noted a few days ago on IRC, the preauth mech protocol can simply require the client to reflect a value provided by the KDC.)

greghudson:
I took another look at how you are preprocessing incoming cookies.  Since the kdcpreauth interface doesn't give access to the full padata list, you appear to be mutating request->padata so that modules can find the decrypted cookie value there.

We have already put significant effort into getting rid of KDC code which mutates incoming requests.  I don't think we would want to go backwards on that.

As a side note, designated initializers are not currently part of our portability assumptions because they aren't supported (or are only supported very recently) by MSVC.  That might change in the future, but for now we need to avoid them.

kaduk:
RFC 6113 is pretty clear that only a single cookie is to be used.  Countermanding that is in effect a protocol change, and I'm not very enthusiastic about making protocol changes solely for (relatively minor) ease of implementation.

In general, I like having file formats and data structure formats be documented in prose independently of the code implementing them, even if there is no (current) need for interoperability.  Having standalone documentation makes it easier to understand the code and consider modifications or extensions to the format; the lack of such documentation has made life much harder for me in parts of my work with OpenAFS.

I agree with Greg that we are trying very hard to get out of the business of mutating the incoming request, so the callback approach seems a pretty natural solution.

As a side note, some of the comments in cookies.c are pretty banal and could probably be removed.

PR #253, opened 2015-02-06
Fix replay issue of AS-REQ with PKINIT

tc-anssi:
AS-REQ packets of clients using PKINIT can be replayed to the KDC. This is
  non-compliant with RFC 4556, since the timestamp in the pkAuthenticator
  is not checked. Hence, for each replayed packet, a fresh AS-REP is sent
  back by the KDC. However, no valid scenario to exploit this issue has been
  identified. This commit adds the missing check.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_srv.c
+++ b/src/plugins/preauth/pkinit/pkinit_srv.c
@@ -429,6 +429,12 @@ pkinit_server_verify_padata(krb5_context context,
             goto cleanup;
         }
 
+        /* check pkAuthenticator timestamp */
+        retval = krb5_check_clockskew(context, auth_pack->pkAuthenticator.ctime)
+        if (retval) {
+            goto cleanup;
+        }
+
         /* check dh parameters */
         if (auth_pack->clientPublicValue != NULL) {
             retval = server_check_dh(context, plgctx->cryptoctx,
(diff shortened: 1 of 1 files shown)


greghudson:
It occurs to me that checking the timestamp on the KDC will create issues like RT #7114 where client clock skew prevents authentication even with kdc_timesync turned on.

We can mitigate this by using cb->get_preauth_time() in pkinit_clnt.c instead of krb5_us_timeofday(), thus using the unauthenticated timestamp in the preauth-required error.  I don't see any security consequences from this; an attacker could induce a legitimate client to create a pkAuthenticator for a future timestamp, but replaying that authenticator in the future would only result in issuing a ticket that the attacker cannot decrypt.  And once freshness tokens are deployed and required, that attack would become impossible without the ability to create a valid KDC freshness token for the future timestamp.

Of course a 1.14 KDC would still have clock skew issues with <1.14 PKINIT clients, but the same is already true of Heimdal and presumably MS KDCs and <1.14 PKINIT clients.

greghudson:
https://github.com/greghudson/krb5/commits/pkinit_time has what I am planning to push.

greghudson:
Pushed to master as 54984d618e01027abe73e6772fe7049c79938518 fcc1076541a3bd9a5fa4db0be6f74888b3f5f193

PR #255, opened 2015-02-23
Test KDB module and S4U2Proxy tests

greghudson:
The first commit adds a new test KDB module.  The second commit uses the module to add successful S4U2Proxy delegation tests.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -18,6 +18,7 @@ SUBDIRS=util include lib \
 	plugins/pwqual/test \
 	plugins/kdb/db2 \
 	@ldap_plugin_dir@ \
+	plugins/kdb/test \
 	plugins/preauth/otp \
 	plugins/preauth/pkinit \
 	plugins/tls/k5tls \
--- a/src/configure.in
+++ b/src/configure.in
@@ -1446,6 +1446,7 @@ dnl	ccapi ccapi/lib ccapi/lib/unix ccapi/server ccapi/server/unix ccapi/test
 	plugins/kdb/db2/libdb2/recno
 	plugins/kdb/db2/libdb2/test
 	plugins/kdb/hdb
+	plugins/kdb/test
 	plugins/preauth/cksum_body
 	plugins/preauth/otp
 	plugins/preauth/securid_sam2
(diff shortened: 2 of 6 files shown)


greghudson:
While writing up regression tests for RT #8139 I realized I would need multiple kvnos for one of the cases.  So, re-pushed with a more complicated make_keys() function which allows multiple keys entries and a leading kvno for each entry.  Also more function comments.

kaduk:
It's unclear that we need $(S) in the Makefile if we don't build the kdc on windows.

kdb_test.c:61 says that all data has a kvno of 1, still.

Is "stype" a common enough abbreviation (for salt type) that it needs no expansion in the comment for make_keyblock()?  make_keyblock() also seems to leak hashstr.

In make_keys()'s inner loop, the kd++ could be moved up into the loop declaration, but I don't have a strong preference either way.  The ks_list's contents seem to be leaked.

testhandle->names is a use-after-free waiting to happen, but this is a test module so it's probably fine.

There might be some logic in the free_principal and decrypt_master_key routines to handle situations that the test module cannot currently produce, but it doesn't seem harmful to have it in.

It seems like we ought to add some tests for unsuccessful S4U2Proxy as well.

greghudson:
We use $(S) everywhere for mydir and BUILDTOP, or at the very least we do in other kdb module makefiles.

I fixed some of the comments related to key values including the one at line 61.  I changed "stype" to "salttype".  I added a free(hashstr); to make_keyblock().

In make_keys(), I do not want to put kd++ in the for loop expression because kd's lifetime is not limited to that loop.  I added code to free the ks_lists contents.

I agree that testhandle->names would be bad memory management for production code, but it's fine for a test module and it makes the code a little bit more concise.

A KDB module's free_principal() routine must be able to handle input constructed by the caller using krb5_db_alloc().  It's kind of terrible but I haven't thought of a better design and a transition path.  That's why there's code there to handle more than we actually generate.  decrypt_key_data() may be in a similar situation but I'm not as certain.

There are already tests for unsuccessful S4U2Proxy using the db2 module (which always says no).  If we added additional tests for unsucessful S4U2Proxy using the test KDB module, it would more fully test the module's check_allowed_to_delegate method, but it wouldn't test the KDC logic any better.

kaduk:
Okay, sounds like a good plan.

kaduk:
I looked at the changed code (but didn't save the old hash so as to get a diff between the old and new trees), and it looks good.

PR #256, opened 2015-02-26
Fix kadmin script mode command-not-found error

greghudson:
In ss_wrapper.c, if ss_execute_command() returns an error, we should
call ss_perror() with *args as the third argument and not request
(which is NULL).  Expand out the conditional into three commented
branches for greater clarity, since the error-handling is no longer
identical for the ss_execute_command() and ss_execute_line() cases.


tlyu:
This looks OK to me.

PR #257, opened 2015-03-03
Load local TGT for KDC requests and fix AD-SIGNEDPATH logic

greghudson:
~~Only the last four commits on the branch are for this pull request.  The preceding commits are from PR #214 (KDC authdata refactoring).  The github pull request page may sort the commits chronologically, so you might have to look at the branch commit chain directly.~~

This branch makes the server realm local TGT available to KDC authdata processing, and uses it to fix several corner cases in AD-SIGNEDPATH checksum construction and verification (RT #8139).  It will also be important for CAMMACs.


greghudson:
Rebased on top of PR #264 as well as PR #214 since there are otherwise merge conflicts between the two.  Also expanded the commit message for the first commit since the diff is more expansive than the message implied.

PR #258, opened 2015-03-05
Import names immediately with COMPOSITE_EXPORT

DirectXMan12:
RFC 6680 specifies that GSS_Export_name_composite() "outputs a token that
"can be imported with GSS_Import_name(), using GSS_C_NT_COMPOSITE_EXPORT
as the name type...".  Therefore, in the gss_import_name mechglue, we
should perform the import process imediately when either
GSS_C_NT_COMPOSITE_EXPORT or GSS_C_NT_EXPORT_NAME are used (not just
for the later, as is the current functionality).

The naming extension test was also updated to display the result
of importing with GSS_C_NT_COMPOSITE_EXPORT in addition to
GSS_C_NT_EXPORT_NAME.

### DIFF
--- a/src/lib/gssapi/mechglue/g_imp_name.c
+++ b/src/lib/gssapi/mechglue/g_imp_name.c
@@ -36,7 +36,7 @@
 #include <errno.h>
 
 /* local function to import GSS_C_EXPORT_NAME names */
-static OM_uint32 importExportName(OM_uint32 *, gss_union_name_t);
+static OM_uint32 importExportName(OM_uint32 *, gss_union_name_t, gss_OID);
 
 static OM_uint32
 val_imp_name_args(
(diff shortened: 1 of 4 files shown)


greghudson:
Mostly nitpicks:

* The three-line comment added to g_imp_name.c isn't formatted according to our usual convention for 3+ line comments, and I also think it's unnecessary; it's almost seems like it's there to justify the change, rather than the code in its new state.

* The one-line // comment added to g_imp_name.c seems incomplete, is probably unnecessary, and we don't use that kind of comment.

* In t_namingexts.c, display_name() would probably be better than display_non_canon_name().  That would also serve to eliminate the >79-character line at line 81.

* In t_namingexts.c, there is no call to check_gsserr() after the import with GSS_C_NT_COMPOSITE_EXPORT.

More significantly, there are no automated tests, since we don't use t_namingexts.c in make check.  I think it shouldn't be too difficult to add automated tests by adding a flag to t_export_name.c to do a composite export instead of a simple export, and then invoke it that way from t_gssapi.py.

greghudson:
Hm, the new test fails for me; whatever goes into a composite export token for an imported name doesn't seem to be consistent.  I will take a closer look.

greghudson:
Oh, of course: you're using realm.host_princ in the test, which will be different for every host.  Can you redo it with p:a@b as in the previous tests?

greghudson:
Pushed to master as 29dec110c43ae9cebdcd935906a3131ca9ac0c99

PR #259, opened 2015-03-09
Larger kvno support

greghudson:
This started out as a plan to just implement the 32-bit keytab kvno extension, but wound up becoming a larger overhaul as I discovered other limitations.  Each of the five commits is still fairly short.  For more background, see:

http://k5wiki.kerberos.org/wiki/Projects/Larger_key_versions

### DIFF
--- a/doc/formats/keytab_file_format.rst
+++ b/doc/formats/keytab_file_format.rst
@@ -31,6 +31,7 @@ the key entry.  Key entries use the following informal grammar::
         enctype (16 bits)
         key length (32 bits)
         key contents
+        key version (32 bits) [in release 1.14 and later]
 
     principal ::=
         count of components (32 bits) [includes realm in version 1]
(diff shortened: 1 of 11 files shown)


kaduk:
As a general comment (relevant for multiple commits), do we want to require that the 8-bit version is the last octet of the 32-bit version and report an error otherwise, or are we happy to just ignore the 8-bit version and make the 32-bit version always an override?

In 2d3b588, I think I will be comfortable not bumping the symbol version of krb5_dbe_def_cpw() for the behavior change introduced by this commit, so not bumping the KDB binary version seems reasonable.  Most of the changes are straightforward, but I will note that due to C's value-preserving integer promotion rules, passing a krb5_ui_2 to printf will get promoted to plain 'int', so that the "%d" format specifier would still be appropriate, since the value will always be within 0 and UINT16_MAX inclusive.  (I do not request a change; there should not be any difference in behavior.)

3643927 looks fine.

d46ebb8 is fine.  I don't know if the commit message should say something about existing implementations of xdr_char() making all the padding octets zero (so that new peers reading from old implementations do not get leading garbage), though.

In 51d6261, "f.write('\x05\x02\x00\x00\x00' + chr(len(record)))" assumes that the length of the test record will always fit in a single octet, which is probably reasonable, and will probably be pretty easy to figure out if someone ever needs to add more stuff to the hardcoded record which pushes it above that boundary.  I might add a comment to the check that klist -k recognizes our hardcoded bits as a valid keytab, though.  The actual code looks fine.

In d27523a, I wonder if we care enough to apply more complicated matching rules to the case where a kvno is supplied just in case there are multiple entries matching the low 8 bits.  My conclusion is that we probably don't need to care, at the moment.  I also gave a little bit of thought as to whether the input (matching) kvno would ever be affected by 8-bit limits, but again, my current conclusion is that we don't need to care.

Thanks for taking this on.

PR #260, opened 2015-03-12
Fix scope of kadmind ACL wildcard back-references

greghudson:
In kadm5int_acl_find_entry(), clear the wildcard back-references list
for each acl entry.  Otherwise the wildcards we process can affect
back-references for later entries.


kaduk:
I'm not immediately seeing how the test change would catch a regression for this -- where do we create a principal that would match one/*/*/five ?

The code change looks good.

greghudson:
We don't create a principal which matches that pattern.  However, one/two/three/four matches it well enough that we process the two wildcard entries trying to match it.  That throws off the back-references in the next line.

kaduk:
Ah, I see; thanks.

tlyu:
This looks good to me.  I checked that the test case seems to catch the pre-patched failure condition.

PR #261, opened 2015-03-13
Improve kadmind ACL restriction docs and diagnostics

greghudson:
The first commit changes kadm5_acl.rst to reference the correct flags namespace.  Tom may be combining these namespaces, but that may not be backportable, whereas the documentation fix definitely is.

The second commit issues a diagnostic to the kadmind logs if we fail to parse restrictions.  It's not as good of a diagnostic as we could get with a better parsing function, but it's way better than nothing.

There are lots and lots of opportunities for cleanup in this code.  I will add some entries to the Cleanups page on the wiki but will not be engaging in home renovations for the moment.

### DIFF
--- a/doc/admin/conf_files/kadm5_acl.rst
+++ b/doc/admin/conf_files/kadm5_acl.rst
@@ -72,8 +72,8 @@ ignored.  Lines containing ACL entries have the format::
 
         {+\|-}\ *flagname*
             flag is forced to the indicated value.  The permissible flags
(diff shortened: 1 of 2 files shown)


tlyu:
This looks OK to me.

greghudson:
Pushed to master as ef21069070c1eb2ab1ade1d1406f5cd3920c83a9 e9eaafeab12b2b62595f4dff2fca3345b2d95b4a

PR #262, opened 2015-03-14
Extend kadmin timeout to one hour

greghudson:
The first commit is intended to be backportable by downstreams, although I'm not sure we need to backport it ourselves as it isn't really a bugfix.  The second commit is mostly just code hygiene, although it affects the behavior of clnt_create() which we don't use within the tree.


kaduk:
That seems like what we talked about and agreed upon; +1

PR #263, opened 2015-03-17
Fix renewable ticket lifetimes

greghudson:
Commit b0661f9176f5eb2644ba459e1b1e87d3dd502174 removed the starttime
hack in the EncTicketPart decoder.  Take this into account when
computing the old lifetime of a ticket we are renewing.

This bug appeared only on master and not as part of any release, so
there is no associated ticket.


tlyu:
Verified that the test case seems to catch the failure condition.  The patch looks good to me.

greghudson:
Pushed to master as 3d9de684ea933c4887a7aebabc71287cbf5a3f3c

PR #264, opened 2015-03-19
Process TGS authdata after transited in KDC

greghudson:
The CAMMAC authorization data container requires a checksum over the
encrypted part of the issued ticket, with the CAMMAC contents
substituted for the authdata field.  For this to work, we must
finalize the non-authdata fields of the encrypted ticket part before
adding authdata.  Call handle_authdata() after checking and modifying
the transited field and potentially setting the
transited-policy-checked flag.

Also remove a redundant and inoperative conditional change to
enc_tkt_reply.times.starttime which happens after the ticket is
encrypted.  We do the same thing right after setting up the ticket
times.


tlyu:
This looks OK to me.

PR #265, opened 2015-03-23
Authentication indicator support

greghudson:
There are a lot of commits on this branch.  Each one is fairly small.  They are:

* ASN.1 UTF-8 string list encoder/decoder
* Add constant declarations
* Filter CAMMACs from non-KDC sources
* CAMMAC and authind utility functions
* kdcpreauth callback for auth indicator
* Insert indicators in AS-REQs
* Extract and propagate indicators in TGS-REQs
* Enforce indicators
* Tests
* OTP support for generating indicator (and tests)
* PKINIT support for generating indicator (and tests)
* Documentation

~~I am using the placeholder value -1169 for the authentication indicator; I don't expect to push this to master until I have replaced that with an assigned value.~~

Not implemented yet:

* Cross-realm support (probably requires an allow_indicator string attribute or similar)
* KDB module interaction; the only plugin support so far is for kdcpreauth modules
* Omitting unnecessary verifiers (e.g. kdc-verifier for cross TGT tickets)
* Allowing ticket modification requests using svc-verifier only

If we don't implement the last bullet item before 1.14 we will be sort of stuck including kdc-verifiers in all CAMMACs, but that wouldn't be the end of the world.


greghudson:
Revised to use the value 97 for auth-indicator.  (This assignment is pending objections on the kitten list.)

There is now a conflict between this PR and PR #241 as both add a test preauth module.  My plan is to get PR #241 pushed to master and change the "Add test preauth module" commit in this PR to instead add indicator functionality to PR #241's test preauth module.

greghudson:
Rebased against master.  I squashed the s4u2proxy test harness, test preauth module changes, and auth indicator tests into one commit, so there are now only twelve commits (plus PR #231) on the branch.

tlyu:
Reviewed in person, with minor changes.  Looks good now.

PR #266, opened 2015-03-26
Disable principal renames for LDAP

greghudson:
The current principal rename procedure does not work with the LDAP KDB
module, instead having the effect of deleting the principal.  The fix
is not easy and requires amending the DAL (see issue #8065).  For now,
detect LDAP and error out when a rename operation is attempted.


kaduk:
This seems okay.

I feel like we ought to mark 0x7FFE as reserved in kdb.h at some point, though.

greghudson:
Pushed to master as 8483243664a289fea142d8a9de61eba30d713871

PR #267, opened 2015-03-31
kpropd -t full dump patch

greghudson:
Here is John Devitofranceschi's commit from RT #8161, unmodified.

It's a good proof of concept; it demonstrates that since the listener child process will exit after accepting a full dump, the master process can simply wait for it to exit.  However, in addition to having some style issues, it is a bit too hackish:

* Most problematically, it contains a recursive call to main(), which is not a  safe practice.

* The meaning of waitfor_do_standalone is not well-stated; it begins at 0, is incremented to 1 when the first full dump is requested, and is incremented again to 2 after the full dump is complete (and possibly again to 3 if another dump is requested).  When the patch uses it, it specifically compares it to 1.  So it's implicitly a tri-state, beginning life at "no dump was requested", incrementing to "a dump was requested and should be waited for," and then from there to "a dump was already waited for and we shouldn't do that again."

* Nothing prevents a second full dump from being requested in the second invocation of main().  If this happens, we won't wait for it and will instead kill the child at some random point during its execution, hopefully (but not necessarily) before it does very much.  That's not very airtight design.

I will see if I can come up with something backportable and more safe.  I will also see about implementing the item from http://k5wiki.kerberos.org/wiki/Cleanups about simplifying kpropd back down to one process, which would make runonce easier to implement correctly.

### DIFF
--- a/doc/admin/admin_commands/kpropd.rst
+++ b/doc/admin/admin_commands/kpropd.rst
@@ -15,6 +15,7 @@ SYNOPSIS
 [**-p** *kdb5_util_prog*]
 [**-P** *port*]
 [**-d**]
+[**-t**]
 
 DESCRIPTION
 -----------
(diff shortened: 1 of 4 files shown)


greghudson:
I replaced John's commit with three commits of my own.

The first commit fixes an old shortcoming of kpropd since Nico's rework in 1.11, by signalling the parent from the fullprop child after a dump is received.

The second commit documents kpropd -t and fixes it to simply avoid exiting after requesting a full resync.  We also don't exit from the fullprop child in case multiple full resyncs are required.

The third commit adds tests.

kaduk:
af2c734 looks plausible, but I don't have the full kprop architecture in my head so I could have missed something architectural.

Similarly, 6a9d96f looks okay.

The tests look fine, too.

PR #268, opened 2015-04-01
Make all Python test scripts executable

greghudson:
For the convenience of developers manually running Python test
scripts, set the executable bits on all of them, and make sure the
first line is always "#!/usr/bin/python".


kaduk:
Seems straightforward enough; +1

PR #269, opened 2015-04-08
Add tests for client principal aliases

greghudson:
Augment the LDAP KDB module tests to include client principal aliases
as well as server principal aliases.  Also revise the server principal
alias tests to include an AS-REQ case.  (This requires adjusting the
subsequent test not to assume a ccache containing a TGT.)


kaduk:
Seems okay.

It will need tweaks if backported since 60516bb111ac68ce0d809043d46c0c1f815a7b30 changed the kadminl semantics, but I don't see a reason to want to backport it, so that's probably irrelevant.

PR #270, opened 2015-04-09
Avoid unnecessary iprop full resyncs after resets

greghudson:
The first three commits are from PR #267.

The last commit resolves the second "related problem" documented at http://k5wiki.kerberos.org/wiki/Projects/Hierarchical_iprop#Related_problems
(the first problem is resolved by PR #267) by creating dummy entries after header resets.

### DIFF
--- a/src/lib/kdb/kdb_log.c
+++ b/src/lib/kdb/kdb_log.c
@@ -51,13 +51,10 @@ time_current(kdbe_time_t *out)
 }
 
 /* Sync update entry to disk. */
-static krb5_error_code
+static void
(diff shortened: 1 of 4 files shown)


kaduk:
I had to think for a bit about the use of set_dummy() in ulog_set_last(), but it seems to be accurate.

The error message in t_iprop.py in the case where we expected a dummy entry could perhaps also print what entry we did get?  Though, I guess that's not entirely consistent with the existing state of things.

I skimmed the test changes and they seem to be as expected for this code change.

So, +1.

PR #271, opened 2015-04-10
Use the appropriate nm via the exported variable to help/fix cross compiling

tgurr:
Reason explained here: https://galileo.mailstation.de/gerrit/#/c/1117/
Idea for the fix taken from: http://opensource.apple.com/source/llvmgcc42/llvmgcc42-2336.9/libstdc++-v3/scripts/make_exports.pl

### DIFF
--- a/src/util/export-check.pl
+++ b/src/util/export-check.pl
@@ -32,13 +32,15 @@ $0 =~ s/^.*?([\w.-]+)$/$1/;
 # This code assumes the GNU version of nm.
 # For now, we'll only run it on GNU/Linux systems, so that's okay.
 
+my $nm = $ENV{'NM'} || "nm";
+
 if ($#ARGV != 1) {
     die "usage: $0 exportfile libfoo.so\n";
 }
 my($exfile, $libfile) = @ARGV;
 
 @missing = ();
-open NM, "nm -Dg --defined-only $libfile |" || die "can't run nm on $libfile: $!";
+open NM, "$nm -Dg --defined-only $libfile |" || die "can't run nm on $libfile: $!";
 open EXPORT, "< $exfile" || die "can't read $exfile: $!";
 
 @export = <EXPORT>;


tgurr:
Is there anything wrong with my commit? For clarification this is needed when the build tools are prefixed like for example we do to allow to cross compile:

"Some build tools are now prefixed with the host architecture, so you can call the one you want to build for, for example you'll find /usr/${host}/bin/${host}-nm and possibly multiple /usr/${host}/bin/${target}-nm"

The only problem within krb5 is the single hardcoded nm in the src/util/export-check.pl file, for everything else autotools are doing the right thing.

tlyu:
I think I recall us being confused as to why you could expect the NM variable to be exported.  As far as I know we currently do nothing to export the NM make variable as an environment variable.

tgurr:
It falls back to nm when nothing is exported so existing working behaviour isn't affected. However I'm not that firm in Perl so if you come up with a better fix you're more happy with then even better.

Another approach would be to sed in our prefixed nm in our package build script downstream, but I'd rather like to see this fixed properly upstream since I'm sure more and more distributions and/or projects will run into such issues in the future and there are plenty of such problems, so we try our best to upstream our fixes so everyone can benefit.

tlyu:
Do you explicitly set the NM environment variable in your downstream build process?  It seems like we don't even set a make variable for it, so we would have to introduce some additional autoconf tests if we wanted to set the make variable "correctly".  Prefixing the command as in "NM=$(NM) .../export-check.pl" could then work.

You might be able to skip the export-check.pl invocation if you're not adding/removing APIs or making any changes to the export list; it's mostly there as a consistency check.

greghudson:
I think I would prefer to conditionalize running export-check on --enable-maintainer-mode, which Tom was starting to get at in his last sentence.  That requires a little bit of restructuring of the shlib.conf outputs, I think.

tgurr:
Yes, we set e.g. NM=x86_64-pc-linux-gnu-nm in our build environment.

Putting the export-check into --enable-maintainer-mode would probably also work for us downstream since we usually don't enable maintainer-mode when building packages.

tlyu:
Closing because I've opened PR #373 with an alternative solution.

PR #272, opened 2015-04-13
Fix LDAP ticket policies on big-endian LP64

greghudson:
krb5_ldap_get_value() takes a pointer to int, and should not be passed
a pointer to any integral type which might have a different width.
Use an intermediate variable for each call.

The erroneous calls in ldap_misc.c were passing pointers to int32_t,
which is harmless on all common platforms.  The calls in
ldap_tkt_policy.c were passing pointers to long; on big-endian LP64
platforms, the result would be written to the high 32 bits of the long
value.


kaduk:
In ldap_tkt_policy.c, val does not need to be initialized, but I guess the prevailing style is to always initialize?

+1 regardless.

PR #273, opened 2015-04-13
Remove STRING_BUFFER() macro in gssapi_generic.c

tlyu:
In gssapi_generic.c, struct mech_attr_info_desc included some
gss_buffer_desc members whose length fields were never used.
Additionally, the STRING_BUFFER() macro's computation of the (unused)
length fields was incorrect, causing warnings in some versions of
clang.  Remove the problematic STRING_BUFFER() macro and adjust the
array and generic_gss_display_mech_attr() accordingly.


greghudson:
The string constant at line 234 no longer needs to be wrapped.

The string constant at line 301 still needs to be wrapped, but the second line now has the wrong indentation amount.  This line and the one at line 357 might read better if they were re-wrapped at a later point in the string, but it's not terribly important.  (There's also a typo at line 357, but that's out of scope and arguably shouldn't be fixed in a non-backported commit.)

tlyu:
Unwrapped the line 234 string, and fixed the indentation on the line 301 string.

greghudson:
Looks good now.

PR #274, opened 2015-04-15
Fix memory leak in DB2 iteration

greghudson:
Use the correct function to free the decoded principal entry in
curs_run_cb().


kaduk:
+1

PR #275, opened 2015-04-16
Ribbon improvements for accessibility

kaduk:
In our testing, the MFC library's ribbon is not very accessible (e.g., for screen readers), and we have been requested to support that use case.

Swap out the MFC ribbon for the standard windows library ribbon, which exposes all of its controls to screen reading software.

Also included are some miscellaneous cleanup encountered along the way while building in various build environments.

These are the first new C++ files added to our tree in a long time, so we should take the opportunity to decide what style we want to have in the future.


kaduk:
I realized that I still have some literal tab characters in the new files; I have a fix in my local branch but will hold off on updating the pull request so as to keep the commit hashes stable for now.

greghudson:
* In "Remove MBCS...": I suggest changing "quiesce" to "quiet" in the commit message; quiescing the build would mean rendering it inactive, not less noisy.

* In "Switch to Windows SDK...":
  * Commit messages for new code shouldn't generally contain code walkthroughs.  Anything non-obvious should be expressed in comments.
  * operator new cannot return NULL (except in VC6, which I don't believe we need to care about), so don't check for that.  Two occurrences.
  * Is there a difference between !SUCCEEDED(ret) and FAILED(ret)?  If not, I suggest consistently using the latter.
  * LeashUIApplication::OnDestroyUICommand() probably should return S_OK.  Either way, it's not very useful to express uncertainty in comments.

* In "Remove another lingering Leash reference", the flags bitmask should probably be wrapped to the next line.

kaduk:
I pushed some updates; move discussion from commit message to code comments where appropriate, replace !SUCCEEDED with FAILED, etc..

greghudson:
I see one test for "ret != S_OK" in the new code.  Should that be "FAILED(ret)"?  (It's in SaveRibbonState().)

In the second-to-last commit, I was suggesting moving the whole bitmask to the next line, since I think it fits on one line.  Not really a big deal though.

Other than those two things, I think everything I said looks to be addressed.

kaduk:
Pushed to master as 0fdbfb1

PR #276, opened 2015-04-21
Use memory cache in gss_acquire_cred_with_password

greghudson:
gss_acquire_cred_with_password() was originally introduced in Solaris.
When we introduced it in 1.9, we unfortunately gave it different and
less useful semantics.  Restore this function to the Solaris
semantics, which are to always get credentials and store them in a
private memory ccache.  The caller can use gss_store_cred() to make
the resulting creds visible to other processes if desired.

### DIFF
--- a/src/appl/gss-sample/t_gss_sample.py
+++ b/src/appl/gss-sample/t_gss_sample.py
@@ -41,7 +41,6 @@ def server_client_test(realm, options):
     if 'Signature verified.' not in output:
         fail('Expected message not seen in gss-client output')
     stop_daemon(server)
-    realm.klist(realm.user_princ, realm.host_princ)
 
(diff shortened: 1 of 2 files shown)


greghudson:
> Why not move this out to the relevant call sites of acquire_cred()?

Primarily because krb5_gss_acquire_cred_with_password() doesn't have a krb5_context on hand, and it would need one to create the memory cache.  It would also be slightly harder to make sure that the cache gets destroyed on both success and failure, but that's manageable.

> Or assert(req_ccache == NULL);?

I can add an assert, sure.

nicowilliams:
I see.  The ugly things we must do to cope with the lack of a krb5_context for GSS...

tlyu:
Looks OK to me.

PR #278, opened 2015-05-04
Add rename method to kadm5_hook

greghudson:
Unmodified patch from John Hascall.  The versioning isn't right, and we should consider adding an automated test.


greghudson:
I fixed the patch to use a minor rather than major version bump, and added a test case.  Should be ready.

tlyu:
Looks OK to me.

greghudson:
Pushed to master as a99e5565e99b83a86002332e39938aa6bed6a26a

PR #279, opened 2015-05-05
Kfw updates

kaduk:
A couple fixes that had been sitting around in my local branch, plus adding support for a registry value for the default realm, and bumping the release number to 4.1.x.  Notably, do not set allow_weak_crypto in the MIT-internal krb5.ini any more.

### DIFF
--- a/src/lib/krb5/ccache/cc_mslsa.c
+++ b/src/lib/krb5/ccache/cc_mslsa.c
@@ -1568,8 +1568,8 @@ krb5_lcc_resolve (krb5_context context, krb5_ccache *id, const char *residual)
         /* Take the first client principal we find; they should all be the
          * same anyway. */
         for (i = 0; i < pResponse->CountOfTickets; i++) {
-            if (UnicodeStringToMITPrinc(&pResponse->Tickets[0].ClientName,
-                                        &pResponse->Tickets[0].ClientRealm,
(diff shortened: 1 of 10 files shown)


greghudson:
KfW shortcut keys: patch looks fine, but the commit message is confusing to me.

Fix loop: the commit message might want to indicate what the consequences of the bug were.

Do not set allow_weak_crypto: looks good.

Supply a hostrealm plugin:
* The summary should say "hostrealm module"; we try to avoid using the term "plugin" as a noun.
* In get_from_registry(), I suggest replacing the -1s with EIO.
* We generally put output arguments last, so get_from_registry() should have its parameters reversed.  Also, get_from_registry() should set *buf_out = NULL at the beginning.  And, "buf" is kind of a weird name for a callee-allocated string; I would be happer with "str_out" or "value_out".
* The comment for get_from_registry() does not need to restate the function name.
* The second paragraph should be written in the imperative, e.g. "On success, return 0 and set *buf_out to....  Return KRB5_PLUGIN_NO_HANDLE if the registry key does not exist, or another code if it cannot be read."
* In registry_default_realm(), /* else */ is not a convention we typically use.  This function can be made less repetitive using:

    ret = get_from_registry(&prof_realm, HKEY_LOCAL_MACHINE);
    if (ret == KRB5_PLUGIN_NO_HANDLE)
        ret = get_from_registry(&prof_realm, HKEY_CURRENT_USER);

The last commit looks okay.

PR #280, opened 2015-05-15
OpenSSL plugin: set hostname TLS extension to support SNI

tiran:
This simple patch adds server name indicator support to the OpenSSL plugin. SNI is required for e.g. virtual hosting of a KDC proxy instance. With SNI the webserver can select the correct vhost and certificate.


greghudson:
Here is what I plan to push: https://github.com/greghudson/krb5/commit/be0410b64da265e7905a7aa68e58ebbb48cfa80f

Same patch, rewritten commit message.

greghudson:
Pushed to master as 4b6045adb7a044cd7ddc3987da2f26bf8a5281fe

PR #281, opened 2015-05-15
Better error messages in krb5_cccol_have_content()

nicowilliams:
GSS users don't get ccache filenames in no credentials error messages.  This fixes that.

### DIFF
--- a/src/lib/krb5/ccache/cccursor.c
+++ b/src/lib/krb5/ccache/cccursor.c
@@ -227,16 +227,22 @@ krb5_cccol_have_content(krb5_context context)
     krb5_ccache cache;
     krb5_creds creds;
     krb5_boolean found = FALSE;
+    krb5_error_code ret;
+    int checked_one = 0;
+    const char *name;
 
-    if (krb5_cccol_cursor_new(context, &col_cursor))
+    if ((ret = krb5_cccol_cursor_new(context, &col_cursor)))
         goto no_entries;
 
-    while (!found && !krb5_cccol_cursor_next(context, col_cursor, &cache) &&
+    while (!found &&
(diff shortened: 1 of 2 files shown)


greghudson:
I guess this supercedes PR #212?

The first commit isn't needed.  err_fmt is already documented in krb5_conf.rst; the man page just hasn't been regenerated yet.

nicowilliams:
@greghudson Oy, I'd completely forgotten about #212.  I've closed it.

greghudson:
Sorry it's taken so long to review this.  I can make any necessary changes so you don't have to swap the code back in.  Aside from some logic and style issues, there are a couple of behavioral issues to discuss:

* When the collection cursor returns no caches, you want to report the value of $KRB5CCNAME in addition to the default cache name.  That's verbose, kind of a layering violation, and (from past discussions) seems very narrowly tailored to a specific set of customer requirements.  Will you settle for just including the default cache name?

* If the cccursor returns some caches, and we get no errors iterating through them but no creds are returned (or all of the creds are ccconfig creds), we have no error to wrap; every operation returned success.  I assume there's nothing better to do in this case than treat it the same way as a collection with no ccaches, i.e. report the default cache name but don't wrap an error.

greghudson:
I resolved these questions with Nico by voice (reporting the default cache name is adequate; if caches are found but no errors are encountered, report the default cache name).

Please have a look at the top commit here: https://github.com/greghudson/krb5/commits/have_content_err

greghudson:
Pushed to master as 7746beda3b0312216ac3ffa18fa3179f252b15f4 after a positive review comment from Nico.

nicowilliams:
Thanks!

PR #283, opened 2015-05-19
Implement possiblitiy to not force conf/integ in gssapi

gladiac:
This is needed to implement the SASL/GSS-SPNEGO client in Samba.

I submit this patchset for discussion. It misses documentation ...


greghudson:
Please include a clear, verifiable description of the need for this change in the commit message.  The justifications given so far in other channels still do not make sense to me.

nicowilliams:
I agree with @greghudson.  Why is this necessary?

nicowilliams:
Simo explained the issue on IRC.  There's a pseudo-SASL mechanism that is called GSS-SPNEGO and which is not RFC4752-compliant: it has no wrap token exchange for negotiating security layers, it has no authz-id support, and no negotiation of maximum SL message size.  The GSS_C_INTEG/CONF_FLAG outputs are taken by the acceptor to reflect what the client wants, and it can't have both.

We can't simply not assert these flags because the mech has historically asserted them regardless of whether the app asked for them.

We can't add new req_flags (GSS_C_NO_INTEG_FLAG) because those bits are scarce.

With the GSS-APIv2u1 init/accept model there's no way to deal with this other than via a cred option.

Therefore matching the Heimdal interface for this should do.

greghudson:
Thanks; the reference to https://groups.yahoo.com/neo/groups/cat-ietf/conversations/topics/575 was helpful.

I will review the details of this patch.  I think we have an open issue in RT from ages ago, so I will also hunt that down and add details to it.

greghudson:
http://krbdev.mit.edu/rt/Ticket/Display.html?id=6938 is the ticket.

gladiac:
I will update the comments in the commit message. However I wonder if the OID I choose is ok, if the function names are fine for you or if I should change it ...

gladiac:
Oh, should I use the same OID constant as in Heimdal?

GSS_KRB5_CRED_NO_CI_FLAGS_X - 1.2.752.43.13.29

greghudson:
Use the same OID value and constant name as Heimdal.  Declare the constant in gssapi_krb5.h.  Do not add a public function for this; applications will use gss_set_cred_option() as they must for Heimdal.  Making this use case easy or pretty for the caller is not a goal.

gladiac:
Ok, the OID definition is wrong, I need to find the right place to add it ...

gladiac:
The implementation segfaults comparing the OID now. I need to find out what I still have wrong and if the symbols are all there ...

gladiac:
Tested and the current patchset works now!

greghudson:
Please make the following changes:

* Squash this down to one commit
* Use a flag field no_ci_flags instead of a field default_init_flags
* Get rid of no_ci_flags.c and put the implementation directly in gssapi_krb5.c as a static function named no_ci_flags().
* The definitions of GSS_KRB5_CRED_NO_CI_FLAGS_X_OID_LENGTH, GSS_KRB5_CRED_NO_CI_FLAGS_X_OID should move to gssapi_krb5.c and can drop the GSS_KRB5_CRED_ prefix.  The definition should have a comment like the ones for the OID declarations in gsssapi_krb5.h.

gladiac:
Greg, I've did all the changes you requested. I'm going on vacation. I might be able to take a look or do changes on Tuesday but I can't promise. Feel free to change the patch. I will be back June 15th.

greghudson:
I started editing this patch and adding tests; see the top two commits at https://github.com/greghudson/krb5/commits/ciflags.

However, I ran into a potentially major snag.  The change to spnego_mech.c was never discussed, is not at all innocuous (SPNEGO needs to be able to compute a MIC using the negotiated mechanism), and should have been called out with a big red flag.  I'm not sure how to make this change work with SPNEGO.  Heimdal seems not to ask for integrity when invoking the lower-level mechanisms, which seems like it can only work by accident.

gladiac:
We could remove the spnego hunk for now. However I saw there was a discussion about it on the Heimdal mailing list ...

greghudson:
krb5 part pushed to master as 7e6965ae33338216650384ca559d49e90312087a c1887eda950dfd84696f4f9bab9098f0bf1fd3c0.  Please create a new pull request for any SPNEGO changes related to this issue.

PR #284, opened 2015-06-06
In gss_release_oid() just return on NULL oid

simo5:
It is useless to call actuall functions if we have no OID, and can
lead to segfaults as some mechanisms do not check for null oid but
immediately dereference it.

See https://github.com/modauthgssapi/mod_auth_gssapi/issues/34
for an example.

### DIFF
--- a/src/lib/gssapi/mechglue/g_initialize.c
+++ b/src/lib/gssapi/mechglue/g_initialize.c
@@ -175,6 +175,9 @@ gss_OID *oid;
 	if (*minor_status != 0)
 		return (GSS_S_FAILURE);
 
+	if (oid == NULL)
+		return (GSS_S_COMPLETE);
+
 	k5_mutex_lock(&g_mechListLock);
 	aMech = g_mechList;
 	while (aMech != NULL) {


greghudson:
I knew we had a conversation about this before, but I guess we didn't make any changes.  I found the conversation here (and on the previous day):

http://colabti.org/irclogger/irclogger_log/krbdev?date=2015-03-06

The bug is really in gss_inquire_name(), but I guess this is okay as a workaround.

More generally, the mechglue should never be calling gss_release_oid(); it should use generic_gss_release_oid() for OIDs it allocates using generic_gss functions, and should leave alone OIDs returned by mechs.  I see some bad memory management around the name_type field of gss_union_name_t (usually it's an allocated OID, sometimes it's an OID returned by a mech's gss_display_name()), so fixing all of the calls requires some analysis.

simo5:
Yes this is a stop-gap measure to avoid crashing for now.
I agree we need to also do rework some the flow, but I think putting this one in for a start would avoid a painful experience to others.

greghudson:
I put a slight variation in https://github.com/greghudson/krb5/commits/release_oid (top commit).  Let me know if it seems okay.

greghudson:
Pushed to master as 4676e823e6ee9a5731872b31c5588c1b5932e0a3

PR #285, opened 2015-06-09
GSSAPI: Allow inquire_context on partial contexts

DirectXMan12:
RFC 2743 states that `gss_inquire_context` must always return
`flags`, `locally_initiated`, and `open` even if a context is not
yet fully established.  Additionally, a partially established
context may also return `mech_type`.

Previously, the krb5 mech raised an error for inquire_context on
partially completed contexts.  It now follows the rules layed
out in RFC 2743.

A new test case was added to test that inquire context works correctly
on both in-progress and established contexts (this involved moving a
function from t_namingexts.c into the common test code as well).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -307,6 +307,7 @@ testlog
 /src/tests/gssapi/t_s4u2proxy_krb5
 /src/tests/gssapi/t_saslname
 /src/tests/gssapi/t_spnego
+/src/tests/gssapi/t_inq_ctx
 
 /src/tests/hammer/kdc5_hammer
 
--- a/src/lib/gssapi/krb5/inq_context.c
+++ b/src/lib/gssapi/krb5/inq_context.c
@@ -104,55 +104,66 @@ krb5_gss_inquire_context(minor_status, context_handle, initiator_name,
         *acceptor_name = (gss_name_t) NULL;
 
     ctx = (krb5_gss_ctx_id_rec *) context_handle;
-
-    if (ctx->terminated || !ctx->established) {
-        *minor_status = KG_CTX_INCOMPLETE;
-        return(GSS_S_NO_CONTEXT);
-    }
-
-    initiator = NULL;
-    acceptor = NULL;
     context = ctx->k5_context;
 
-    if ((code = krb5_timeofday(context, &now))) {
-        *minor_status = code;
-        save_error_info(*minor_status, context);
-        return(GSS_S_FAILURE);
-    }
+    /* RFC 2743 states that a partially completed context must return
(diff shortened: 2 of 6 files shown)


greghudson:
You added t_inq_ctx to the dependencies for check-pytests in tests/gssapi, but I don't see any changes to actually run the program as part of that rule.

DirectXMan12:
> You added t_inq_ctx to the dependencies for check-pytests in tests/gssapi, but I don't see any changes to actually run the program as part of that rule.

It's in `t_gssapi.py`: https://github.com/krb5/krb5/pull/285/files#diff-a150ab7ac4d643a2572931bebdfdb84eR216

greghudson:
Oh, I missed it, my bad.

greghudson:
The license statement in t_inq_ctx.c mentions Red Hat at the top, but uses text from the old MIT license statement (including MIT's name) below.  I can't change the license on your code; please resubmit this with a correct Red Hat license.

While I'm asking for changes, here are some changes I would have just made myself:

* Rebase against master (which had a GSSAPI test program added).

* The list of GSSAPI test programs was alphabetized.  Maintain that order rather than adding to the end out of order.

* Indenting inq_context.c created some long lines.  Fix those by moving the assignments out of the conditional where it happens.

* t_inq_ctx.c has a line (I think just one) longer than 79 characters.

There are also a few changes which could simplify the test program:

* Can you just discard the partially established initiator context and use establish_contexts() from scratch to test inquiring established contexts?  That gets rid of the big finish_establish_contexts() helper, which is sort-of-not-quite duplicated from common.c:establish_contexts().

* start_init_context() is pretty trivial and can be inlined into main().

* Current C test harnesses either tend to be silent (if they do their own checks), or display output to be checked by the Python script, usually on successive runs with different parameters.  Since this program does its own checks and its output is ignored by t_gssapi.py, it should probably be silent.  That means you don't need to move display_name() into common.c, since you won't need it.

DirectXMan12:
Ok, I made all the changes except the inlining of `start_init_context` (if it's a big deal, I'll change that, too, but IMO, leaving it as a separate function improves readability).

tlyu:
Is there some reason that start_init_context() goes to the trouble of saving the major status, but main() doesn't bother to check it?

greghudson:
Here's what I plan to push: https://github.com/greghudson/krb5/commits/inqctx

It fixes the inconsistency noted by Tom, among other minor things.

DirectXMan12:
:+1:

greghudson:
Pushed to master as 9f095e1aab582e5a94c93d587f6d09e9d8f7479e

PR #288, opened 2015-06-23
Suppress ci flags for SPNEGO mech

gladiac:
Hi Greg,

here is the changeset you discussed with Simo some days ago. Let me know if this is what you wanted.

 -- andreas

### DIFF
--- a/src/lib/gssapi/spnego/gssapiP_spnego.h
+++ b/src/lib/gssapi/spnego/gssapiP_spnego.h
@@ -85,6 +85,7 @@ typedef struct {
 typedef struct {
 	gss_cred_id_t mcred;	/* mechglue union of obtainable creds */
 	gss_OID_set neg_mechs;	/* app-specified list of allowable mechs */
+	unsigned int no_ci_flags;	/* Do not set CONF or INTEG flags */
 } spnego_gss_cred_id_rec, *spnego_gss_cred_id_t;
 
 /* Structure for context handle */
--- a/src/lib/gssapi/spnego/spnego_mech.c
+++ b/src/lib/gssapi/spnego/spnego_mech.c
@@ -894,14 +894,20 @@ init_ctx_call_init(OM_uint32 *minor_status,
 {
 	OM_uint32 ret, tmpret, tmpmin;
 	gss_cred_id_t mcred;
+	OM_uint32 init_sec_ctx_req_flags = req_flags;
 
 	mcred = (spcred == NULL) ? GSS_C_NO_CREDENTIAL : spcred->mcred;
+
+	if (spcred == NULL || spcred->no_ci_flags == 0) {
+		init_sec_ctx_req_flags |= GSS_C_INTEG_FLAG;
+	}
+
 	ret = gss_init_sec_context(minor_status,
 				   mcred,
 				   &sc->ctx_handle,
 				   target_name,
 				   sc->internal_mech,
-				   (req_flags | GSS_C_INTEG_FLAG),
+				   init_sec_ctx_req_flags,
 				   time_req,
 				   GSS_C_NO_CHANNEL_BINDINGS,
 				   mechtok_in,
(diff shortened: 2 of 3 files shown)


greghudson:
That doesn't look right.  There should be no need to add a new OID (definitely not one which stomps on the Heimdal OID arc) and no need to change the mechglue union cred structure.  Just make SPNEGO honor the KRB5_GSS_CRED_NO_CI_FLAGS_X option by not requesting integrity from the underlying mech.

Also, this can't possibly work; what kind of testing did you do?  gssapi_krb5.c:no_ci_flags() is trying to get at the union cred just by casting the pointer gss_cred_id_t instead of krb5_gss_cred_id_t type.  C casts aren't magic; a simple comparison of krb5_no_ci_flags() and no_ci_flags() clearly indicates a type violation.

nicowilliams:
Doesn't the acceptor side of SPNEGO also need to pretend that if the mechanism is krb5 then GSS_C_INTEG_FLAG is set in ret_flags?

greghudson:
We do not currently know that it is necessary to do anything like that.

gladiac:
I was not sure what you meant in the first place. I've discussed it with Simo today. I haven't had the time to test it yet ...

nicowilliams:
There has never been an implementation of the Kerberos GSS mechanism that didn't include support for integrity protection (wraps and MICs).  There was one implementation (to my knowledge) that did not support confidentiality protection in some distributions.

I don't think SPNEGO should ever skip the MIC if it would want it just because the mechanism did not signal GSS_C_INTEG_FLAG in ret_flags in gss_accept_sec_context().

gladiac:
This patchset adds what you requested and I've added some tests too.

greghudson:
* "Add krb5_prefix...": is this still required or useful?

* "Add function to allocate...": the helper function should not have a "spnego_" prefix.  The commit message should have an explanatory paragraph.  We do not use Signed-off-by: or Pair-Programmed-With: in commit messages; please do not apply those conventions to krb5 contributions.

* "Implement GSS_KRB5_CRED_NO_CI_FLAGS_X...": there is too much unfiltered cut-and-paste here.  SPNEGO does not need a dispatch table while there is only one cred option to handle, and it definitely does not need g_OID_prefix_equal().  Adding the invalid-argument checks is out of scope for this commit, and isn't particularly necessary since the mechglue is doing all of the same checks.

* "Add tests...": there should be no need to cast &mechset_spnego.  "krb5" should generally not be capitalized.

gladiac:
I address these points tomorrow.

gladiac:
I've addressed all your points. The cast in the tests is needed to avoid the following warning:

```

```gcc -DHAVE_CONFIG_H -DUSE_AUTOCONF_H -I../../include -I../../../src/include -I../../../src/tests/gssapi/../../lib/gssapi/mechglue -I../../../src/tests/gssapi/../../lib/gssapi/krb5 -I..
/../../src/tests/gssapi/../../lib/gssapi/generic -I../../lib/gssapi/krb5 -I../../lib/gssapi/generic -DKRB5_DEPRECATED=1 -DKRB5_PRIVATE  -g -O2 -Wall -Wcast-align -Wshadow -Wmissing-pro
totypes -Wno-format-zero-length -Woverflow -Wstrict-overflow -Wmissing-format-attribute -Wmissing-prototypes -Wreturn-type -Wmissing-braces -Wparentheses -Wswitch -Wunused-function -Wu
nused-label -Wunused-variable -Wunused-value -Wunknown-pragmas -Wsign-compare -Werror=uninitialized -Werror=pointer-arith -Werror=declaration-after-statement -Werror-implicit-function-
declaration -pthread -c ../../../src/tests/gssapi/t_ciflags.c
../../../src/tests/gssapi/t_ciflags.c: In function ‘main’:
../../../src/tests/gssapi/t_ciflags.c:121:33: warning: passing argument 3 of ‘gss_set_cred_option’ discards ‘const’ qualifier from pointer target type
                                 GSS_KRB5_CRED_NO_CI_FLAGS_X,
                                 ^
In file included from ../../include/gssapi/gssapi_krb5.h:28:0,
                 from ../../../src/tests/gssapi/common.h:36,
                 from ../../../src/tests/gssapi/t_ciflags.c:37:
../../include/gssapi/gssapi_ext.h:172:11: note: expected ‘gss_OID’ but argument is of type ‘const struct gss_OID_desc * const’
 OM_uint32 KRB5_CALLCONV gss_set_cred_option
           ^

greghudson:
I said there was no need to cast &mechset_spnego.  Why are you quoting a warning related to GSS_KRB5_CRED_NO_CI_FLAGS_X?

greghudson:
The structure of the added tests is confused.  We do not need to get a SPNEGO initiator cred; the cred we already acquire in the test program will work as a SPNEGO cred since it was acquired with no desired mech list.

Don't make any more changes to the pull request at this point, as I am already in the process of editing them.

greghudson:
Also, it's not appropriate to break the tests in one commit and fix them in the next.  I will make the (very small) necessary changes to t_ciflags.c part of the second commit.

greghudson:
My edits are at https://github.com/greghudson/krb5/commits/spnego_ciflags

Please verify that these two commits are acceptable.

gladiac:
@greghudson Fine for me :)

greghudson:
Pushed to master as 98092b003a730fa490bf2fff1b50a2339c1b2811 cf39ed349976908626cad3e05e17788f8334bce9

gladiac:
Thanks you very much! With which version of krb5 will this be released?

greghudson:
1.14.  It's a new feature so it won't show up in a 1.13.x release.

PR #290, opened 2015-06-26
Deindent krb5_string_to_keysalts

tlyu:
Remove a level of indentation for the list-appending part of the
krb5_string_to_keysalts() loop body by consolidating the strtok_r()
calls into the controlling expreession of the loop.

### DIFF
--- a/src/lib/kadm5/str_conv.c
+++ b/src/lib/kadm5/str_conv.c
@@ -279,7 +279,7 @@ krb5_string_to_keysalts(const char *string, const char *tupleseps,
                         const char *ksaltseps, krb5_boolean dups,
                         krb5_key_salt_tuple **ksaltp, krb5_int32 *nksaltp)
 {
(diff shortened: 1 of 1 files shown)


greghudson:
Seems okay.  I think a for loop could also work to eliminate the need for p, but I don't have a strong preference.

    for (ksp = strtok_r(copy, tseps, &tlasts); ksp != NULL;
          ksp = strtok_r(NULL, tsteps, &tlasts)) {

tlyu:
I prefer to avoid splitting the clauses of a for-statement across multiple lines.

tlyu:
Pushed as f2f95f490e405c6e2889a0630915a9e729b6906c

tlyu:
Actually pushed as e110ce6ed19f5349e304e826e6b8066312c6c15c because I forgot to push it before closing this pull request and then other commits landed after.

PR #292, opened 2015-07-06
Unify KDB principal flag conversion functions

tlyu:
These changes unify the KDB prinicpal flag specifiers used by kadmin,
kdc.conf default_principal_flags, and kadm5.acl.  Each of those
interfaces will accept any of the historically accepted input forms of
any of those interfaces.  Additionally, accept specifiers in the form
that kadmin prints them in, as well as hexadecimal numbers.

Supersede krb5_string_to_flags() with krb5_flagspec_to_mask().  The
latter has a pseudo-ternary output, allowing different pointers for
flags to set versus flags to clear.  Additional functionality includes
parsing hexadecimal numbers for flag settings.

Remove krb5_input_flag_to_string(), which nothing in the tree used,
and probably hasn't ever worked properly due to long-standing gaps in
the flag number sequence.

Remove krb5_flags_to_string(), which nothing in the tree used.
Verbose flag output can be added back through another interface if
there is demand.

Add krb5_flagnum_to_string(), which produces a string representation
of a flag number.  Additional functionality includes output of
hexadecimal numbers for unknown flags.

Add krb5_flags_to_string(), which produces an array of strings
describing the flags, using the output from krb5_flagnum_to_string().


tlyu:
I'll probably squash the last two commits, if that makes sense.

greghudson:
Since I'm being a bit slow to finish reviewing this, I want to note that princflags.py uses ternary operators (A if B else C), which require Python 2.5.  There are several options, all of which I am okay with:

* Require a newer version of Python in configure.in to run the test suite.  I chose Python 2.4 five years ago; these days I think it's rare to see a version older than 2.6, and even requiring 2.7 would be reasonable.

* Rewrite those expressions to avoid ternary operators.

* Skip t_princflags.py (using skip_rest()) if the Python version is less than 2.5.  It's probably necessary to do the check before importing princflags.

greghudson:
Squashing the second and third commits should be reasonable for this change.  If the new facilities are going to be in a separate commit, that commit should also be separate from removing the old stuff.

We should not wind up with an unused compatibility stub for krb5_string_to_flags().  It is prototyped in adm_proto.h, which is not a public header file.  Also, the comment in server_acl.c for kadm5int_acl_parse_restrictions() should be updated with the new name.

In princflags.py in the first commit:

* The global variable _di is undocumented and overly terse.

* flagspec2mask() can use "if is in pflags" instead of try/accept.  Also, the end can be made more precise with a ternary expression.

* speclist2mask() doesn't really need the intermediate "a" variable.  More importantly, it is unused, and has poorly defined behavior if a flag name is repeated--the same flag bit can appear in both toset and toclear, with no memory of the order from the speclist.  I would suggest removing it.

* The two functions to print out C code should be made private.  A comment should be added to the top to indicate what happens if the script is run standalone.

In the second commit, there is a typo "prinicpal" in the commit message.  Also, the krb5_flagspec_to_mask() definition over-wraps its parameters, and the krb5_string_to_flags() definition has weird formatting after its ansification.  (As I noted above, the latter function should probably just be removed.)

In the third commit, kadmin_parse_princ_args() has an "if (retval) return -1; else ..." bit.  There is no need for the "else".

I didn't find any other issues.

tlyu:
I agree with most of your comments.

re speclist2mask() in princflags.py, I guess it doesn't quite map to the way krb5_flagspec_to_mask() actually gets called using the same pointer for toset and toclear.  There is a pre-existing related corner case in how kadm5.acl fails to distinguish between orderings when the same flag appears with both positive and negative senses.

Maybe both flagspec2mask() and speclist2mask() should go away due to lack of callers, but I found them useful when experimenting with things in an interactive interpreter.

tlyu:
I think these latest changes address the issues.  I could squash the comment change in k5test.py with the test suite commit.

greghudson:
There is a version test in configure.in which controls whether Python tests get run; it needs to be updated for the Python 2.5 requirement.  That change (along with the k5test.py comment change) should either be squashed with the first commit or should precede the first commit.

I think everything else has been addressed.

PR #293, opened 2015-07-09
Fix kdcpreauth counting bug

greghudson:
Commit be20a5f5cee8d6c4072d1b81712520dbf9f6eefd made
load_preauth_plugins() handle negative preauth type numbers.
get_plugin_vtables() also needs to handle negative preauth type
numbers, or it returns the wrong count.

### DIFF
--- a/src/kdc/kdc_preauth.c
+++ b/src/kdc/kdc_preauth.c
@@ -235,7 +235,7 @@ get_plugin_vtables(krb5_context context,
             n_tables++;
     }
     for (i = 0, n_systems = 0; i < n_tables; i++) {
-        for (count = 0; vtables[i].pa_type_list[count] > 0; count++);
+        for (count = 0; vtables[i].pa_type_list[count] != 0; count++);
         n_systems += count;
     }
     *vtables_out = vtables;


tlyu:
I'm somewhat concerned that there is a 32/64 bit differential in the symptoms.  I can't see an obvious cause from the code -- the negative number goes into an int32_t.

greghudson:
The environmental factor is actually whether PKINIT is built.

On my dev machine, PKINIT is built.  In the sometimes-failing test scenario, get_plugin_vtables() counts a maximum of 7 systems (where the correct value is 8), including four for PKINIT.  load_preauth_plugins() adds 5 for the static preauth systems and allocates 12+1 entries in the array.  When it goes to fill the array, PKINIT fails in vt->init() because the test scenario isn't configured for PKINIT, so we only wind up adding 9 systems to the array, and the termination element gets written to preauth_systems[9].

On the buildbot machine, PKINIT is not built.  get_plugin_vtables() counts 3 systems (the correct value is 4).  load_preauth_plugins() allocates 8+1 entries.  Every module succeeds in vt->init() so 9 systems are added to the array, and the termination element gets written to preauth_systems[9] which is out of bounds.

tlyu:
Thanks for the explanation.  That was considerably more complicated than I expected.

tlyu:
Looks OK to me.

PR #294, opened 2015-07-13
Fix memleak in libverto:vfree

tkuthan:
verto_set_allocator(resize, hierarchical) has an unducumented
requirement on allocator function resize, that resize(ptr, 0) is
equivalent to free(ptr).

Some implementations of the default allocator realloc don't meet this
requirement. realloc(ptr, 0) may return a unique pointer that can be
successfully passed to free. This new pointer never gets freed and the
memory overhead leaks.

This fix replaces realloc(ptr, 0) with free(ptr) in vresize and
documents the allocator requirement.

### DIFF
--- a/src/util/verto/verto.c
+++ b/src/util/verto/verto.c
@@ -132,6 +132,11 @@ vresize(void *mem, size_t size)
 {
     if (!resize_cb)
         resize_cb = &realloc;
+    if (size == 0 && resize_cb == &realloc) {
+        /* avoid memleak as realloc(X,0) can return a free-able pointer */
+        free(mem);
+        return NULL;
+    }
     return (*resize_cb)(mem, size);
 }
 
--- a/src/util/verto/verto.h
+++ b/src/util/verto/verto.h
@@ -195,7 +195,8 @@ verto_set_default(const char *impl, verto_ev_type reqtypes);
  * @see verto_add_idle()
  * @see verto_add_signal()
  * @see verto_add_child()
- * @param resize The allocator to use (behaves like realloc())
+ * @param resize The allocator to use (behaves like realloc(),
+ *        resize(ptr, 0) frees memory pointed by ptr.)
  * @param hierarchical Zero if the allocator is not hierarchical
  */
 int


greghudson:
src/util/verto is a bundled copy of https://fedorahosted.org/libverto/, so the issue should be filed there instead.

The requirement that realloc(ptr, 0) act like free(ptr) is "documented" in the sense that it's in the Linux man page, but is not part of C99 (or, I presume, POSIX), so the term "non-standardized" might be more precise.

greghudson:
Actually, this realloc() requirement appears to be standardized in the Single Unix specification:

http://pubs.opengroup.org/onlinepubs/007908799/xsh/realloc.html

tlyu:
A more recent reference is http://pubs.opengroup.org/onlinepubs/9699919799/functions/realloc.html which seems to revise the POSIX behavior to match the C standard, in that if realloc(ptr, 0) returns a null pointer, it doesn't free ptr.

greghudson:
Interesting.  So I agree that libverto should change, but it remains the case that this should be filed upstream.

tkuthan:
Greg, Tom, thank you for your comments.
I will report the bug to (further) upstream. I will reference this pull request (because our legal approval for contributing bug fixes upstream is for MIT Kerberos).

tkuthan:
I have filed https://fedorahosted.org/libverto/ticket/13

greghudson:
Some additional notes:

* At least on Ubuntu 14.04, realloc(NULL, 0) returns non-null--that is, it behaves like malloc(0) rather than free(NULL).  This is consistent with the man page.

* The version of libev we currently have in the tree assumes that realloc(x, 0) works like free(x) under glibc, and can leak memory when freeing null pointers.  libev 4.22 has given up on this assumption and checks the length in ev_realloc_emul() on all platforms.

frozencemetery:
@tkuthan would you mind submitting your patch as a PR against [libverto](https://github.com/latchset/libverto) to fix latchset/libverto#14?  Thanks!

frozencemetery:
@tkuthan would you mind submitting your patch as a PR against [libverto](https://github.com/latchset/libverto) to fix latchset/libverto#14?  Thanks!

tkuthan:
@frozencemetery Sure, thank you for pointing out the repo. At this moment I'm traveling; I believe I could submit it on Monday. Sorry for the delay.

tkuthan:
@frozencemetery Done.
https://github.com/latchset/libverto/pull/18

frozencemetery:
Thanks @tkuthan!  Fix merged upstream.

frozencemetery:
(I believe this fix was merged into master with e690f1b5691edf87b6c8a92eab95db5cfaee01d3).

PR #295, opened 2015-07-13
Fix princflags memory management

tlyu:
Fix some out of memory error cases (found by Coverity) that could
cause multiple frees or freeing of invalid pointers.  In
krb5_flagnum_to_string(), don't assume that asprintf() stores a null
pointer on failure (it does in BSD but not in glibc).  In
krb5_flags_to_strings(), free the correct pointer in the cleanup loop
in on error.

ticket: 8215

### DIFF
--- a/src/lib/kadm5/str_conv.c
+++ b/src/lib/kadm5/str_conv.c
@@ -196,7 +196,8 @@ krb5_flagspec_to_mask(const char *spec, krb5_flags *toset, krb5_flags *toclear)
 }
 
 /*
- * Copy the flag name of flagnum to outstr.
+ * Copy the flag name of flagnum to outstr.  On error, outstr points to a null
+ * pointer.
  */
 krb5_error_code
 krb5_flagnum_to_string(int flagnum, char **outstr)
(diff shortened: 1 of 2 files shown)


greghudson:
The sense of the asprintf check appears to be backwards.  (Is there no test coverage for this?)  Most of our asprintf() invocations check for the result being < 0 to indicate an error, though that might be needlessly imprecise.

A number of people seem to really dislike having braces around one arm of an if-else and not the other, so it might be good to put braces around the else arm of the krb5_flagnum_to_string() test.

tlyu:
You're right about the asprintf check; thanks.  There is no automated test coverage for that case; I will add one.  I was manually testing the hexadecimal output stuff, and apparently forgot to do so after making the change that had the wrong sense of the asprintf return value check.

tlyu:
Updated according to your suggestions, including an additional test case.

greghudson:
Looks okay to me.  (The term "lamp test" isn't familiar to me and googling it isn't enlightening, but it doesn't really make the test script harder to understand.)

tlyu:
Clarified the comment somewhat, and added a call to delprinc to make the test a little cleaner.

A "lamp test" is a process, typically as part of a self-test (sometimes manually activated), that simultaneously activates every light or display element on a device.  It's somewhat interesting that Google doesn't seem to readily find that definition.  It was probably more useful when most indicator lamps were incandescent and thus likely to burn out while in service.  It seems that "lamp test mode" (with quotes) finds more useful references.

greghudson:
I figured it was something like that.  Anyway, it looks fine.

tlyu:
Pushed as commit dd5f948614b6662fc40dc8de3f567078cfe6295e

PR #296, opened 2015-07-14
Limit use of deprecated krb5 mech OIDs

greghudson:
Filter out mechs with the GSS_C_MA_DEPRECATED attribute from the set of mechanisms obtained by SPNEGO, and from the set used when gss_acquire_cred() is called with no desired_mechs attribute.
    
SPNEGO acceptors will still accept the old and wrong krb5 OIDs, but SPNEGO initiators will not offer them.  According to [MS-SPNG], only Windows 2000 does not recognize the standard krb5 OID, and it is client-only.
    
In gss-client.c, use the standard krb5 OID for the -krb5 option, as acceptors who call gss_acquire_cred()  with no desired_mechs to create an acceptor cred will no longer accept the old or wrong krb5 OIDs.

[There is also a first commit to make t_spnego.c work under the new regime.  Unfortunately it requires even more knowledge of SPNEGO encoding details in the test program.]

### DIFF
--- a/src/appl/gss-sample/gss-client.c
+++ b/src/appl/gss-sample/gss-client.c
@@ -816,7 +816,7 @@ main(argc, argv)
         } else if (strcmp(*argv, "-spnego") == 0) {
             spnego = 1;
         } else if (strcmp(*argv, "-krb5") == 0) {
-            mechanism = "{ 1 3 5 1 5 2 }";
+            mechanism = "{ 1 2 840 113554 1 2 2 }";
 #ifdef _WIN32
         } else if (strcmp(*argv, "-threads") == 0) {
             argc--;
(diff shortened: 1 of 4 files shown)


tlyu:
I'm not sure I see how the wrong OID is getting excluded.

greghudson:
In gssapi_krb5.c, there is (already) code to set GSS_C_MA_DEPRECATED for any krb5 oid which isn't gss_mech_iakerb or gss_mech_krb5.  These changes are just excluding deprecated mechs from SPNEGO and from the default gss_acquire_cred mech list.

tlyu:
Oh, sorry.  I somehow misread one of those conditionals as setting GSS_C_MA_DEPRECATED only for the old OID.  Now it looks OK to me.

PR #297, opened 2015-07-16
Conditionalize iprop stderr output in kadmind

greghudson:
kadmind should be quiet in nofork mode after it prints the
"starting..." sentinel line, or it can fill the pipe buffer when run
from k5test.py.  Since there is currently no run-time debuf flag,
conditionalize the DPRINT macro in ipropd_svc.c on DEBUG at compile
time.


tlyu:
Looks OK to me.

PR #298, opened 2015-07-17
Fix compatibility with pre-1.11 iprop dumps

greghudson:
Ticket #7223 added new policy fields and a new dump format version to
marshal them, but did not add a new iprop dump format.  As a result,
slave KDCs running 1.11 or later cannot receive full resyncs from
master KDCs running 1.10 or earlier.

Retroactively add support for pre-1.11 policy entries by making
process_r1_11_policy() read the first ten fields, check whether the
next whitespace character is a newline, and then read the rest.

### DIFF
--- a/src/kadmin/dbutil/dump.c
+++ b/src/kadmin/dbutil/dump.c
@@ -958,41 +958,61 @@ process_r1_11_policy(krb5_context context, const char *fname, FILE *filep,
     char namebuf[1024];
     char keysaltbuf[KRB5_KDB_MAX_ALLOWED_KS_LEN + 1];
     unsigned int refcnt;
-    int nread, ret = 0;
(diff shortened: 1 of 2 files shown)


tlyu:
We should credit John Devitofranceschi in the commit message or RT ticket for reporting this.  Did you want to add documentation about the upgrade issue in this pull request, or another one later?

greghudson:
I edited the commit message to credit John.

I don't think we currently have documentation on upgrades, and I'm not sure I presently have time to write it.  I can file a new ticket for the lack of upgrade documentation, with a reference to this issue so that we remember to document that this bug affects upgrades to affected versions.

tlyu:
Thanks; looks good.  Opening a new ticket for the upgrade documentation seems like a reasonable plan.

PR #299, opened 2015-07-23
Simplify test preauth module use

greghudson:
I made an outdated assumption when writing and using the test preauth module, leading to a small amount of unnecessary complexity.  This commit fixes that.


tlyu:
Looks OK to me.

greghudson:
Pushed to master as 8b6d2227be777c4ef62bec796e864020db03031b

PR #300, opened 2015-07-26
Fix uncommon leak in krb5_init_creds_step()

greghudson:
Release any previous value of ctx->err_padata before setting it in
init_creds_step_reply().  It could have a prior value after a realm
referral or retriable error.


tlyu:
Looks OK to me.

greghudson:
Pushed to master as 997eb174f5fd81747ad0ecb671f00c25951931b1

PR #301, opened 2015-07-27
KDC_ERR_PREAUTH_EXPIRED support

greghudson:
init_creds_step_reply() is beyond the acceptable threshold of function length, so I attempted to reduce some of its tech debt before adding more to it.  I have some ideas for restructuring the initial creds code, but it would be risky without comprehensive automated testing (which is difficult with current infrastructure).

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -391,6 +391,7 @@ typedef unsigned char   u_char;
                                                       not find a KDC */
 #define KRB_AP_ERR_IAKERB_KDC_NO_RESPONSE       86 /* The KDC did not respond
                                                       to the IAKERB proxy */
+#define KDC_ERR_PREAUTH_EXPIRED                 90 /* RFC 6113 */
 #define KDC_ERR_MORE_PREAUTH_DATA_REQUIRED      91 /* RFC 6113 */
 #define KRB_ERR_MAX 127 /* err table base max offset for protocol err codes */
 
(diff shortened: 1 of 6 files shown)


tlyu:
Seems mostly OK.  Last sentence of the first commit message seems to be missing a word in "since we don't that the KDC".  Also I'm not sure that the other changes in the first commit make the large removed block comment for negotiation_requests_restart completely redundant, but I'm willing to be convinced.

greghudson:
Commit message fixed.  I also added a parenthetical to the comment in the upgrade-to-FAST restart case to note that FAST negotiation normally prevents that restart.

The big comment removed in the first commit includes:

1. A brief summary of how FAST negotiation prevents the need for an upgrade-to-FAST restart.
2. A description of what it means to restart for a FAST upgrade.
3. A more verbose description of why we need to restart without informational padata against pre-1.7 MIT KDCs.
4. A summary of the state variables involved in these particular restarts.

For (1), an explanatory comment for fast.c:krb5int_fast_verify_nego() would probably be sufficient, and there should be one anyway.  I will add one as part of the first commit.

For (2), I think the explanatory comment for restart_init_creds_loop() (already present) should suffice.

For (3), I think the short description is adequate.  We are missing the tidbit that pre-1.7 KDCs don't include method-data in PREAUTH_FAILED errors so there's no need to look for it, but since we currently don't have that facility (see http://k5wiki.kerberos.org/wiki/Projects/Initial_creds_improvements) I don't think the discussion is needed yet.

For (4), I think the state variables are pretty self-explanatory.  In particular, the adjacency of the upgrade-to-FAST and downgrade-to-no-informational-padata cases makes it pretty clear what the restarted field is doing.

tlyu:
Thanks; looks good now.

greghudson:
Pushed to master as 9914d38658e5612db5b2847892b5ddce2b73c344 d3e0af0774dd100f00fbc8895b99355d82d86bf1

PR #305, opened 2015-08-04
Make cross-realm S4U2Self work

greghudson:
When sending a S4U2Self query to a foreign realm, send an enterprise
server principal so that the foreign KDC can identify the home realm
of the server principal.

To make this work, adjust the memory management of
krb5_get_self_cred_from_kdc().  s4u_creds is now a shallow copy of
in_creds which owns no memory.  A new variable eprinc owns the
enterprise form of the server principal, constructed using a new
helper function convert_to_enterprise().  Since we have to set the
server realm for KDC-REQ encoding to work, a new temporary variable
sprinc holds a shallow copy of *eprinc with the realm pointing to the
realm we are currently querying.

Based on a patch by Sumit Bose.

### DIFF
--- a/src/lib/krb5/krb/s4u_creds.c
+++ b/src/lib/krb5/krb/s4u_creds.c
@@ -440,6 +440,24 @@ cleanup:
     return code;
 }
 
+/* Unparse princ and re-parse it as an enterprise principal. */
+static krb5_error_code
+convert_to_enterprise(krb5_context context, krb5_principal princ,
+                      krb5_principal *eprinc_out)
+{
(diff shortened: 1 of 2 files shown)


tlyu:
Looks reasonable.  The added memory aliasing isn't great, but there is precedent in this very file within krb5_get_credentials_for_proxy().  The body of the loop in krb5_get_self_cred_from_kdc() is about 90 lines long, with variables declared at the top, and could therefore use refactoring.  However, that was a preexisting condition and can probably be done separately.

PR #307, opened 2015-08-06
Expose krb5_c_prfplus() and krb5_c_derive_prfplus()

npmccallum:
This commit permits the external use of the RFC 6113 PRF+ function.
It also creates a function used to derive an input key with a string
to make a new output key.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -643,6 +643,40 @@ krb5_error_code KRB5_CALLCONV
 krb5_c_prf_length(krb5_context context, krb5_enctype enctype, size_t *len);
 
 /**
+ * Generate pseudo-random bytes using RFC 6113 PRF+.
+ *
+ * The caller must allocate the output buffer. The length of the pseudo-random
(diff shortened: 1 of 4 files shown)


kaduk:
I think that the style prefers 'cleanup' as the exit label to 'egress'; I'm not sure why it was changed.

npmccallum:
Yeah, that was my mistake. It was an accident of the history of the patch. I can fix it tomorrow.

npmccallum:
Fixed.

greghudson:
Adjusted version: https://github.com/greghudson/krb5/commits/prfplus

The changes are mostly inoccuous, but there's one thing I want to red-flag: your versions of krb5_c_derive_prfplus() and krb5_c_fx_cf2_simple() can leave a dangling pointer in *out on error.  We try hard to avoid this as it can lead to high-impact bugs.  See http://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling

greghudson:
Pushed to master as 35f288092f0df7f4aca92e1f51db3611a3b32ada

PR #308, opened 2015-08-07
Do not allow retries on SOCK_STREAM sockets in libkrad

npmccallum:
Before this patch, libkrad would follow the same exact logic for
all socket types when the retries parameter was non-zero. This
meant that when connecting with SOCK_STREAM, multiple requests
were sent in case of packet drops, which, of course, cannot
happen for SOCK_STREAM.

Instead, we will just disable retries for SOCK_STREAM sockets.


greghudson:
Pushed to master with minor changes as 25e0656fdf9862faf9aa91288023776e9a47caad

PR #311, opened 2015-08-11
Fix new doxygen parameter lists

greghudson:
Use [out] instead of [in] for the output parameters of
krb5_c_prfplus() and krb5_c_derive_prfplus().  Also use "out" instead
of "output" for krb5_c_derive_prfplus() to match the parameter name in
the definition.


tlyu:
Looks OK to me.

PR #312, opened 2015-08-12
Vs2015 fixes - second batch

tsondergaard:
The first three commits 67d1d01, 27051ce and 4488321 in this pull request I consider ready to merge and uncontroversial. The fourth commit 98d290b "Remove windows/gina" is a suggestion - GINA modules are no longer supported since Windows Vista and the code doesn't seem to be complete anyway.

The last commit 7ed67d9 is not a fix at all, it just works around the problem that the FILE is an opaque struct in the new standard library shipped with VS2015, so it doesn't have a visible _flag member anymore. I am wondering if the fp->_flag actually matters at all. The fp, a FILE *, is not touched anymore after 

*stdout = *fp;
fp->_flag = 0;

at least not by user code, but the c library may have some at-exit handling for open FILEs.

Note that with all five commits I can build krb5 with VS2015 without errors.


tsondergaard:
After googling a bit, I think the redirection of stdout should handle at the file descriptor level, leaving the stdio.h FILE level mostly oblivious to the change. It can be done with dup2. See the example for dup2 on msdn:

https://msdn.microsoft.com/en-us/library/aa246796%28v=vs.60%29.aspx

kaduk:
I am a little nervous about some of these changes, given that we know of people who are still building with VS 2008 and possibly even VS 2005.  I can't spin up the VM farm right now to check (e.g.) what VISUALSTUDIOVERSION those leave in the environment, though.

tsondergaard:
I assume it is commit 67d1d01 that bothers you. If you look at the old behaviour it uses either MFC100(D) or MFC110(D) - Visual Studio 2005 ships with MFC80(D) and Visual Studio 2008 with MFC90(D). I can improve the patch by falling back to MFC100(D).dll as it did before unless VisualStudioVersion is set - I don't believe that VisualStudioVersion is set by older versions of Visual Studio. That way the patch will leave behaviour unchanged for those older Visual Studio versions.

tsondergaard:
f4ea555 should now behave exactly as before for older versions of Visual Studio.

kaduk:
Thanks for the updates in f4ea5551; that's a nice simplification (though it would be nice to double-check that the old versions don't set VISUALSTUDIOVERSION).  Unfortunately, the added code in 2d1a97cf adds a version comparison against "14.0" that will presumably need updating for a 15.0 and later.  I don't know if there's some nmake cleverness we can do that will just do the right thing for later versions (assuming they retain the new CRT behavior).

tsondergaard:
I've updated the check against "14.0" - there is no operator for comparing version numbers in nmake, but while not bullet proof the check should work as long as they don't remove the .0 in future versions of VisualStudio.

I have also removed the Work-in-progress commit, which I hope means that you will accept this pull request

tlyu:
I haven't tested yet, but as I recall, the construct $(VISUALSTUDIOVERSION:.=) wouldn't work in POSIX Make to delete the decimal point in strings such as "14.0", because POSIX specifies that construct to work only on suffixes.  nmake might allow it, but it might be safer to use $(VISUALSTUDIOVERSION:.0=0).

tsondergaard:
According to the reference documentation it doesn't only work with suffixes as it states that "each occurrence of string1 in its definition string is replaced by string2" and there can be only one suffix in a string:

https://msdn.microsoft.com/en-us/library/bsd42ets.aspx?f=255&MSPPError=-2147217396

Also we cannot be sure that version numbers will end with .0 - Visual Studio .net 2003 had version number 7.1.

tlyu:
It does seem that nmake exhibits that non-POSIX variant behavior.  What sets VISUALSTUDIOVERSION?  I don't see it being set by the SDK setenv.cmd that I have (SDK 7.1, VS 10.0).

kaduk:
It sounded like vcvars.bat would do so, but I haven't gotten to test yet.

tsondergaard:
The following is observed from the Visual Studio installations on my own computer and the CI build systems I have access to: vcvarsall.bat in Visual Studio 2012 (11.0), Visual Studio 2013 (12.0) and Visual Studio 2015 (14.0) sets it. The SetEnv.cmd from Windows SDK 7.1 does not set it. I don't know if Visual Studio 2010 set it - my guess would be no, since SDK 7.1 doesn't.

tlyu:
Looks like vcvarsall.bat on Visual Studio 2010 doesn't set it either.

tsondergaard:
Note that regardless if vs2010 set VisualStudioVersion or not, the patches behave correctly for vs2010

tsondergaard:
Let me know if there is anything I can do to help land these small changes.

greghudson:
Pushed to master as f9e79bfdfaa6eb7c2059fb496b296b4316ca6352 and preceding.

PR #313, opened 2015-08-12
Indicate false when name is not MN at gss_inquire_name

frenche:
Currently name_is_MN is left uninitialized which seem like a typo.

Thanks,
Isaac B.


greghudson:
Pushed to master as a3f3f4069858ea795b732ec1d96fae20d5fafe24

PR #314, opened 2015-08-16
Secure cookie suport

greghudson:
These commits add secure cookie support, with tests and documentation.

This work is ready for review, but there is one small issue to address.  We're using key usage value 513 when encrypting with the cookie key; that's in the range 512-1023 which is reserved for internal use within a Kerberos implementation.  Right now the value is defined in fast_util.c; that doesn't seem very conducive to avoiding conflicts within our implementation, should we ever need another key usage value with similar characteristics.  I am wondering if we should instead define it in krb5.hin, or perhaps in k5-int.h.


tlyu:
Defining the keyusage in either krb5.hin or k5-int.h is fine.  I have a slight preference for krb5.hin, so it's easier to find.

greghudson:
Amended to define the key usage constant in krb5.hin.

tlyu:
Thanks; looks good now.

PR #315, opened 2015-08-17
Resolve credentials if time_rec is requested

simo5:
The code normally tries to defer credential acquisition to a later
time. However if the application requests the lifetime the code
needs to resolve the credential and return the actual expiration
time. Returning 0 would cause the application to think credentials
are expired.

### DIFF
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -825,8 +825,14 @@ acquire_cred_context(krb5_context context, OM_uint32 *minor_status,
         if (code != 0)
             goto krb_error_out;
 
-        if (time_rec)
+        if (time_rec) {
+            ret = kg_cred_resolve(minor_status, context, (gss_cred_id_t)cred,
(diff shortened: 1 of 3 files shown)


greghudson:
What testing did you do on this?  I get a compiler warning when I build it, and the test suite fails in a number of places because the mechglue always passes a time_rec pointer to the mech.

simo5:
I did a quick make check I saw no compiler warning, but I may have missed it in the noise.
I'll recheck, hold on

simo5:
This ios a more comprehensive fix, it passes make check and I've seen no warnings.
I haven't tested it agains SAP's gsstest though, should be able to do that later today.

simo5:
Ok, with the latest push I have a patch that properly passes make check, as well as it makes gsstest from SAP happy.

simo5:
gsstest here for the curious: ftp://ftp.sap.com/pub/icc/bc-snc40/

greghudson:
Pushed to master as 50f426ac17a81ff5b7c212c24645b9874ea911f0

PR #317, opened 2015-08-21
Keep valgrind happy after time_rec change

simo5:
Initialize the time variables so valgrind does not complain.
all these values are ignored if time_rec is NULL, so not having those
variables initialized is harmless, but it is annoying to get noise in
the valgrind output.


greghudson:
Pushed to master as 042e9fc95a662acb54dc9168749c6725f17ae34a

PR #319, opened 2015-08-22
Correct SPNEGO NegHints hintName value and add test case

greghudson:
One of the features added in 1.7 was support for NegTokenInit2 messages contaning NegHints values.  These tokens are sent by the acceptor in response to a completely empty token from the initiator.  While investigating the SPNEGO code for PR #318, I realized that we have no test coverage for NegHints, and that it's a fair amount of code in several different places (in the mechglue to route empty tokens in SPNEGO, in the SPNEGO code which calls acc_ctx_hint(), and in make_spnego_tokenInit_msg()).  I also noticed that our code is going to somewhat heroic efforts to make up a name to put inside NegHints, when the MSDN spec and Heimdal have moved on to just inserting a dummy string there.  I wanted to fix those things sooner rather than later.

The first two commits on the branch are from PR #318; only the last commit is relevant.  It changes the hintName value to be the dummy string, without making any invasive changes to function contracts (although there are plenty of cleanup opportunities there), and adds a test case.

The test case will fail if any other mechanisms are configured in /etc/gss/mech.  I haven't decided yet whether that's worth worrying about.  Other GSS tests could conceivably fail in that case as well, although I can't think of any off the top of my head.  Simo says he rarely runs the test suite on a system with NTLMSSP configured.  If we want to make the test case more resilient, we could look for specific substrings in the token, or perhaps just the [3] SEQUENCE [0] GeneralString "not_defined_in_RFC4178@please_ignore" at the end.

### DIFF
--- a/src/lib/gssapi/spnego/spnego_mech.c
+++ b/src/lib/gssapi/spnego/spnego_mech.c
@@ -191,7 +191,7 @@ static const gss_OID_set_desc spnego_oidsets[] = {
 };
 const gss_OID_set_desc * const gss_mech_set_spnego = spnego_oidsets+0;
 
-static int make_NegHints(OM_uint32 *, spnego_gss_cred_id_t, gss_buffer_t *);
+static int make_NegHints(OM_uint32 *, gss_buffer_t *);
 static int put_neg_hints(unsigned char **, gss_buffer_t, unsigned int);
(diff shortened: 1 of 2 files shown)


tlyu:
This looks OK to me.

PR #320, opened 2015-08-27
gss_display_name_ext: check for NULL name_type

DirectXMan12:
It is possible for the input name's name_type to be GSS_C_NO_OID.
`g_OID_equal` does not account for GSS_C_NO_OID, so we have to
manually check before use to prevent null pointer dereferences.


greghudson:
Shouldn't that second hunk use an or, not an and?

DirectXMan12:
Yep, good catch.  Never git while tired :-P

greghudson:
Pushed to master as 3fdf09ac9a36581b47f40c9d177e463cc12687ff

PR #321, opened 2015-09-02
Fix krb5_rd_req() memory leak

greghudson:
In release 1.13, commit eba8c4909ec7ba0d7054d5d1b1061319e9970cc7 (ticket #7232) introduced a memory leak when skipping keytab entries which do not match the application-provided server specification.  Fix it by freeing the keytab entry before continuing the loop on a failure to match.

The second commit is a less important fix to a test program, noticed while verifying the first commit.


greghudson:
Pushed to master as 3aa8506ee9e1f564e3f396eed5ac5616d7c54b34 5b156650dca05ee5d41813cd757706520b427a18

PR #322, opened 2015-09-02
Fix error handling in gss_export_sec_context()

greghudson:
In the mechglue gss_export_sec_context(), make sure to delete the
union context if the underlying mech context has been deleted.  This
can happen if the mech's gss_export_sec_context() returns a failure
and deletes the context (not a behavior exhibited by any of our
in-tree mechanisms, but an allowed behavior for other mechs), or if we
fail to allocate space for the wrapped token.


greghudson:
Pushed to master as 4f35b27a9ee38ca0b557ce8e6d059924a63d4eff

PR #323, opened 2015-09-03
client referral: return correct client principal in case of a wrong realm

abbra:
prepare_error_as_req() gets protocol error code but compares it with
table code and misses to add proper client principal as returned by the
DAL driver.

Rework the code to avoid errcode comparison within prepare_error_as_req().

### DIFF
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -181,6 +181,7 @@ finish_process_as_req(struct as_req_state *state, krb5_error_code errcode)
     void *oldarg;
     kdc_realm_t *kdc_active_realm = state->active_realm;
     krb5_audit_state *au_state = state->au_state;
+    krb5_principal client_princ = NULL;
 
     assert(state);
     oldrespond = state->respond;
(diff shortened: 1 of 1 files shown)


greghudson:
Please review the top two commits at https://github.com/greghudson/krb5/commits/client_referral as a counterproposal.

I'm no longer sure that I like the idea of changing the code to pass an unconditional client princ to prepare_error_as(); it adds too much complexity to finish_as_req() which is in dire need of refactoring (a perennial problem for the KDC code).  It might be better to push more responsibility down into prepare_error_as(), perhaps passing it the state pointer instead of a half-dozen fields of it.  I will defer that work to a later time.

abbra:
I'm fine with your patch If you want to do refactoring of prepare_error_as() later.

tlyu:
@greghudson your patches look OK to me.

greghudson:
Pushed to master as 4e036ef4127a9b09d1a567472da1df24c55cdb89 1e4a0394d0085e48d732ab318f9cbe08e9359587

PR #324, opened 2015-09-07
Improve PKINIT OpenSSL error reporting

greghudson:
When a non-trivial OpenSSL function fails during PKINIT processing,
try to ensure that the error message includes an indication of the
what PKINIT was doing and the reason for the first queued OpenSSL
error, and flush all queued OpenSSL errors to the trace log.  For
certificate verification failures, also include the higher-level error
from the cert store.  Add new helper functions oerr() and oerr_cert()
to minimize the amount of code needed to handle each error.

(Inspired by http://mailman.mit.edu/pipermail/kerberos/2015-September/020946.html and http://mailman.mit.edu/pipermail/kerberos/2015-September/020951.html)

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -409,6 +409,71 @@ unsigned char pkinit_4096_dhprime[4096/8] = {
 
 MAKE_INIT_FUNCTION(pkinit_openssl_init);
 
+static krb5_error_code oerr(krb5_context context, krb5_error_code code,
+                            const char *fmt, ...)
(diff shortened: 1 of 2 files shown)


tlyu:
This looks OK to me.

tlyu:
We should check how much (if any) of this overlaps with RT 7149.

greghudson:
This doesn't solve the specific case mentioned in RT 7149; if the anchors file contains garbage, you'll get an error like "Failed to verify own certificate (depth 0): unable to get local issuer certificate while getting initial credentials," which may provide clues but doesn't directly indicate the problem.

It does make progress on the general issue described in RT 7149, but more work is needed before we can say we're done with that.

PR #325, opened 2015-09-09
Add tabular dump capability to kdb5_util

tlyu:
This new kdb5_util tabdump command provides a reporting-friendly
tabular dump format for the KDC database.  This format is also
suitable for importing into relational databases for queries.  Output
is in tab-separated or CSV format.  The user can select an output
table with a fixed number of columns.

Currently, only provide tables for a subset of available principal
data.  This includes making visible some data that is hidden in
hexadecimal strings in the tl_data of the ordinary dump format.

### DIFF
--- a/doc/admin/admin_commands/kdb5_util.rst
+++ b/doc/admin/admin_commands/kdb5_util.rst
@@ -330,6 +330,158 @@ principal processed to be listed, with an indication as to whether it
 needed updating or not.  The **-n** option performs a dry run, only
 showing the actions which would have been taken.
 
+tabdump
+~~~~~~~
+
+    **tabdump** [**-H**] [**-c**] [**-e**] [**-n**] [**-o** *outfile*]
+    *dumptype*
+
+Dump selected fields of the database in a tabular format suitable for
+reporting (e.g., using traditional Unix text processing tools) or
+importing into relational databases.  The data format is tab-separated
+(default), or optionally comma-separated (CSV), with a fixed number of
+columns.  The output begins with a header line containing field names,
+unless suppression is requested using the **-H** option.
+
+The *dumptype* parameter specifies the name of an output table (see
+below).
+
+Options:
+
+**-H**
+    suppress writing the field names in a header line
+
+**-c**
+    use comma separated values (CSV) format, with minimal quoting,
+    instead of the default tab-separated (unquoted, unescaped) format
+
+**-e**
+    write empty hexadecimal string fields as empty fields instead of
+    as "-1".
+
+**-n**
+    produce numeric output for fields that normally have symbolic
+    output, such as enctypes and flag names.  Also requests output of
+    time stamps as decimal POSIX time_t values.
+
+**-o** *outfile*
+    write the dump to the specified output file instead of to standard
+    output
+
+Dump types:
+
+**keydata**
+    principal encryption key information, including actual key data
+    (which is still encrypted in the master key)
+
+    **name**
+        principal name
+    **keyindex**
+        index of this key in the principal's key list
+    **kvno**
+        key version number
+    **enctype**
+        encryption type
+    **key**
+        key data as a hexadecimal string
+    **salttype**
+        salt type
+    **salt**
+        salt data as a hexadecimal string
+
+**keyinfo**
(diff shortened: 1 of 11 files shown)


greghudson:
(First commit)

t_tdumputil.py could use a helper function for comparing out/expected, or just less informative messages on failure.  Also, success() is only intended to be called once per test script.

qquote() does not check for malloc() failure.

I would get rid of alwaysquote if it's not currently used.

The comment before rechandle_common() is missing a '*' on the second line.

writequoted() doesn't behave well if quotechar is '\0'.  Its callers don't call it if quotechar is '\0', but I think it should either assert that quotechar is set, or call vfprintf if quotchar is '\0' (simplifying the callers).  Probably the former, since the latter behavior doesn't really agree with the name.

In writequoted(), the else arm of the if-else should be braced since the if arm is.  We also generally put a blank line before cleanup labels.

The external functions, and probably the others as well, should have brief explanatory comments before them.

(Second commit)

The third paragraph of the commit message is missing a noun in the first sentence.

Line 60 of tabdump.c is too long.

The declarations of keydata_field and friends are inconsistent about "char *const" vs. "char * const".  (Either is justifiable since we don't have that nailed down in the coding practices.)

NTDTYPES could be a "const size_t n_tdtypes" to minimize preprocessor use.

The functions in tabdump.c should have brief explanatory comments.

(Third commit)

t_tabdump.py should only call success() once.  (I have been meaning to add some kind of progress indication to k5test.py, to make it easier to tell what part of a script we're in when failures happen.  But success() isn't it, at least currently.)

I didn't find anything to comment on in the fourth commit.

tlyu:
Thanks.  I've addressed most of these.  I think a close-brace serves the same visual role as a blank line before the cleanup label.  I thought NDTYPES as a preprocessor macro fits with the majority of our existing practice, but I haven't done a thorough survey.

I also added a rechandle_free() function and improved the documentation.

greghudson:
A few notes on the descriptions, and also one thing I didn't pick up on before.

(First commit)

k5-platform.h ensures SIZE_MAX; you can use that in qquote() intead of ~(size_t)0.  Alternatively, this function could use k5buf:

    struct k5buf buf;
    const char *sp;

    k5_buf_init_dynamic(&buf);
    for (sp = s; *sp != '\0'; sp++) {
        k5_buf_add_len(&buf, sp, 1);
        if (*sp == fl->quotechar)
            k5_buf_add_len(&buf, sp, 1);
    }
    return buf.data;

gcc gives a warning about writequoted(), which is annoying since I don't think there's much preventative value in giving __format__ attributes to functions taking va_lists.

The description for writeheader() should say "is null" rather than "is non-null".

(Second commit)

Several of the function descriptions (e.g. keydata() and keyinfo()) are noun phrases which basically expand out the function name a little bit.  Saying things like "Write a principal's key data" would be clearer while still fitting on one line.

write_data() is another candidate for k5buf, but it wouldn't save as many lines of code as for qquote().

tlyu:
Thanks, I'll try k5buf and make the comment improvements.  Also it seems that clang doesn't support the vprintf signature for its format attribute, despite defining __GNUC__, so shutting up that warning portably might be challenging.

tlyu:
Updated.

greghudson:
For vasprintf in k5-platform.h, we have:

    extern int krb5int_vasprintf(char **, const char *, va_list)
    #if !defined(__cplusplus) && (__GNUC__ > 2)
        __attribute__((__format__(__printf__, 2, 0)))
    #endif
        ;

Does that work for shutting up the warning?

greghudson:
In the second commit, the descriptive comments which are now sentence (e.g. "Write a principal's ticket policy") should be punctuated with a period.  Also, write_data() needs a check for buf.data == NULL or a call to k5_buf_status() after it finishes the for loop.  I think that's all for the current patch set.

tlyu:
The gcc-annotated krb5int_vasprintf() prototype is under a !defined(HAVE_VASPRINTF), so it probably doesn't get compiled on OS X, which defines HAVE_VASPRINTF=1.

tlyu:
Updated.  How does this look?

greghudson:
I think this is okay now.

PR #326, opened 2015-09-16
Add missing error mappings to new mechglue func

tkuthan:
mechglue functions gss_get_mic_iov, gss_get_mic_iov_length and
gss_verify_mic_iov don't call map_error to map mechanism specific error
code.  As a result, on error subsequent call to gss_display_status fails
with GSS_S_BAD_MECH, because no translation for the error code is found
in the error table.

This patch adds the missing map_error call.


greghudson:
Pushed to master as 969c976ff021db7e409ce9a38d28316d9acfa2cc

PR #329, opened 2015-09-21
Fail during configure if stdint.h missing

tlyu:
We now require stdint.h to build this software.  Gracefully fail
during configure time if stdint.h is missing.

ticket: 8221
target_version: 1.14
tags: pullup


greghudson:
Looks okay to me.

PR #330, opened 2015-09-23
Fix kadmin with e2fsprogs libss

greghudson:
The libss in e2fsprogs exports ss_execute_command(), but does not
prototype it (as of this writing; a patch has been submitted
upstream).  When using the system ss library, check if a prototype is
needed and provide one if so.


frozencemetery:
This works for me (Fedora & Debian).  Thanks much!

tlyu:
Looks OK to me too.

PR #331, opened 2015-09-25
Fix minor utf8-to-ucs2s read overrun bug

greghudson:
k5_utf8s_to_ucs2s() reads and discards one extra byte from the input
string before terminating its loop, possibly overrunning the input
buffer of its caller.  This overrun is typically without consequence,
but can show up in tools like asan or valgrind during RC4
string-to-key operations.  Fix the bug by swapping the order of the
loop conditions.


tlyu:
This looks OK to me.

greghudson:
Pushed to master as eb52da21d72faa3d00b1205a5a0fdbabc45c9e6d

PR #332, opened 2015-09-29
Don't depend on system krb5.conf in test suite

frozencemetery:
It's my own fault that I ended up caring about this behavior, but this may be more broadly useful in case anyone hits it again.  Anyway, this fixes tests which break if there are problems parsing /etc/krb5.conf from the system.

### DIFF
--- a/src/lib/krad/Makefile.in
+++ b/src/lib/krad/Makefile.in
@@ -32,13 +32,15 @@ install-unix:: install-libs
 
 clean-unix:: clean-liblinks clean-libs clean-libobjs
 
+HIDE_CONFIG = KRB5_CONFIG=/dev/null
+
 check-unix:: t_attr t_attrset t_code t_packet t_remote t_client
-	$(RUN_SETUP) $(VALGRIND) ./t_attr
-	$(RUN_SETUP) $(VALGRIND) ./t_attrset
-	$(RUN_SETUP) $(VALGRIND) ./t_code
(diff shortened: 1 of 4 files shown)


greghudson:
I agree that tests should not access /etc/krb5.conf.

The idiom you used is taken from lib/krb5/krb/Makefile.in, which actually has a t_krb5.conf file, but is used in directories which do not.  An empty krb5.conf probably does not cause any trouble for the tests, but it's weird to use such a specific pathname for a file which does not exist.  Also, I think there are less verbose solutions available; "KRB5_CONF=pathname command ..." could be used instead of explicitly exporting KRB5_CONF, and a make variable could be used to avoid the repetition.

frozencemetery:
Oh, that makes sense.  And it probably looked like it worked okay because empty is superior to broken...

How would you prefer that the cases where t_krb5.conf doesn't exist be handled?  `KRB5_CONF=/dev/null` or some such?

greghudson:
/dev/null is probably reasonable.  Even on platforms where it doesn't exist, it won't affect the tests.

frozencemetery:
New version hopefully addressing comments.  The expansion of `RUN_SETUP` uses the `VARIABLE=value; export $VARIABLE` idiom, so setting `KRB5_CONFIG` without using this idiom has to occur after `RUN_SETUP`.

greghudson:
Please see PR #341; I decided to make some deeper build system changes to make running C test programs from Makefiles cleaner.

PR #333, opened 2015-09-30
Use krb5_sname_to_princ for hostbased princs

tkuthan:
Currently add_admin_princs creates hostbased service principal names
directly from naming queries.

To avoid code duplication and potential inconsistencies in hostname
canonicalization, this patch introduces add_admin_sname_princ as a
hostbased alternative to add_admin_princ and canonicalizes hostname
using krb5_sname_to_princ.

### DIFF
--- a/src/kadmin/dbutil/kadm5_create.c
+++ b/src/kadmin/dbutil/kadm5_create.c
@@ -44,8 +44,14 @@
 #include <kdb.h>
 #include "kdb5_util.h"
 
-static int add_admin_princ(void *handle, krb5_context context,
-                           char *name, char *realm, int attrs, int lifetime);
+static int add_admin_princ_common(void *handle, krb5_context context,
+                                  krb5_principal principal, int attrs,
+                                  int lifetime);
+static int add_admin_princ(void *handle, krb5_context context, char *name,
+                           char *realm, int attrs, int lifetime);
+static int add_admin_sname_princ(void *handle, krb5_context context,
+                                 char *sname, char *realm, int attrs,
+                                 int lifetime);
 static int add_admin_princs(void *handle, krb5_context context, char *realm);
 
 #define ERR 1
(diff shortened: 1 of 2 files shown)


greghudson:
Were you encountering an actual inconsistency in canonicalization which motivated this change, and if so, what were the circumstances?

tlyu:
I think this is related to https://krbdev.mit.edu/rt/Ticket/Display.html?id=6522
and I recall experiencing some test suite problems that might be attributed to this.

greghudson:
It can't be very closely related; that ticket refers to a completely different facility, and all of the behaviors it mentions are already implemented (prior to the 2009 ticket creation date) in add_admin_princs().

greghudson:
I haven't done anything approximating a complete review, but I spotted a problem in add_admin_sname_princ().  It should use krb5_set_principal_realm(); what it does now could leak memory.  A zero-length realm does not necessarily imply a lack of allocated memory in the data field.

tkuthan:
Updated the code to use krb5_set_principal_realm. (How did I not find it originally?)

Our motivation for this change is in the fact, that we have modified krb5_sname_to_princ to force DNS for naming queries. (There are known deployments, were NIS and DNS don't agree on canonical hostnames.) Obviously we need created admin SPNs' hostname part to match  krb5_sname_to_princ output.

In general, re-implementing krb5_sname_to_princ logic in add_admin_princs doesn't feel quite right.

tlyu:
The add_admin_princ(... KADM5_ADMIN_HOST_SERVICE ...) adds a KRB5_KDB_PWCHANGE_SERVICE flag that wasn't there before.  Only the changepw service principal should have that set, not the general kadmin service principal.

tkuthan:
Thank you for caching that, fixed. Together with the wrong lifetime.

greghudson:
Thanks for explaining the motivation behind the change.  I'm thinking we should take this change (after I review it more carefully) for 1.15 and not put it into 1.14, since it is just a code quality improvement from our perspective and it carries some risk.

When looking over the existing code, I did find one behavior difference between add_admin_princs() and sn2princ: we still use AI_ADDRCONFIG in add_admin_princs(), but have stopped doing so in sn2princ.  Changing that would be a simple, low-risk patch.

tkuthan:
OK, that is still convenient for us.
For the time being we will maintain this as a patch and drop it once it makes it to the released code.

greghudson:
Huh.  There are some more significant behavior differences between add_admin_princ() and sn2princ:

* add_admin_princ() never uses reverse DNS, while sn2princ does by default.
* sn2princ is affected by the rdns and dns_canonicalize_hostname profile variables (the latter added in #7703); add_admin_princ() is not.
* sn2princ handles :port suffixes (#7795).  Not really an issue since local hostnames aren't likely to have them.
* If forward resolution fails, sn2princ falls back to the un-canonicalized hostname (a change by Sam in Oct 2011, no ticket).  add_admin_princ() gives up.
* sn2princ removes a trailing dot from the result; add_admin_princ() does not.  I think only buggy resolvers would put a trailing dot there, so not a big deal.
* sn2princ does a realm lookup with krb5_get_host_realm(); add_admin_princ() always uses the realm being created.

I'm almost wondering if we want to add a new API for canonicalizing hostnames (I think Heimdal has one we could consider copying) and use that instead.  At a minimum, if we use sn2princ, we should always ignore the realm it returns, not just if it's empty.  It would never make sense to create admin princes for realm Y when creating a database for realm X.

tkuthan:
Updated the patch to always ignore realm from krb5_sname_to_principal as recommended.

greghudson:
While thinking about this, I noticed that libkadm5clnt also doesn't use krb5_sname_to_principal() to construct the service principal name from the configured admin hostname.  Instead, kadm5_get_admin_service_name() contains yet another piece of canonicalization code--also using AI_ADDRCONFIG, unaffected by rdns or dns_canonicalize_hostname.

greghudson:
Just to avoid duplication of effort: I am working on an alternative pull request which will address this and the other two issues I noted.  Adding a krb5_expand_hostname() API means we don't need structural changes to add_admin_princs(), so that patch becomes simpler.

greghudson:
I am closing this in favor of PR #349.

PR #334, opened 2015-09-30
Correct major code for non-default QOP

tkuthan:
This patch fixes krb5_gss_wrap_size_limit return code to comply with
RFC 2743; non default QOP argument should result in GSS_S_BAD_QOP, not
GSS_S_FAILURE.

(This is the first of a batch of 6 fixes (mostly nits) that were reported by Solaris gss_api regression testsuite. Sorry for the (mini) flood.)

### DIFF
--- a/src/lib/gssapi/krb5/k5seal.c
+++ b/src/lib/gssapi/krb5/k5seal.c
@@ -337,7 +337,7 @@ kg_seal(minor_status, context_handle, conf_req_flag, qop_req,
        them later.  */
     if (qop_req != 0) {
         *minor_status = (OM_uint32) G_UNKNOWN_QOP;
-        return GSS_S_FAILURE;
+        return GSS_S_BAD_QOP;
     }
 
     ctx = (krb5_gss_ctx_id_rec *) context_handle;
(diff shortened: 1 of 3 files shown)


greghudson:
GSS_S_BAD_QOP was added in GSS-API v2, and is currently unused by the krb5 mech; instead we return GSS_S_FAILURE with a minor status of G_UNKNOWN_QOP.  Changing these returns to GSS_S_BAD_QOP would improve conformance with the current standard, but it's very piecemeal to do it for gss_wrap_size_limit() and not gss_wrap/gss_get_mic.  Please search for G_UNKNOWN_QOP and change the other five places as well.

Also, please review all submitted patches for unrelated formatting changes and remove them.  We won't take them as part of bugfixes, and it just adds extra work for us to reverse them.

tkuthan:
Will do.

tkuthan:
Updated the patch to fix the remaining occurrences + ditched the unrelated formating change.

greghudson:
Pushed to master as 45ccc1c85f42e4f41f2042df8a51dd7826533029

I unfortunately missed this in the final review, but there was still a formatting change in wrap_size_limit.c after your change, which renders the changed code stylistically inconsistent with the adjacent if block.

tkuthan:
Sorry for that.
For future reference, by 'formatting change' you mean my removing of () from return clause?
Cstyle was complaiing about that; should I ignore these complains next time?

greghudson:
Thank you for using cstyle!  But yes, its output should be taken as advisory rather than mandatory, and the prevailing style of nearby code that you isn't being modified should take precedence.  (In theory we will gradually modernize code so that it conforms to our practices, in commits which don't change the code's behavior, and this dissonance will become less of an issue over time.)

PR #335, opened 2015-09-30
Correct minor status on cred usage mismatch

tkuthan:
In krb5_gss_store_cred_into if actual_usage differs from expected cred
usage, minor status should be set to G_CRED_USAGE_MISMATCH instead of
G_BAD_USAGE.

### DIFF
--- a/src/lib/gssapi/krb5/store_cred.c
+++ b/src/lib/gssapi/krb5/store_cred.c
@@ -217,7 +217,7 @@ krb5_gss_store_cred_into(OM_uint32 *minor_status,
         return GSS_S_CREDENTIALS_EXPIRED;
 
     if (actual_usage != GSS_C_INITIATE && actual_usage != GSS_C_BOTH) {
-        *minor_status = G_BAD_USAGE;
+        *minor_status = G_STORE_ACCEPTOR_CRED_NOSUPP;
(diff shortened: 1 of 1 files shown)


greghudson:
I think G_STORE_ACCEPTOR_CRED_NOSUPP is more applicable.  If the caller says GSS_C_BOTH and the cred is an initiator cred, we proceed normally, so we aren't normally picky about the caller asking for extra.  Therefore, if the caller asks for GSS_C_BOTH and the cred is an acceptor cred, the problem isn't a mismatch with what the caller asked, but that we can't store an acceptor cred.

tkuthan:
Makes sense, fixed.

tkuthan:
Makes sense, fixed.

greghudson:
Looking at this more closely, I think we can simplify the change down to returning G_STORE_ACCEPTOR_CRED_NOSUPP with the existing conditional, without adding a new conditional block.  We just got actual_usage from gss_inquire_cred(), not from the caller, so we can be confident that there are only three possible values.

tkuthan:
That is also true, fixed again.

greghudson:
Pushed to master as c0e16bb2f654038ad81602e89851f232916da051

PR #336, opened 2015-09-30
Defines error status GSS_S_BAD_MIC

tkuthan:
GSS_S_BAD_MIC is preferred alias for GSS_S_BAD_SIG according to RFC
2743.


greghudson:
Pushed to master as be87852a12737eed448032875fb74f23e9cbf26b

PR #337, opened 2015-09-30
inquire names for gss_mech_krb5_wrong

tkuthan:
The GSS-API library indicates Kerberos mechanism under multiple OIDs
including gss_mech_krb5_wrong. But it is not able to inquire names for
this OID.

This patch modifies krb5_gss_inquire_names_for_mech not to fail with
GSS_S_BAD_MECH for mechanism = gss_mech_krb5_wrong.


greghudson:
Pushed to master as 95736f7d51cce7551c3c20450ff56831a71043df

PR #338, opened 2015-09-30
Check input of oid_set manipulating functions

tkuthan:
This patch adds additional argument sanity checks to oid_set
manipulating functions generic_gss_create_empty_oid_set and
generic_gss_add_oid_set_member.

### DIFF
--- a/src/lib/gssapi/generic/oid_ops.c
+++ b/src/lib/gssapi/generic/oid_ops.c
@@ -137,6 +137,9 @@ generic_gss_create_empty_oid_set(OM_uint32 *minor_status, gss_OID_set *oid_set)
 {
     *minor_status = 0;
 
+    if (oid_set == NULL)
+        return GSS_S_CALL_INACCESSIBLE_WRITE;
+
     if ((*oid_set = (gss_OID_set) gssalloc_malloc(sizeof(gss_OID_set_desc)))) {
         memset(*oid_set, 0, sizeof(gss_OID_set_desc));
         return(GSS_S_COMPLETE);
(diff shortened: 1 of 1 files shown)


greghudson:
What's the motivation for this change?  If it's fixing a bug, we need to know for the RT issue, and we need to consider other ways of fixing the bug.  It's not our practice to put null pointer checks in every internal utility function.

tkuthan:
This test was added to Solaris gss_api testsuite in 2005 after a bug, where gss_create_empty_oid_set was segfaulting on NULL oid_set. Sadly neither that bugID nor the testsuite bugID give any background on the use case.

For a public function it is reasonable to sanitize input as opposed to just crashing on invalid input.

But looking at it now I am in two minds about the return value. GSS_S_CALL_INACCESSIBLE_WRITE is the logical choice, but RFC 2743 says this function returns either just GSS_S_COMPLETE or GSS_S_FAILURE.

greghudson:
It's normal for public GSS-API functions to check input arguments.  What threw me is that the patch modifies the internal generic_gss_... functions rather than the public APIs, and I hadn't yet realized that the API functions were trivial wrappers around the internal functions.

I see that generic_gss_add_oid_set_member() already does some invalid argument checking on one of its parameters, which isn't my preference, but does present a consistency argument for placing additional checking in the same place.

RFC 2743 doesn't talk about GSS_S_CALL_ return values at all, and RFC 2744 never enumerates them as possible results.  I think RFC 2743 is only enumerating "routine errors" as defined in 2744 section 3.9.1, and calling errors are always a possibility.  So I think the return value is fine.

tkuthan:
Oh, right. Even better then.

greghudson:
Pushed to master as c9e035794caa784b6cdf416e2b3f1d641d011390

PR #340, opened 2015-10-01
Avoid using system libraries during testing

frozencemetery:
We want to ensure that we test what we have built, rather than the system configuration.

This is a fix originally from [rhbz#1164304](https://bugzilla.redhat.com/show_bug.cgi?id=1164304).

### DIFF
--- a/src/kadmin/testing/proto/krb5.conf.proto
+++ b/src/kadmin/testing/proto/krb5.conf.proto
@@ -2,6 +2,7 @@
 	default_realm = __REALM__
 	default_keytab_name = FILE:__K5ROOT__/v5srvtab
 	dns_fallback = no
+	plugin_base_dir = __PLUGIN_DIR__
 
 [realms]
 	__REALM__ = {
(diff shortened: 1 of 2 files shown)


greghudson:
From the comments in the Red Hat bugzilla entry, it sounds like this fix may be incomplete.  I did an strace over make check to see what other references we make to the install tree.  I found:

* We load KDB modules from the install tree, but don't use them (as long as there is one in the build tree).  To fix this, we should modify kdb_load_library() to use only the profile paths when they are specified.

* We reference the message catalog ten times from kdb5_util invocations run from the kdb_check target in tests.  I guess we should add LC_ALL=C to RUN_SETUP.

The bugzilla report suggests that the test suite opens libkadm5srv_mit.so.8 from the install tree, which would be more concerning, but I couldn't find any evidence of that.  This could be something we have fixed since 1.12, but I don't know how we could have; we have been using --enable-new-dtags since 1.9.

greghudson:
Pushed to master as 3db59d81bbbe389553d13efeee1c99117b459b24.  I will open RT tickets for the remaining issues.

frozencemetery:
Sounds good, thanks!

PR #341, opened 2015-10-04
Fix some uses of installed files in the test suite

greghudson:
These four commits supersede PR #332 and also fix RT issue 8264.  The first three commits are intended for the release branch; the fourth is not.


frozencemetery:
Tested good & looks reasonable, thanks!

tlyu:
Fixup for RUN_ENV simplifications is in e767d9fe054a9976d3d7ef25348688fc54995832

PR #342, opened 2015-10-04
Make ksu work when unsetenv() returns NULL

greghudson:
Some older platforms (OS X 10.4, glibc 2.2.1) declare unsetenv() as
returning void, as does ksu's compatibility definition of unsetenv().
Don't use the return value in get_configured_defccname().

### DIFF
--- a/src/clients/ksu/main.c
+++ b/src/clients/ksu/main.c
@@ -810,12 +810,7 @@ get_configured_defccname(krb5_context context, char **target_out)
 
     *target_out = NULL;
 
-    if (unsetenv(KRB5_ENV_CCNAME) != 0) {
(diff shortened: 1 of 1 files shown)


kaduk:
Did we try this on a platform with aggressive unused-result warnings, such as gcc 5 with recent glibc?  More and more functions are getting marked to trigger warnings if their return value is not checked.

greghudson:
Well, we already have an unchecked use of unsetenv(), right at the beginning of main().  So it wouldn't be a new problem.

Also, I don't think unsetenv() can fail if you pass it non-null input, in any reasonable implementation.  So it seems unlikely that this particular function would get marked with that warning.

PR #343, opened 2015-10-06
Pass down the new password as a const char pointer

gladiac:
The new password parameter which is passed down or passed around internally should be const. It gets never modified and shouldn't.


greghudson:
This patch creates a >79-character line in krb5.hin (line 5111).

Also, what about krb5_change_password()?  It seems incomplete to constify two out of three newpw parameters.

gladiac:
You're right, I forgot it ...

greghudson:
Pushed to master as 03c5058f2a1315aa718f0a083f5568bedaf187e8

PR #345, opened 2015-10-21
Secure cookie memory hygiene and a couple of memory leak fixes

greghudson:
I noticed the lack of sanitization when making the prototype SPAKE preauth module, and noticed a couple of memory leaks while fixing the sanitization issue.


tlyu:
This looks OK to me.

PR #346, opened 2015-11-10
Rename two test programs for consistency

greghudson:
I've been meaning to correct this minor annoyance for a while, and since I will be amending t_init_creds.c for SPAKE preauth tests, I figured now is a reasonable time.


tlyu:
This looks ok to me.

PR #347, opened 2015-11-13
Check for the correct SPI function.

simo5:
Checking for the generic gss_acquire_cred function is no guarantee that
gss_acquire_cre_impersonate_name is also implemented.


greghudson:
Pushed to master as 46a4e225d2ecaa4077aa65f12f64273bf4911d3a

PR #348, opened 2015-11-13
Fix impersonate_name to work with interposers

simo5:
This follows the same modifications applied to gss_acquire_cred_with_pw
when interposer plugins were introduced.

### DIFF
--- a/src/lib/gssapi/mechglue/g_acquire_cred_imp_name.c
+++ b/src/lib/gssapi/mechglue/g_acquire_cred_imp_name.c
@@ -334,6 +334,8 @@ gss_add_cred_impersonate_name(OM_uint32 *minor_status,
     gss_cred_id_t	cred = NULL;
     gss_OID		new_mechs_array = NULL;
     gss_cred_id_t *	new_cred_array = NULL;
+    gss_OID_set		target_mechs = GSS_C_NO_OID_SET;
+    gss_OID		selected_mech = GSS_C_NO_OID;
 
     status = val_add_cred_impersonate_name_args(minor_status,
 						input_cred_handle,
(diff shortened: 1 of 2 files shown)


greghudson:
Did you run the test suite on this?  I get a failure in t_s4u.py.

greghudson:
Notes on what goes wrong in t_s4u.py:

* At command 15, we run "./t_s4u p:user@KRBTEST.COM p:service/2@KRBTEST.COM".  Everything works as expected: service/1 makes an S4U2Self request and gets a ticket from user to service/1, but it's not forwardable because service/1 doesn't have ok-to-auth-as-delegate set.  So gss_accept_sec_context() doesn't produce a delegated cred and t_s4u doesn't try constrained delegation.

* At command 16, we run the same command with the --spnego option.  Something goes wrong in gss_acquire_cred_impersonate_name(), and the client winds up making a regular TGS request and getting tickets from service/1 to service/1.    t_s4u tries constrained delegation with this evidence ticket and gets a NOT_ALLOWED_TO_DELEGATE error from the KDC.

simo5:
Sorry, I was testing with gssproxy's framework and did not actually run krb5's test suite, my bad.
I'll ping again once I have it fixed.

simo5:
I pushed a new patch that fixes the issue with Spnego.
It simply does the same thing that's already done in acquire_cred_with_password

However  I did not for see this issue when testing the interposer w/o SPNEGO, so please hold off merging until I am finished testing against the interposer code I have in gss-proxy.

simo5:
@greghudson  fwiw I did test with krb5 but not spnego interpodsed by gssproxy, but that's fine as the interposer normally does not interpose spnego.
So if you agree with the patch here please feel free to merge.

greghudson:
Pushed to master as b3901af6970fb7bde88eb16d51c8d05db6f37746

PR #349, opened 2015-11-23
Add and use krb5_expand_hostname()

greghudson:
This supersedes PR #333.

I have elected not to solve RT issue 8277 in this pull request, because solving that issue is complicated.


tlyu:
The bit about trailing dots makes me a little nervous, but I'm willing to believe it's not too important in practice. I had to fix up conflicts in the exports file when attempting to rebase onto master. Otherwise, it looks ok to me.

PR #350, opened 2015-11-24
Ensure pwsize is initialized in chpass_util.c

frozencemetery:
This got lost in the CVE shuffle, but one of the things needed to get 1.14 building happily was initialization of pwsize.  This is how I'm currently doing it in Fedora.

### DIFF
--- a/src/lib/kadm5/chpass_util.c
+++ b/src/lib/kadm5/chpass_util.c
@@ -74,6 +74,7 @@ kadm5_ret_t _kadm5_chpass_principal_util(void *server_handle,
     if (ret_pw)
         *ret_pw = NULL;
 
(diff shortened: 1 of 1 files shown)


greghudson:
What compiler version gives this error?  (It's a false positive, and I want to note the compiler version in the commit message.)

frozencemetery:
So it is.  That function reads oddly to me (to the point where I did my own cleanup [over here](https://github.com/krb5/krb5/compare/master...frozencemetery:chpass_cleanup)), so I did not realize.  I did check now, though; 4.9.2, 5.1.1, and 5.2.1 all seem okay.  Must have been in an older version of something.

Sorry for the noise.

PR #351, opened 2015-11-25
Fix memory leak in SPNEGO gss_init_sec_context()

greghudson:
After the initial call to spnego_gss_init_sec_context(), the context
handle can leak if init_ctx_cont() returns an error, because the
cleanup handler assumes that spnego_ctx contains the value of
*context_handle.  Fix this leak by setting spnego_ctx before the if
block which contains that call.  Reported by Adam Bernstein.

### DIFF
--- a/src/lib/gssapi/spnego/spnego_mech.c
+++ b/src/lib/gssapi/spnego/spnego_mech.c
@@ -115,14 +115,14 @@ handle_mic(OM_uint32 *, gss_buffer_t, int, spnego_gss_ctx_id_t,
 	   gss_buffer_t *, OM_uint32 *, send_token_flag *);
 
 static OM_uint32
-init_ctx_new(OM_uint32 *, spnego_gss_cred_id_t, gss_ctx_id_t *,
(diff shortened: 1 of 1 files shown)


greghudson:
It's not redundant; init_ctx_new() may have altered *context_handle.

A more aggressive change would adjust the contract of init_ctx_new() to operate on a spnego_gss_ctx_id_t * and init_ctx_cont() to operate on a spnego_gss_ctx_id_t, instead of operating on the generic GSS types; then there would be no need for the apparent redundancy.  If you'd like, I can prepare that change as a second commit for the master branch only.

tlyu:
I'm ok with you making the proposed improvement patch for master only.

greghudson:
I rebased against the current master and pushed a second commit to clean up the helpers which currently operate on gss_ctx_id_t parameters.  The adjustments to spnego_gss_accept_sec_context() weren't completely trivial because of the violence done to that function to add NegHints support.

tlyu:
This looks good to me now.

PR #352, opened 2015-11-28
iprop notification

greghudson:
This is Richard Basch's iprop notification.  At a minimum it needs:

* Style fixes
* All of the XXX comments addressed
* Documentation changes to the rst sources, not the generated man page
* Automated tests

I have not given it a careful review, so it may need additional changes as well.


tlyu:
Edit subject line to link to JIRA.

tlyu:
Really edit the subject line this time, and bump so JIRA notices. Sorry for the noise.

greghudson:
I'll make a note here when I'm making a non-trivial change to Richard's patch.  The original is commit 83e42bb94f33a44e97f6903cc8734486cfa2ca2c.

The first change: the patch adds code in kpropd.c around line 991 to "Poll immediately after a resync to retrieve missing updates," if the iprop_get_updates_1() result is UPDATE_FULL_RESYNC_NEEDED and the full_resync() result is UPDATE_OK.  This code isn't useful, because the full resync probably hasn't happened yet; that happens asynchronously after kadmind launches kprop and connects to the child process.  This code change causes t_iprop.py to fail with "kadmind gave consecutive full resyncs".  We already have a working facility to poll immediately after a resync in commit c19fee0a748dd26ba6ac62118cf4b9ebec36ba2f (RT issue 8161).  I will remove this code change.

greghudson:
Next change: Richard's patch contains changes to kproplog, which the commit message describes as "Fix kproplog when ulog is not full and does not begin with sno=1".  I believe these changes are functionally redundant with 2743fdee849be4a00a9ea72d601004c95b22dd56 and will remove them from the commit.  Looking closer, there are three kproplog.c changes:

* The bulk of the code changes are to adjust internal contracts to pass the ulog and ulogsize values to print_update() via context->kdblog_context instead of passing them directly.  I don't know why this would be desirable; it adds code and obscures the data flow.

* There is a change to avoid using kdb_first_sno when calculating the starting index, instead using kdb_last_sno - kdb_num.  This change was probably necessary in Richard's implementation of hierarchical iprop, because he redefined kdb_first_sno to be the serial number state preceding the initial update, but it should not be a problem for us because I chose to use dummy update entries to solve the same problem (see item 2 in http://k5wiki.kerberos.org/wiki/Projects/Hierarchical_iprop#Related_problems).

* There is a change to treat index entries as corrupt if the serial number doesn't match the expected value.  This change is fine but as far as I know unnecessary, and it would probably be better to print an annotation to the "Update serial #" line if we decide to care.

greghudson:
t_iprop.py is failing under Travis due to the 0491c778064e80aaf2aaeb4475a0db333542fed9 problem (because of new code introduced in kpropd.c).  The error-handling when getclhoststr() fails includes a double-free (iprop_svc_principal was just freed before the if statement); the code should free iprop_svc_princstr instead.  In general there is too much code added to a function which already needed refactoring, and I have some concerns about the functional semantics of the kpropd changes.  For now I will fix the double-free and use NI_MAXHOST for the buffer size.

greghudson:
In kpropd.c, I changed getclhoststr() to operate on the iprop principal name rather than its unparsed form.  This is more elegant, but still pretty roundabout.  The goal of this function is to get the canonicalized local hostname.  We know it's in the second component of iprop_svc_principal because we created that principal with sn2princ with a NULL host argument.  I don't think there's currently a better way to do this, but it's confusing to read.

I'm still thinking through my reservations about how the notification from kpropd to kadmind works on intermediate KDCs.  The current method seems fragile, and probably can't be tested in our test suite.   It operates by temporarily replacing params.admin_server with the local hostname and calling kadm5_init_with_skey(), just as we would to talk to the upstream kadmind.  This can only work if the upstream and local kadmind are running on the same port, which can't be true in our automated tests.  It is also unconditional, which means that leaf kpropds (the usual case) will be repeatedly trying and failing to connect to their local port 749s.  That's usually harmless and pretty cheap, but it may be less so in the age of apparmor and selinux and IDS systems.

There are also some issues with the code at the detail level.  As I previously noted, the notification code needs to be factored into its own function.  It triggers when incr_ret is UPDATE_OK, which means it won't happen after a full resync followed by an UPDATE_NIL response to the subsequent incremental request.  The trigger point could be moved into the switch statement on incr_ret rather than being a conditional afterwards.  There's a call to kadm5_get_kiprop_host_srv_name() which appears to be redundant with the one to acquire master_svc_princstr.

greghudson:
After some discussion with Sarah and Tom, I think we're not comfortable with using the kadmin protocol to communicate from kpropd to the kadmind -proponly on the same host.  A Unix-domain socket in the KDC directory seems like it would make more sense.  Sending a signal would also work, since kpropd and kadmind should be running as the same user, but it seems hard to make kpropd aware of the kadmind pid in a robust way.

PR #353, opened 2015-11-29
Avoiding directives that break statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/src/lib/kdb/kdb_default.c
+++ b/src/lib/kdb/kdb_default.c
@@ -249,13 +249,15 @@ krb5_db_def_fetch_mkey_stash(krb5_context   context,
     krb5_ui_2 enctype;
     krb5_ui_4 keylength;
     FILE *kf = NULL;
+    int io_error;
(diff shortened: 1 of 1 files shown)


greghudson:
I think at this point we can just assume that "b" is supported, and get rid of the ANSI_STDIO conditional.  (It's used in a few other places in the tree where we open binary files.)  Pre-C89 compatibility was probably a legitimate concern when that code was first written, but not any more.

greghudson:
PR #354 supersedes this, so I'm closing it.  Thanks for pointing out the problem area.

PR #354, opened 2015-11-29
Remove ANSI_STDIO conditional

greghudson:
As we no longer support pre-C89 platforms, assume that the fopen "b"
flag is always allowed.

Supersedes PR #353.


tlyu:
This looks OK to me.

PR #355, opened 2015-11-30
Remove unused SPNEGO context fields

greghudson:
Nico's IAKERB/SPNEGO patches contained two small SPNEGO code cleanup commits; this pull request combines them into one commit since they are thematically similar.  The unused fields are also present and unused in 1.5 and in my checkout of the Solaris code.  I assume they are development vestiges from when the code was first written.


tlyu:
This looks ok to me.

PR #357, opened 2015-12-04
Updating .gitignore

SaharahSarah:
Adding some files that I generally add to all .gitignore, as well as some eclipse files and the build directory.


greghudson:
It seems like this could be a single commit.  Also, please see http://k5wiki.kerberos.org/wiki/Coding_style/Version_control_practices for how we write commit messages.

The changes themselves seem fine.  I personally use a .git/info/exclude containing "/*build" since I often have multiple build directories, but I think it's fine to privilege a top-level build directory.  (We have talked about moving the build system up to the top level to make the doc build more sensible, at which point having a "build" subdir in the source tree would no longer work.  But we can always remove the .gitignore line if we ever do that.)

tlyu:
To better conform with advice like http://chris.beams.io/posts/git-commit/, I suggest changing the summary line to "Ignore some common files" or maybe "Add some common files to .gitignore" (note the lack of period at the end).

SaharahSarah:
Done

tlyu:
Hm, the commit metadata didn't work right on this pull request for reasons I don't quite understand. Pushed as 652da869592b1ef4b96694db07ed178add1f80e2

PR #359, opened 2015-12-07
Renaming reconf script to align more closely with autotools conventions

SaharahSarah:
There are certain conventions users expect when working with a project
that uses autotools. Deviating from those conventions adds confusion and
complexity to a project, and let's face it, krb5 is not exactly a
shining example of a convention-conforming project. In an effort to try
to simplify the krb5 build, I am moving and renaming the src/util/reconf
script to src/bootstrap. The bootstrap name was chosen since the
autotools documentation says it is the preferred name. For more
information on this please see:

https://goo.gl/JRvpYK
https://goo.gl/V1i521

ticket: 8299


SaharahSarah:
Did you hear that? That's the sound of build script maintainers all groaning at once. I'm sure we'll hear that much more in the coming months.

greghudson:
At a minimum, the documentation should be updated.  We should probably discuss this on Tuesday.

SaharahSarah:
I updated all references to the old script.

SaharahSarah:
Let's wait on this. There seems to be a lot of opinion on it, and we might be leaning towards just deleting the script completely.

tlyu:
Closing because I just opened PR #374 as an alternative.

PR #360, opened 2015-12-07
Fix NSS k5crypto iov procesing bug

greghudson:
I ran across a bug in the NSS k5crypto back end while running it through the test suite.  I don't think anyone uses this back end, but here is a fix.


kaduk:
Looks okay from here.

PR #361, opened 2015-12-07
Add IPv6 address logging support to audit plugin

SaharahSarah:
The jsonwalker.py test was failing due to the audit plugin only logging
the ip for IPv4. The audit plugin should log the ip address for both
IPv4 and IPv6. This commit renames the json "ipv4" field in the
"fromaddr" json object to "ip" and sets the field to both IPv4 and IPv6
addresses. Any parsers of the audit log can tell which address is stored
in the "ip" field by checking the "type" field in "fromaddr". The
jsonwalker.py reference json, au_dict.json, is also updated to handle
this new field name.

###### old audit:
```javascript
{"fromaddr":{"type":<int>,"length":<int>,"ipv4":[<4 int>]}
```
where "ipv4" only occurs when type is ADDRTYPE_INET

###### new audit:
```javascript
{"fromaddr":{"type":<int>,"length":<int>,"ip":[<<length> int>]}
```
where "ip" only occurs when type is ADDRTYPE_INET or ADDRTYPE_INET6

[ticket: 8298](https://krbdev.mit.edu/rt/Ticket/Display.html?id=8298)

### DIFF
--- a/src/plugins/audit/j_dict.h
+++ b/src/plugins/audit/j_dict.h
@@ -44,7 +44,7 @@
 #define AU_FROMPORT           "fromport"
 #define AU_FROMADDR           "fromaddr"
 #define AU_TYPE               "type" /* used by fromaddr */
-#define AU_IPV4               "ipv4" /* used by fromaddr */
+#define AU_IP                 "ip" /* used by fromaddr */
 #define AU_SESS_ETYPE         "sess_etype"
 #define AU_SRV_ETYPE          "srv_etype"
 #define AU_REP_ETYPE          "rep_etype"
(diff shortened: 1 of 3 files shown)


tlyu:
This change isn't backward compatible, but it's an experimental plugin, so we might not care about that so much. Alternatively, we could add an "ipv6" key that can be present instead of "ipv4" when there's a v6 address. I suspect the latter approach would require adding specialized logic to jsonwalker.py though.

SaharahSarah:
The [audit plugin wiki entry](http://k5wiki.kerberos.org/wiki/Projects/Audit#Description) states:

> The interface is considered "experimental", in that API stability is not guaranteed to future major releases.

Second, having an ipv6 and ipv4 key is redundant since the type field already says what type it is. There's no reason to add complexity when the needed information is already present.

tlyu:
Good point; it's a fairly direct translation of krb5_address, which itself a close translation of struct sockaddr.

tlyu:
Pushed as 5e063638032912f93df1426dc92cdae5e65c5e5a.

PR #362, opened 2015-12-08
Fix FILE ccache negative enctype unmarshalling

greghudson:
Opinions are welcome on whether this should get a ticket and be marked for backport.


tlyu:
This looks ok to me. We probably won't need to back port it unless we intend to back port tests that need negative enctypes. Let's wait to open a ticket until someone shows up with a use case that requires back porting.

PR #363, opened 2015-12-09
Correctly use k5_wrapmsg() in ldap_principal2.c

tlyu:
Commit ebcdf02f8ec212555b1762007fa8454615900f36 incorrectly used
k5_prependmsg() in an error handling clause in
krb5_ldap_get_principal().  Use k5_wrapmsg() instead.

ticket: 8301
target_version: 1.14-next
tags: pullup

PR #364, opened 2015-12-09
Error on int-conversion with clang

greghudson:
gcc has no option (short of -Werror) to error on "makes pointer from
integer without a cast" warnings, but clang does.  In aclocal.m4, add
"error=int-conversion" to the warning flags we try to use.

### DIFF
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -526,7 +526,7 @@ if test "$GCC" = yes ; then
     TRY_WARN_CC_FLAG(-Wno-format-zero-length)
     # Other flags here may not be supported on some versions of
     # gcc that people want to use.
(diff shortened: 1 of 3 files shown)


greghudson:
It looks like gcc 5.0 has at least -Wno-int-conversion, and therefore might have -Werror=int-conversion.  gcc 5.0 also adds -Wno-incompatible-pointer-types (and therefore perhaps also -Werror=incompatible-pointer-types), which would catch a similar class of errors.  I'm try to determine if clang has that as well.

greghudson:
Okay, updated the message to make note of gcc 5.0, and added -Werror=incompatible-pointer-types.  I had to clean up some const qualifier warnings for the second one.

I won't push this until PR #363 is pushed, since it would of course cause a build failure.

tlyu:
This looks ok to me.

PR #365, opened 2015-12-09
kdb5_util must set the TL_DATA mask flag

simo5:
When the kdb5_util adds a new key it ends up modifying tl_data but
the KADM5_TL_DATA mask flag is not set causing databases that rely
on this signaling (like the LDAP database) to fail to store the
changes made to the entry.

### DIFF
--- a/src/kadmin/dbutil/kdb5_mkey.c
+++ b/src/kadmin/dbutil/kdb5_mkey.c
@@ -178,7 +178,7 @@ add_new_mkey(krb5_context context, krb5_db_entry *master_entry,
                                            mkey_aux_data_head))) {
         goto clean_n_exit;
     }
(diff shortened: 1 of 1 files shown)


greghudson:
On IRC you said, "I was thinking it needs to be set in krb5_dbe_update_tl_data() but looking at it more makes me dubious."  What made you dubious about that approach?

greghudson:
Oh, nevermind, krb5_db_update_tl_data() doesn't take the entry, just the tl-data list.

simo5:
All those functions are in src/lib/kdb which does not include kadmin.h where the mask macros are defined.
That's why I've just changed the kdb5_util function that call them instead.

greghudson:
Pushed to master as c877f13c8985d820583b0d7ac1bb4c5dc36e677e

PR #366, opened 2015-12-11
Add libkrb5support dependencies to test plugins

greghudson:
In some build environments, dependencies on libkrb5support can be
generated just from static inline functions in our header files, even
if those functions aren't used.  In two test plugin modules, use
$(KRB5_BASE_DEPLIBS) and $(KRB5_BASE_LIBS) to depend on libkrb5support
as well as libkrb5.  (This also pulls in libk5crypto, which is
unnecessary for these modules, but is inconsequential for a test
module.)  Reported by Will Fiveash.


tlyu:
This looks ok to me.

greghudson:
Pushed to master as 5568d31f45fb78f505340a5b520b22d4dd3f6522

PR #367, opened 2015-12-11
Make krb5_ucs2 and krb5_ucs4 unsigned

greghudson:
[This should hopefully fix the build under gcc 5.x, which appears to have been broken by PR #364.]

krb5 contains some Unicode code adapted from OpenLDAP.  OpenLDAP uses
unsigned types for UCS-2 and UCS-4 code points, but the adapted code
uses signed types, which creates pointer-sign warnings when we call
uccompatdecomp() and similar functions.  To match the OpenLDAP code,
use unsigned types.

### DIFF
--- a/src/include/k5-utf8.h
+++ b/src/include/k5-utf8.h
@@ -68,37 +68,10 @@
 #ifndef K5_UTF8_H
 #define K5_UTF8_H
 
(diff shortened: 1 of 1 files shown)


tlyu:
This looks ok to me (once #368 is applied).

PR #368, opened 2015-12-11
Cast away constness when assigning to krb5_data

tlyu:
Some password-changing library functions take a const char * parameter
but try to assign it to krb5_data.data, which isn't const.  PR #364
causes some compilers to produce errors in such situations, so cast
away the constness.  This is almost certainly safe because of the
nature of the code that consumes these krb5_data values.

The first commit in this PR is already included in #367.


greghudson:
Looks good to me.

PR #369, opened 2015-12-16
Fix kdb5_ldap_util stashsrvpw password file logic

greghudson:
kdb5_ldap_util stashsrvpw has several inconsistencies with the
password file determination in libkdb_ldap, and could try to fopen() a
NULL filename in some cases.  Factor out the determination of the
configured password file and make it consistent with libkdb_ldap.
DEF_SERVICE_PASSWD_FILE is no longer used after these changes, as it
is not respected by libkdb_ldap.


tlyu:
At a previous dev call, I think we talked about this and were ok with this change as is. We would have a separate feature request to add a more proper default path.

PR #370, opened 2015-12-17
Fix mechglue on gss_inquire_attrs_for_mech()

frozencemetery:
This includes proper mechanism selection in `gss_inquire_attrs_for_mech()`
itself as well as passing the correct mech down from `gss_accept_sec_context()`
through `allow_mech_by_default()`.

1.14 introduces the new allow_mech_by_default() call which caused all sorts of excitement in GSS-Proxy land.

### DIFF
--- a/src/lib/gssapi/mechglue/g_accept_sec_context.c
+++ b/src/lib/gssapi/mechglue/g_accept_sec_context.c
@@ -94,6 +94,12 @@ allow_mech_by_default(gss_OID mech)
     gss_OID_set attrs;
     int reject = 0, p;
 
+    /* Whether we accept an interposer mech depends on whether we accept the
+     * mech it interposes. */
+    mech = gssint_get_public_oid(mech);
+    if (mech == GSS_C_NO_OID)
+        return 0;
+
     status = gss_inquire_attrs_for_mech(&minor, mech, &attrs, NULL);
     if (status)
 	return 0;
--- a/src/lib/gssapi/mechglue/g_mechattr.c
+++ b/src/lib/gssapi/mechglue/g_mechattr.c
@@ -161,6 +161,7 @@ gss_inquire_attrs_for_mech(
 {
     OM_uint32       status, tmpMinor;
     gss_mechanism   mech;
+    gss_OID         selected_mech;
 
     if (minor == NULL)
         return GSS_S_CALL_INACCESSIBLE_WRITE;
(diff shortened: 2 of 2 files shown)


greghudson:
gssint_get_public_oid() can return GSS_C_NO_OID if the oid is not found in the global mech list.  It doesn't look like the right thing happens in that case, in gss_accept_sec_context().  (Even if that is unlikely to happen in practice, it's not very defensive to assume it cannot.)

frozencemetery:
Is there a case where the oid would not be in the global mech list?  I can't immediately think of one, so should we just error out if it's not?

greghudson:
I think it could only happen if the global mech list changes during execution.  If it happens, then yes, we should error out just as we do when allow_mech_by_default() returns false.  I think what happens in the current patch is that we try to figure out whether the default mech is allowed by default.

It may be easiest to change allow_mech_by_default() to operate on either public or interposer OIDs, rather than modify the calling code.

frozencemetery:
You're right, that is easier.  It also makes the failure path clearer, I think.

greghudson:
Shouldn't allow_mech_by_default() return 0 if gssint_get_public_oid() returns GSS_C_NO_OID, not 1?

frozencemetery:
... yes it should.  Sorry about that.

greghudson:
The code changes now look good for what they do, but I'm still trying to work through the design.

First, the attributes for a mech do not depend on credentials, so I don't see a need to ever interpose gss_inquire_attrs_for_mech().  But I'm not 100% sure of the interposer design; are there any other GSS functions that we deliberately don't interpose, and if so, do they need any interposer logic?  (There are other functions which have no interposer logic, such as the ones in g_saslname.c, but I don't know if that's just an oversight.)

Second, if we do allow interposition in gss_inquire_attrs_for_mech(), I'm trying to figure out why the gss_accept_sec_context() change is necessary.  In IRC:

    rharwood: when interposing, gss_inquire_attrs_for_mech gets called from allow_mech_by_default
      using the interposer mech, and we enter the interposer with mech set incorrectly                                                                                  
    rharwood: if the interposer then tries to call back into gssapi, the result is recur until stack depth
      exceeded

My understanding is that if gss_inquire_attrs_for_mech() is called with the interposer OID, and it uses gssint_select_mech_type(), it should enter the real mech, not the interposer, and there should be no room for recursion.  Why is this not the case?

frozencemetery:
Apologies, I was not correct that it can loop forever; that was the other half of the change.

Without the accept_sec_context change, allow_mech_by_default gets called with the interposer mech, which means that so to does gss_inquire_attrs_for_mech.  gss_inquire_attrs_for_mech then calls into the underlying mech (krb5) with mech_oid set to the interposer mech.  Once in krb5, it proceeds to fail the check `g_OID_equal(mech, gss_mech_krb5)` and is marked depricated.

After gss_inquire_attrs_for_mech returns, allow_mech_by_default then checks whether the mech is marked depricated and, finding that it is, rejects it.

greghudson:
That seems like a bug in gss_inquire_attrs_for_mech() as patched by this commit.  Other interposable functions, like gss_init_sec_context(), pass the public OID into the mech.

My first question still stands: does inquire_attrs_for_mech needs to be interposable?

simo5:
I think all functions need to be interposable to stay on the safe side

greghudson:
After discussion with Simo: I will take the first hunk, which I believe will fix the bug by itself.  Making inquire_attrs_for_mech interposable is okay, but should be done along with (at least) the g_saslname.c functions at the same time, and should be done properly, passing the public OID to the mechanism rather than the interposer OID.

greghudson:
Pushed to master as 0b43d10333f4c4b29896cebc9447d8866b661217

PR #371, opened 2015-12-17
Add support for enabling or disabling the pac request with kinit

gladiac:
This adds the function krb5_get_init_creds_opt_set_pac_request() which is already present in Heimdal. This function is used in the Samba source code while requesting tickets. Samba requires often a PAC so we force the request.

It might be useful if the KDC is not serving TCP and the reply over UDP would be too large. So you can disable adding the PAC to the kinit request ...


greghudson:
Can you elaborate on "Samba requires often a PAC"?

gladiac:
If a user logs in on a system joined to AD with Kerberos we want a PAC, because it includes the group memberships etc. So our function which do the Kerberos kinit tell the KDC we want a PAC returned. If the KDC can't send us a back we want a failure reported so we can fall back to Netlogon.

Afaik the KDC might include it but it isn't an error if it isn't returned when it hasn't been explicitly requested.

greghudson:
My understanding (which is not very confident) is that AD KDCs include the PAC unless it is explicitly un-requested (that is, by sending PA-PAC-REQUEST with include-pac false).  Are there real-world situations where KDCs will send a pac only if a positive PA-PAC-REQUEST is included?

This isn't an objection to the change, but I would like to know as much as possible about the need for the functionality.

gladiac:
You are probably right, we have code path which requests it and code path which doesn't request it. I need to dive deeper into the code to find out in which cases we turn in of. However winbind turns it on in case of authentication using kerberos.

gladiac:
The winbind code code requests the PAC using krb5_get_init_creds_opt_set_pac_request() because it requires the groups the user is a member of. And code which expects a PAC sets the request.

'net ads kerberos kinit' requests it and you can dump it with 'net ads kerberos pac'

The Samba client functions disable the PAC request:

* Changing the password with 'net ads password' which uses the kpasswd protocol.
* The client spnego session setup used by most of our client tools
* The ntlm_auth tool using spnego authtication

greghudson:
Sorry for the long delay in reviewing this.  Here are my notes:

* If I read the Heimdal kinit code right, it has long options for requesting a PAC, but not -q/-Q options.  Do you actually need the kinit flags for anything?  I would prefer not to chew up short option space for it, but long option support is still on our todo list.  If it's just about testing, we can do that in other ways.

* The gic_opt accessors will crash if passed a pointer to a non-extended gic_opt structure.  See other accessors for how to check.

* The get_in_tkt.c code will leak memory if allocating req succeeds but request_padata_type() fails to allocate pad or pa.  It's probably best if request_padata_type() copies its argument rather than claiming the memory.

* One gic_opt accessor will do rather than two, I think.

* Representing a tristate as -1 or a boolean seems reasonable given appropriate comments, and is less verbose than an enum.  (However, in the gic_opt setter function, make sure to force the structure field to 0 or 1 by using !!req_pac.)

* A switch statement is a very wordy way of saying "if request_pac isn't unset".

* We put braces around if-else bodies if they are more than one line long, even if they are a single statement.  Also if there are braces around the other arm of the if-else.

gladiac:
We use 'kinit' with the request-pac option in the Samba testsuite for testing. I would prefer to use a long option but they are not supported on LInux.

I think I addressed all your other concerns.

greghudson:
I didn't communicate a couple of points clearly enough:

* I said we put braces around if-else bodies if they exceed one line; you appear to have started putting them around one-line bodies as well, which isn't our usual practice.

* When I said one accessor would suffice, I was imagining one that returns -1, 0, or 1, not a function with two outputs.  (The accessor's prototype in int-proto.h should have a comment explaining the return values.)  You also still appear to have the enum in get_in_tkt.c.

I will talk to Tom about whether we can do long-option support in kinit.  I'd really prefer not to burn two short options.

obnoxxx:
I think every project written in C should mandate braces around one-line if-else blocks.
It not only reads more easily (my personal opinion), but it also protects from severe bugs... (see Apple's 'goto fail'...). Especially important for security-critical code. Imho, it can't afford not to mandate braces - simply too dangerous.
Just giving my 2 cents here ... :-)

gladiac:
I'm sorry for the long delay but FOSDEM is ahead and I needed to prepare for that. I will submit an updated patchset next week. Support for long options in kinit would be great ...

gladiac:
Updated patchset, still waiting for long option support ... :-)

greghudson:
Long option support is now on master.  Can you update this PR to make the new kinit options compatible with Heimdal's, and not to burn kinit -q/-Q?

gladiac:
Here we go :)

greghudson:
Why make the options gratuitously different?

Is there a reason not to use the flags field of the long option structure to simplify the getopt loop?  And if so, why is there a cast in the initializer expression for opt?

gladiac:
I've used --pac-request and --no-pac-request now.

greghudson:
Have you tested this beyond making sure that the kinit options don't crash the client?  [MS-KILE] says that the KERB-PA-PAC-REQUEST should be a DER-encoded ASN.1 sequence; you seem to be sending a single octet.

greghudson:
I've pushed a modified version of the PR branch here: https://github.com/greghudson/krb5/commits/pr371

There are some important differences which will require review:
* krb5_get_init_creds_opt_set_pac_request() matches the Heimdal function signature.
* An encoded KERB-PA-PAC-REQUEST is sent, not a single octet.
* Proper tests are included using the t_authdata.py facilities, in place of the testing commit in the PR.

gladiac:
Thanks for the explanation and modifications.

In the first commit you write:

[ghudson@mit.edu: use a shorter name, add a descriptive comment,
rewrite commit message]

What do you want a shorter name for, add the descriptive comment in the code or in the commit message?

greghudson:
In the code, I used add_padata() as the helper function name, instead of request_padata_type().  I've re-pushed the branch with a hopefully clearer parenthetical statement in that commit message.

gladiac:
Can I help with anything, should I work on the comments you added? I'm currently drowned with badlock ...

greghudson:
Oh, no, those bracketed annotations are to indicate stuff that I changed relative to your commit.  I believe it's a common git convention.  They aren't requests for you to change anything.  All I need at this point is review from someone else on the team, particularly of the final commit which is all new.

simo5:
@greghudson who can make this final review ?

greghudson:
I'd be happy with review from anyone with enough Kerberos experience to have a reasonable chance of spotting bugs.  I'm not as worried about coding practice issues for review of changes I wrote, but pointing out typos or lack of clarity in naming or comments is also valid feedback.

tlyu:
The commits in greghudson/krb5@eebe1ff456d3ef15287231869ffb1fea6f429ad7 (pr371 branch) look ok to me.

greghudson:
Pushed to master as 12cc2b9dab45b05c13642c2b4b0ce0d3191663c7 and the four previous commits.

PR #372, opened 2015-12-17
Make kdc logs not readable by normal users

frozencemetery:
Default umask is 022, so with these changes the expected permissions on the logs becomes 0640 instead of 0644.

### DIFF
--- a/src/lib/kadm5/logger.c
+++ b/src/lib/kadm5/logger.c
@@ -354,7 +354,7 @@ krb5_klog_init(krb5_context kcontext, char *ename, char *whoami, krb5_boolean do
     const char  *logging_profent[3];
     const char  *logging_defent[3];
     char        **logging_specs;
-    int         i, ngood;
+    int         i, ngood, fd, append;
     char        *cp, *cp2;
     char        savec = '\0';
     int         error;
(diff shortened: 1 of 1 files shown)


greghudson:
Using umask() in a library function is inherently not thread-safe.  While we don't call the logging functions from threaded processes now, I don't want to make it harder to do so in the future.

I am also not sure it is the business of the KDC or kadmind code to determine the mode of the log file independent of the process umask.  Are there comparable decisions in other daemons?

frozencemetery:
When things log to syslog, this is left as a decision on the part of the syslogger (Debian sets `$FileCreateMode 0640`, for instance).  I think it's reasonable for kdc/kadmind to ensure that unprivileged users can't read this information, but it's also not unreasonable for us to just set `umask=...` in our unit files if you're opposed to this change.

An example of a daemon doing this is [libvirt](https://libvirt.org/git/?p=libvirt.git;a=blob;f=src/logging/log_daemon.c;h=6da5876c499fb20bfb222d350d2b435503dba224;hb=HEAD#l422) (starting at line 437); their mask is slightly different, though the idea is the same.

tlyu:
Arguably, the umask setting belongs in the individual daemon programs.

nicowilliams:
I agree, don't use `umask()` for this.  Use `open()` then `fdopen()`, or set the umask in the various daemons' `main()` functions.  But note that umask is currently configurable by inheritance; forcing it in `main()` will either make it non-configurable, or will require additional configuration in `kdc.conf`.  Finally, KDCs tend to be (should be!) highly constrained as to who can login, so I don't see much, if any value at all in setting log file permissions to be more restrictive than they are now.

greghudson:
I am okay with using open() with mode 660 and fdopen().  I'm also okay with doing nothing.  I don't think we have cause to muck with the umask either in the library or in daemons.  I'm still curious what the common practice is for daemons which create log files, but I'm not going to run off and do a survey.

frozencemetery:
As requested, this has been updated with the open()+fdopen() approach.  I haven't changed the file mode from the original since I wasn't sure whether you wanted it changed to 0660 or whether that was a typo.  (I'm fine with 0660, 0640, and 0600; I picked 0640 since every other daemon I could find was either 0644 or 0640.)

greghudson:
I'm confused by your note about modes.  I said mode 0660 because it is equivalent to the old version of the patch, which called "umask(old_mask | S_IROTH | S_IWOTH);".  When you use open(), the mode you specify is combined with the process umask, so with a umask of 022 you will of course wind up with mode 0640.

The new version of your patch appears to specify mode 0660, contrary to what you said.  Again, with a process umask of 022 you will wind up with file mode 0640.

Separately, the expression for the second argument to open() is too complicated.  You'll need to introduce some intermediates; I suggest one for (cp[4] == ':') and one for the open flags.  (Note that we do not generally declare variable inside inner scopes.)

frozencemetery:
Good thing I wrote that out then because I was also confused.  Should be 0640 now.

Call to open() has been adjusted.

greghudson:
There's no error-checking on the open() call; if it fails, fdopen() will probably rewrite errno with EINVAL and the log message will contain the wrong reason.  To fix this cleanly, you may need to pull the set_cloexec_file() call and subsequent assignments out of the if-else block.

Also, the way you set up the "append" intermediate variable, I don't think you really need an open_flags intermediate variable, so you can get rid of that.

frozencemetery:
Adjusted error handling and removed open_flags variable.

greghudson:
Pushed to master as 9914b93516bbce9b1123ed5f9f796b7028944892 with changes.  Please have a look at that commit to see what I meant by pulling code out of the if-else block (always look for opportunities to express code at lower levels of indentation).

frozencemetery:
Understood.  The control flow in your version is less clear to me, but your point is taken.

PR #373, opened 2015-12-17
Only run export-check.pl in maintainer mode

tlyu:
Cross compilation environments might not be using the default nm
command.  Rather than try to communicate the correct value to
util/export-check.pl, only run export-check.pl in maintainer mode.

This supersedes PR #271.


greghudson:
I have a mild preference to name the variables using EXPORT_CHECK instead of LINUX_EXPORT_CHECK, since the script is technically dependent on gnu nm rather than Linux.  Other than that, this looks fine.

tlyu:
I agree, even though we only run it on Linux now. It also makes the variable names shorter.

greghudson:
Looks good now.  Having "&&" at the beginning of EXPORT_CMD is a bit ugly, but having a trailing && at the end of LDCOMBINE_TAIL in the non-maintainer case wouldn't work.

PR #374, opened 2015-12-17
Remove util/reconf

tlyu:
The bulk of the util/reconf script is to detect old versions of
autoconf and possibly work around bugs in them.  Require
autoconf>=2.63, and remove the util/reconf script.  Adjust
documentation and other scripts accordingly.

Also remove Makefile rules that remove autom4te.cache.  Such rules
were supposedly necessary when mixing of different versions of
autoconf was common; this seems to be less of an issue now.

This supersedes PR #359.


greghudson:
I think we should remove autom4te.cache in mkrel.  Otherwise this looks good.  (I kind of thought autotools had gotten rid of autom4te.cache; apparently they haven't, which is disappointing.)

tlyu:
I think the existing util/mkrel already removes autom4te.cache; verifying that now.

greghudson:
You're right, it does.  No objections, then.

tlyu:
There is a minor cosmetic flaw: we should use find -depth or else the rm command will cause find to output a spurious error message like "krb5-current/src/autom4te.cache: no such file or directory".

PR #375, opened 2015-12-19
Kadmin set and extract keys with lockdown

simo5:
This patchset adds a number of features around key material in kdamin.
1. A new setkey operation that allows to properly store salts.
2. A new get_principal_keys operation protected by a new ACL_EXTRACT acl to retrieve stored keys
3. A new principal attribute named KRB5_KDB_LOCKDOWN_KEYS that loks down keys and prevent
any change or operation that would allow an admin to retrieve keys or set arbitrary keys including delete or renaming of a principal.

Additionally some tests have been added/modified to check these operations.


greghudson:
I did a brief scan of these commits and they generally look good.  I won't have time to do a careful review for a while because I have a bunch of bugs to work through.  Things I did notice:

* Documentation for the extract ACL needs to be fleshed out, explaining the risks.

* Documentation for the lockdown flag needs to be written.

* You should be able to modify the flags on a locked-down principal as long as you aren't unsetting the lockdown flag.

* There should be more comprehensive tests for the lockdown flag semantics, and they should be in tests/t_kadmin_acl.py rather than the dejagnu tests.  t_kadmin_acl.py should also have tests for the extract acl bit.

* The kadm5_get_principal_keys() change fixes some corner cases in xst -norandkey when there are multiple kvnos in the key data; we should test those fixes.

simo5:
On Sat, 2015-12-19 at 12:00 -0800, Greg Hudson wrote:
> I did a brief scan of these commits and they generally look good.  I
> won't have time to do a careful review for a while because I have a
> bunch of bugs to work through.

Thanks, I will try to work on these soon, but holidays are looming so I
may not be able to get to the more time requiring aspects until new
year.

>   Things I did notice:
> 
> * Documentation for the extract ACL needs to be fleshed out,
> explaining the risks.

I added something to the kamd5.acl (not much), is that the place I
should expand on it ?

> * Documentation for the lockdown flag needs to be written.

Agreed.

> * You should be able to modify the flags on a locked-down principal as
> long as you aren't unsetting the lockdown flag.

This is already done in the modify operation. The operation is denied
exclusively if the incoming request has the lockdown flag turned off and
the principal has it on.

> * There should be more comprehensive tests for the lockdown flag
> semantics, and they should be in tests/t_kadmin_acl.py rather than the
> dejagnu tests.  t_kadmin_acl.py should also have tests for the extract
> acl bit.

Agreed.

> * The kadm5_get_principal_keys() change fixes some corner cases in xst
> -norandkey when there are multiple kvnos in the key data; we should
> test those fixes.

It's not clear to me that anything really changed here, I will take a
deeper look later.

Simo.

-- 
Simo Sorce * Red Hat, Inc * New York

simo5:
Added documentation and tests, pus a fix for the chrand behavior caught by the tests.

doc/admin/conf_files/kadm5_acl.rst
```diff
@@ -47,15 +47,26 @@ ignored.  Lines containing ACL entries have the format::
...
     p  [Dis]allows the propagation of the principal database (used in :ref:`incr_db_prop`)
     s  [Dis]allows the explicit setting of the key for a principal
-    x  Short for admcilsp. All privileges
+    x  Short for admcilsp. All privileges (except ``e``)
     \* Same as x.
     == ======================================================
 
+.. note::
+
+          The ``extract`` privileges is never include in wildcard privileges.
```
nicowilliams: s/include/included/

doc/admin/conf_files/kadm5_acl.rst
```diff
@@ -47,15 +47,26 @@ ignored.  Lines containing ACL entries have the format::
...
+          It must be explicitly assigned. This privilege allows to extract
```
nicowilliams: s/allows/allows the user/

doc/admin/conf_files/kadm5_acl.rst
```diff
@@ -47,15 +47,26 @@ ignored.  Lines containing ACL entries have the format::
...
+          keys from the database and must be handled with great care to avoid
+          disclosure of important keys like those of the kadmin/* or krbtgt/*
+          principals, as that may lead to total compromise of the security of
+          the realm. See the lockdown_keys principal attribute for a method
+          to prevent the extraction of keys of specific principals regradless
```
nicowilliams: s/to prevent/for preventing/

src/kadmin/cli/keytab.c
```diff
@@ -277,11 +266,23 @@ add_principal(void *lhandle, char *keytab_str, krb5_keytab keytab,
...
+                com_err(whoami, ENOMEM, _("while extracting keys"));
+                goto cleanup;
+            }
+            for (i = 0; i < nkeys; i++) {
+                keys[i] = key_data[i].key;
+                memset(&key_data[i].key, 0 , sizeof(krb5_keyblock));
+            }
+            kadm5_free_kadm5_key_data(context, nkeys, key_data);
+        }
+    } else
```
nicowilliams: Braces around multi-line else body missing.

src/kadmin/cli/keytab.c
```diff
@@ -277,11 +266,23 @@ add_principal(void *lhandle, char *keytab_str, krb5_keytab keytab,
...
-    if (norandkey)
-        code = kadm5_get_principal_keys(handle, princ, &keys, &nkeys);
-    else
-#endif
+    if (norandkey) {
+        kadm5_key_data *key_data;
+
+        code = kadm5_get_principal_keys(handle, princ, 0, &key_data, &nkeys);
+        if (code == 0) {
+            keys = malloc(nkeys * sizeof(krb5_keyblock));
```
nicowilliams: Use `calloc()`.

src/kadmin/cli/keytab.c
```diff
@@ -277,11 +266,23 @@ add_principal(void *lhandle, char *keytab_str, krb5_keytab keytab,
...
+            if (!keys) {
+                com_err(whoami, ENOMEM, _("while extracting keys"));
+                goto cleanup;
+            }
+            for (i = 0; i < nkeys; i++) {
+                keys[i] = key_data[i].key;
+                memset(&key_data[i].key, 0 , sizeof(krb5_keyblock));
```
nicowilliams: Use `sizeof(key_data[i].key[0])` or similar.

src/kadmin/server/kadm_rpc_svc.c
```diff
@@ -222,6 +223,18 @@ void kadm_1(rqstp, transp)
 	  local = (char *(*)()) set_string_2_svc;
 	  break;
 
+     case SETKEY_PRINCIPAL4:
+	  xdr_argument = xdr_setkey4_arg;
```
nicowilliams: Does `setkey4_arg` need to be added to the `argument` union?

src/kadmin/server/server_stubs.c
```diff
@@ -1099,6 +1181,94 @@ setkey_principal3_2_svc(setkey3_arg *arg, struct svc_req *rqstp)
...
+
+    free(prime_arg);
+    gss_release_buffer(&minor_stat, &client_name);
+    gss_release_buffer(&minor_stat, &service_name);
+exit_func:
+    free_server_handle(handle);
+    return &ret;
+}
+
+static kadm5_ret_t
```
nicowilliams: A comment would be nice.

    /*
     * Utility function that clears the principal's keys before it's returned
     * to the client if the principal is not allowed to have its keys extracted.
     */

src/kadmin/testing/proto/krb5.conf.proto
```diff
@@ -3,6 +3,7 @@
 	default_keytab_name = FILE:__K5ROOT__/v5srvtab
 	dns_fallback = no
 	plugin_base_dir = __PLUGIN_DIR__
+        allow_weak_crypto = true
```
nicowilliams: Why?

src/lib/kadm5/admin.h
```diff
@@ -455,16 +467,11 @@ krb5_error_code kadm5_init_krb5_context (krb5_context *);
...
- * kadm5_get_principal_keys is used only by kadmin.local to extract existing
- * keys from the database without changing them.  It should never be exposed
- * to the network protocol.
- */
 kadm5_ret_t    kadm5_get_principal_keys(void *server_handle,
                                         krb5_principal principal,
-                                        krb5_keyblock **keyblocks,
-                                        int *n_keys);
-
+                                        krb5_kvno kvno,
```
nicowilliams: You can't do this.  `kadm5_get_principal_keys()` has been in `src/lib/kadm5/srv/libkadm5srv_mit.exports` since 2010.  Wallet and who knows what else might be using this function.  You'll need a new one.

src/lib/kadm5/kadm_rpc.h
```diff
@@ -108,6 +108,15 @@ struct setkey3_arg {
 };
 typedef struct setkey3_arg setkey3_arg;
 
+struct setkey4_arg {
+	krb5_ui_4 api_version;
+	krb5_principal princ;
+	krb5_boolean keepold;
+	kadm5_key_data *key_data;
+	int n_key_data;
+};
```
nicowilliams: Any other arguments we might need so we don't have to add yet another RPC later?

One that I can think of: do/don't change the principal's kvno.

src/lib/kadm5/kadm_rpc.h
```diff
@@ -229,6 +238,21 @@ struct sstring_arg {
 };
 typedef struct sstring_arg sstring_arg;
 
+struct getpkeys_arg {
+	krb5_ui_4 api_version;
+	krb5_principal princ;
+	krb5_kvno kvno;
```
nicowilliams: Is there a magic value that means "all kvnos"?

src/lib/kadm5/clnt/libkadm5clnt_mit.exports
```diff
@@ -22,6 +23,7 @@ kadm5_get_config_params
 kadm5_get_policies
 kadm5_get_policy
 kadm5_get_principal
+kadm5_get_principal_keys
```
nicowilliams: Whatever new name you add for this for `libkadm5srv` (see above comment about ABI stability), use that on the `libkadm5clnt` side too.

src/lib/kadm5/admin.h
nicowilliams: Also, just in case, I believe MIT (and Heimdal) must commit -and implicitly have committed- to a stable API/ABI for kadm5.  (MIT and Heimdal may not have the same kadm5 _ABI_, but a subset of the kadm5 API is source compatible, and within each project, the ABI is mostly stable.)  Regardless of whatever level of API/ABI stability MIT is willing to accord to kadm5, minimizing gratuitous changes will be welcomed by projects like Russ' Wallet and Roland's krb5_admin.

src/lib/kadm5/srv/libkadm5srv_mit.exports
```diff
@@ -19,6 +19,7 @@ kadm5_delete_principal
 kadm5_destroy
 kadm5_flush
 kadm5_free_config_params
+kadm5_free_kadm5_key_data
 kadm5_free_key_data
 kadm5_free_name_list
 kadm5_free_policy_ent
```
nicowilliams: I wish github would let me add comments on lines outside diff hunks.  `kadm5_get_principal_keys` appears below if you expand lines between this and the next hunk.

src/lib/kadm5/kadm_rpc.h
nicowilliams: I strongly encourage you to add that one argument (do/don't change the principal's current kvno).

Also, will it be the case that the `key_data` argument can contain keys of unspecified kvno to mean "add this as the next kvno"?

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -2176,34 +2162,46 @@ kadm5_setkey_principal_3(void *server_handle,
 kadm5_ret_t
 kadm5_get_principal_keys(void *server_handle /* IN */,
                          krb5_principal principal /* IN */,
-                         krb5_keyblock **keyblocks /* OUT */,
-                         int *n_keys /* OUT */)
+                         krb5_kvno kvno /* IN */,
```
nicowilliams: See above commentary about API/ABI stability.

src/lib/kadm5/admin.h
nicowilliams: Lastly, incompatible changes are one thing, but adding incompatible differences between MIT and Heimdal is a bit pain, as application code has to have `#ifdef MIT .. #elif defined(HEIMDAL) ... #endif` nastiness.

greghudson: We have explicitly not committed to a stable API for kadm5; see the comment at the top of admin.h.  We were asked to install the kadmin headers even if we could not guarantee stability.  Having people come back later and say "well, now that you're installing the header, you've implicitly guaranteed API stability" will make us a lot less likely to agree to that kind of compromise in the future.  We have made changes to the API in the past, as in commit 1b8c72b7ef064eabb37d726e831b4618cb37d2c7.

I thought we had a #define in <kadmin/admin.h> so that source files could work with multiple versions of the API, but apparently that's only true for <kdb.h>.  We can add one, although the historically misleading api_version could be confusing.  We definitely need to bump the soname if we modify the API.

I don't see this function in Heimdal, so I'm not sure what you're talking about there.  I also can't find any other uses of this function outside of our source tree in a Google search, although that of course isn't conclusive.

src/kadmin/cli/keytab.c
simo5: The body is not multiline (in C sense), but I'll add braces

src/kadmin/cli/keytab.c
simo5: Why ?

src/kadmin/server/kadm_rpc_svc.c
simo5: Probably, but I noticed there have been other omissions in the past too ...

src/kadmin/testing/proto/krb5.conf.proto
simo5: The old setkey/getkey test I revived uses DES keys, and fails if DES is not allowed ...

src/lib/kadm5/admin.h
simo5: Just let me know how you want to handle it,
I initially proposed a separate kadm5_extract_keys() function, but either way I am fine.

src/lib/kadm5/kadm_rpc.h
simo5: The API has that already built in.
If you pass a kvno of 0 the KDC will generate a new one, if you pass your own KVNO then that will be set.
I do not think a way to change the keys but retain whatever KVNO happens to be there is useful, and for those that may have some contrived reason to do something like that they can simply query the princ to find the current KVNO and then setkey using that KVNO number in the data.

src/lib/kadm5/kadm_rpc.h
simo5: Ah thanks for asking this question, the idea was to be able to specify a kvno or "get the last keys" with a kvno of 0, but it seem I forgot to enforce that and kvno is ignored ...
I'll fix that.

src/lib/kadm5/kadm_rpc.h
greghudson: So, Nico is advocating the addition of an extra flag argument, in anticipation of future support for a kvno tl-data item in principal entries.  The purpose of the anticipated kvno tl-data item is to make the KDC continue serving existing keys for a window of time until we are sure that new keys have been stored in all of the relevant keytabs.  I tentatively agree with Simo that adding this argument is not warranted.  Further thoughts:

* If we decide to add such an argument for setkey, we probably need it for chrand as well, and possibly for chpass.

* I can think of a plausible design which does not need this argument.  To wit: if a kvno tl-data value is present, it remains unchanged by setkey/chpass/chrand when keepold is set, and you just have to update it explicitly afterwards.

simo5: A tl-data argument would have to be changed via modprinc, it is possible to set it in advance of calling setkey4 and then the set keys will not activate until that tl-data is changed again.

nicowilliams: @greghudson Sure, but then you'll need kadmin to follow a chrand/chpass/setkey with a modprinc in order to change the kvno.  I can live with that.  Can you?

nicowilliams: Another thought: don't we want to be able to add/delete keys for existing kvnos?  Heimdal can do this.  setkey4 seems like a good way to add keys for an existing keyset, though maybe not for deleting them.

src/kadmin/cli/keytab.c
nicowilliams: Because it's easier to see that there's no mistake here.  At first I thought `key_data` might not be a krb5_keyblock, and I had to go check.  Using a `sizeof()` expression in a `memset()` call such that the expression in the `sizeof()` is derived from the first argument to `memset()` is a time-honored idiom for avoiding using the wrong size.  You might think "who cares, I got it right in this case", but other people will have to read this code later (making it easier to read the code is a nice courtesy to extend to them), and some might have to change it, and they might screw it up.

src/lib/kadm5/kadm_rpc.h
greghudson: We have a purgekeys API for deleting principals.

We don't have to completely design this kvno feature; the point is that it's entirely reasonable to demand an explicit kvno update after (1) setting an explicit kvno on a principle, and (2) using chrand/chpass/setkey -keepold.

We have a purgekeys API for removing whole kvnos.  When it was developed, we noted at the time that Heimdal could remove kvnos from an existing key set, but Tom elected not to provide that functionality.  My understanding is that in Heimdal it's normal to create keys for a principal and then prune its enctypes of any values you don't want for that service, while in MIT krb5 you normally create the keys with the desired enctypes in the first place.  So we may just not have any use cases for that functionality.

nicowilliams: I imagined that the Heimdal delete individual keys functionality was for 1DES/RC4 migration, but I admit I don't know.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1977,176 +2011,122 @@ kadm5_setkey_principal_3(void *server_handle,
...
     if (keepold) {
-        old_key_data = kdb->key_data;
-        n_old_keys = kdb->n_key_data;
+        n_new_key_data = kdb->n_key_data + n_key_data;
     } else {
-        if (kdb->key_data != NULL)
-            cleanup_key_data(handle->context, kdb->n_key_data, kdb->key_data);
-        n_old_keys = 0;
-        old_key_data = NULL;
+        n_new_key_data = kdb->n_key_data;
```
greghudson: This value assigned should be n_key_data, shouldn't it?

greghudson:
Notes on the setkey_4 semantics:

* It disallows n_key_data == 0, while the current setkey_3 does allow n_keys == 0.  While purgekeys is the normal way to discard all of the keys in a principal, I think we should probably allow n_key_data to be zero.

* I don't think I had realized that we were only going to accept key data with identical kvno values.  If there's not a good reason for this restriction, I'm not sure we should have it.

* When keepold is true, nothing seems to prevent adding keys with the same enctype and salttype as existing keys of the same kvno, even though we are check for duplicates within the keys being added.  I'm actually not sure what restrictions we want here.

* We have a new requirement that key data be sorted in descending order by kvno.  We should call krb5_dbe_sort_key_data() after constructing the new key data to preserve this invariant.

* I notice that chpass/chrand/cprinc sort the requested ks_tuple by the order in the principal's keysalt policy when one is set, while setkey_3 and setkey_4 do not sort the incoming keys that way.  Since it's a pre-existing problem, it's not strictly necessary that we fix it in this PR, but it seems like a shortcoming.

Also, make_ks_from_key_data() seems like the wrong helper function abstraction for this function.  The right abstraction would be one which checks the key data against the keysalt policy by folding in the call to apply_keysalt_policy()--or perhaps one which solves the above problem by re-sorting the incoming key data.

src/lib/kadm5/srv/svr_principal.c
simo5: Yeah looks like it should be.

greghudson:
The "Change kadm5_get_principal_keys" commit is going to a fair amount of effort to throw away the kvno information from the extracted key_data, so that it can work with the remainder of add_principal() which substitutes princ_rec.kvno from the gprinc result.  This obviously fails for "kadmin.local cpw -randkey -keepold princ; kadmin ktadd -norandkey princ" (as does the current code, of course).

With a little bit of restructuring and about the same amount of total code, we could use the correct kvno for the entries we add.

greghudson:
In both of the new RPCs, you omit some _2_ name components which are present in all of the existing RPCs, including ones we recently added.  I can't fully explain why the _2_ name component is there, but I would be more comfortable being consistent.

greghudson:
Relaying decisions from voice conversation:

* It's okay to continue restricting setkey_4 to a single kvno for now.  It effectively pushes complexity onto a hypothetical future client which wants to set multiple kvnos at once.  We can always relax the restriction in the future.

* setkey_4 should reject a request if keepold is true and the requested kvno collides with existing key data.

* setkey_4 does not need to fix the lack of sorting by keysalt policy, as it is not easy to do so.  We can always fix that in the future if it becomes important.

simo5:
NOTE: (based on IRC discussion) n_keys == 0 was never functional (make_ks_from_keys() would return ENOMEM)  so I am going to keep enforcing n_keys != 0

simo5:
I think I have addressed all concerns raised, please take a look.

greghudson:
Pushed to master as b47c99e3fb6c6c41e2f03ce3695c9f945985665f and prior.

greghudson:
Just a note: in the keytab.c changes for the new kadm5_get_principal_keys() signature, I factored out the randkey_princ case into a helper function.  Whenever flow control and memory management starts getting that awkward, it's time to think about creating a helper function.

tlyu:
Update title to link to JIRA.

PR #376, opened 2015-12-21
Remove a bunch of static variables in the kadmin client and server libraries

simo5:
Also fixes a couple memleaks in the client, and removes a to of duplicated code in the server.

This patchset depends (and incorporates until pushed) on the one in PR #375

### DIFF
--- a/src/kadmin/server/kadm_rpc_svc.c
+++ b/src/kadmin/server/kadm_rpc_svc.c
@@ -61,9 +61,20 @@ void kadm_1(rqstp, transp)
 	  setkey4_arg setkey_principal4_2_arg;
 	  getpkeys_arg get_principal_keys_2_arg;
      } argument;
-     char *result;
+     union {
+	  generic_ret gen_ret;
+	  gprinc_ret get_principal_2_ret;
+	  chrand_ret chrand_principal_2_ret;
+	  gpol_ret get_policy_2_ret;
+	  getprivs_ret get_privs_2_ret;
+	  gprincs_ret get_princs_2_ret;
+	  gpols_ret get_pols_2_ret;
+	  chrand_ret chrand_principal3_2_ret;
+	  gstrings_ret get_string_2_ret;
+	  getpkeys_ret get_principal_keys_ret;
+     } result;
      bool_t (*xdr_argument)(), (*xdr_result)();
-     char *(*local)();
+     bool_t (*local)();
 
      if (rqstp->rq_cred.oa_flavor != AUTH_GSSAPI &&
 	 !check_rpcsec_auth(rqstp)) {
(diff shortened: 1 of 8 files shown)


simo5:
Rebased and squashed the commits.
Note: this PR depends on #375 so it includes all its commits until it is pushed, only the last 2 commits are actual code changes belonging to this PR

greghudson:
The recent CVE commits changed the server stubs, so this will need to be rebased.  Also, setup_stub_srv() should return an error if princ_str is non-null but princ is null, as it means the client unexpectedly passed a null principal name.

simo5:
Ok rebased to fix the conflicts

greghudson:
Our practice is to put output parameters at the end; this PR seems to add a whole bunch to the beginning of RPC stub functions.

greghudson:
Also, the SERVER typedef seems weirdly incongruous.  Is there really a problem with continuing to use struct svc_req?

simo5:
I think I used the SERVER typedef to avoid blowing through the 80 col limit mostly.

greghudson:
From voice discussion:

* The SERVER typedef should be undone; it's deceptive because it's a server request object, not a server handle.

* Simo should leave alone the output parameter positioning and stub function renaming for now; Tom and I will figure out whether we want that to change.

simo5:
Rebased on #375 and renamed back SERVER to struct svc_req as requested.

greghudson:
I took a look at the rpcgen state of the art for thread-safe RPC stubs.  If you run rpcgen -M on a .x file, you get client stubs which look like:

```
enum clnt_stat
iprop_get_updates_1(kdb_last_t *argp, kdb_incr_result_t *clnt_res, CLIENT *clnt)
{
        return (clnt_call(clnt, IPROP_GET_UPDATES,
                (xdrproc_t) xdr_kdb_last_t, (caddr_t) argp,
                (xdrproc_t) xdr_kdb_incr_result_t, (caddr_t) clnt_res,
                TIMEOUT));
}
```

and server stubs which look like:

```
bool_t
iprop_get_updates_1_svc(kdb_last_t *argp, kdb_incr_result_t *result, struct svc_req *rqstp)
{
        bool_t retval;

        /*
         * insert server code here
         */

        return retval;
}
```

Perhaps we should follow this convention rather than creating our own?  bool_t and enum clnt_stat are existing RPC types.

simo5:
So your request would be to return a bool_t for the server and invert arg, and result arguments for both client and server calls ?
Or something else ?

greghudson:
In addition to what you said:

* The client stubs should return the result of clnt_call() rather than setting a code in the result structure and returning void.
* The client and server stubs should go back to the names they have in master.

For the boolean return value of the server stubs, it looks like false means not to send a reply, just like returning NULL would mean on master.  I think for our existing stubs that means we always return true, since we never return NULL at present.

simo5:
Do you want to change all return structures to not include a status code, or should we just set it as well as return it from the client stubs ?
Changing all the xx_ret structures is a bigger change than what we have in this PR currently.

simo5:
I think I have addressed the requests (clarified on IRC) with this new patchset.

greghudson:
Pushed to master as 0937ccdd4e97c474bc1b79e059f1301ae6c851ea 01d2bd6ab3f39163f6185ac4c69f6f87535212b4.

PR #377, opened 2015-12-21
Add Github provided OS and IDE files to .gitignore.

SaharahSarah:
It is a good idea to have a fairly complete .gitignore file. It
is helpful for devs to know just what shouldn't end up in the repo,
as well as preventing accidental commits and even random trash
ending up in master.

Many of the files were added from the [github gitignore project.](https://github.com/github/gitignore)


tlyu:
Mostly looks good. It might be better to make the reordering be a separate commit prior to the commit that adds patterns. The OSX pattern "Icon" seems to be "Icon" followed by \r\n\n in the linked repository, but there's no \r in this patch. I'm wary of committing files with unusual line ending characters; is this one safe to leave out?

SaharahSarah:
Leaving out the Icon folder seems like it would be fine. This is just a general list I copied and pasted that seemed more relevant. I'll go ahead and remove separate out the reordering and remove the Icon folder.

tlyu:
Pushed as 6cd1ed92898387e307a2ae777c7473126eddb605.

PR #378, opened 2015-12-22
Add kdc_tcp_ports config option to kdc configuration for kadmin test.

SaharahSarah:
The kadmin rpc tests were starting the kdc without setting the tcp
ports for the kdc. This was then defaulting to port 88, which caused
the socket to fail setting up if the test was not run as root.

### DIFF
--- a/src/kadmin/testing/proto/kdc.conf.proto
+++ b/src/kadmin/testing/proto/kdc.conf.proto
@@ -1,6 +1,6 @@
 [kdcdefaults]
 	kdc_ports = 1750 
-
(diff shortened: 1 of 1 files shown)


greghudson:
Pushed to master as 50ffad3016dd5484656fbb93ad84f849bced2207 with minor changes.

The first line of commit message summaries must be 50 characters or less, and should not end with a period.  We try to follow all of the recommendations in http://chris.beams.io/posts/git-commit/ .  Also, we use two spaces after sentence-ending periods.

PR #379, opened 2015-12-22
Modernize UTF-8 conversions

tlyu:
Unicode and ISO/IEC 10646 now agree that the maximum valid code point
is U+10FFFF.  (Unicode 8.0 Appendix C.2 implies that ISO/IEC
10646:2011 aligned the extent of the code space.)

Adjust src/util/support/utf8.c accordingly, mostly by
deleting code.

Also delete the unused macro KRB5_UCS_UTF8LEN().

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -492,5 +492,6 @@ local.properties
 /src/util/support/t_path
 /src/util/support/t_path_win
 /src/util/support/t_unal
+/src/util/support/t_utf8
 
 /src/util/verto/rename.h
--- a/src/util/support/Makefile.in
+++ b/src/util/support/Makefile.in
@@ -127,7 +127,8 @@ SRCS=\
 	$(srcdir)/base64.c \
 	$(srcdir)/json.c \
 	$(srcdir)/bcmp.c \
-	$(srcdir)/strerror_r.c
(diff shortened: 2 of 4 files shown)


greghudson:
Looks good to me.  I checked the semantics of all of the altered data tables and they all appear to be modified appropriately.

tlyu:
Working on adding test cases and being stricter about the modern U+10FFFF maximum code point.

tlyu:
Updated commits exclude code points above U+10FFFF and add tests.

tlyu:
One small additional change to make utf8_lentab consistent with utf8_mintab. (There is no change in tested behavior.)

greghudson:
t_utf8 needs to be added to SRCS, to the clean rule (along with t_utf8.o), and to the top-level .gitignore file.

My preference is to get rid of the OLDENCODINGS gunk in the test, as I don't see it being useful in the future, but it's a weak preference.

Otherwise this looks good.

tlyu:
Added t_utf8 files to lists. I'd rather keep OLDENCODINGS, or at least remove it in a separate commit. It was useful to have around to verify the consistency of the old implementation with the tests.

greghudson:
This looks good now.

PR #380, opened 2015-12-23
Add configuration options for restricting kadmind and krb5kdc  to certain interfaces. [KRB-50]

SaharahSarah:
krb5kdc and kadmind have always only supported binding to the wildcard
addresses.  2 configuration options have been added that will restrict
krb5kdc and kadmind to only binding to the specified interfaces.
    
The configuration option expects all specified addresses to be local
and be abled to be bound to. If any of them fail then krb5kdc
and kadmind will fail to start. This in turn means that if kdc.conf
is copied from the master to the slaves then it needs to be updated
for each individual slave's interfaces.
    
Technically hostnames can be specified, though this will complicate
things if a DNS query has to be done to find the hostname, but any
issues that would arise from that would be a configuration problem.
    
[ticket: 1093](https://krbdev.mit.edu/rt/Ticket/Display.html?id=1093)
[ticket: 6742](https://krbdev.mit.edu/rt/Ticket/Display.html?id=6742)


greghudson:
The arguments against KRB5_SUCCESS are:

* Unix C programmers are generally familiar with returns-zero-on-success interfaces because that's what most system calls do.  It is also a behavior common to libraries using com_err.

* It is more succinct to check for errors using "if (ret)" (read as "if ret is an error code") than "if (ret != KRB5_SUCCESS)", and we use the former pattern frequently.  For that pattern to make sense, it needs to be implicitly understood that zero means success.

Obviously, these arguments aren't compelling to every API designer, as witnessed by GSS_S_COMPLETE in the GSS-API.  Still, my preference is not to modify the libkrb5 error code model as part of this pull request.

I'm also not really a fan of using doxygen markup when documenting internal interfaces; I prefer just using comments without any markup or formal structure, as it is more concise and more readable in the source code.  I'm not sure whether Tom has the same preference, so this may bear more discussion after the holidays.

These are just my own opinions, and just first impressions; I haven't read carefully over the code changes yet.

SaharahSarah:
"Generally familiar" is not a great way to really program. 0 meaning success is just an assumption in krb5, and isn't even always right. Assumptions are bad. They cause bugs and don't "enforce" a specific coding style. krb5_error_code has no documented success, so a programmer could easily just put "returns -1 on success" but then someone else could see krb5_error_code as the return value and assume 0 means success. There needs to be some unifying return code for krb5_error_code that means success, and KRB5_SUCCESS definitely achieves that. There could be other ways of doing so, such as documenting that '0' means success, but it needs to be visible and consistent.

As for doxygen comments, I have to admit that despite how I feel about the language I come from 5 years as a java developer. One of Java's few good qualities are javadocs. A lot of IDE's have tooltips that show the documentation for a function, typedev, struct, etc when you mouse over them. It's a very handy feature (though, unfortunately, it seems that the IDE I'm currently using doesn't compile the doxygen comments into an html structure like Eclipse would do for the tooltips). Also, one of the frustrating things I've been having to deal with in krb5 is this lack of documentation for every function in a clear, concise, and consistent manner. Doxygen markup might be a bit extra but it does enforce a pattern making the code MUCH easier to read and follow. Having to sit and read through the code on a function just to figure out what all the arguments are is annoying and inefficient.

Not only that, but "internal interfaces" are still public. This is an OSS project so many people will still easily be reading through the code.  Documentation like that will decrease the barrier for new contributers, which is nothing but a good thing, as well as make the code much easier to follow for those that just want to read over it. Honestly, this is the first time I've heard someone say they aren't a fan of documentation...

greghudson:
I would be happy to see a doxygen comment added for krb5_error code noting that 0 means success, that other values are com_err codes, and that krb5_get_error_message() can be used to convert the code to a string.

I did not mean to imply that I was not a fan of documentation in general; I am specifically talking about using doxygen markup and a fixed form.  To me, doxygen markup makes comments incrementally harder to read within the source code; its value is only realized when it is formatted into a human-readable form, as we do with our libkrb5 API documentation.  In some universe, the majority of krb5 developers could be using tools which format and display doxygen comments in a nice human-readable way, but in this universe that is not true as far as I know.

Likewise, adhering to a fixed form when documenting internal functions (such as documenting every parameter in a separate bullet point) tends to lead to a lot of repetition and boilerplate, making it harder to find the important parts.  Concise documentation is more readable for me.

SaharahSarah:
A comment for krb5_error_code would work here as well. I had added the KRB5_SUCCESS for the sole reason that it is the convention I have been used to seeing when working with other projects. I can go ahead and amend that commit with just a comment on krb5_error_code.

Instead of using doxygen comments, what's the best way to specify what arguments are, return codes, error codes, etc.? Once again, in this case, I was writing those based off prior habits and conventions I was used to. I'd be happy to change it to what is considered an easier to read format, but I still think that the same information should be present, especially with detailed descriptions on what the function does, its expectations, any comments on why certain internals were chosen, and argument and return descriptions.

greghudson:
All of that information is valuable, when it isn't obvious or repetitive.  For example, a very large number of our internal functions take krb5_context parameters, so it even briefly describing the context parameter for every function would just contribute to people's eyes glazing over.  For loop_add_udp_port(), the meaning of port is obvious, but the meaning of addresses is not, so we only really have to talk about the latter.  Since several functions use the same string form of addresses, we can talk about it in just one place without repeating ourselves; that's an advantage of not adhering to a tool that only understands per-function documentation comments.

If by "any comments on why certain internals are chosen" you mean discussion of the function's implementation, that material typically belongs inside the function definition, unless it somehow affects the caller.

In general, my preference is to decide what we want to say about an internal function and say it in plain sentences as concisely as possible.  The comment above lib/krad/internal.h:kr_remote_send() is a good example of how a moderately complex internal function can be documented without falling into boilerplate.

SaharahSarah:
Assuming a function argument's purpose can cause a lot of headaches. There's a lot of good articles out there on this. A good quote I found looking it up really quick:

> Argument/parameter names are never self-explanatory. You always think they are when you've spent the past day working on the code, but try coming back to it after a 2-week vacation and you'll see just how unhelpful they really are.

Explaining arguments and their purpose, no matter how trivial, is incredibly helpful. There have been many circumstances where I got upset when an argument didn't have an explanation, even though to the programmers it may have been obvious. Yes, in this case the port number seems obvious to us, but documenting that is very helpful. Also, moving that documentation to another place requires hunting for it when you are just looking at the function itself and want to see what the coders intentions were. It's very simple to just look at which argument you want information on. I fail to see why it's such an issue to have the documentation above each function is such an issue.

greghudson:
In my experience, having a practice of describing every argument to every function yields a lot of descriptions which don't convey any new information.  If we think port is obvious now but some developer finds it non-obvious later, there's likely a subtle point we're not thinking of at the moment--and if we're not thinking about it now, a description like "the port for the UDP socket(s)" probably won't help them.  In krb5.hin we describe every argument because we are generating formatted API documentation, but I'm not sure that the description "Library context" for every context parameter actually helps anyone.  (A doxygen comment for the krb5_context type might help, of course, since the generated documentation is cross-referenced.)

There is a lot of conflicting advice out there on documenting code, including people who say you should avoid it (http://blog.codinghorror.com/coding-without-comments/), which I don't agree with either.  My experience leads me to value conciseness while still providing clarity; other people's experience may lead them to be more verbose.

SaharahSarah:
So, I asked a bunch of developers over the weekend about this problem. I linked them to this pull request and asked their opinion on everything. Everyone I have talked to has said that the comments are wonderful and they see absolutely no reason to change them. The most important person I asked was my mentor through college. In his words:

> Yeah, those comments are entirely appropriate. People who say they aren't are probably too familiar with the code, so they think it's obvious or something, I guess.

greghudson:
The quoted response makes me wonder if my views were clearly understood.  For clarity, here's how I might document those functions as they are currently implemented in the PR:

    /*
     * Configure the loop to include UDP, TCP, or RPC listener sockets on the
     * specified port.  If addresses is null, listener sockets will be bound to all
     * interface addresses or to the wildcard address.  Otherwise, divide addresses
     * into tokens using any of ADDRESSES_DELIM as delimiters; each token will be
     * converted into an address using getaddrinfo(), and listener sockets will be
     * bound to the resulting addresses.
     *                                                                              
     * [XXX describe prognum, versnum, and dispatchfn for loop_add_rpc_service()]   
     */
    krb5_error_code loop_add_udp_port(int port, const char *addresses);
    krb5_error_code loop_add_tcp_port(int port, const char *addresses);
    krb5_error_code loop_add_rpc_service(int port, const char *addresses,
                                         u_long prognum, u_long versnum,
                                         void (*dispatchfn)());

The salient points are: no doxygen markup; document several similar functions at once rather than repeat content; use precision when a parameter is handled in a complex way but only briefly mention a parameter whose meaning is obvious; and waste no space on boilerplate.

We don't need absolute consistency in our practices here, but I at least want us all to be on the same page for whether we use doxygen markup for internal interfaces.  If we do decide to use doxygen markup for this purpose, I would like to have a plan for how we validate and make use of it.  Getting Tom's input would be helpful; if necessary we can discuss this on 1/5.

nicowilliams:
Bikeshedding alert: I don't think the word "restrict" belongs in the new parameters.  Just `kdc_interfaces` and so on should do.

greghudson:
I took a look at bind9 and httpd, and found that they allow the administrator to specify any combination of addresses and ports.  httpd uses a syntax similar to the one used in our "kdc" variable for address/port combinations: https://httpd.apache.org/docs/2.4/bind.html

If I read this pull request correctly, it only allows the administrator to specify a cartesian product of {port1 port2 ...} * {addr1 addr2 ... }.  I don't have a lot of information about the use cases for specifying listener addresses, so I don't know if that would be sufficient.

nicowilliams:
@greghudson Nowadays it's extremely unlikely to run the KDC daemons on non-default ports for any purpose other than testing, thus cartesian product semantics are sufficient.  But! I can't be certain of this, and the `kdc = ` in krb.conf allows for better than cartesian product semantics, so I'm inclined to disfavor cartesian product as the only method here.

nicowilliams:
IMO and FWIW:

Re: comments.  I don't favor extensive doxygen style comments for non-public APIs.  I don't disfavor them either, but my approach when looking for and using these functions is this: RTFS every time rather than RTFM.  My reason for this is simple: comments and code diverge after a while, but there's much more pressure for stability on public APIs, so I may trust their docs but maybe not so much docs on private APIs.

Sometimes code is the best documentation, and this is generally the case for private APIs.  Of course, when the functions in question are huge but the description of what the function does is not, then docs are handy -- not only might they save me the bother of RTFS, but they might also guide the process of RTFS.

Private APIs might be more stable if they were more like public APIs.  The private APIs in question have global state, thus they are not much like public APIs.  Make of that what you will.

In any case, in this case I agree that Doxygen-style commentary on those functions (`loop_add_*()`) is not particularly useful, though knowing whether port information is taken from `addresses` is taken (and whether service names are allowed) *is* important.

As for comments in the code, I've come to dislike comments that restate what code is obviously doing.  Instead I favor comments that help the reader understand things that are not obvious.

Re: Doxygen.  I agree that there should be no need to describe krb5_context arguments, but Doxygen will make you do it...  I don't agree that Doxygen style comments on private interfaces must be validated though, as they should not be extracted, though making sure that they aren't extracted may be tricky (well, no, just don't use `/**`).

Re: error codes.  I don't think we need a KRB5_SUCCESS.  Knowing that 0 is success is not a large cognitive load.

Re: `if (ret)` and `if (!ret)` vs. `if (ret != 0)` and `if (ret == 0)`.  I don't care, but consistency is nice (I am aware of hobgoblins).  Pick a style and stick to it.  Don't just change, as it might mean changing lots of code to be consistent or else being inconsistent.  Consistency is not all that important, but less noise when reading the code -> better; for some inconsistency is noise, and even if that makes them narrow minded, it doesn't mean we should ignore them.  These sorts of preferences should be documented in a style guide.

When I come to a new codebase I _try_ to stick to their preferred styles.  This can be difficult at times, but when I manage it it reduces upstream integration friction.

greghudson:
Proposed configuration, based on conversation today.  Subject to bikeshedding.

* Listener addresses can be specified as *port* or *address* or *address*:*port* or [*address*]:*port* (brackets are necessary for IPv6 addresses); this is very similar to the syntax we use for client KDC configuration.  If no port is given, the default port is used (e.g. 88 for krb5kdc).  If no address is given, the wildcard address is used.

* New variables "kdc_listen" and "kdc_listen_tcp" replace "kdc_ports" and "kdc_tcp_ports"; the old names work as aliases if the new names are not defined.  The values are tokenized and interpreted as listener addresses for UDP and TCP respectively.

* If the new variable "kadmind_listen" is defined, it gives listener addresses for kadmind RPC connections.  If it is not defined, we continue using kadmind_port (which is also used for client configuration).

* If the new variable "kpasswd_listen" is defined, it gives listener addresses for password changes on both TCP and UDP.  If it is not defined, kpasswd_port from the alt_prof realm config is used.

SaharahSarah:
OK, the changes we discussed have been made. kdc_ports and kdc_tcp_ports has been removed, so
configuration files will have to be updated.

greghudson:
kdc_ports and kdc_tcp_ports should not be removed; in my proposal, I said "the old names work as aliases if the new names are not defined."

SaharahSarah:
Missed that.

SaharahSarah:
OK, kdc_ports and kdc_tcp_ports are now aliases and will be checked if kdc_listen or kdc_tcp_listen are not present.

SaharahSarah:
[As discussed in the mailing list](http://mailman.mit.edu/pipermail/krbdev/2016-January/012515.html) the next step will be adding in support for IP_RECVDSTADDR as well as pktinfo.

SaharahSarah:
Latest commit adds IPV4 pktinfo support for BSD. Go ahead and take a look and let me know what you think.

greghudson:
Initial reactions:

* At least for now, I think the pktinfo support should be contained within lib/apputils, not in lib/krb5/os.  The prototypes can go in a new header file in lib/apputils.

* All of the OS conditionalization for pktinfo support should be inside the .c file implementing the pktinfo functions, not exposed to source files consuming the header which prototypes those functions.

* Moving the code from net-server.c to the new file should be a separate commit from adding FreeBSD support.

* I would prefer to see these changes early in the commit series, or in a separate PR, rather than having them go at the end of the listener address support.

greghudson:
Looking over udppktinfo.c in more detail:

* I don't like defining IP_RECVPKTINFO to IP_RECVDSTADDR, as they are not exact equivalents.  Better to change set_pktinfo() to handle them separately.

* On NetBSD, IP_PKTINFO and IP_RECVDSTADDR/IP_SENDSRCADDR are both defined.  We might want to use "#if defined(IP_PKTINFO) / #elif defined(IP_RECVDSTADDR)" blocks to avoid having to think about what happens if we do both.

Other than what I've noted, I don't see any problems given that the new code is a derivative of the current code.  At some point it might be nice to modernize the code and refactor it to avoid preprocessor conditionalization inside functions (see how lib/krb5/ccache/cc_kcm.c conditionalizes on __APPLE__ for an example of how we like to do that).  But having all of that confined to a separate file is good enough for now.

SaharahSarah:
OK, the IP_SENDSRCADDR was moved to [PR 404](https://github.com/krb5/krb5/pull/404).

SaharahSarah:
A request has come in to also add the same support to kpropd.

SaharahSarah:
After some discussion we have decided to postpone kpropd, so it will not be included in this PR.

greghudson:
I just started skimming this PR again and I noticed some common issues.  Please go through and make these changes to each of the commits:

* Limit the summary line to 50 characters, per http://chris.beams.io/posts/git-commit/ and git-commit(1).
* In commit messages, put two spaces after sentence-ending periods.
* Don't add or update copyrights in files where there is less than 20 lines of new code in the changes.  (Don't count changed existing statements as new code.)
* When you add a copyright statement, use the current one from src/prototype/prototype.c, not the old one with the export control statement.

greghudson:
Looking at the second commit, I'm not sure we're on the same page on copyright years and statements yet:

* locate_kdc.c has only mechanical changes to existing lines of code and some deletions, which I don't think amounts to a copyrightable change.

* It hasn't been our practice to change the license statement from the old one to the new one when updating copyright years.  If we want to do that for existing files, my preference is to do it for the whole tree at once rather than incrementally.

* t_locate_kdc.c has only a single partial-line deletion, but has a copyright statement added.

SaharahSarah:
No we are, I honestly don't know why I did that with locate_kdc, but I was already removing it. I could have sworn I had removed the copyright notice from t_locate_kdc... I'll go ahead and fix that.

SaharahSarah:
Should be good now.

greghudson:
If I understand the second commit properly, the only change that affects the KDC process is the removal of port 750 from DEFAULT_KDC_UDP_PORTLIST, and the remainder of the change affects the default ports contacted by the client when contacting a KDC, right?  The commit message gives me the impression that the change is only about the KDC process; perhaps "as well as remove the secondary port from the KDC" could be clarified.

There are some additional port 750 vestiges:

* KDC_SECONDARY_PORTNAME is defined in osconf.hin but no longer used.
* src/config-files/services.append can lose the definitions of kerberos-sec and most (all?) of the comment at the top.
* There are documentation references in krb5kdc.rst, install_kdc.rst, krb5_conf.rst, kdc_conf.rst, and mitK5defaults.rst.
* There is a comment in sendto_kdc.c which uses port 750 as an example; getting rid of the parenthetical seems reasonable.
* config-files/kdc.conf still lists port 750 for kdc_ports.
* lib/krb5/os/td_krb5.conf and util/profile/test.ini use it.  Not sure if any tests would have to be changed if we stopped doing that.

SaharahSarah:
OK, pretty sure I got everything you mentioned.

greghudson:
For the first commit:

* The commit message can just be a one-liner, I think; adding documentation rarely needs justification or explanation.

* Suggested edit to the comment: "Used to conveying an operation status.  The value 0 indicates success;  any other values are com_err codes.  Use krb5_get_error_message() to obtain a string describing the error."

Second commit:

* In the commit message, what does "as well as remove the secondary port from the KDC" refer to?  (I was unclear in my feedback before; I see that you added a sentence about the client change, but I still can't understand what the existing text means.)

* The commit message should note the removal of port 750 from example config fragments.

* In locate_kdc.c, the variable name "i_port" should not use Hungarian notation.  Where you change the line "i_port = htons (l);", get rid of the space before paren.

Third commit:

* Since we're extracting logic from the middle of a function and expanding on its functionality, I think it makes sense to clean up the code.  The style checker finds several issues, and there are also interior variable declarations and implicit null pointer conditionals.  "cleanup:" should not be indented.  The first two conditions in k5_parse_host_string() can be simplified by just returning values since nothing has been allocated yet.  The parameter name "host" should probably change.  There are some Hungarian notation variable names.  k5memdup0() can be used to simplify the copying of the hostname.

* In locate_kdc.c, the call to k5_parse_host_string() could produce success with a null hostname if the string contains just a port.  This needs a check to avoid a crash in add_host_to_list().

* There are bad continuation line indents in k5-int.h and the definition header of k5_parse_host_string().

* k5-int.h should have a blank line after the new prototypes before the comment.

* "[host][[:][port]]" is too loosely specified; it matches "hostport" or "host:" or ":port".  I think the syntax has to be expressed as "host[:port] or port".

* k5_is_string_numeric() does not need to use strlen() before iterating over its input.  It should have braces around the for loop body.

tsitkov:
Consider adding an example for kadmind_listen documentation.

SaharahSarah:
I think I've covered everything...

greghudson:
Third commit: k5_parse_host_string() treats address == NULL the same as an empty string, but there's no contractual guarantee that it is allowed (nor should there be).  It's reasonable to check and return EINVAL, but treating it like a valid argument seems would just conceal bugs.  Similarly, I don't think it's great to make k5_is_string_numeric(NULL) work; I'd rather it crash than return a specific value.

greghudson:
I pushed the first two commits to master (4d02def02a172bdfc5c5c2c9059cef559d8c1feb 624476e0350cde6c37078a808c7b6bceb6046c53).  I made some minor changes, so you'll need to manually exclude them when rebasing.

src/lib/krb5/os/locate_kdc.c
```diff
@@ -259,43 +259,26 @@ locate_srv_conf_1(krb5_context context, const krb5_data *realm,
...
-        /* If the hostname was in brackets, strip those off now. */
-        if (*host == '[' && (cp = strchr(host, ']'))) {
-            host++;
-            *cp = '\0';
+        code = k5_parse_host_string(hostlist[i], default_port, &host,
+                                    &port_num);
+        if (code)
+            goto cleanup;
+        port_num = htons(port_num);
+        if (host != NULL) {
```
greghudson: If this happens (someone wrote "kdc = 88" or similar), it seems better to throw an error than silently ignore the entry.

(The current behavior for "kdc = 88" is to try to send to 0.0.0.0 on port 88, which is clearly wrong, but is better than silently ignoring the entry.  Also, right now our error reporting is bad for invalid KDC addresses--"Invalid argument while getting initial credentials"--but we can fix that separately.)

src/lib/apputils/net-server.c
```diff
@@ -807,20 +799,19 @@ setup_udp_port_1(struct socksetup *data, struct sockaddr *addr, int pktinfo)
...
-                          "%s port %d"), paddr(addr), port);
-                close(sock);
-                return 1;
-            }
+        r = set_pktinfo(sock, addr->sa_family);
+        if (r) {
+            com_err(data->prog, r,
+                    _("Cannot request packet info for udp socket address "
+                        "%s port %d"), paddr(addr), port);
+            krb5_klog_syslog(LOG_INFO, _("System does not support pktinfo "
```
greghudson: I'd make this a LOG_WARNING.  The continuation lines for the string are indented to the wrong open paren; this may format better if you drop to the next line for the string argument, so that the string fragments can be longer.  Also, two spaces after the period.

src/lib/apputils/net-server.c
```diff
@@ -1194,9 +1004,9 @@ process_packet(verto_ctx *ctx, verto_ev *ev)
 
     if (state->daddr_len == 0 && conn->type == CONN_UDP) {
         /*
-         * If the PKTINFO option isn't set, this socket should be bound to a
-         * specific local address.  This info probably should've been saved in
-         * our socket data structure at setup time.
+         * An address couldn't be obtain, so the PKTINFO option probably isn't
```
greghudson: I suggest "A receiving address couldn't be obtained".  (Definite correct "obtain" to "obtained" either way.)  You also have a single space after a period later on in the comment, and I would suggest "If we can't get the address, keep going." for that final sentence.

greghudson:
Okay, looking at the big net-server.c commit:

* It's not our practice to use /* NULL */ designations in function parameters.  (It's important to communicate when a pointer argument may be NULL, but we generally do that in function comments.)

* Where we use enums, we use all-uppercase identifiers for the constants--so UDP, TCP, RPC rather than udp, tcp, rpc.  On the other hand, BIND_FAMILIES and BIND_CONN_TYPES are variable names and should be all-lowercase.  Likewise VERTO_CALLBACK, FD_ADDED_CALLBACK, and any others I missed.

* It doesn't make much sense to say that something is "up to the implementer" in a comment for an internal function, especially for user-visible behavior.  If we allow hostnames in listener addresses now, it won't be safe to reverse that decision later.  I'm okay with allowing hostnames (httpd does as well), but we shouldn't hedge about it.

* In the loop that checks for conflicting addresses in loop_add_address():
** The loop could crash if val.address is NULL but address is not.
** The loop could be deindented by reversing the first conditional and continuing.
** The continue statement at line 338 doesn't do anything.
** The current code is silent about adding redundant ports; the new code would be noisier (at the LOG_DEBUG level, but I think even those go into log files) when running krb5kdc with multiple realms.

* To check for invalid ports, it is more correct to explicitly check the range (port < 0 || port > 65535)  than to check if the value changed after assignment.  C doesn't guarantee that "short" is 16 bytes, and signed vs. unsigned comparisons have tricky semantics.

* In loop_add_addresses(), use strtok_r() rather than strtok().  strtok() can be dangerous even in non-threaded code (the caller could also be using strtok(), for instance).

* There are out-of-scope changes to add_fd(), create_server_socket(), setnbio(), setkeepalive(), and setnolinger(), and no in-scope changes to those functions.  In create_server_socket() the call to set_cloexec_fd() seems to have been removed, which introduces a bug.

* Why does setup_socket() take a callback parameter and call it at the end?  Why can't the caller just call that function after a successful return?

Taking a break; I will need to finish reviewing the important part of the commit later.

SaharahSarah:
> * It's not our practice to use /* NULL */ designations in function parameters. (It's important to communicate when a pointer argument may be NULL, but we generally do that in function comments.)

Removed the /* NULL */ designations.

> * Where we use enums, we use all-uppercase identifiers for the constants--so UDP, TCP, RPC rather than udp, tcp, rpc. On the other hand, BIND_FAMILIES and BIND_CONN_TYPES are variable names and should be all-lowercase. Likewise VERTO_CALLBACK, FD_ADDED_CALLBACK, and any others I missed.

Changed the enum constants to uppercase and the variable names to lower case.

> * It doesn't make much sense to say that something is "up to the implementer" in a comment for an internal function, especially for user-visible behavior. If we allow hostnames in listener addresses now, it won't be safe to reverse that decision later. I'm okay with allowing hostnames (httpd does as well), but we shouldn't hedge about it.

I removed the "up to the implementer" part of the comment

> * In the loop that checks for conflicting addresses in loop_add_address():
>  * The loop could crash if val.address is NULL but address is not.

Fixed. We also check if something is NULL using a full var == NULL and the code was just checking it by equality directly, so I changed it to the correct format.

>  * The loop could be deindented by reversing the first conditional and continuing.

Done.

>  * The continue statement at line 338 doesn't do anything.

Removed.

>  * The current code is silent about adding redundant ports; the new code would be noisier (at the LOG_DEBUG level, but I think even those go into log files) when running krb5kdc with multiple realms.

I'd say that if debugging is turned on, then they're trying to debug something. This would be a relevant message in some case where debugging is necessary, so I kept the call.

> * To check for invalid ports, it is more correct to explicitly check the range (port < 0 || port > 65535)  than to check if the value changed after assignment.  C doesn't guarantee that "short" is 16 bytes, and signed vs. unsigned comparisons have tricky semantics.

This actually was taken from the code that was there beforehand. I changed it anyways.

> * In loop_add_addresses(), use strtok_r() rather than strtok().  strtok() can be dangerous even in non-threaded code (the caller could also be using strtok(), for instance).

Changed strtok to strtok_r.

> * There are out-of-scope changes to add_fd(), create_server_socket(), setnbio(), setkeepalive(), and setnolinger(), and no in-scope changes to those functions.  In create_server_socket() the call to set_cloexec_fd() seems to have been removed, which introduces a bug.

Reverted all the out-of-scope changes.

> * Why does setup_socket() take a callback parameter and call it at the end?  Why can't the caller just call that function after a successful return?

The callback function requires the socket and the verto_ev. setup_socket is in charge of closing the socket and deleting the verto_ev in the case of an error. To have the calling function call the callback function after a successful return the socket and verto_ev would have to be passed back to the caller, who would then have to ensure that in case of an error that they are closed and deleted. IMO the callback function is much simpler and easier than having all the mess that would be required to not have the callback function.

greghudson:
I see that the on_fd_added callback has access to values which the caller does not, but I still find it disturbingly complicated to see a synchronously-invoked callback within the design of an internal function.  Looking harder at the second half of the net-server.c changes, I would suggest moving some intelligence out of setup_addresses() into setup_socket().  It can decide whether to call on_rpc_listener_added() (with a different name) based on ba->type.  The flags argument can also be eliminated and replaced with conditionals based on ba.

I'm not sure if this is codified in our coding style page, but in trinary expressions, it's our practice to put parentheses around the conditional expression unless it's just a variable with a boolean value:

    a = b ? x : y;
    a = (b == c) ? x : y;

In setup_addresses(), I suggest renaming "val" to "addr".

Output from the style-checker as amended by PR #433:

    787ce4a Move logic for parsing host strings into new file
      lib/krb5/krb/parse_host_string.c:
           57  Indentation may be incorrect
    8902640 Remove workaround when pktinfo is unsupported
      lib/apputils/net-server.c:
          806  Indentation may be incorrect
          808  Indentation may be incorrect
          809  Indentation may be incorrect
          810  Indentation may be incorrect
    1980456 Add ability to bind addresses to the net-server
      kadmin/server/ovsec_kadmd.c:
          160  Indentation may be incorrect
          167  Indentation may be incorrect
          168  Indentation may be incorrect
      lib/apputils/net-server.c:
          790  Comment is 3+ lines but is not formatted as block comment
          850  Indentation may be incorrect
          851  Indentation may be incorrect
          852  Indentation may be incorrect
          912  Indentation may be incorrect
    36eff05 Allow user to restrict kadmind bind addresses
      kadmin/server/ovsec_kadmd.c:
          163  Indentation may be incorrect
          572  Trailing blank line
    0378d63 Allow user to restrict KDC to specific addresses
      kdc/main.c:
          270  Indentation may be incorrect

greghudson:
Pushed to master as 5f53d6cfb2cdc2e666a3fd2fe4f3ef21aa8258ae and preceding.

PR #381, opened 2015-12-28
Fix Kerberos Constrained Delegation in GSSAPI to re-use TGS

frenche:
Currently, if we have impersonator credentials we insist on
getting proxy credentials via TGS exchange, but we might as well
already have cached credentials so we need to check for it first.

Signed-off-by: Isaac Boukris <iboukris@gmail.com>

### DIFF
--- a/src/lib/gssapi/krb5/init_sec_context.c
+++ b/src/lib/gssapi/krb5/init_sec_context.c
@@ -118,13 +118,14 @@ int krb5_gss_dbg_client_expcreds = 0;
  * ccache.
  */
 static krb5_error_code get_credentials(context, cred, server, now,
-                                       endtime, out_creds)
+                                       endtime, out_creds, probe)
     krb5_context context;
     krb5_gss_cred_id_t cred;
     krb5_gss_name_t server;
     krb5_timestamp now;
     krb5_timestamp endtime;
     krb5_creds **out_creds;
+    int probe;
 {
     krb5_error_code     code;
     krb5_creds          in_creds, evidence_creds, *result_creds = NULL;
@@ -144,7 +145,11 @@ static krb5_error_code get_credentials(context, cred, server, now,
      * we're not trying to get a ticket to ourselves (in which case
      * we can just use the S4U2Self or evidence ticket directly).
      */
-    if (cred->impersonator &&
+    if (probe) {
+        flags |= KRB5_GC_CACHED;
+        in_creds.client = cred->name->princ;
+    }
+    else if (cred->impersonator &&
         !krb5_principal_compare(context, cred->impersonator, server->princ)) {
         krb5_creds mcreds;
 
@@ -582,8 +587,18 @@ kg_new_connection(
                                   &ctx->there)))
         goto cleanup;
 
-    code = get_credentials(context, cred, ctx->there, now,
-                           ctx->krb_times.endtime, &k_cred);
+    if (cred->impersonator) {
+        code = get_credentials(context, cred, ctx->there, now,
+                           ctx->krb_times.endtime, &k_cred, 1);
+
+        if (code == KRB5_CC_NOTFOUND)
+            code = get_credentials(context, cred, ctx->there, now,
+                               ctx->krb_times.endtime, &k_cred, 0);
+    }
+    else {
+        code = get_credentials(context, cred, ctx->there, now,
+                           ctx->krb_times.endtime, &k_cred, 0);
+    }
     if (code)
         goto cleanup;
 
--- a/src/lib/gssapi/krb5/init_sec_context.c
+++ b/src/lib/gssapi/krb5/init_sec_context.c
@@ -139,37 +139,7 @@ static krb5_error_code get_credentials(context, cred, server, now,
 
     assert(cred->name != NULL);
 
-    /*
-     * Do constrained delegation if we have proxy credentials and
-     * we're not trying to get a ticket to ourselves (in which case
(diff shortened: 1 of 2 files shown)


frenche:
Hi @greghudson - did you get a chance to look at this issue?

Thanks!

greghudson:
Sorry, not yet.  When I do look at this, I will very likely want to push the caching down closer to the layer where it is done for regular TGS requests, rather than adding a probe flag.  I don't currently know exactly what changes that will entail.

The automated test case is very much appreciated.

src/lib/gssapi/krb5/init_sec_context.c
```diff
@@ -118,13 +118,14 @@ int krb5_gss_dbg_client_expcreds = 0;
  * ccache.
  */
 static krb5_error_code get_credentials(context, cred, server, now,
-                                       endtime, out_creds)
```
nicowilliams: IMO, the sooner we start fixing the K&R prototypes to C89+ prototypes, the better...

Maybe instead of `probe` you could pass in `flags`.

src/lib/gssapi/krb5/init_sec_context.c
```diff
@@ -582,8 +587,18 @@ kg_new_connection(
                                   &ctx->there)))
         goto cleanup;
 
-    code = get_credentials(context, cred, ctx->there, now,
-                           ctx->krb_times.endtime, &k_cred);
+    if (cred->impersonator) {
```
nicowilliams: If instead of `probe` you pass in `flags`, then this chunk of code could have just two calls to `get_credentials()`.  You could have a local `flags` initialized to 0 then:

    if (cred->impersonator)
        flags |= KRB5_GC_CACHED;
    code = get_credentials(context, cred, ctx->there, now, ctx->krb_times.endtime, &k_cred, flags);
    if (code == KRB5_CC_NOTFOUND) {
        flags &= ~KRB5_GC_CACHED;
        code = get_credentials(context, cred, ctx->there, now, ctx->krb_times.endtime, &k_cred, flags);
    }
    if (code)
        goto cleanup;

nicowilliams: Actually, on second thought, mutating a local `flags` here is silly, so just:

    if (cred->impersonator)
        code = get_credentials(context, cred, ctx->there, now, ctx->krb_times.endtime, &k_cred, KRB5_GC_CACHED);
    if (code == KRB5_CC_NOTFOUND)
        code = get_credentials(context, cred, ctx->there, now, ctx->krb_times.endtime, &k_cred, 0);
    if (code)
        goto cleanup;

src/lib/gssapi/krb5/init_sec_context.c
frenche: Yea, prototype was weird, first time I seen this :)

frenche:
Thanks for the feedback and comments - I'll try to see if I can improve it.

nicowilliams:
Wait for @greghudson 's say so on the K&R C vs. C89 prototypes.  I don't know that they MIT is ready to do this yet (to help Oracle keep up with diffs).

greghudson:
Don't bother changing the declaration style for the purpose of this patch.  We'll do a sweep at some point.
The important issue for me is handling caching at the best layer.

frenche:
The more I look at it, the more I think, here, is actually the right place for the fix.

See the comment at:
https://github.com/krb5/krb5/blob/master/src/lib/krb5/krb/get_creds.c#L1032

To quote:
    /* For constrained delegation, the expected result is in second_ticket, so
     * we can't really do a cache check here. */
But I we don't seem to check the cache anywhere else.

An older version of that comment is a little more explicit, quote:
    /*See if we already have the ticket cached. To do this usefully
     * for constrained delegation, we would need to look inside
     * second_ticket, which we can't do. */

Looks like constrained delegation won't use cached tickets.
So let's avoid doing constrained delegation unless we really have to.

I've been able to move the changes into get_credentials() and to get rid of the probe.
I added it as a new commit, I'll rebase later on (unless prefered otherwise).

frenche:
Hi again,

I had another look and I think the main problem is because once we call krb5_get_credentials() we don't know what is the client name (we only know the impersonator) so we can't construct matching credentials to lookup a cached ticket.

If we could pass *both* client name and impersonator name, let's say by adding impersonator name to krb5_creds, then we could wait till after we call check_cache() before we switch to impersonator (+2ed ticket), but I doubt this would be desirable.

So I still think that trying cache-only before starting constrained-delegation (as suggested) may be not such a bad idea after all.

Thanks!

frenche:
One more note, since we pass the ccache to krb5_get_credentials() so perhaps we still could get the client name later with krb5_cc_get_principal() but this does not seem right either.

greghudson:
Once again, sorry for the slow response time.  I like the approach of the current commit much better, and will try to take a closer look and integrate it soon.

greghudson:
Just a note: in the tests you added, you are passing buf.value as a C string parameter to check_ticket_count(), where buf is the result of gss_display_name().  GSS-API buffers are counted strings; there is no guarantee that they have a trailing zero byte.  I will fix this up, but there is a whole class of security bugs at the interface between counted and zero-terminated strings, so I thought it was worth mentioning.

greghudson:
Pushed to master as f149b9cc1e0f8c6e7cca2ee0a5fd8feff7deaf58

frenche:
Thanks a lot for working with me on this!

PR #382, opened 2015-12-29
Sanitize context pointer in gss_export_sec_context

tkuthan:
After 4f35b27a context pointer in gss_export_sec_context() is first
dereferenced before arguments are sanitized in val_exp_sec_ctx_args().
With context == NULL the new code segfaults instead of failing
gracefully.

Revert this part of 4f35b27a and only dereference context if not NULL.

### DIFF
--- a/src/lib/gssapi/mechglue/g_exp_sec_context.c
+++ b/src/lib/gssapi/mechglue/g_exp_sec_context.c
@@ -79,7 +79,7 @@ gss_buffer_t		interprocess_token;
 {
     OM_uint32		status;
     OM_uint32 		length;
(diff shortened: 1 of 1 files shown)


greghudson:
This looks correct.  However, we generally try to make sure that the contents of cleanup labels are always safe, in case more jump points to them are handled later.  So, please initialize ctx to NULL (not GSS_C_NO_CONTEXT because it isn't an abstract context handle) and, in the cleanup handler, add a "ctx != NULL" clause to the conditional.

tkuthan:
Done.

greghudson:
Pushed to master as b6f29cbd2ab132e336b5435447348400e9a9e241

PR #383, opened 2015-12-30
Replace MD5 use in rcache with SHA1

tkuthan:
rcache uses an unkeyed MD5 hash of the authenticator to distinguish
between different request with equal client principal, server principal
and microsecond time. When OpenSSL crypto provider is used and
underlying OpenSSL is run in FIPS mode, MD5 algorithm is disabled and
gss_accept_sec_context() results in an abort in rcache processing.

This change effectively implements a different rcache extension.
The new extension identifier is 'SHA1:' (instead of 'HASH:')
and the checksum type is CKSUMTYPE_NIST_SHA (instead of CKSUMTYPE_RSA_MD5).

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -623,6 +623,12 @@ krb5int_arcfour_gsscrypt(const krb5_keyblock *keyblock, krb5_keyusage usage,
                          const krb5_data *kd_data, krb5_crypto_iov *data,
                          size_t num_data);
 
+#define K5_SHA256_HASHLEN (256/8)
+
+/* Write the SHA-256 hash of in to out. */
+krb5_error_code
+k5_sha256(const krb5_data *in, uint8_t out[K5_SHA256_HASHLEN]);
+
 /*
  * Attempt to zero memory in a way that compilers won't optimize out.
  *
--- a/src/lib/crypto/builtin/sha2/sha256.c
+++ b/src/lib/crypto/builtin/sha2/sha256.c
@@ -255,3 +255,14 @@ k5_sha256_final(void *res, SHA256_CTX *m)
 	}
     }
 }
+
+krb5_error_code
+k5_sha256(const krb5_data *in, uint8_t out[K5_SHA256_HASHLEN])
+{
+    SHA256_CTX ctx;
+
+    k5_sha256_init(&ctx);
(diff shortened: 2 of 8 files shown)


greghudson:
I filed a request with IANA to register unkeyed SHA-2 checksum types, and am waiting to hear back.  (The registry requires expert review, and Ken Raeburn is the expert; although there's very little to review in this case, it is the holidays.)  I'm still hoping to avoid migrating to another dying hash function.

tkuthan:
Thank you for the update. I didn't imagine, that SHA-2 family based checksum could be available in the short-term. 
I'll be happy to update this patch to use modern hash function, but I wouldn't mind this being solved separately and this pull request being closed either. Whatever is more convenient for you.
Thanks.

greghudson:
If the checksum type registration hasn't gone through after a while, I can also add a tree-private libk5crypto API (akin to krb5int_arcfour_gsscrypt()) for SHA-256 and we can use that.

greghudson:
After feedback from Ken, I dropped my IANA registration request, on the belief that the checksum number space should not be used for any new unkeyed checksums.

I still want to avoid migrating to another dead hash, so I will define a tree-internal libk5crypto interface for SHA-256 and we can use that from the replay cache code.  It may be a while before I can get to it.

tkuthan:
By 'tree-internal libk5crypto interface'  you mean that the SHA-256 code would be directly in MIT krb5 source code and would not be part of crypto plugin SPI?
This approach would have performance corollaries on platforms that have hardware accelerated SHA-256...

greghudson:
No, I mean we wouldn't be providing the API to consumers of libkrb5 or libk5crypto, only to code within the tree.  The implementation would be in the crypto back-end.

tkuthan:
Oh good. Thanks for explaining that to me.

greghudson:
Okay, please cherry-pick the commit from PR #392 (should be commit 0ecdbe9aafc2e9914ee82890048ee7c26f2a7be1 from the main github repository) and redo this patch using k5_sha256() to compute the hash.

tkuthan:
Rewritten using k5_sha256.
Sorry for the delay, I have been working on a different technology...

greghudson:
Thanks.  Pushed to master as c546a30c7c9299a419f757768a3349bde09c9cd4.

PR #385, opened 2016-01-05
Check internal context in init_context mechglue call on error.

simo5:
Seen crashes in the wild without this check as some code seem to end up
calling gss_inquire_context via the spnego mech before while the internal
context is still NULL.

### DIFF
--- a/src/lib/gssapi/mechglue/g_init_sec_context.c
+++ b/src/lib/gssapi/mechglue/g_init_sec_context.c
@@ -228,8 +228,13 @@ OM_uint32 *		time_rec;
 	 * context info on the first call to init, and on all
 	 * subsequent calls make the caller responsible for
 	 * calling gss_delete_sec_context
+	 * However if the mechanism decided to null the internal context
+	 * we should also null the union context to avoid dereferencing
+	 * a null pointer in other gssapi calls
 	 */
 	map_error(minor_status, mech);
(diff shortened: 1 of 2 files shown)


greghudson:
I need a better understanding of the bug here.  I didn't think it was possible to create a mechglue context with a NULL internal_ctx_id.

simo5:
On Tue, 2016-01-05 at 09:36 -0800, Greg Hudson wrote:
> I need a better understanding of the bug here.  I didn't think it was possible to create a mechglue context with a NULL internal_ctx_id.

This is all I have for now:
https://bugzilla.redhat.com/show_bug.cgi?id=1295893

greghudson:
Simo and I believe we worked out what must have happened: on a second call to gss_init_sec_context(), SPNEGO gets an error and deletes its SPNEGO context, but the mechglue preserves the union context with a NULL internal_ctx_id pointer.  I would like to see gss_init_sec_context() delete and null out the union context in this circumstance.  gss_accept_sec_context() might need a similar change.

simo5:
Changed as requested, code now checks for internal_ctx_id on errors and clears the whole union context. Accept_sec_context already clears it all unconditioanlly on errors.

greghudson:
Pushed to master as 3beb564cea3d219efcf71682b6576cad548c2d23

PR #386, opened 2016-01-06
Work around uninitialized warning in cc_kcm.c

tlyu:
Some versions of clang erroneously detect use of an uninitialized
variable reply_len in kcmio_call() when building on non-Mac platforms.
Initialize it to work around this warning.

ticket: 8335 (new)


greghudson:
This looks okay to me.

PR #387, opened 2016-01-07
MAXHOSTNAMELEN is too short for some FQDNs

tlyu:
Testing in Travis CI reveals that sim_client and iprop don't deal well with FQDNs longer than MAXHOSTNAMELEN.

### DIFF
--- a/src/appl/simple/client/sim_client.c
+++ b/src/appl/simple/client/sim_client.c
@@ -67,7 +67,6 @@ main(int argc, char *argv[])
     struct servent *serv;
     struct hostent *host;
     char *cp;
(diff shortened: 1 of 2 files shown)


greghudson:
For the first commit, it's true that the test suite will pass in a canonicalized hostname, but more importantly, krb5_mk_req() will canonicalize the hostname passed to it by sim_client, making it unnecessary for sim_client to canonicalize the hostname even if it were used in some other way.  The code changes look good.

The second commit looks fine.

tlyu:
Rebased and adjusted commit message on the first commit.

greghudson:
The first commit message looks good now, so I think this PR is ready to push.

PR #388, opened 2016-01-07
Add .travis.yml [KRB-42]

tlyu:
Do Travis CI testing with clang and gcc, on 64-bit Ubuntu Trusty.
Performance would probably be better using the container-based Travis
infrastructure, but that is currently limited to Precise, and we would
need some important apt packages whitelisted, e.g., dejagnu.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: c
+
+sudo: required
+
+dist: trusty
+
+compiler:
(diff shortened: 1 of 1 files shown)


tlyu:
Linking JIRA issue KRB-42.

greghudson:
Looks okay to me.

PR #389, opened 2016-01-08
Make ksu work with prompting clpreauth modules

greghudson:
Commit 5fd5a67c5a93514e7d0a64425baa007ad91f57de switched ksu from
using krb5_get_in_tkt_with_password() to
krb5_get_init_creds_password(), but did not supply a prompter
argument.  Pass krb5_prompter_posix so that clpreauth modules can
prompt for additional information during authentication.


tlyu:
This looks OK to me.

PR #390, opened 2016-01-12
Remove nss libk5crypto implementation

greghudson:
Delete the nss libk5crypto implementation because: (1) the original
contributor no longer requires it; (2) as far as we know it is rarely
or never used; (3) NSS has significant impedance mismatch with MIT
krb5's requirements, such as the inability to preserve crypto state
across fork(); (4) NSS cannot do PBKDF2 with SHA-256 or SHA-384, as
will be required for the aes-sha2 enctypes.


tlyu:
This looks OK to me.

tsitkov:
You need to update the documentation http://web.mit.edu/kerberos/krb5-devel/doc/mitK5features.html#quick-facts. In "Crypto backends" subsection one should mention something in lines with  NSS release <1.15.  Of course, this can be done at the later stage, but it is easy to forget.

greghudson:
The PR already changes the crypto backends subsection to avoid mentioning NSS at all.  I don't think it's important to state that it was there in previous releases, since we don't think anyone is using it.

PR #391, opened 2016-01-13
Expand interposer and some fixups

frozencemetery:
This is the continuation of #370.  gss_inquire_names_for_mech() should now be interposed correctly, and gss_inquire_attrs_for_mech() and gss_inquire_saslname_for_mech() are newly interposable.  As gss_inquire_mech_for_saslname() cannot be made interposable per rfc5801, this is all of the functions in g_saslname.c as requested.

Thanks!

### DIFF
--- a/src/lib/gssapi/mechglue/g_inq_cred.c
+++ b/src/lib/gssapi/mechglue/g_inq_cred.c
@@ -169,7 +169,7 @@ gss_inquire_cred_by_mech(minor_status, cred_handle, mech_type, name,
     gss_mechanism	mech;
     OM_uint32		status, temp_minor_status;
     gss_name_t		internal_name;
-    gss_OID		selected_mech;
+    gss_OID		selected_mech, public_mech;
 
     if (minor_status != NULL)
 	*minor_status = 0;
@@ -198,8 +198,9 @@ gss_inquire_cred_by_mech(minor_status, cred_handle, mech_type, name,
 	return (GSS_S_DEFECTIVE_CREDENTIAL);
 #endif
 
+    public_mech = gssint_get_public_oid(selected_mech);
     status = mech->gss_inquire_cred_by_mech(minor_status,
-					    mech_cred, selected_mech,
+					    mech_cred, public_mech,
 					    name ? &internal_name : NULL,
 					    initiator_lifetime,
 					    acceptor_lifetime, cred_usage);
--- a/src/lib/gssapi/mechglue/g_inq_names.c
+++ b/src/lib/gssapi/mechglue/g_inq_names.c
@@ -40,7 +40,7 @@ gss_OID_set *	name_types;
 
 {
     OM_uint32		status;
-    gss_OID		selected_mech = GSS_C_NO_OID;
+    gss_OID		selected_mech = GSS_C_NO_OID, public_mech;
     gss_mechanism	mech;
 
     /* Initialize outputs. */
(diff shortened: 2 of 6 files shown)


greghudson:
Did you check for other non-interposable mechglue functions?  Is there a reason why the third commit is separate from the first?

frozencemetery:
I went through the other functions and didn't immediately see anything, though my interposer tests are not done yet so I cannot say for certain.

The first and third commits are separate because in my head they are doing fundamentally different things (enabling something to work vs. fixing something that looked like it should have worked) to different functions.  I am not attached to this separation and if you prefer it to be different I can make it so.

greghudson:
Oh, I see now that they do two separate things.  The third commit needs more than just a one-line explanation to explain what it is fixing, though.  I think you also need to fix gss_inquire_cred_by_mech(), gss_store_cred(), and gss_localname().

I also wonder if &mech->mech_type would work as an alternative to the more verbose gssint_get_public_oid(selected_mech).  Of course then we'd want to change everything to match.

greghudson:
The first commit changes the behavior when mech is NULL or does not provide gss_inquire_attrs_for_mech().  I think the behavior change might be desirable; returning no attributes but a base set of known attributes doesn't seem likely to produce correct application behavior.  But the change should perhaps be in a separate commit, and you absolutely need more than a one-line commit message when you do this sort of thing.  Oversimplified commit messages has been a common theme in recent PRs from you; please have a look at some commits to the master branch to see the level of detail we aim for.

Also, you should de-indent the call to mech->function by doing:

    if (mech == NULL)
        return ...
    if (mech->function == NULL)
        return ...
    status = mech->function(...)

We never like to see if (expected) { do mainline logic } else { return error; } flow control.

In the second commit, the behavior change when the mech does not provide gss_inquire_saslname_for_mech() seems undesirable; it defeats the purpose of the oidToSaslNameAlloc() fallback.  You should remove the initialization of status to GSS_S_BAD_MECH; it is inoperative after the changes.  The commit message should explain why gss_inquire_mech_for_saslname() cannot be interposed; just having the flat assertion there is confusing.  (The explanation of "per rfc5801" in the PR description is too vague.)

frozencemetery:
New version.  Thanks for pointing out the additional functions.

There is already existing code with the flow control you criticize; if it is such a problem, it may be good to remove it.

greghudson:
In my previous comment, I wrote, "In the second commit, the behavior change when the mech does not provide gss_inquire_saslname_for_mech() seems undesirable; it defeats the purpose of the oidToSaslNameAlloc() fallback."  Did you address this feedback?  The behavior change still seems to be present.

I don't like the new formatting of the calls into mech->function(), with the argument list dropped down to the next line and overwrapped to one argument per line.  I am increasingly dissatisfied with "gssint_get_public_oid(desired_mech)" as an argument expression in a call to a function with a (usually) really long name.  I think we should either switch to &mech->mech_type everywhere, if that will work, or add a public_mech local variable everywhere.

frozencemetery:
New version.  Fallback actually restored this time.

`&mech->mech_type` is not equivalent to a call to `gssint_get_public_oid(desired_mech)`.  I've added a local variable where needed for wrapping.

greghudson:
Can you explain how the two aren't equivalent?

greghudson:
In the second commit ("Report interposer failures..."), the commit message seems wrong.  We're changing the behavior if the caller's mech oid doesn't resolve to a mechanism, or if the mechanism has no gss_inquire_attrs_for_mech() method.  These cases have nothing to do with the interposer.

In the third commit ("Enable interposing gss_inquire_saslname_for_mech()"), I think the fallback is only restored in an uninteresting case.  If the mech is unsupported, RFC 5801 says we can return GSS_S_BAD_MECH and I think it's fine to do so.  If the mech is supported but has no gss_inquire_saslname_for_mech() method, then we want to do the OID conversion (see RFC 5801 section 3.1); GSS_S_UNAVAILABLE isn't even a specified return code for this function.

frozencemetery:
Fixed commit message on second commit, and I believe also fixed the fallback in the third commit.

As for `&mech->mech_type`: I checked with Simo, and I think it is the case that if `mech` is an interposer, then `&mech->mech_type` will be the interposed OID, not the public OID.  I believe this is why we skip interposer mechs in [gssint_get_public_oid](https://github.com/krb5/krb5/blob/master/src/lib/gssapi/mechglue/g_initialize.c#L1008-L1009).

greghudson:
That makes sense; thanks for explaining.

greghudson:
Pushed to master as 3be2b486058758cfcd16c8af0a8f560159e77cda 030a4a03a0480969d6acf1591f39fd194642805a 92dbcf2eb436933f769c17e6a10f671992636e5f fe73f1130695880bd83cf811c37131b12711be23.  (I squashed the last two commits.)

PR #392, opened 2016-01-13
Add internal SHA-256 interface

greghudson:
Add a simple private libk5crypto interface for computing SHA-256
hashes.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -623,6 +623,12 @@ krb5int_arcfour_gsscrypt(const krb5_keyblock *keyblock, krb5_keyusage usage,
                          const krb5_data *kd_data, krb5_crypto_iov *data,
                          size_t num_data);
 
+#define K5_SHA256_HASHLEN (256 / 8)
+
+/* Write the SHA-256 hash of in to out. */
+krb5_error_code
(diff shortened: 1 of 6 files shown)


tlyu:
This looks mostly OK to me. There's no test coverage for the new interface, but we do have testing for the underlying interfaces (init/update/fini) that it uses. I don't have a strong opinion on whether to add a test specifically for the new interface. What do you think?

greghudson:
I think it's okay to leave that specific interface untested for now.  (Obviously it will get some coverage testing when the rcache code is changed to use it, but that won't test for correct SHA-256 results.)  This code will also be affected by the aes-sha2 enctype work (see https://github.com/greghudson/krb5/commit/aba206c654b6b4292ec7d9866ff2cad1b295cda8).  We could conceivably, in that commit, put t_sha2.c in crypto_tests rather than builtin/sha2 and test the k5_sha256() interface.

tlyu:
That's tempting, but we probably should avoid unrolling the "one million a's" test if we're going to do that.

greghudson:
Does the test program in https://github.com/greghudson/krb5/commit/eea0bd75432254317af035d701067db804d23bbe look reasonable?  It uses the hash_provider interface with a thousand IOVs for the "one million a's" test, which results in a thousand k5_sha256_update() calls in the builtin module.

PR #395, opened 2016-01-14
Work around uninitialized warning in logger.c

greghudson:
gcc 4.6.3 erroneously detects uninitialized use of the variable f
after commit 9914b93516bbce9b1123ed5f9f796b7028944892.  Initialize it
to work around this warning.


tlyu:
This looks OK to me.

PR #396, opened 2016-01-14
Fix iprop server stub error management

greghudson:
The ipropd stubs free client_name and server_name in the cleanup
handler, so should not free them in out-of-memory conditions.
Reported by Will Fiveash.


kaduk:
Looks good to me.

PR #397, opened 2016-01-14
Fix ksetpwd password reading loop

greghudson:
In ksetpwd (which we do not install), fix the loop which reads the new
password twice until they match.  Previously it would stop with a
dangling pointer to freed memory in new_password if they don't match
on the first try.  Reported by Will Fiveash.

### DIFF
--- a/src/clients/kpasswd/ksetpwd.c
+++ b/src/clients/kpasswd/ksetpwd.c
@@ -227,7 +227,7 @@ static int init_creds()
 
 int main( int argc, char ** argv )
 {
-    char * new_password = NULL;
+    char * new_password;
     char * new_password2;
     krb5_context    kcontext;
     krb5_error_code kerr;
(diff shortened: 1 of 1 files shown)


src/clients/kpasswd/ksetpwd.c
```diff
@@ -266,17 +266,15 @@ int main( int argc, char ** argv )
...
         new_password = getpass("Enter new password: ");
         new_password2 = getpass("Verify new password: ");
-        if( strcmp( new_password, new_password2 ) )
-        {
-            printf("Passwords do not match\n");
-            free( new_password );
-            free( new_password2 );
-            continue;
-        }
+        if( strcmp( new_password, new_password2 ) == 0)
```
kaduk: I guess we're not supposed to tighten up the parens since that's prevailing style in this file (ugh)?

kaduk:
This looks good to me.

PR #398, opened 2016-01-18
Fix ticket renewal of enterprise principals

gladiac:
While writing tests for Samba AD using the MIT KDC I discover that it isn't possible to renew tickets with enterprise principals. With the attached patch I can successfully do:

kinit -E user@ENTERPRISE.REALM
kinit -R

I was not able to get the test working in 'make check' of the MIT KRB5 source code. I'm posting it so maybe you can tell me if this can be tested or not.

### DIFF
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -418,6 +418,7 @@ k5_pac_validate_client(krb5_context context,
     krb5_ui_2 pac_princname_length;
     int64_t pac_nt_authtime;
     krb5_principal pac_principal;
+    int flags;
 
     ret = k5_pac_locate_buffer(context, pac, KRB5_PAC_CLIENT_INFO,
                                &client_info);
@@ -446,8 +447,13 @@ k5_pac_validate_client(krb5_context context,
     if (ret != 0)
         return ret;
 
+    flags = KRB5_PRINCIPAL_PARSE_NO_REALM;
+    if (principal->type == KRB5_NT_ENTERPRISE_PRINCIPAL) {
+        flags |= KRB5_PRINCIPAL_PARSE_ENTERPRISE;
+    }
+
     ret = krb5_parse_name_flags(context, pac_princname,
-                                KRB5_PRINCIPAL_PARSE_NO_REALM, &pac_principal);
+                                flags, &pac_principal);
     if (ret != 0) {
         free(pac_princname);
         return ret;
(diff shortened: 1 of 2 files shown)


greghudson:
I need a lot more information.  For starters, what exactly goes wrong without this patch?

gladiac:
I have a patchset which implements a Samba kdb driver to access our database. The request comes in and the KDC calls the sign_authdata function form the kdb driver:

kdb_samba_db_sign_auth_data() [1] -> as it is not a AS req we call ks_verify_pac() [1] -> krb5_pac_verify() -> k5_pac_validate_client() -> krb5_parse_name_flags(KRB5_PRINCIPAL_PARSE_NO_REALM)

krb5_parse_name_flags() fails in line 214:

ret = KRB5_PARSE_MALFORMED;
k5_setmsg(context, ret, _("Principal %s has realm present"), name);

We call allocate_princ() with enterpise set to 0 as we do not pass the information to krb5_parse_name_flags() that we deal with an enterprise principal.

---

The error returned from the KDC is in process_tgs_req() after calling handle_authdata():

krb5_klog_syslog(LOG_INFO, _("TGS_REQ : handle_authdata (%d)"), errcode);
status = "HANDLE_AUTHDATA";


[1] https://git.samba.org/?p=asn/samba.git;a=blob;f=source4/kdc/mit-kdb/kdb_samba_policies.c;h=189a9296227bc7d2227129295c14aaa09d519e42;hb=refs/heads/master-mit-kdc

gladiac:
As described above. I'm on an AD server using MIT KDC with a Samba kdb driver, to reproduce the issue I do the following:

kinit -E user@ENTERPRISE.REALM
kinit -R

greghudson:
Okay.  The test case you added (aside from failing because it tries to renew non-renewable tickets) doesn't exercise the code you changed; I think the only way to test this would be to write C code to create a PAC and call krb5_pac_verify().

While I understand better what's going wrong, I would like to better understand why there is an unparsed enterprise principal name in the PAC_CLIENT_INFO buffer of the PAC.  My understanding of enterprise principal names is that they are mapped onto real principal names during an AS-REQ, so I would expect there to be a real principal name inside the PAC.  Is Samba's behavior consistent with Microsoft's in this respect?  If it were common to see enterprise principal names in a PAC buffer, I would have expected Luke's code to allow for it.

gladiac:
You're right, it was a bug in the samba kdb driver which did not handle AS requests correctly ... Sorry for the noise.

PR #399, opened 2016-01-26
Implement password history in LDAP plugin [KRB-49]

SaharahSarah:
The password history is stored in the kerberos LDAP schema attribute 'krbPwdHistory', with one history entry per attribute.  When the history is decoded the history entries are sorted by kvno in descending order, with the next replacement key set to the last key entry, or the key history entry with the smallest kvno.

This commit is based on code commited by Tomas Kuthan <tkuthan@gmail.com>.
 
ticket: [5889](https://krbdev.mit.edu/rt/Ticket/Display.html?id=5889)

### DIFF
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -1,6 +1,6 @@
 /* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 /*
- * Copyright 1990,1991 by the Massachusetts Institute of Technology.
+ * Copyright 1990, 1991, 2016 by the Massachusetts Institute of Technology.
  * All Rights Reserved.
  *
  * Export of this software from the United States of America may
@@ -209,6 +209,8 @@ typedef struct _krb5_db_entry_new {
 
     krb5_principal        princ;                /* Length, data */
     krb5_tl_data        * tl_data;              /* Linked list */
+
+    /* key_data must be sorted by kvno in descending order */
     krb5_key_data       * key_data;             /* Array */
 } krb5_db_entry;
 
@@ -682,6 +684,18 @@ krb5_error_code krb5_db_check_allowed_to_delegate(krb5_context kcontext,
                                                   krb5_const_principal client,
                                                   const krb5_db_entry *server,
                                                   krb5_const_principal proxy);
+
+/**
+ * Sort an array of @a krb5_key_data keys in descending order by their kvno.
+ *
+ * @param key_data
+ *     The @a krb5_key_data array to sort.  This is sorted in place so the
+ *     array will be modified. Key data order within a kvno is preserved.
+ * @param key_data_length
+ *     The legnth of @a key_data.
+ */
+void
+krb5_dbe_sort_key_data(krb5_key_data *key_data, size_t key_data_length);
 
 /* default functions. Should not be directly called */
 /*
--- a/src/lib/kadm5/srv/svr_principal.c
+++ b/src/lib/kadm5/srv/svr_principal.c
@@ -1614,6 +1614,9 @@ kadm5_chpass_principal_3(void *server_handle,
         KADM5_FAIL_AUTH_COUNT;
     /* | KADM5_CPW_FUNCTION */
 
+    if (hist_added == 1)
+        kdb->mask |= KADM5_KEY_HIST;
+
     ret = k5_kadm5_hook_chpass(handle->context, handle->hook_handles,
                                KADM5_HOOK_STAGE_PRECOMMIT, principal, keepold,
                                new_n_ks_tuple, new_ks_tuple, password);
--- a/src/lib/kadm5/srv/svr_principal.c
+++ b/src/lib/kadm5/srv/svr_principal.c
@@ -1084,6 +1084,16 @@ check_pw_reuse(krb5_context context,
     return(0);
 }
 
+static
+void free_history_entry(krb5_context context, osa_pw_hist_ent *hist)
+{
+    int i;
+
+    for (i = 0; i < hist->n_key_data; i++)
+        krb5_free_key_data_contents(context, &hist->key_data[i]);
+    free(hist->key_data);
+}
+
 /*
  * Function: create_history_entry
  *
@@ -1097,7 +1107,7 @@ check_pw_reuse(krb5_context context,
  *      hist_key        (r) history keyblock to encrypt key data with
  *      n_key_data      (r) number of elements in key_data
  *      key_data        (r) keys to add to the history entry
- *      hist            (w) history entry to fill in
+ *      hist_out        (w) history entry to fill in
  *
  * Effects:
  *
@@ -1109,45 +1119,64 @@ check_pw_reuse(krb5_context context,
 static
 int create_history_entry(krb5_context context,
                          krb5_keyblock *hist_key, int n_key_data,
-                         krb5_key_data *key_data, osa_pw_hist_ent *hist)
+                         krb5_key_data *key_data, osa_pw_hist_ent *hist_out)
 {
-    krb5_error_code ret;
+    int i;
+    krb5_error_code ret = 0;
     krb5_keyblock key;
     krb5_keysalt salt;
-    int i;
+    krb5_ui_2 kvno;
+    osa_pw_hist_ent hist;
 
-    hist->key_data = k5calloc(n_key_data, sizeof(krb5_key_data), &ret);
-    if (hist->key_data == NULL)
-        return ret;
+    hist_out->key_data = NULL;
+    hist_out->n_key_data = 0;
+
+    if (n_key_data < 0)
+        return EINVAL;
+
+    memset(&key, 0, sizeof(key));
+    memset(&hist, 0, sizeof(hist));
+
+    if (n_key_data == 0)
+        goto cleanup;
+
+    hist.key_data = k5calloc(n_key_data, sizeof(krb5_key_data), &ret);
+    if (hist.key_data == NULL)
+        goto cleanup;
+
+    /*
+     * We only want to store the most recent kvno, and key_data should already
+     * be sorted in descending order by kvno.
+     */
+    kvno = key_data[0].key_data_kvno;
 
     for (i = 0; i < n_key_data; i++) {
-        ret = krb5_dbe_decrypt_key_data(context, NULL, &key_data[i], &key,
+        if (key_data[i].key_data_kvno < kvno)
+            break;
+        ret = krb5_dbe_decrypt_key_data(context, NULL,
+                                        &key_data[i], &key,
                                         &salt);
         if (ret)
-            return ret;
+            goto cleanup;
 
         ret = krb5_dbe_encrypt_key_data(context, hist_key, &key, &salt,
                                         key_data[i].key_data_kvno,
-                                        &hist->key_data[i]);
+                                        &hist.key_data[hist.n_key_data]);
         if (ret)
-            return ret;
-
+            goto cleanup;
+        hist.n_key_data++;
         krb5_free_keyblock_contents(context, &key);
         /* krb5_free_keysalt(context, &salt); */
     }
 
-    hist->n_key_data = n_key_data;
-    return 0;
-}
+    *hist_out = hist;
+    hist.n_key_data = 0;
+    hist.key_data = NULL;
 
-static
-void free_history_entry(krb5_context context, osa_pw_hist_ent *hist)
-{
-    int i;
-
-    for (i = 0; i < hist->n_key_data; i++)
-        krb5_free_key_data_contents(context, &hist->key_data[i]);
-    free(hist->key_data);
+cleanup:
+    krb5_free_keyblock_contents(context, &key);
+    free_history_entry(context, &hist);
+    return ret;
 }
 
 /*
@@ -1526,11 +1555,14 @@ kadm5_chpass_principal_3(void *server_handle,
                     goto done;
             }
 
-            ret = add_to_history(handle->context, hist_kvno, &adb, &pol,
-                                 &hist);
-            if (ret)
-                goto done;
-            hist_added = 1;
+            /* Don't save empty history. */
+            if (hist.n_key_data > 0) {
+                ret = add_to_history(handle->context, hist_kvno, &adb, &pol,
+                                     &hist);
+                if (ret)
+                    goto done;
+                hist_added = 1;
+            }
         }
 
         if (pol.pw_max_life)
--- a/src/lib/kdb/kdb5.c
+++ b/src/lib/kdb/kdb5.c
@@ -1,6 +1,7 @@
 /* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 /*
- * Copyright 2006, 2009, 2010 by the Massachusetts Institute of Technology.
+ * Copyright 2006, 2009, 2010, 2016 by the Massachusetts Institute of
+ * Technology.
  * All Rights Reserved.
  *
  * Export of this software from the United States of America may
@@ -752,7 +753,15 @@ krb5_db_get_principal(krb5_context kcontext, krb5_const_principal search_for,
         return status;
     if (v->get_principal == NULL)
         return KRB5_PLUGIN_OP_NOTSUPP;
-    return v->get_principal(kcontext, search_for, flags, entry);
+    status = v->get_principal(kcontext, search_for, flags, entry);
+    if (status)
+        return status;
+
+    /* Sort the keys in the db entry as some parts of krb5 expect it to be. */
+    if ((*entry)->key_data != NULL)
+        krb5_dbe_sort_key_data((*entry)->key_data, (*entry)->n_key_data);
+
+    return 0;
 }
 
 void
@@ -942,6 +951,26 @@ krb5_db_delete_principal(krb5_context kcontext, krb5_principal search_for)
     return status;
 }
 
+/*
+ * Use a proxy function for iterate so that we can sort the keys before sending
+ * them to the callback.
+ */
+struct callback_proxy_args {
+    int (*func)(krb5_pointer, krb5_db_entry *);
+    krb5_pointer func_arg;
+};
+
+static int
+sort_entry_callback_proxy(krb5_pointer func_arg, krb5_db_entry *entry)
+{
+    struct callback_proxy_args *args = (struct callback_proxy_args *)func_arg;
+
+    /* Sort the keys in the db entry as some parts of krb5 expect it to be. */
+    if (entry && entry->key_data)
+        krb5_dbe_sort_key_data(entry->key_data, entry->n_key_data);
+    return args->func(args->func_arg, entry);
+}
+
 krb5_error_code
 krb5_db_iterate(krb5_context kcontext, char *match_entry,
                 int (*func)(krb5_pointer, krb5_db_entry *),
@@ -949,13 +978,22 @@ krb5_db_iterate(krb5_context kcontext, char *match_entry,
 {
     krb5_error_code status = 0;
     kdb_vftabl *v;
+    struct callback_proxy_args proxy_args;
 
     status = get_vftabl(kcontext, &v);
     if (status)
         return status;
     if (v->iterate == NULL)
         return KRB5_PLUGIN_OP_NOTSUPP;
-    return v->iterate(kcontext, match_entry, func, func_arg, iterflags);
+
+    /*
+     * Use the proxy function so that the keys will be sorted before passed to
+     * the callback.
+     */
+    proxy_args.func = func;
+    proxy_args.func_arg = func_arg;
+    return v->iterate(kcontext, match_entry, &sort_entry_callback_proxy,
+                      &proxy_args, iterflags);
 }
 
 /* Return a read only pointer alias to mkey list.  Do not free this! */
@@ -2563,4 +2601,24 @@ krb5_db_check_allowed_to_delegate(krb5_context kcontext,
     if (v->check_allowed_to_delegate == NULL)
         return KRB5_PLUGIN_OP_NOTSUPP;
     return v->check_allowed_to_delegate(kcontext, client, server, proxy);
+}
+
+void
+krb5_dbe_sort_key_data(krb5_key_data *key_data, size_t key_data_length)
+{
+    size_t i, j;
+    krb5_key_data tmp;
+
+    /* Use insertion sort as a stable sort. */
+    for (i = 1; i < key_data_length; i++) {
+        j = i;
+        while (j > 0 &&
+               key_data[j - 1].key_data_kvno <
+               key_data[j].key_data_kvno) {
+            tmp = key_data[j];
+            key_data[j] = key_data[j - 1];
+            key_data[j - 1] = tmp;
+            j--;
+        }
+    }
 }
--- a/src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
+++ b/src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
@@ -40,6 +40,7 @@
 #include "ldap_pwd_policy.h"
 #include <time.h>
 #include <ctype.h>
+#include <kadm5/admin.h>
 
 #ifdef NEED_STRPTIME_PROTO
 extern char *strptime(const char *, const char *, struct tm *);
@@ -1313,6 +1314,20 @@ remove_overlapping_subtrees(char **list, int *subtcount, int sscope)
     *subtcount = count;
 }
 
+void
+free_princ_ent(osa_princ_ent_t princ_ent)
+{
+    unsigned int i;
+
+    for (i = 0; i < princ_ent->old_key_len; i++) {
+        k5_free_key_data(princ_ent->old_keys[i].n_key_data,
+                         princ_ent->old_keys[i].key_data);
+        princ_ent->old_keys[i].n_key_data = 0;
+        princ_ent->old_keys[i].key_data = NULL;
+    }
+    free(princ_ent->old_keys);
+}
+
 /*
  * Fill out a krb5_db_entry princ entry struct given a LDAP message containing
  * the results of a principal search of the directory.
@@ -1333,6 +1348,7 @@ populate_krb5_db_entry(krb5_context context, krb5_ldap_context *ldap_context,
     char **pnvalues = NULL, **ocvalues = NULL, **a2d2 = NULL;
     struct berval **ber_key_data = NULL, **ber_tl_data = NULL;
     krb5_tl_data userinfo_tl_data = { NULL }, **endp, *tl;
+    osa_princ_ent_rec princ_ent;
 
     ret = krb5_copy_principal(context, princ, &entry->princ);
     if (ret)
@@ -1440,6 +1456,8 @@ populate_krb5_db_entry(krb5_context context, krb5_ldap_context *ldap_context,
             goto cleanup;
     }
 
+    memset(&princ_ent, 0, sizeof(osa_princ_ent_rec));
+
     ret = krb5_ldap_get_string(ld, ent, "krbpwdpolicyreference", &pwdpolicydn,
                                &attr_present);
     if (ret)
@@ -1451,8 +1469,21 @@ populate_krb5_db_entry(krb5_context context, krb5_ldap_context *ldap_context,
         ret = krb5_ldap_policydn_to_name(context, pwdpolicydn, &polname);
         if (ret)
             goto cleanup;
+        princ_ent.policy = polname;
+        princ_ent.aux_attributes |= KADM5_POLICY;
+    }
 
-        ret = krb5_update_tl_kadm_data(context, entry, polname);
+    ber_key_data = ldap_get_values_len(ld, ent, "krbpwdhistory");
+    if (ber_key_data != NULL) {
+        mask |= KDB_PWD_HISTORY_ATTR;
+        ret = krb5_decode_histkey(context, ber_key_data, &princ_ent);
+        if (ret)
+            goto cleanup;
+        ldap_value_free_len(ber_key_data);
+    }
+
+    if (princ_ent.aux_attributes) {
+        ret = krb5_update_tl_kadm_data(context, entry, &princ_ent);
         if (ret)
             goto cleanup;
     }
@@ -1460,8 +1491,7 @@ populate_krb5_db_entry(krb5_context context, krb5_ldap_context *ldap_context,
     ber_key_data = ldap_get_values_len(ld, ent, "krbprincipalkey");
     if (ber_key_data != NULL) {
         mask |= KDB_SECRET_KEY_ATTR;
-        ret = krb5_decode_krbsecretkey(context, entry, ber_key_data,
-                                       &userinfo_tl_data, &mkvno);
+        ret = krb5_decode_krbsecretkey(context, entry, ber_key_data, &mkvno);
         if (ret)
             goto cleanup;
         if (mkvno != 0) {
@@ -1567,6 +1597,7 @@ cleanup:
     free(tktpolname);
     free(policydn);
     krb5_free_unparsed_name(context, user);
+    free_princ_ent(&princ_ent);
     return ret;
 }
 
--- a/src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
+++ b/src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
@@ -1,6 +1,29 @@
 /* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 /* plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c */
 /*
+ * Copyright 2016 by the Massachusetts Institute of Technology.
+ * All Rights Reserved.
+ *
+ * Export of this software from the United States of America may
+ *   require a specific license from the United States Government.
+ *   It is the responsibility of any person or organization contemplating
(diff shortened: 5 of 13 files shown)


greghudson:
(Summarizing voice discussion.)  For the second commit, the old code leaves hist in a problematic state if the loop fails partway through, and the new code can leak memory in that case.  I think the approach most consistent with our current practices would be:

* Rename the parameter hist to hist_out.

* Create a local variable hist, initialized to { 0, NULL }.

* Keep hist.n_key_data consistent with the number of valid encrypted key data items as we go through the loop.

* On success, assign "*hist_out = hist;".  On failure, free_history_entry(context, &hist).  The definition of free_history_entry() should be moved up above create_history_entry() so we don't have to prototype it.

src/include/kdb.h
```diff
@@ -209,6 +209,16 @@ typedef struct _krb5_db_entry_new {
...
     krb5_tl_data        * tl_data;              /* Linked list */
+
+    /*
+     * Currently some parts of krb5 expect key_data to be sorted in reverse
+     * kvno order. Make sure this is sorted to prevent incorrect behavior from
+     * occuring.
+     *
+     * For more info please see here:
+     *
+     * https://github.com/krb5/krb5/pull/132#issuecomment-52522419
```
greghudson: I don't think we want to be linking to github PR comments from source code comments.  Also, I suggest cutting this comment down to "key_data must be sorted by kvno in descending order"; there is no need to hedge with "currently" or give a rationale.

src/include/kdb.h
```diff
@@ -683,6 +693,18 @@ krb5_error_code krb5_db_check_allowed_to_delegate(krb5_context kcontext,
                                                   const krb5_db_entry *server,
                                                   krb5_const_principal proxy);
 
+/**
+ * Sort an array of @a krb5_key_data keys in descending order by their kvno.
```
greghudson: It would be good to state that key data order within a kvno is preserved.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1105,45 +1115,59 @@ check_pw_reuse(krb5_context context,
...
 
-    hist->key_data = k5calloc(n_key_data, sizeof(krb5_key_data), &ret);
-    if (hist->key_data == NULL)
-        return ret;
+    hist_out->key_data = NULL;
+    hist_out->n_key_data = 0;
 
-    for (i = 0; i < n_key_data; i++) {
-        ret = krb5_dbe_decrypt_key_data(context, NULL, &key_data[i], &key,
+    if (n_key_data <= 0)
```
greghudson: I see why you are adding this, but it's not a valid assumption that there are always existing keys in the key data.  Try "addprinc -nokey -policy polname test" (with an existing policy name) followed by "cpw -pw pw test".

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1105,45 +1115,59 @@ check_pw_reuse(krb5_context context,
...
+    if (hist.key_data == NULL)
+        goto cleanup;
+
+    /*
+     * We only want to store the most recent kvno, and key_data should already
+     * be sorted in descending order by kvno.
+     */
+    kvno = key_data[0].key_data_kvno;
+
+    for (hist.n_key_data = 0; hist.n_key_data < n_key_data; hist.n_key_data++) {
```
greghudson: I think the loop read a lot better with "i" as the loop index, since it is used six times within the loop.  hist.n_key_data can still be used as the index into hist.key_data, and incremented each time we encrypt a key into the array.

src/lib/kdb/kdb5.c
```diff
@@ -752,7 +753,15 @@ krb5_db_get_principal(krb5_context kcontext, krb5_const_principal search_for,
         return status;
     if (v->get_principal == NULL)
         return KRB5_PLUGIN_OP_NOTSUPP;
-    return v->get_principal(kcontext, search_for, flags, entry);
+    status = v->get_principal(kcontext, search_for, flags, entry);
+    if (status)
+        return status;
+
+    /* Sort the keys in the db entry as some parts of krb5 expect it to be. */
+    if ((*entry)->key_data)
```
greghudson: This is an implicit null check, which we avoid in new code; use "!= NULL" to make it explicit.  (Alternatively, check for n_key_data > 0.)

src/lib/kdb/kdb5.c
```diff
@@ -2564,3 +2602,28 @@ krb5_db_check_allowed_to_delegate(krb5_context kcontext,
         return KRB5_PLUGIN_OP_NOTSUPP;
     return v->check_allowed_to_delegate(kcontext, client, server, proxy);
 }
+
+#define COMPARE_KEY_DATA(k1, k2) k1.key_data_kvno < k2.key_data_kvno
```
greghudson: A static inline function would probably be better than a macro here.  If you must use a macro, the result should be enclosed in parentheses as should k1 and k2 (see http://www.cs.yale.edu/homes/aspnes/pinewiki/C(2f)Macros.html ).  "compare" is also a somewhat deceptive name for a less-than check; idiomatically C comparison functions behave like strcmp() in their return values.

src/lib/kdb/kdb5.c
```diff
@@ -2564,3 +2602,28 @@ krb5_db_check_allowed_to_delegate(krb5_context kcontext,
...
+
+void
+krb5_dbe_sort_key_data(krb5_key_data *key_data, size_t key_data_length)
+{
+    size_t counter, indx;
```
greghudson: "i" and "j" are idiomatic names for loop variables acting as array indexes, and we generally prefer them to longer names.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -446,43 +473,111 @@ krb5_encode_krbsecretkey(krb5_key_data *key_data_in, int n_key_data,
...
 
-    free(key_data);
+    return ret;
+}
+
+/*
+ * Encode a principal's key history for insertion into ldap.
+ */
+static struct berval **
+krb5_encode_histkey(osa_princ_ent_rec *princ_ent) {
```
greghudson: Open brace goes on the next line.  (This was in Tomas's PR but still needs to be fixed.)

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1482,207 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+ *  - mkvno
+ *      The master kvno that the keys were encrypted with.
+ *  - keys_out
+ *      The decoded keys in a key_entry struct. Must be freed.
+ *  - n_keys_out
+ *      The number of key_entry's in keys_out.
+ *  - total_keys_out
+ *      An optional argument that if given will be set to the total number of
+ *      keys found throughout all the entries: sum(keys_out.n_key_data)
+ *      May be NULL.
```
greghudson: It would be simpler to make this a mandatory argument and pass a dummy variable in the caller that doesn't need it.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1482,207 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+    *n_keys_out = 0;
+    if (total_keys_out)
+        *total_keys_out = 0;
 
-    if ((st=krb5_unparse_name(context, entries->princ, &user)) != 0)
+    /*
+     * Get the number of keys. Previously, similar implementations used realloc
+     * instead of pre-counting but I think this will have better performance.
+     */
+    for (n_keys = 0, i = 0; bvalues[i] != NULL; i++)
```
greghudson: Performance is irrelevant here; the cost of LDAP queries will dominate any time we spend on memory bookkeeping in this function.  If you want to pre-count for simplicity that's fine, but performance is not a meaningful justification for it.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1482,207 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
         if (bvalues[i]->bv_len == 0)
             continue;
+
         in.length = bvalues[i]->bv_len;
         in.data = bvalues[i]->bv_val;
 
-        st = asn1_decode_sequence_of_keys (&in,
-                                           &kd,
-                                           &n_kd,
+        err = asn1_decode_sequence_of_keys(&in,
```
greghudson: Minimize the function argument overwrapping here; I believe this call can fit on two lines.  (The old code was just as much at fault, but it should be fixed while we're changing it.)

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1482,207 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
     return st;
 }
 
-krb5_error_code
-krb5_decode_krbsecretkey(krb5_context context, krb5_db_entry *entries,
-                         struct berval **bvalues,
-                         krb5_tl_data *userinfo_tl_data, krb5_kvno *mkvno)
+struct key_entry {
```
greghudson: The name "key_entry" is deceptive since this structure holds multiple key data entries.  This should probably be named "struct keyset" after KrbKeySet, and the arguments which refer to it should use "keyset" rather than "key" as nouns (e.g. "keysets_out" instead of "keys_out" in decode_keys() below).

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1482,207 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+            struct key_entry **keys_out, unsigned int *n_keys_out,
+            /* NULL */ unsigned int *total_keys_out)
 {
-    char                        *user=NULL;
-    int                         i=0, j=0, noofkeys=0;
-    krb5_key_data               *key_data=NULL, *tmp;
-    krb5_error_code             st=0;
+    krb5_error_code err = 0;
+    unsigned int n_keys, i, ki, total_keys;
+    struct key_entry *p_keys = NULL;
```
greghudson: As we discussed, avoid using "p_" as a prefix for local variable names just because they're of pointer type.

src/tests/kdbtest.c
```diff
@@ -231,7 +231,7 @@ check_entry(krb5_db_entry *ent)
     }
     for (tl = sample_entry.tl_data; tl != NULL; tl = tl->tl_data_next) {
         etl.tl_data_type = tl->tl_data_type;
-        CHECK(krb5_dbe_lookup_tl_data(ctx, ent, &etl));
+        /*HERE*/CHECK(krb5_dbe_lookup_tl_data(ctx, ent, &etl));
```
greghudson: What's this?

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: I rather like it as an optional argument, and think it is simpler. No need to have some useless variable in the caller, that could just be confusing seeing it being set but never used.

tlyu:
Linking to prior pull request #132.

src/lib/kdb/kdb5.c
nicowilliams: I learned long ago to not use `i` and `j`, nor `p` and `q`: they are easy to confuse (and think of dyslexics).  Instead I tend to use `i` and `k`, for example.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -446,43 +473,116 @@ krb5_encode_krbsecretkey(krb5_key_data *key_data_in, int n_key_data,
...
-            free(code);
-            j++;
-            last = i + 1;
-
-            if (i < n_key_data - 1)
-                currkvno = key_data[i + 1].key_data_kvno;
+    ret[num_versions] = NULL;
+
+    /* n_key_data may be 0 if a policy is created without a key. */
+    if (n_key_data > 0) {
```
greghudson: We should de-indent the for loop by doing "goto cleanup" if n_key_data is 0.  Also, don't you mean "principal" rather than "policy" in the comment?

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -415,19 +438,23 @@ free_berdata(struct berval **array)
...
+/*
+ * Encode krb5_key_data into a berval struct for ldap insertion.
+ */
+static krb5_error_code
+encode_keys(krb5_key_data *key_data_in, int n_key_data,
+                 krb5_kvno mkvno, struct berval **ret) {
     krb5_error_code err = 0;
+    int i;
     krb5_key_data *key_data = NULL;
+    struct berval *p_ret = NULL;
```
greghudson: Here's another p_ prefix on a pointer variable.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1109,45 +1119,64 @@ check_pw_reuse(krb5_context context,
...
+    if (hist.key_data == NULL)
+        goto cleanup;
+
+    /*
+     * We only want to store the most recent kvno, and key_data should already
+     * be sorted in descending order by kvno.
+     */
+    kvno = key_data[0].key_data_kvno;
+
+    for (i = 0, hist.n_key_data = 0; i < n_key_data; i++) {
```
greghudson: hist.n_key_data is already 0 from the memset() call above, so I don't think we need it in the for loop initializer.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1492,198 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
-    entries->n_key_data = noofkeys;
+    if (total_keys_out)
+        *total_keys_out = total_keys;
+
+    *n_keysets_out = n_keys;
+    *keysets_out = keysets;
+    keysets = NULL;
+
+cleanup:
+    free(keysets);
```
greghudson: free() is not sufficient to free all of a deep structure like an array of struct keysets.  This function leaks memory on failure, and its callers leak memory on success.  Try running a kadmin.local command through valgrind --tool=memcheck --leak-check=full.  (There's a one-time leak from krb5_ldap_read_realm_params() which you can for now; we should fix that but it's harmless in practice.)

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1492,198 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+    unsigned int i, n_key_entries, total_keys;
+
+    err = krb5_unparse_name(context, entries->princ, &user);
+    if (err)
+        goto cleanup;
+
+    err = decode_keys(bvalues, mkvno, &keys, &n_key_entries, &total_keys);
+    if (err != 0) {
+        const char *msg = error_message(err);
+        k5_setmsg(context, err,
```
greghudson: Since you're changing this code, make it use k5_prependmsg() (an internal alias for krb5_prepend_error_message()).

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: Good catch. I'll get it fixed ASAP.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1492,198 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+
+krb5_error_code
+krb5_decode_krbsecretkey(krb5_context context, krb5_db_entry *entries,
+                         struct berval **bvalues, krb5_kvno *mkvno)
+{
+    char                        *user=NULL;
+    krb5_key_data               *key_data=NULL, *tmp;
+    krb5_error_code             err=0;
+
+    struct keyset *keys = NULL;
```
greghudson: "keys" should be "keysets".  Also, please remove the blank line in the middle of the variable declarations.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1492,198 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+ * Sort the old keys for a principal entry by kvno in descending order. Expects
+ * each key history entry to only contain keys with the same kvno.
+ */
+static void
+sort_old_keys(osa_princ_ent_rec *adb)
+{
+    size_t counter, indx;
+    osa_pw_hist_ent tmp;
+
+    /* Use insertion sort. */
```
greghudson: As discussed in the August 20 comments on PR #132, you can use qsort() here.

src/tests/t_kdb.py
```diff
@@ -336,6 +336,31 @@ def ldap_add(dn, objectclass, attrs=[]):
...
+                             chain([0], repeat(1)))
+    # Tries passwords, with one password change that should succeed followed by
+    # all previous passwords that should fail
+    # (0, 0) (0, 1) (1, 0) (0, 1), (1, 1)...
+    list(starmap(cpw, chain.from_iterable(imap(cmb, islice(count(), nhist)))))
+
+    # Test that keys are rotated out correctly
+    cpw(nhist, 0)
+    cpw(0, 0)
+map(test_pwhist, imap(pow, repeat(2), range(5)))
```
greghudson: These tests run hundreds of commands, way more than is necessary to hit the edge cases of password history.  I also think it's hard to read this testing code because it overuses itertools functions, although I expect that to be a topic of some debate.

SaharahSarah:
OK, large amount of memory leaks fixed. I can't find anymore now. decode_histkey and encode_histkey were leaking basically everything.

greghudson:
In the current state of the PR, the old function asn1_decode_sequence_of_keys() decodes from ASN.1 into an ldap_seqof_key_data structure, then parts out that structure into separate variables and throws it away.  decode_keys() reassembles the key_data and n_key_data values into a struct keyset, which is facially similar to ldap_seqof_key_data.  (The mkvno field, meanwhile, just gets overwritten for each bvalue until the last one is used.  This is okay because we expect it to be the same for every keyset, but isn't a great interface definition.)  I wonder if it would be simpler to make asn1_decode_sequence_of_keys() preserve the ldap_seqof_key_data structure, and use that instead of struct keyset.

SaharahSarah:
Hmm, so you're suggesting just passing in the ldap_seqof_keys to asn1_decode_sequence_of_keys? I can do that, it'd be pretty simple.

SaharahSarah:
OK, I removed the keyset struct, if you like it I'll throw it into the previous comit.

greghudson:
I think that should be squashed with the previous commit, yes.

src/tests/t_kdb.py
SaharahSarah: > These tests run hundreds of commands, way more than is necessary to hit the edge cases of password history. 

Yes? What's wrong with good tests for it? It tests up to a key history policy of 16, I think that's pretty fair. A lot better than the original one I had that was 64. That took like 5 minutes to run.

> I also think it's hard to read this testing code because it overuses itertools functions, although I expect that to be a topic of some debate.

It should be fine. It's well documented on what it does and it does it fast and efficiently. There might be a better way to do it with less iterators... I'm skilled at itertools, but I still have room to improve.

SaharahSarah:
Okay, it's been added to the commit.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: Well, this works, so, unless there's some pressing need to switch to qsort I don't see the point.

src/include/kdb.h
SaharahSarah: This has been added.

src/tests/t_kdb.py
greghudson: Testing a history size of 16 isn't likely to expose any more problems than testing a history size of 3; there are no additional edge cases.  Running hundreds of commands when a dozen would suffice makes it more difficult to sort through the testing sequence and debug problems when a test in t_kdb.py fails.  And while the ten extra seconds of runtime might not be a problem in isolation, when multiplied by the hundreds of similarly-sized pieces of functionality in krb5, it could make the test suite take so long that developers will stop wanting to run it before submitting changes.

Most Python resources I can find discourage using lambda for non-anonymous functions (use def instead), encourage using list comprehensions and generator expressions over using map and filter, and discourage using the backtick operator (which I had actually never heard of before).  For instance:

  http://stackoverflow.com/questions/134626/which-is-more-preferable-to-use-in-python-lambda-functions-or-nested-functions
  http://stackoverflow.com/questions/1673071/what-do-backticks-mean-to-the-python-interpreter-num

In this case, I think the edge cases can be hit with a history size of 2 or 3 and no iteration or functions in the Python code, just a straightforward list of commands with expected outcomes.  If any of the commands fail, it's much easier for a developer to figure out where the failure is in the test script and what was supposed to happen.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
greghudson: qsort() is the idiomatic way to sort when a stable sort isn't needed.  It's efficient for large amounts of data (although that probably isn't all that important to us here), and developers looking at the code don't have to think about how an insertion sort works.

src/tests/t_kdb.py
SaharahSarah: This treat is specifically designed to test that the key rotation is working right. I had originally written it because I had some concerns over that functionality specifically, and I saw no reason to not include it after that. 3 password histories won't really be enough for that, and anything beyond three will have a lot of commands which would be better to put in a loop. I could add basic history tests so any issues that doing have to do with the rotation will show up in something far easier to read, but I would hesitate to remove this test. I can drop it down to 8 tests which are fairly quick as a compromise.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: How about adding a function pointer to get the krb5_key_data to the already existing method that sorts the keys in the DIL?

src/tests/t_kdb.py
greghudson: Here is what I would consider a more reasonable version of the same tests, in the same number of lines of Python code:

    # Test password history.                                                    
    def test_pwhist(nhist):
        def cpw(n):
            realm.run([kadminl, 'cpw', '-pw', str(n), princ])
        def cpw_fail(n):
            realm.run([kadminl, 'cpw', '-pw', str(n), princ], expected_code=1)
        output('*** Testing password history of size %d\n' % nhist)
        princ = 'pwhistprinc' + str(nhist)
        pol = 'pwhistpol' + str(nhist)
        realm.run([kadminl, 'addpol', '-history', str(nhist), pol])
        realm.run([kadminl, 'addprinc', '-policy', pol, '-nokey', princ])
        for i in range(nhist):
            # Set a password, then check that all previous passwords fail.      
            cpw(i)
            for j in range(i + 1):
                cpw_fail(i)
        # Set one more new password, and make sure the oldest key is            
        # rotated out.                                                          
        cpw(nhist)
        # I would add: cpw_fail(1)
        cpw(0)

    for n in (1, 2, 4, 8, 16):
        test_pwhist(n)

I can understand being concerned by the complexity of add_to_history(); that function is grossly overcomplicated because it uses a circular queue for no reason.  (The old keys are marshalled and unmarshalled on each password change, which is O(n) in the number of old keys, so trying to keep the insertion cost to O(1) is pointless.)  I would be happy to see it rewritten to be simpler.    But for its current incarnation, running a big quadratic test doesn't hit any more of its edge cases than a small quadratic test.  To cover that function properly you would need to do things like change the password policy after loading up the principal with some old keys.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
greghudson: I don't see why that would be better than using qsort().

SaharahSarah: It's using qsort now.

src/tests/t_kdb.py
SaharahSarah: Well, another issue besides the complexity of add_to_history is that ldap does not return the same osa_princ_ent_rec that was passed into it. add_to_history seems to assume that the database will store them in a certain order, which is why it has the circular queue. If I were to look at add_to_history without knowing anything about the ldap plugin I would start having the same assumption, and those types of disconnects in the code could potentially lead to some developer committing code in the future that could break the ldap history.

SaharahSarah: Oh, also, good call on the lambda. I had originally written a big long line in the python interpreter to test this myself but then afterwards copied the line. It was long so I split it up really quick so it slipped my mind setting the lambda to a function.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -415,19 +436,24 @@ free_berdata(struct berval **array)
...
-    struct berval **ret = NULL;
-    int currkvno;
-    int num_versions = 1;
-    int i, j, last;
+/*
+ * Encode krb5_key_data into a berval struct for ldap insertion.
+ */
+static krb5_error_code
+encode_keys(krb5_key_data *key_data_in, int n_key_data,
+                 krb5_kvno mkvno, struct berval **ret_out)
```
greghudson: Please rename "ret_out" to "bval_out" and "ret" to "bval".  I'm not super-attached to "bval", but "ret" doesn't make sense when we aren't actually returning it as we were in the old code, and "ret_out" is just confusing.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -446,31 +472,68 @@ krb5_encode_krbsecretkey(krb5_key_data *key_data_in, int n_key_data,
...
+        goto cleanup;
+
+    ret->bv_len = code->length;
+    ret->bv_val = code->data;
+
+    *ret_out = ret;
+    ret = NULL;
+
+cleanup:
+    free(code);
```
greghudson: This free() should be moved up to just after we transfer ownership of code->data to ret.  A simple free() would not be sufficient to free a populated krb5_data * object if we haven't already stolen the memory from it, so it doesn't make sense to put it in the cleanup handler.  You can also remove the initialization of code to NULL after this change.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1493,221 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+free_ldap_seqof_key_data(ldap_seqof_key_data *keysets, krb5_int16 n_keysets,
+                         int full_free)
 {
-    char                        *user=NULL;
-    int                         i=0, j=0, noofkeys=0;
-    krb5_key_data               *key_data=NULL, *tmp;
-    krb5_error_code             st=0;
+    int i;
+
+    if (keysets != NULL) {
```
greghudson: In situations like this, it's usually our preference to de-indent the contents of the if block by returning if keysets == NULL.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1493,221 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+compare_descending_osa_pw_hist_ent(const void *left_in, const void *right_in)
+{
+    osa_pw_hist_ent *left = (osa_pw_hist_ent *) left_in;
+    osa_pw_hist_ent *right = (osa_pw_hist_ent *) right_in;
+
+    /*
+     * qsort sorts in ascending order, but we need descending. Subtracting the
+     * left from the right will give us the desired sorting order.
+     */
+    return right->key_data[0].key_data_kvno - left->key_data[0].key_data_kvno;
```
greghudson: I am nervous at best, about assuming that the osa_pw_hist_ent pointers are non-empty.  I don't think decode_keys() or krb5_decode_krbsecretkey() enforces this, so in theory it's possible to read in zero-length.

Treating any empty entries as having kvno 0 would be reasonable:

    kvno_left = left->n_key_data ? left->key_data[0].key_data_kvno[0] : 0;
    kvno_right = (same thing for right)
    return kvno_right - kvno_left;

Also, if you take the above, make sure to use "int" as the type of kvno_left and kvno_right.  Right now you're subtracting two uint16_t values, so you never produce a negative result.  I'm not sure why the tests are passing.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: if `asn1_encode_sequence_of_keys` ever returns an error, can `code` still be allocated memory?

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: Hmm, I had thought about that but I thought I was checking that with bv_len=0, but it seems that that doesn't give the result I thought it did.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
greghudson: No; the asn1_encode_* functions conform to our usual standard of setting output pointers to NULL on failure (see k5_asn1_full_encode() in lib/krb5/asn.1/asn1_encode.c).  That's a good thing to check, though.

src/tests/t_kdb.py
```diff
@@ -336,6 +336,31 @@ def ldap_add(dn, objectclass, attrs=[]):
...
+    for i in range(nhist):
+        # Set a password, then check that all previous passwords fail.
+        cpw(i)
+        for j in range(i + 1):
+            cpw_fail(j)
+    # Set one more new password, and make sure the oldest key is
+    # rotated out.
+    cpw(nhist)
+    cpw(0)
+    cpw_fail(0)
```
greghudson: If this cpw_fail(0) call was added based on my suggestion, then I was actually suggesting adding a cpw_fail(1) before the cpw(0) to test that the second-oldest key is still present.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -1375,64 +1493,221 @@ krb5_read_tkt_policy(krb5_context context, krb5_ldap_context *ldap_context,
...
+                    osa_princ_ent_rec *princ_ent)
+{
+    krb5_error_code err = 0;
+    krb5_int16 i, n_keysets = 0;
+    ldap_seqof_key_data *keysets = NULL;
+
+    err = decode_keys(bvalues, &keysets,
+                      &n_keysets, NULL);
+    if (err != 0) {
+        k5_prependmsg(context, err, error_message(err));
```
greghudson: This is not how k5_prependmsg() works.  Replace both of these calls with:

    k5_prependmsg(context, err, _("unable to ..."));

and similarly for the other case.

src/tests/t_kdb.py
SaharahSarah: First off, when 0 is added as a password 1 will be removed, so setting the password to 1 will still work. Also, when the policy only has a password history of 1 then only the current password will fail.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
SaharahSarah: Done

src/tests/t_kdb.py
greghudson: I suggested adding cpw_fail(1) before cpw(0), not after.  In the nhist == 1 case, this fails as expected because we just set a password of 1.

SaharahSarah: Ahh, gotcha

src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
```diff
@@ -1313,6 +1314,20 @@ remove_overlapping_subtrees(char **list, int *subtcount, int sscope)
     *subtcount = count;
 }
 
+void
```
greghudson: This function needs to be static, or we get a missing-prototype warning.

greghudson:
Pushed to master as 44ad57d8d38efc944f64536354435f5b721c0ee0 d7f91ac2f6655e77bb3658c2c8cc6132f958a340 b46cce2ea8c0841f7f93db73eefcd180c87a3eae.

PR #400, opened 2016-01-26
Add hints for -A flag to klist and kdestroy

mrogers950:
When using DIR/keyring ccache, a user accustomed to the FILE ccache behavior
may not be aware of all active caches and the default klist/kdestroy commands
could make it seem like there is no active cache left. If there are multiple
caches, print a warning with the number of caches and hint at the -A flag.

### DIFF
--- a/src/clients/kdestroy/kdestroy.c
+++ b/src/clients/kdestroy/kdestroy.c
@@ -66,6 +66,29 @@ static void usage()
     exit(2);
 }
 
+void
+print_remaining_cc_warning(krb5_context context)
+{
+    krb5_error_code code;
+    krb5_ccache cache;
+    krb5_cccol_cursor cursor;
+
+    code = krb5_cccol_cursor_new(context, &cursor);
+    if (code) {
+        com_err(progname, code, _("while listing credential caches"));
+        exit(1);
+    }
+
+    code = krb5_cccol_cursor_next(context, cursor, &cache);
+    if (code == 0 && cache != NULL) {
+        fprintf(stderr,
+                _("Other credential caches present, use -A to destroy all\n"));
+        krb5_cc_close(context, cache);
+    }
+
+    krb5_cccol_cursor_free(context, &cursor);
+}
+
 int
 main(argc, argv)
     int argc;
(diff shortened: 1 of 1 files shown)


greghudson:
Sorry for the slow response; we've been behind on pull request review and I've also wanted to think about this.  It's an interesting idea to improve usability, but also runs the risk of being annoying (or even breaking scripts) when people know what's going on.

Please squash the two commits.  We want to review what's going onto the master branch, and we try to keep our master branch clean of development artifacts.

kdestroy -q probably shouldn't print the warning, although the currently meaning of that "quiet" flag is pretty antiquated.

There's an argument for printing this warning to stderr rather than stdout.  kdestroy currently only prints anything to stderr, and in klist, printing the warning to stderr would separate it from information about the primary credential cache.

Finally, the code to count credential caches in the collection should be in a separate function, not inlined into already long functions.  It should also perhaps stop counting after two, since it doesn't really care at that point.

mrogers950:
Thanks, I squashed them with my revisions. 
kdestroy -q only avoids the bell on error, so perhaps there should be an -s flag to silence everything.

Also, the functions I added could make use of a ccache counter function, i.e. "krb5_cccol_count()" but it seems that would need to be a part of the public interface. Thoughts?

mrogers950:
It's not urgent but if there's anything else needed to finish this up let me know.

mrogers950:
Related RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1273271

greghudson:
After reading the bugzilla report, I think it would be sufficient to do this in kdestroy alone.

As I noted in the call, I would like to see the kdestroy change respect the -q (quiet) flag.

mrogers950:
I've pushed an update with those changes.

greghudson:
Pushed to master as d4cc39674b407a0d9698097b27788b3ffdbc6fca with a few small changes.  For the future:

* I made print_remaining_cc_warning() static and added a comment.  Helper functions should almost always have descriptive comments, unless they are trivial.

* I renamed "code" to "retval" to be consistent with other usage in the file.  When there is nothing to be consistent with, the preferred name for a variable of this purpose is "ret".

PR #401, opened 2016-02-04
Fix populate_krb5_db_entry() princ_ent init

greghudson:
The most recent commit introduced a new variable princ_ent in
populate_krb5_db_entry().  princ_ent is cleaned up by the function's
cleanup label, so it must be initialized before any "goto cleanup"
statements.

(Discovered by Coverity after PR #399 was pushed to master.)


kaduk:
The location change looks good, but should it use sizeof(princ_ent) instead of sizeof(osa_princ_ent_rec)?

greghudson:
Sure, that's usually our preference, so I changed it.

PR #402, opened 2016-02-09
added check for write permissions to kdb5_create.c

krichter722:


greghudson:
Sorry for the slow response time to this PR; we have been behind on PR review of late.

In Unix programs, it is generally discouraged to check permissions before attempting a filesystem operation, because it can introduce time-of-check/time-of-use race conditions.  It's better to just try the operation and see if it fails.

In addition, kdb5_util create should not be making use of global_params.dbname; the fact that it currently does so in some diagnostic messages is a bug.  Since release 1.5, the database layer has been abstracted with multiple database modules, and the DB2 module doesn't necessarily use global_params.dbname to figure out where the database will live.  Right now we sometimes display incorrect diagnostic messages when global_params.dbname is wrong; adding an access check would compound the situation by causing kdb5_util create to spuriously fail in some configurations.

PR #403, opened 2016-02-10
Fix and adjust t_kprop.py

greghudson:
The listprincs check was at the wrong indentation level and had the
wrong argument grouping; fix it so we actually verify the propagation.

Stop using the -t (runonce) flag to kpropd, so that kpropd continues
to run until k5test.py terminates it.  Quit out of the read loop when
we see that the load process is completed, instead of looking for end
of input.  This change is needed in order to add hooks in k5test.py
for checking daemons for memory leaks before terminating them.


mrogers950:
LGTM.

PR #404, opened 2016-02-11
Add support for IP_SENDSRCADDR for ip pktinfo functionality.

SaharahSarah:
BSD uses IP_SENDSRCADDR instead of IP_PKTINFO for ipv4 pktinfo
functionality.  Add support for using this when IP_PKTINFO is not
available.

Put into a separate PR as discussed in [PR 380](https://github.com/krb5/krb5/pull/380#issuecomment-180628692).

### DIFF
--- a/src/lib/apputils/Makefile.in
+++ b/src/lib/apputils/Makefile.in
@@ -12,19 +12,20 @@ SED = sed
 ##DOS##XTRA=
 ##DOS##OBJFILE=$(OUTPRE)apputils.lst
 
-STLIBOBJS=net-server.o @LIBOBJS@
+STLIBOBJS=net-server.o udppktinfo.o @LIBOBJS@
 LIBBASE=apputils
 
 all-unix:: all-liblinks
 clean-unix:: clean-liblinks clean-libs clean-libobjs
 install-unix:: install-libs
 
-LINTFLAGS=-uhvb 
+LINTFLAGS=-uhvb
 LINTFILES= daemon.c
 LIBOBJS=$(OUTPRE)daemon.$(OBJEXT)
 
 SRCS=	$(srcdir)/daemon.c \
-	$(srcdir)/net-server.c
+	$(srcdir)/net-server.c \
+	$(srcdir)/udppktinfo.c
 
 @libpriv_frag@
 @lib_frag@
--- a/src/lib/apputils/net-server.c
+++ b/src/lib/apputils/net-server.c
@@ -60,6 +60,8 @@
 #include "net-server.h"
 #include <signal.h>
 
+#include "udppktinfo.h"
+
 /* XXX */
 #define KDC5_NONET                               (-1779992062L)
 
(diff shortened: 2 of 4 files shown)


greghudson:
Thanks for redoing against the current net-server.c code.  My notes:

* Remember that commit message summary lines don't get trailing periods; they are like email subjects.

In the first commit:

* The commit message should mention that setup_udp_pktinfo_ports() will now always try to set up pktinfo wildcard listener sockets, and will fall back at runtime if the system doesn't support it.

* libapputils is a private library, statically linked into krb5kdc and kadmind.  As such, it does not need to use a namespace prefix.  The moved functions shouldn't need to be renamed, nor should aux_addressing_info need a name change.

* Your editor seems to have removed the form-feed characters.

* The prototypes in udppktinfo.h have a space between "*" and "auxaddr".  There's also no newline at the end of this file.

In the second commit, I like the refactored code much better, but have some nitpicks:

* Should the set_ip*_pktinfo() functions really have sockopt parameters?  You also have a mistake in set_ipv6_pktinfo() in this commit which is corrected in the third commit.

* Where you move code into helper functions, fix the multi-line conditionals to break after the operators rather than before, and move the interior variable declarations to function scope.  (Likewise for the new functions in the third commit.)

* We don't use /* NULL */ designations inside parameter lists, and I think where you do use them, the parameters aren't allowed to be null.

* I think you have a bad indent at line 366 of udppktinfo.c (",goto use_sendto;") and the braces can be removed.

* We should have a brief conversation including Tom on tomorrow's status call about how to name conditionalized helper functions; I think Tom prefers a slightly different approach.

* You add a blank line to udppktinfo.h in this commit; that should be moved to the first commit.

In the third commit:

* The commit message should probably say FreeBSD.  Current NetBSD implements IP_PKTINFO as well as IP_RECVDSTADDR/IP_SENDSRCADDR, and OpenBSD requires a bit more investigation as it only defines IP_RECVDSTADDR.

SaharahSarah:
OK, pretty sure I fixed everything you mentioned except the form feed. My editor removes it automatically because it removes spaces at the end of lines. I don't really know a good way to add it back.

SaharahSarah:
FreeBSD doesn't work, apparently IP_SENDSRCADDR doesn't work if the socket isn't bound to a wildcard i.e. bound to a specific address. Fixing it now.

greghudson:
The helper functions still need to be changed to break lines after binary operators rather than before.

In the unsupported cases, instead of:

    #define check_cmsg_v6_pktinfo check_cmsg_v6_pktinfo_unsupported

you can just write:

    #define check_cmsg_v6_pktinfo check(c, t, l, a) 0

and similarly for the other helpers.

I don't particularly like how the conditionalizaton of send_to_from() and recv_from_to() turned out; in this specific case, I'd be happier just defining the function with its exported name twice.  We definitely avoid using leading underscores in symbol names.

SaharahSarah:
Got rid of the unsupported functions and turned them into macros. Also got rid of the ugly conditionalization functions. Oh, lines are broken after binary operators too.

Also, added in a check to not use pktinfo when not bound to a wildcard address. This fixes the issue in FreeBSD when binding to specific addresses.

greghudson:
is_socket_bound_to_wildcard() needs to use "struct sockaddr_storage" for its local variable to ensure that there is enough space for every socket family.

Also, when lines get long enough that they wrap awkwardly, look for ways to shorten them.  In this case, intermediate variables of type struct sockaddr_in * or struct sockaddr_in6 * would help a lot with readability as well as length.  Also, (x) ? 1 : 0 can be simplified down to just x if x is already a boolean value.  So, for instance, "return sin6->sin_addr.s_addr == INADDR_ANY;"

greghudson:
I just recalled that we have some inline functions in socket-utils.h (included by k5-int.h) which can improve type safety and brevity when casting between different socket types: sa2sin(), sa2sin6(), ss2sa(), ss2sin(), and ss2sin6().  I think it would be good to use those functions in the second commit when refactoring udppktinfo.c.

SaharahSarah:
Refactored is_socket_bound_to_wildcard to use socket-utils.h helper functions.

greghudson:
Thanks.  Can you also do that for the refactored helper functions, like set_msg_from_ip_pktinfo() and check_cmsg_ipv6_pktinfo()?

SaharahSarah:
OK those have been switched over.

greghudson:
Pushed to master with minor changes (mostly whitespace) as b6396a5f7c720f51bb7daaf8624324d49163babb 076d6dbc50d683be540274654b35270fa8cfc289 457a4ac5057d44c0c947efc3c047c32897b54747

PR #405, opened 2016-02-12
Remove form feed characters

greghudson:
Some older code in the tree uses form feed characters.  kdevelop does
not appear to preserve them, and it is not our current practice to use
them, so get rid of them in almost all files under src.  Leave alone
lib/gssapi/krb5/3des.txt, which is a formatted internet draft.


SaharahSarah:
Looks good to me.

PR #406, opened 2016-02-13
Fix build failure of hdb plugin due to internal API changes

pefoley2:
The hdb plugin has bit-rotted somewhat.
Update the code to conform to the new internal APIs.

### DIFF
--- a/src/plugins/kdb/hdb/kdb_hdb.c
+++ b/src/plugins/kdb/hdb/kdb_hdb.c
@@ -220,7 +220,7 @@ cleanup:
         kh = NULL;
     }
 
-    krb5int_free_error(&errinfo, NULL);
+    k5_free_error(&errinfo, NULL);
 
     *pkh = kh;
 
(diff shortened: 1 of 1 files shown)


greghudson:
Thanks; I hope to get to this shortly, as it looks pretty straightforward.  Have you tested the code after these changes, or just gotten it to compile?

pefoley2:
I've gotten it to compile, I'm still working on trying to migrate a heimdal realm over.

pefoley2:
I didn't manage to get the code to do much more then build successfully.
It still targets HDB_INTERFACE_VERSION 4, while the latest version of heimdal targets HDB_INTERFACE_VERSION 8.
I wound up using heimdal 1.16's mit dump export support instead, so I'm unlikely to look much more closely at the hdb plugin.
The code builds now, but you'd need a heimdal from before https://github.com/heimdal/heimdal/commit/ba04bad361de3377e0c33da5c3c23eb02394aee5 to use it in its current state.
I'm not sure how much work would be involved in updating the code to target a later interface version, but regardless it's probably still worthwhile to merge these changes as they're fairly straightforward build fixes.
Maybe add a disclaimer about interface version mismatch too?

greghudson:
Thanks for investigating.  We might consider removing the hdb module; in hindsight, a Heimdal-to-MIT database shim isn't a very future-proof or testable design, compared to a Heimdal dump import feature.  If Heimdal implements an MIT dump export, then perhaps we don't need either.

tlyu:
Removal of the hdb module is in #414.

greghudson:
I'm closing this PR because we decided to remove the hdb KDB module, but I do want to say thanks again for the work and for the information about the changes to Heimdal.

PR #407, opened 2016-02-17
Harmonize struct packing for gic_opt.c

tlyu:
struct extended_options in gic_opt.c extends krb5_get_init_creds_opt.
On Mac OS X, for historical reasons, we define krb5_get_init_creds_opt
with an alignment/packing of 2, conflicting with the compiler default.
This results in alignment change warnings from clang on Mac OS X.
Ensure that extended_options has the same packing.


greghudson:
Looks okay to me.

kaduk:
That seems okay, since the structure is only used within the one file, eliminating any risk of breaking the ABI.

PR #408, opened 2016-02-18
Fix compilation issues in udppktinfo.c

greghudson:
Move is_socket_bound_to_wildcard() into an #if block so it doesn't get
build when it isn't used.  Avoid using the identifier "socket" as it
can produce shadowed declaration warnings.  Make the definition of the
fallback send_to_from() conform to the declaration.


SaharahSarah:
I don't see any problems.

tlyu:
s/get build/get built/ in the commit message. Otherwise, it looks ok to me.

PR #409, opened 2016-02-22
Increase initial DNS buffer size

greghudson:
In dnsglue.c (which is used to look up SRV and TXT records), increase
the initial buffer size guess from 2048 to 4096 to accomodate DNSSEC
signatures.  Suggested by Daniel Colascione.


tlyu:
This looks ok to me.

tlyu:
This looks ok to me.

PR #410, opened 2016-02-22
Add CAMMAC and authentication indicator helper functions to libkrb5

mrogers950:
This adds krb5_unwrap_cammac_svc() to verify and decode a CAMMAC,
and krb5_authind_decode() to obtain an array of auth indicator strings
from the resulting authdata.

Also move the kdc function cammac_check_svcver() to the internal lib
as k5_cammac_check_svcver() to use in the above code.

### DIFF
--- a/doc/admin/auth_indicator.rst
+++ b/doc/admin/auth_indicator.rst
@@ -51,3 +51,6 @@ but a user who authenticates with a password would not::
     $ kvno host/high.value.server
     kvno: KDC policy rejects request while getting credentials for
       host/high.value.server@KRBTEST.COM
+
+Indicators present in a ticket are exposed to GSSAPI through the
+:ref:`auth-indicator <gssapi_authind_attr>` name attribute.
--- a/doc/appdev/gssapi.rst
+++ b/doc/appdev/gssapi.rst
@@ -513,6 +513,39 @@ gss_get_mic_iov_length and gss_get_mic_iov::
     if (GSS_ERROR(major))
         handle_error(major, minor);
 
+Name Attributes
+---------------
+
+The following extensions (declared in ``<gssapi/gssapi_ext.h>``) can
+be used in release 1.8 or later to examine name attributes described
+in :rfc:`6680`::
+
+    OM_uint32 gss_inquire_name(OM_uint32 minor_status,
+                               gss_name_t name,
+                               int *name_is_MN,
+                               gss_OID *MN_mech,
+                               gss_buffer_set_t *attrs);
+
+    OM_uint32 gss_get_name_attribute(OM_uint32 minor_status,
+                                     gss_name_t name,
+                                     gss_buffer_t attr,
+                                     int *authenticated,
+                                     int *complete,
+                                     gss_buffer_t value,
+                                     gss_buffer_t display_value,
+                                     int *more);
+
+.. _gssapi_authind_attr:
+
+* "auth-indicator" attribute:
+
+In release 1.15 this name attribute type will be included in the
+gss_inquire_name output if the ticket contains :ref:`authentication
+indicators <auth_indicator>`.  When a client performs
+pre-authentication, the KDC may include authentication indicator values
+into the authorization data of a ticket.  These are site-defined strings
+that are intended to convey the type or strength of pre-authentication,
+allowing an application to make additional authorization policy checks.
 
 .. _gss_accept_sec_context: http://tools.ietf.org/html/rfc2744.html#section-5.1
 .. _gss_acquire_cred: http://tools.ietf.org/html/rfc2744.html#section-5.2
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -956,6 +956,12 @@ void k5_free_kkdcp_message(krb5_context context, krb5_kkdcp_message *val);
 void k5_free_cammac(krb5_context context, krb5_cammac *val);
 void k5_free_secure_cookie(krb5_context context, krb5_secure_cookie *val);
 
+krb5_error_code
+k5_unwrap_cammac_svc(krb5_context context, const krb5_authdata *ad,
+                     const krb5_keyblock *key, krb5_authdata ***adata_out);
+krb5_error_code
+k5_authind_decode(const krb5_authdata *ad, krb5_data ***indicators);
+
 /* #include "krb5/wordsize.h" -- comes in through base-defs.h. */
 #include "com_err.h"
 #include "k5-plugin.h"
--- a/src/include/krb5/authdata_plugin.h
+++ b/src/include/krb5/authdata_plugin.h
@@ -55,8 +55,9 @@ typedef krb5_error_code
 #define AD_USAGE_TGS_REQ        0x02
 #define AD_USAGE_AP_REQ         0x04
 #define AD_USAGE_KDC_ISSUED     0x08
-#define AD_USAGE_MASK           0x0F
 #define AD_INFORMATIONAL        0x10
+#define AD_CAMMAC_PROTECTED     0x20
+#define AD_USAGE_MASK           0x2F
 
 struct _krb5_authdata_context;
 
--- a/src/kdc/cammac.c
+++ b/src/kdc/cammac.c
@@ -171,24 +171,3 @@ cleanup:
     krb5_free_data(context, der_enctkt);
     return valid;
 }
-
-/* Return true if cammac's service verifier is valid for server_key. */
-krb5_boolean
-cammac_check_svcver(krb5_context context, krb5_cammac *cammac,
-                    krb5_keyblock *server_key)
-{
-    krb5_error_code ret;
-    krb5_verifier_mac *ver = cammac->svc_verifier;
-    krb5_boolean valid = FALSE;
-    krb5_data *der_authdata = NULL;
-
-    if (ver == NULL)
-        return FALSE;
-    ret = encode_krb5_authdata(cammac->elements, &der_authdata);
-    if (ret)
-        return FALSE;
-    ret = krb5_c_verify_checksum(context, server_key, KRB5_KEYUSAGE_CAMMAC,
-                                 der_authdata, &ver->checksum, &valid);
-    krb5_free_data(context, der_authdata);
-    return valid;
-}
--- a/src/kdc/kdc_util.h
+++ b/src/kdc/kdc_util.h
@@ -137,10 +137,6 @@ krb5_boolean
 cammac_check_kdcver(krb5_context context, krb5_cammac *cammac,
                     krb5_enc_tkt_part *enc_tkt, krb5_db_entry *krbtgt);
 
-krb5_boolean
-cammac_check_svcver(krb5_context context, krb5_cammac *cammac,
-                    krb5_keyblock *server_key);
-
 /* do_as_req.c */
 void
 process_as_req (krb5_kdc_req *, krb5_data *,
--- a/src/lib/krb5/krb/Makefile.in
+++ b/src/lib/krb5/krb/Makefile.in
@@ -18,6 +18,7 @@ STLIBOBJS= \
 	allow_weak.o	\
 	appdefault.o	\
 	auth_con.o	\
+	cammac_util.o	\
 	authdata.o	\
 	authdata_exp.o	\
 	authdata_enc.o	\
@@ -128,6 +129,7 @@ OBJS=	$(OUTPRE)addr_comp.$(OBJEXT)	\
 	$(OUTPRE)allow_weak.$(OBJEXT)	\
 	$(OUTPRE)appdefault.$(OBJEXT)	\
 	$(OUTPRE)auth_con.$(OBJEXT)	\
+	$(OUTPRE)cammac_util.$(OBJEXT)	\
 	$(OUTPRE)authdata.$(OBJEXT)	\
 	$(OUTPRE)authdata_exp.$(OBJEXT)	\
 	$(OUTPRE)authdata_enc.$(OBJEXT)	\
@@ -237,6 +239,7 @@ SRCS=	$(srcdir)/addr_comp.c	\
 	$(srcdir)/addr_srch.c	\
 	$(srcdir)/appdefault.c	\
 	$(srcdir)/auth_con.c	\
+	$(srcdir)/cammac_util.c	\
 	$(srcdir)/authdata.c	\
 	$(srcdir)/authdata_exp.c	\
 	$(srcdir)/authdata_enc.c	\
--- a/src/lib/krb5/krb/Makefile.in
+++ b/src/lib/krb5/krb/Makefile.in
@@ -17,6 +17,7 @@ STLIBOBJS= \
 	addr_srch.o	\
 	allow_weak.o	\
 	appdefault.o	\
+	ai_authdata.o   \
 	auth_con.o	\
 	cammac_util.o	\
 	authdata.o	\
@@ -128,6 +129,7 @@ OBJS=	$(OUTPRE)addr_comp.$(OBJEXT)	\
 	$(OUTPRE)addr_srch.$(OBJEXT)	\
 	$(OUTPRE)allow_weak.$(OBJEXT)	\
 	$(OUTPRE)appdefault.$(OBJEXT)	\
+	$(OUTPRE)ai_authdata.$(OBJEXT)  \
 	$(OUTPRE)auth_con.$(OBJEXT)	\
 	$(OUTPRE)cammac_util.$(OBJEXT)	\
 	$(OUTPRE)authdata.$(OBJEXT)	\
@@ -240,6 +242,7 @@ SRCS=	$(srcdir)/addr_comp.c	\
 	$(srcdir)/appdefault.c	\
 	$(srcdir)/auth_con.c	\
 	$(srcdir)/cammac_util.c	\
+	$(srcdir)/ai_authdata.c \
 	$(srcdir)/authdata.c	\
 	$(srcdir)/authdata_exp.c	\
 	$(srcdir)/authdata_enc.c	\
@@ -385,7 +388,7 @@ T_KERB_OBJS= t_kerb.o conv_princ.o unparse.o set_realm.o str_conv.o
 
 T_SER_OBJS= t_ser.o ser_actx.o ser_adata.o ser_addr.o ser_auth.o ser_cksum.o \
 	ser_ctx.o ser_key.o ser_princ.o serialize.o authdata.o pac.o \
-	pac_sign.o authdata_exp.o s4u_authdata.o copy_data.o etype_list.o
+	pac_sign.o ai_authdata.o authdata_exp.o s4u_authdata.o copy_data.o etype_list.o
 
 T_DELTAT_OBJS= t_deltat.o deltat.o
 
--- a/src/lib/krb5/krb/ai_authdata.c
+++ b/src/lib/krb5/krb/ai_authdata.c
@@ -1,0 +1,376 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* src/lib/krb5/krb/ai_authdata.c - Auth indicator AD backend */
+/*
+ * Copyright (C) 2016 by the Massachusetts Institute of Technology.
+ * Copyright (C) 2016 by Red Hat, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * Authdata backend for authentication indicators.
+ */
+
+#include "k5-int.h"
+#include "authdata.h"
+#include "auth_con.h"
+#include "int-proto.h"
+
+struct authind_context {
+    krb5_data **indicators;      /* Decoded. */
+    int count;
+};
+
+static krb5_error_code
+authind_init(krb5_context kcontext, void **plugin_context)
+{
+    *plugin_context = NULL;
+    return 0;
+}
+
+static void
+authind_flags(krb5_context kcontext, void *plugin_context,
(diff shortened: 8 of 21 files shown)


mrogers950:
Two things:
a. Should krb5_authind_decode()'s authdata argument take a single authdata or an array? It seemed natural to do this since krb5_unwrap_cammac_svc returns the cammac->elements array.

b. The behavior of k5_cammac_check_svcver() krb5_unwrap_cammac_svc() on a verification failure is to return 0 but with no string output. Should it return an error instead?

greghudson:
Hello!  I'm told you'll be working with us a lot, so I'm going to be extra-nitpicky in an effort to get closer to agreement on coding practices.  Some initial notes:

Do these need to be public APIs?  If so, they need doxygen comments, and to be added to lib/krb5_32.def and doc/appdev/refs/api/index.rst.  If not, they can have k5_ prefixes and be prototyped in k5-int.h.  They can always be made public APIs later, of course.

Please see http://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling and apply that to krb5_authind_decode().  Also, some nitpicks: (1) I'm not fond of the way you use "count" in that function; it seems more natural for maintain it so that it is a count of the indicators in the list; (2) I don't see a need to initialize i or j in that function, or to declare them on separate lines.

There are tabs in the changes to krb5.hin.  Please either make sure your editor can be configured not to use tabs in files with "indent-tabs-mode: nil" in the initial comment; if you can't, then be paranoid about that when making changes.  We do, unfortunately, have some files in the tree which do use tabs, for historical reasons.  Our style-checking script (http://k5wiki.kerberos.org/wiki/Coding_style/Style_checker) should alert you to issues.

In several places you have misaligned continuation lines, such as in the definitions of k5_cammac_check_svcver() and krb5_unwrap_cammac_svc().  Please make sure your editor can help you get this right, or be careful to get it right manually.  The style-checking script currently isn't clever enough to detect indentation issues.

strndup() is a somewhat recent POSIX addition, and we do not assume its presence on target platforms.  k5memdup0() can be used instead.

The new test in t_authdata.py contains a long line.  Please make sure your editor makes it obvious to you when you are exceeding 79 characters, and avoid that pretty much everywhere.  (There are a few exceptions.)

> Should krb5_authind_decode()'s authdata argument take a single authdata or an array? It seemed natural to do this since krb5_unwrap_cammac_svc returns the cammac->elements array.

Taking an array seems the most convenient to the caller, who doesn't care whether the indicators are expressed in separate containers.  However, the current APIs still seem a little inconvenient for a caller who doesn't care whether the indicators are in one or many CAMMAC containers.  I'm not immediately sure how to improve that.

> The behavior of k5_cammac_check_svcver() krb5_unwrap_cammac_svc() on a verification failure is to return 0 but with no string output. Should it return an error instead?

It should return an error; the caller can decide whether to ignore it.

npmccallum:
Having krb5_authind_decode() take a single authdata is my preference. It adds only minor inconvenience to people who need to assemble a whole set. Most users will be fine with incremental parsing.

Also, I need this API to be public so that I can use it in an out-of-tree kdb module's check_policy_tgs() callback.

mrogers950:
Thanks for the feedback! I've updated and addressed the following:
- Changed to k5_ functions with k5-int.h prototypes
- Fixed continuation lines, indents and other style issues
- Use k5memdup0()
- Refined indicator count/allocation loop in k5_authind_decode()
- k5_authind_decode() takes a single authdata
- k5_cammac_check_svcver() return an error on verification failure. simplifies k5_unwrap_cammac_svc()

npmccallum:
For the record, I no longer believe I need these APIs to be public.

@mrogers950, just glancing through the new patches reveals the new APIs to be not const correct.

greghudson:
I think I'd like to see more of the name attributes work before merging this, following the general principle that it's really hard to know whether you've added the right enabling functions until you start using them.  For example, I'm not sure we need all three of those functions to be visible to the tree outside of libkrb5.  (Functions purely internal to libkrb5 can be prototyped in lib/krb5/krb/int-proto.h, or other internal header files for different parts of libkrb5, rather than in k5-int.h.)

More comments on the PR as it exists now:

* Commit message summary lines must be limited to 50 characters or less, per http://chris.beams.io/posts/git-commit/ and git-commit(1).  Some projects treat this as a soft limit; we treat it as a hard limit.

* Use the imperative mood in commit messages and in comments describing functions.

* We don't use "@parameter" notation in comments describing functions (except for Doxygen comments in krb5.hin).  Just use the unadorned variable name.

* You haven't applied http://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling to k5_authind_decode().  After *authinds is set to NULL at the beginning, it should not be touched until the end of the function, and there should be no conditionalization on ret in the cleanup code.

* We prefer to have mainline logic (as opposed to error-handling) at the lowest possible indentation level, so having the entire mainline logic of k5_authind_decode() in a conditional is undesirable.  De-indent this code by jumping to cleanup if ad doesn't meet the conditions.

* It seems like we need a function to free the result of k5_authind_decode(); it's not good for the cleanup handler and every caller to have to use a for loop to free the list.

mrogers950:
I've updated this branch to include my progress on the name attribute support and the revised k5 functions.

tlyu:
Typo: ai_authdata.$(OBJECT) should be ai_authdata.$(OBJEXT); this broke the Windows build in AppVeyor.

greghudson:
I haven't had a chance to look at this yet, but here is output from the style checker as amended by PR #433:

    5bd5570 Add authentication indicator authdata plugin
      lib/krb5/krb/ai_authdata.c:
           68  Indentation may be incorrect
           69  Indentation may be incorrect
           70  Indentation may be incorrect
           91  Indentation may be incorrect
           92  Indentation may be incorrect
           93  Indentation may be incorrect
           94  Indentation may be incorrect
           95  Indentation may be incorrect
           96  Indentation may be incorrect
          113  Indentation may be incorrect
          114  Indentation may be incorrect
          115  Indentation may be incorrect
          116  Indentation may be incorrect
          117  Indentation may be incorrect
          209  Indentation may be incorrect
          210  Indentation may be incorrect
          211  Indentation may be incorrect
          212  Indentation may be incorrect
          213  Indentation may be incorrect
          214  Indentation may be incorrect
          234  Indentation may be incorrect
          235  Indentation may be incorrect
          236  Indentation may be incorrect
          292  Indentation may be incorrect
          293  Indentation may be incorrect
          294  Indentation may be incorrect
          295  Indentation may be incorrect
          296  Indentation may be incorrect
          297  Indentation may be incorrect
          298  Indentation may be incorrect
          299  Indentation may be incorrect
          300  Indentation may be incorrect
          337  Indentation may be incorrect
          338  Indentation may be incorrect
          339  Indentation may be incorrect
          340  Indentation may be incorrect
          341  Indentation may be incorrect
          342  Indentation may be incorrect
          353  Indentation may be incorrect
          354  Indentation may be incorrect
          355  Indentation may be incorrect
          356  Indentation may be incorrect
          357  Indentation may be incorrect
          396  Indentation may be incorrect
          397  Indentation may be incorrect
          398  Indentation may be incorrect
          399  Indentation may be incorrect
          410  Indentation may be incorrect
          419  Indentation may be incorrect
          420  Indentation may be incorrect
          421  Indentation may be incorrect
          422  Indentation may be incorrect
          423  Indentation may be incorrect
          436  Indentation may be incorrect
          464  Indentation may be incorrect
          465  Indentation may be incorrect
          466  Indentation may be incorrect
          467  Indentation may be incorrect
          468  Indentation may be incorrect
          545  Indentation may be incorrect
          546  Indentation may be incorrect
          547  Indentation may be incorrect
          548  Indentation may be incorrect
          549  Indentation may be incorrect
          556  Indentation may be incorrect
          557  Indentation may be incorrect
      lib/krb5/krb/authdata.h:
           81  Length exceeds 79 characters

I also see a lot of overwrapped parameter lists (one parameter per line when more than one would fit).

tlyu:
License statements for new files should be 2-clause BSD-style (see src/prototype). authind_internalize() in ai_authdata.c seems somewhat long. I haven't had a chance to examine the commit in detail yet.

mrogers950:
I've corrected the parameter lists, and the style checker complaint on authdata.h was for the length of the authdata ftable names, so I've shortened them.

greghudson:
This PR's changes are too expansive for one commit.  A possible commit breakdown:

* Remove the unused cammac_check_kdcver() from the KDC code
* Rename authdata ftable symbols for brevity
* Add libkrb5 CAMMAC functions
* Add AD_CAMMAC_PROTECTED authdata module flag
* Add ai_authdata module
* Tests

I haven't gone through all of the changes in detail, but I have some notes:

This functionality needs to be documented.  Unfortunately the analagous PAC functionality does not appear to be documented, so there's no guide.  We can add a name attributes section to doc/appdev/gssapi.rst.

cammac_check_svcver() should be removed from kdc_util.h as well as cammac.c.

The renamed authdata ftable symbols should use a k5_ rather than krb5int_ prefix.

k5_unwrap_cammac_svc() seems like it belongs in cammac.c, not auth_ind.c.  k5_authind_decode() could go inside ai_authdata.c, eliminating auth_ind.c.

krb5int_authdata_verify() currently does not check the return status of k5_get_kdc_issued_authdata(), and I believe that's intentional, if badly communicated--if there is faulty kdc-issued authdata in the ticket, we ignore it rather than causing rd_req to error out.  This PR changes that behavior with no comment, with potentially severe consequences.  As a coding practice, "just fix this lack of error-checking as we go" should usually be avoided; if it's wrong it can cause a serious problem which slips through code review, and if it's right the change belongs in a separate commit which can be backported.

I am confused about the AD_CAMMAC_PROTECTED flag and the division of responsibility between authdata.c and the authdata module in unwrapping and verifying CAMMACs.  It seems underdocumented and maybe problematic.

We have two existing authdata modules, in pac.c and s4u_authdata.c.  The filename ai_authdata.c is consistent with s4u_authdata.c, but in-tree modules for other pluggable interfaces would invert the name (e.g. localauth_rule.c, localauth_an2ln.c, etc.).  No need to make a change, I think.

authind_fini() can simply be omitted from the vtable; there should be no need for a stub function.

"urn:authentication-indicators:indicators" seems redundant.

In t_authind.py:
* use "out" instead of "output" to hold program output.  "output" is the name of a k5test.py function.  (Some other test scripts use "output"; that's an old mistake that just hasn't been corrected.)
* Where you look for hex strings, add comments for what they are hex encodings of.
* Use blank lines to communicate where one test ends and another begins.
* There's a missing space after a comma in the first realm.kinit() invocation.

mrogers950:
Thanks, I've pushed another update.

With this revision the verification has been moved outside of the authind module, with the verified authdata being passed to the module based on the KRB5_AUTHDATA_AUTH_INDICATOR module type, rather than a CAMMAC usage flag, (the usage flag remains AD_USAGE_AP_REQ). It should be clearer now when reading krb5int_verify_authdata() that some specific authind processing is being done before importing the authdata, in the same vein as KDC-issued authdata. 

Also included a commit to change the k5_get_kdc_issued_authdata() return code to return 0 on verification failure and success, and non-zero on other failures, rather than an assert(). This way KDC-issued authdata that is rejected is silently ignored and other failures will cause krb5int_authdata_verify() to error out.

Other changes:
- k5_extract_indicators(), and k5_get_indicator_authdata() are new functions in authdata.c to find and verify auth indicators.  
- The fini() and verify() modules are removed.
- The attribute being used now a non-URN style of 'auth-indicators'. We may want to pursue registering a URN namespace in the meanwhile.
- I've re-organized the new code with new commits. I eliminated auth_ind.c but moved k5_authind_decode() to authdata.c and added a prototype to k5-int.h; there is a direct test for this function in adata.c, so I did not make it private to ai_authdata.c.
- Added the name attribute documentation entry - there are probably some adjustments needed.

greghudson:
Looking at the fourth and fifth commits, I don't think we are on the same page on design.

authdata.c, the authorization data framework, should handle CAMMAC containers similarly to how it handles KDC-ISSUED containers.  A module should be able to say that it is interested in any type of authorization data element within a CAMMAC container, just as it can currently say that it is interested in any type of element within a KDC-ISSUED container.  The framework should take care of decoding CAMMAC containers.  If a module includes the CAMMAC usage flag, the framework should look within decoded CAMMAC containers for elements of the module's ad_type.  The framework should not have specific handling for the KRB5_AUTHDATA_AUTH_INDICATOR type; the whole point of a generic framework is that it does not need specific code for each leaf authdata type.

In the first commit:

* Use "if (ret)" instead of "if (ret != 0)" and "if (!ret)" instead of "if (ret == 0)".
* Do not use the "@variablename" convention in explanatory comments (as you did with "@server_key"); just use the variable name. 
* EINVAL is not a good return value for the verifier being invalid.  Consider making the validity value separate from the return value, to make it easier for callers to distinguish processing errors from invalid verifiers.
* There is no need to include krb5.h after including k5-int.h.

In the second commit:

* Use full sentences in the explanatory part of the commit message, not sentence fragments.
* The k5_authind_decode() comment should explicitly say that the caller must initialize *indicators to NULL before the first call.
* k5_authind_decode() is out of place in authdata.c, which implements the authorization data framework.  authdata_dec.c would be okay.  ai_authdata.c would also be okay, although it isn't present yet in the current commit order.
* The comments "Count existing" and "Count new" should say what they are counting (i.e. they should be complete sentences).
* Use two spaces after periods in comments, but not at the end of a comment before the "*/".

In the third commit:

* Are the adata.c changes required or related to the stated purpose of the commit?  I'm okay with having these changes in the same commit, but the commit message should describe both changes.
* Why change the loop in display_auth_indicator(), and why free the list incrementally?  (I also think you're leaking the data pointer element of each list element, and you're not checking the return value of malloc().  Test programs can just abort on malloc failure, but should still check.)

mrogers950:
> Looking at the fourth and fifth commits, I don't think we are on the same page on design.
> 

Thank you for the clarification. These changes should be closer to this design. I've addressed your other points as well.

> EINVAL is not a good return value for the verifier being invalid. Consider making the validity value separate from the return value, to make it easier for callers to distinguish processing errors from invalid verifiers.

I opted for returning KRB5KRB_AP_ERR_BAD_INTEGRITY based on the krb5_c_verify_checksum() 'valid' output, and look for this error in k5_extract_cammac().

> k5_authind_decode() is out of place in authdata.c, which implements the authorization data framework. authdata_dec.c would be okay. ai_authdata.c would also be okay, although it isn't present yet in the current commit order.

I've moved it to authdata_dec.c. I did this in a new commit just for the moment and I can ultimately squash it once I resolve the conflicts with other commits to authdata.c

> Why change the loop in display_auth_indicator(), and why free the list incrementally? (I also think you're leaking the data pointer element of each list element, and you're not checking the return value of malloc(). Test programs can just abort on malloc failure, but should still check.)

I can't recall if there was a good reason for the incremental free. I went back to using k5_free_data_ptr_list() at the end. Just now double checking my changes I forgot the malloc() check here. I'll correct that.

greghudson:
In "Authdata test adjustments," I still don't see any reason to change display_auth_indicator(), except to call k5_authind_decode() instead of decode_utf8_strings().  The old for loop should work fine.

I would prefer that "Handle CAMMAC authdata" comes before the auth indicator module itself in the commit sequence.  The addition of k5_authind_ad_client_ftable to authdata_systems can be part of the commit which adds the auth indicator module.  The addition of the AD_CAMMAC_PROTECTED flag should of course be part of the "Handle CAMMAC authdata" commit.

AD_CAMMAC_PROTECTED should be in AD_USAGE_MASK.  (Or maybe AD_USAGE_MASK should be removed, but you don't need to make that change in this PR.)

"Handle CAMMAC authdata" adds three static helper functions to authdata.c.  Static helper functions should not use k5_ or krb5int_ or krb5_ prefixes.  They should have explanatory comments before their definitions.  k5_find_cammac_authdata() and k5_get_cammac_authdata() have confusingly similar names.  I would suggest folding find into get, as it's making just one function call.

mrogers950:
I've pushed another update with those corrections.

greghudson:
libkrb5.exports is sorted (well, mostly; it appears people have made some mistakes).  Please don't make it less so.

display_auth_indicator() in adata.c needs to initialize strs to NULL; I believe that's why the Travis check failed.

mrogers950:
Updated.

greghudson:
In the "pr410" branch of my fork, I have two commits which replace the first four commits of this PR.  Please take a look at what I did there and rebase the other five commits onto it if it looks okay.  You'll need to get rid of the changes to authdata_plugin.h in your fifth commit as they were moved into the authdata.c commit.

For the ai_authdata.c commit, please:

* Remove the AD_USAGE_AP_REQ flag, as it could allow unprotected auth indicators to be accepted.
* Get rid of the "authenticated" flag in authind_context and the serialization, since we don't traffic in unauthenticated indicators.
* Fold update_authind_data() into its only caller, which does very little besides call it.
* In update_authind_data(), either remove the count == 0 ENOENT check or explain it.
* In update_authind_data(), rename "tmp".  All local variables are temporary.
* Explain in a comment what you're doing with the "more" value in authind_get_attribute(), or simplify it.

mrogers950:
Your changes are fine with me and there were no problems in test.  I've made the following adjustments:
- AD_USAGE_AP_REQ is removed from the module flags.
- My understanding of the use of more is the caller of gss_get_name_attribute() sets it to -1 on the first invocation and is called until more reaches 0. The more flag logic in authind_get_attribute() was taken from the s4u authdata code and I agree was difficult to read (it stores a number less than -1 for the count, which could also be unexpected by callers reading the more value).  My changes and comment to this should make things clearer since more reflects the actual number of attributes left, and simply ends on 0.
- Folded update_authind_data() into authind_import_authdata() and changed the ENOENT to return 0, to continue when there's nothing to import.
- Removed the authenticated flag from aictx and from serialization, size, and copy operations.
- Removed the restrict_authenticated check from authind_export_internal. All of the indicators are authenticated, and restrict_authenticated == FALSE does not (to me) imply that it should only return un-authenticated authdata; so I dont see the need for this check.

greghudson:
In authind_import_authdata(), after the k5_authind_decode() loop, you are going to a bunch of effort to copy all of the values into a new array (of krb5_data instead of krb5_data *), adding zero bytes.  But nothing that uses the resulting array seems to need trailing zero bytes.  Why not just store the indps pointer in the context?

authind_copy() is doing something peculiar and non-standards-conformant with the cast to void **.  I need to look more carefully at this and the export/import stuff.

greghudson:
After reviewing how the authdata framework uses the export_authdata method, I believe the idea is that if a GSS caller sets name attributes on a GSS name and acquires initiator creds with that name, krb5 authdata can be generated for TGS-REQs and AP-REQs based on those attributes.  I don't think this functionality is very well exercised by current code.

It doesn't seem like we need this for auth indicators.  Therefore, I think authind_export_authdata can be removed and set to NULL in the vtable, and export_ind can be removed from struct authind_context.

I will confess to being confused as to why pac.c and s4u_authdata.c have export_authdata functions:  s4u_authdata.c doesn't allow any attributes to be set, and while pac.c does, I don't think it ever makes sense to include WIN2K_PAC authdata in a TGS-REQ or AP-REQ.  Even given my confusion, I think it's still clear that we don't need this method for ai_authdata.c.

(I still need to review the usage of export_internal and copy, but I'm turning my attention back to the iprop improvements since those should be just about ready for merging.)

greghudson:
I think we can also null out export_internal in ai_authdata.c.  This method is only used for gss_map_name_to_any(), which was present in version 5 of draft-ietf-kitten-gssapi-naming-exts but isn't present in the final standard.  (So it really shouldn't be in our code, we should have caught that during review when naming extensions implemented, and going forward we should stub out the GSS entry point and remove all of the internal complexity it generates.)

greghudson:
Finally, I think we can null out the copy method in ai_authdata.c.  The authdata framework falls back to externalizing and internalizing, which should be roughly as efficient as the current copy method implementation.

mrogers950:
I've been testing the removal of those methods one at a time - when I test after only removing the copy method, there's a failure in the s4u tests.  
`*** [15] Executing: ./t_s4u p:user@KRBTEST.COM p:service/2@KRBTEST.COM
gss_get_name_attribute: The operation or option is not available or unsupported
gss_get_name_attribute: No such file or directory
Protocol transition tests follow
-----------------------------------

Target name:    service/1@KRBTEST.COM
Target mech:    { 1 2 840 113554 1 2 2 }
Source name:    user@KRBTEST.COM
Source mech:    { 1 2 840 113554 1 2 2 }
*** [15] Completed with return code 1
*** Failure: ./t_s4u failed with code 1.
`
I've been debugging, comparing this with the code when authind_copy() is present in the table, but I haven't figured out how that change alone affects the s4u results.

greghudson:
That test failure goes away if, in authind_get_attribute_types(), "if (aictx->indicators == NULL)" is replaced with "if (aictx->count == 0)".  For a counted array, always test the count, not the pointer.

While I was looking at this, I noticed that internal_unpack_ind_data() allocates an extra krb5_data element and sets it to empty_data().  Why?

mrogers950:
I've NULLed and removed the copy and export functions, and corrected that check in authind_get_attribute_types(). The context now stores the krb5_data pointers from k5_authind_decode and the export authdata copy is removed, and the rest of the code is updated to accommodate the access and freeing of the pointer list.
Along with this, internal_unpack_ind_data() now uses a krb5_data pointer list instead of the bounded array of krb5_data.

greghudson:
Okay, the module is much easier to read now.  I think the following changes should suffice for that commit:

* Remove the version number from the serialized format.

* Fold the internal_unpack_ functions into the caller.  There isn't enough there to need helper functions.  (If we did need helper functions, there would be no need for an "internal_" prefix on their names.)

* Where you check the length against 65535, check it against remain (the number of remaining bytes) instead.  That's a loose limit, but it should prevent absurd allocations without imposing a hardcoded limit on the number of indicators.  When unserializing each indicator, check the length against remain before calling alloc_data(), for similar reasons.

* When fetching the length for each authdata element, check the krb5_ser_unpack_int32 return code.  Otherwise ibuf could be used uninitialized.

* When checking the result of k5alloc(), just check for the pointer being null, as you do elsewhere.

* Use uint8_t instead of krb5_octet throghout the file.  (We're in a slow transition to the use of C99 type names over krb5-specific type names.)

* Abbreviate the identifier "authentication_indicators_attr" to "authind_attr".

* Make sure to double-check this commit with the style checker; it currently finds four issues which should be fixed.

As noted in voice discussion, I am somewhat concerned about each authdata module adding to the representation of an exported composite name, even if there is no authdata for the module in the ticket.  But we can deal with that problem separately from this PR.

greghudson:
Looking at the ~3 commit:

* "Ignore rejected KDC-issued authdata" isn't a great commit summary.  We're already ignoring it; the net effect of this change is to fail authentication to some kinds of errors which are currently ignored.  (ENOMEM is an obvious possibility, but we could also run into decoding errors for IF-RELEVANT and AD-KDC-ISSUED values which aren't valid DER representations of the appropriate ASN.1 type.)

* In addition to KRB5KRB_AP_ERR_BAD_INTEGRITY and KRB5KRB_AP_ERR_INAPP_CKSUM, a bad checksum could induce KRB5_BAD_ENCTYPE or KRB5_BAD_MSIZE.  It's probably fine to check only for KRB5KRB_AP_ERR_BAD_INTEGRITY (the wrong key was used to compute the checksum, but the authdata generator otherwise obeyed protocol constraints); if you're going to check for KRB5KRB_AP_ERR_INAPP_CKSUM, then consistency suggests also checking for the other two.

The ~1 commit is simple and looks good.

In the final documentation commit:

* I don't think we want to reprise RFC 6680 function descriptions in gssapi.rst.  Having man pages for GSS functions would be useful, but we haven't designed the appropriate doc build infrastructure for that, and it should be separate from our main long-form documentation.  So, I would prefer that the documentation just briefly reference RFC 6680 and then go on to talk about the auth-indicator attribute.

* "In release 1.15 this attribute type will be included in the gss_inquire_name output" is overstating things; the attribute type will only appear when there was an auth indicator in the ticket.

* Reference the existing auth indicator documentation in the new gssapi.rst contents (with ":ref:`auth_indicator`"), and add a brief mention of this name attribute, with a reference, to auth_indicator.rst.

mrogers950:
- Removed version number from serialization
- Inlined unpack helper functions
- Added the length > remain checks
- Check the missed krb5_ser_unpack_int32 return
- Fix k5alloc checks
- Changed authentication_indicators_attr -> authind_attr
- Resolve style check issues, minor formatting  and comment cleanup
- Changed local krb5_octet * to uint8_t *. 
- Revised commit message for KDC-issued failure change and added KRB5_BAD_MSIZE and KRB5_BAD_ENCTYPE to the failure check, to be thorough.
- Removed RFC text from doc commit, added references to and from auth_indicators.rst

greghudson:
I will make this change myself, but to be clear: the krb5 type names for fixed-width types (krb5_int32, krb5_octet, etc.) should never be used in new code, even to match existing prototypes.  The equivalent C99 type names are interchangeable with the krb5 type names and should be used instead.

greghudson:
Pushed to master as af2442f04e408bfa04cbe2e317be982ec47af674 and preceding.

greghudson:
I apparently let a ton of stuff past my review.

Matt, please review PR #469 and PR #472, and then prepare a PR fixing the following problems in authdata.c:

* krb5int_authdata_verify() does not free cammac_authdata.

* Despite several rounds of back and forth on the design of f4f619e7905d762204695dd66450e586c183c9fd, krb5int_authdata_verify() still does not use module->ad_type when module->flags contains AD_CAMMAC_PROTECTED and there is a valid CAMMAC in the ticket.  The function should filter cammac_authdata by module->ad_type just as it does for KDC-issued authdata.

PR #411, opened 2016-02-23
Use blocking lock when creating db2 KDB

greghudson:
In 1.11 we switched from non-blocking to blocking locks in the DB2
module, but we missed one call to krb5_lock_file() in ctx_create_db().
This non-blocking lock can cause krb5_db_promote() to fail if the
database is locked when we try to promote the DB, in turn causing
kdb5_util load to fail.  Correct this call to make krb5_db_promote()
more robust.

ticket: 8367 (new)
target_version: 1.14-next
target_version: 1.13-next
tags: pullup


tlyu:
This looks ok to me.

PR #412, opened 2016-02-25
Enable kinit long options on linux [KRB-9]

mrogers950:
Add a configure check for getopt.h


greghudson:
Why check for the getopt.h header and not for getopt_long() directly?  I don't know if there are any systems with getopt.h but no getopt_long().

I am somewhat uncomfortable with having long option support on some Unix platforms but not others, and I don't have a sense of how extensive native support for getopt_long() is.  (It's a Gnu extension, which I believe has been adopted by the BSDs, but it may not be present on commercial Unix platforms since it doesn't appear to be in POSIX.)

We have an implementation of getopt_long() from NetBSD in src/util/windows.  We could move this to src/util/support and build it conditionally like we do some of the other files there.

mrogers950:
Thanks, I'll change it to check for getopt_long() directly, and if it's not available build the implementation from src/util/support.

tlyu:
Linking to JIRA.

mrogers950:
I've pushed an update. It would be good to get some confirmation that I'm not missing any required changes to the Windows build.

greghudson:
I think these adjustments should be made for better consistency with how we handle other functions like strlcpy:

* Please remove the "k5" prefix from k5getopt.c and k5getopt_long.c.
* Place required declarations in k5-platform.h.  All conditionalization should be done there; kinit.c should be able to just use getopt_long() without ifdefs or special includes, and clients/kinit/Makefile.in should not need a LOCALINCLUDES.
* Maintain the lexical sorting order of BSDFILES in the top-level Makefile.in.

In getopt.c and getopt_long.c, you should be able to use #defines (probably just the ones already done in k5-platform.h) to avoid modifying the identifiers in the BSD source.

libkrb5support-fixed.exports should not contain the getopt functions, as they are conditionally defined.

mrogers950:
Thanks! I've pushed an update.

greghudson:
Please use k5_ rather than krb5int_ for all new symbols.  (I appreciate the consistency with strlcpy, but this is a case where we are transitioning away from krb5int.)  In getopt.c and getopt_long.c, you shouldn't need to include the k5_ prefixes on the symbols you're defining as you can rely on the defines in k5-platform.h.

In k5-platform.h, I don't understand why there would be a specific need to protect these new symbols with extern "C" but not the other ones.  I also don't know what's going on with the throw() in the getopt prototype for C++.

On Unix platforms without getopt_long(), you're using the in-tree getopt() as well as the in-tree getopt_long().  This might be necessary, but I'm not certain.  If we kill optreset (a BSD extension not present in the Linux implementation or POSIX), then our in-tree getopt_long can just modify the libc optind/optarg/optopt values, I think.

mrogers950:
> In k5-platform.h, I don't understand why there would be a specific need to protect these new symbols with extern "C" but not the other ones. I also don't know what's going on with the throw() in the getopt prototype for C++.

Ah, I should have mentioned this. There are some c++ library compatibility tests (src/tests/misc) that were failing to compile unless I did that. Same with the throw(), g++ complained about the declaration having a different exception specifier.  

> On Unix platforms without getopt_long(), you're using the in-tree getopt() as well as the in-tree getopt_long(). This might be necessary, but I'm not certain. If we kill optreset (a BSD extension not present in the Linux implementation or POSIX), then our in-tree getopt_long can just modify the libc optind/optarg/optopt values, I think.

That's right, and it seemed necessary for the sake of simplicity. The system getopt() and associated vars are from unistd.h, which is included at different places and ends up conflicting with the in-tree variables if you're using the system getopt() and the in-tree getopt_long(). I originally went down this path but I didn't see a good way to do it without code rearrangement and splitting up compilation of both in-tree functions, so I opted for this. It was also my understanding that the Win build used both in-tree functions anyways.

greghudson:
> Ah, I should have mentioned this. There are some c++ library compatibility tests (src/tests/misc) that were failing to compile unless I did that. Same with the throw(), g++ complained about the declaration having a different exception specifier.

It's not okay for the compiler to see our prototype of getopt() and the system's prototype of getopt() under the same symbol.  Making the prototypes match is not a solution; we cannot possibly match the system's prototype on every platform.  This is, I think, a compelling reason not to use our in-tree getopt() on Unix platforms which don't have getopt_long().

greghudson:
For Unix platforms, assume that unistd.h exists, and declares getopt(), optind, optarg, opterr, and optopt.  Do not #define any of those symbols.  The in-tree getopt_long implementation can modify the libc optind/optind/optarg/opterr variables (no #defines for those).

For Windows, do #define getopt/optind/etc. and use the in-tree getopt implementation.

SaharahSarah:
The Windows build fails. Build output here: http://pastebin.com/3dsUYhw0

greghudson:
Sorry if I'm slow to respond.  github notifies me when there are new PR comments but not when there are changes to the branch.

Are the extern "C" bits in k5-platform.h still required for some reason?

You have some odd spacing in the k5_getopt() and k5_getopt_long() prototypes, like "char**" instead of "char **".  Also, we usually put parameter names in prototypes.

It seems like k5-platform.h should include <unistd.h> near the top (conditional on HAVE_UNISTD_H so we don't include it on Windows), so that getopt() is available to all consumers.  Right now it's only included if HAVE_GETOPT_LONG is defined.

greghudson:
It looks like the Windows build error can be resolved by fixing the ##DOS## comments after the definitions of GETOPT_LONG_ST_OBJ and GETOPT_LONG_OBJ.  Right now they erroneously define GETOPT_ST_OBJ and GETOPT_OBJ.

mrogers950:
Thanks for reviewing! Updates are pushed.

> Are the extern "C" bits in k5-platform.h still required for some reason?

After the suggested changes it appears that these are not needed anymore, so I've removed them. 

> You have some odd spacing in the k5_getopt() and k5_getopt_long() prototypes, like "char*" instead of "char *". Also, we usually put parameter names in prototypes.

Fixed.

> It seems like k5-platform.h should include near the top (conditional on HAVE_UNISTD_H so we don't include it on Windows), so that getopt() is available to all consumers. Right now it's only included if HAVE_GETOPT_LONG is defined.

I believe you meant to say unistd.h instead of k5-platform.h here.. I've moved that include to the top of k5-platform.h, outside of the HAVE_GETOPT_LONG conditional.

> It looks like the Windows build error can be resolved by fixing the ##DOS## comments after the definitions of GETOPT_LONG_ST_OBJ and GETOPT_LONG_OBJ. Right now they erroneously define GETOPT_ST_OBJ and GETOPT_OBJ.

Good catch, thank you.

SaharahSarah:
New build failure in KfW: http://pastebin.com/6CjVVJUc

greghudson:
Thanks, Sarah.  In util/support/Makefile.in, it looks like there is a typo in the definition of EXTRA_SUPPORT_SYMS for Windows; k5_optop should be k5_optopt.

SaharahSarah:
Failure in ms2mit. Can't find getopt: http://pastebin.com/D5B19nvc

greghudson:
I think everywhere this patch removes $(BUILDTOP)\util\windows\$(OUTPRE)getopt.obj, it needs to add $(SLIB) in its place, to link in libkrb5support.

SaharahSarah:
Would it be possible to rebase the branch on top of the current master? I have to do that every time to get KfW to build since a fix for the KfW build was pushed to master recently.

SaharahSarah:
New KfW error: klist can't find getopt.h

http://pastebin.com/FP8R8J2z

greghudson:
klist.c, kvno.c, and kdestroy.c need to remove "#include <getopt.h>" and the surrounding #ifdef/#endif.

mrogers950:
I've rebased and updated.

mrogers950:
Are there any issues left?

greghudson:
I don't believe we have retried the Windows build since the last update.  (We broke the Windows build with the KDC hooks commit, which may be delaying things.)

tlyu:
Currently, #419 is a higher priority, but we should be able to merge this PR after we merge that one.

greghudson:
Using appveyor, I found what are hopefully the last Windows build issues.  I submitted fixes in the form of a PR against the PR branch, which I hope works reasonably well.  (I don't have appveyor testing the installer build yet, so those changes I made by inspection.)

greghudson:
Pushed to master as 8f9ade8ec50cde1176411085294f85ecfb2820a4.  (I took care of merging in the changes from the PR I submitted.)

PR #414, opened 2016-02-25
Remove hdb KDB module

greghudson:
The hdb module has bitrotted, both against our internal interfaces and
against Heimdal's database layer.  It is now possible to dump Heimdal
databases in the MIT krb5 dump format, which is a better option than
bridging between the database layers.


npmccallum:
FWIW, I support this patch and it LGTM. :+1:

tlyu:
This also looks ok to me.

PR #415, opened 2016-02-25
Fix Makefiles for VS2010 KfW build

tlyu:
The new Makefile conditionals in commit
4552159e97007a45370dd49fa6b9fb963bb7d160 don't behave properly if
VISUALSTUDIOVERSION isn't set, probably due to the way nmake orders
macro expansion and boolean short circuiting.  Use nested conditionals
instead.


greghudson:
Looks fine to me.

kaduk:
Also looks fine to me.

SaharahSarah:
Just tested the build. It definitely fixes the problem :)

PR #417, opened 2016-02-26
Admit servers sending incomplete acceptor tokens

simo5:
We've found at least one HTTP/Negotiate implementation in Java that does
not set anything but the responseToken field in the SPNEGO acceptor response
token.
This is technically a violation of RFC 4178 (4.2.2. negTokenResp), but
it is harmless to support as we know what is the mechanism we were trying
to negotiate, and can use that mechanism to check the token.

These implementations are probably not supporting any real negotiation,
as the missing negState precludes any mechanism negotiation on failure.
In case a supportedMech is actually returned that differs from the
opportunistic one but no negState is provided the reselection will fail
with GSS_S_DEFECIVE_TOKEN as it should.


greghudson:
Pushed to master as da748e7621ad20237f105eb1167832d4898fde66

PR #418, opened 2016-02-29
Skip unnecessary mech calls in gss_inquire_cred()

greghudson:
[The diff is best viewed with "git diff -w"]

If the caller does not request a name, lifetime, or cred_usage when
calling gss_inquire_cred(), service the call by copying the mechanism
list (if requested) but do not call into the mech.

This change alleviates an issue where SPNEGO can fail in the presence
of expired krb5 credentials rather than proceeding with a different
mechanism, or can resolve a krb5 credential with the benefit of the
target name.


tlyu:
This looks ok to me.

PR #419, opened 2016-03-02
Default to LSA when TGT in LSA is inaccessible [KRB-13]

SaharahSarah:
When UAC is enabled and a domain user with Administrator privileges logs
in the TGT is inaccessible.  This is because if the TGT is accessible it
may be possible for a non-elevated user session to get a ticket with an
unfiltered PAC and bypass the UAC.  This causes a bug where the current
tickets in the LSA cache are all copied to the ccache except the TGT,
which prevents a user session from getting new tickets.  In this case,
ms2mit will now set the LSA cache to the default cache to workaround
this issue.

### DIFF
--- a/src/windows/ms2mit/ms2mit.c
+++ b/src/windows/ms2mit/ms2mit.c
@@ -39,19 +39,16 @@ xusage(void)
     exit(1);
 }
 
-void
-main(
-    int argc,
-    char *argv[]
-    )
+int
+main(int argc, char *argv[])
 {
-    krb5_context kcontext;
+    krb5_context kcontext = NULL;
     krb5_error_code code;
     krb5_ccache ccache=NULL;
     krb5_ccache mslsa_ccache=NULL;
     krb5_cc_cursor cursor;
     krb5_creds creds;
-    krb5_principal princ;
+    krb5_principal princ = NULL;
     int initial_ticket = 0;
     int option;
     char * ccachestr = 0;
@@ -73,28 +70,23 @@ main(
 
     if (code = krb5_init_context(&kcontext)) {
         com_err(argv[0], code, "while initializing kerberos library");
-        exit(1);
+        goto cleanup;
     }
 
     if (code = krb5_cc_resolve(kcontext, "MSLSA:", &mslsa_ccache)) {
         com_err(argv[0], code, "while opening MS LSA ccache");
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
     if (code = krb5_cc_set_flags(kcontext, mslsa_ccache, KRB5_TC_NOTICKET)) {
         com_err(argv[0], code, "while setting KRB5_TC_NOTICKET flag");
-        krb5_cc_close(kcontext, mslsa_ccache);
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
     /* Enumerate tickets from cache looking for an initial ticket */
     if ((code = krb5_cc_start_seq_get(kcontext, mslsa_ccache, &cursor))) {
         com_err(argv[0], code, "while initiating the cred sequence of MS LSA ccache");
-        krb5_cc_close(kcontext, mslsa_ccache);
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
     while (!(code = krb5_cc_next_cred(kcontext, mslsa_ccache, &cursor, &creds)))
@@ -110,24 +102,18 @@ main(
 
     if (code = krb5_cc_set_flags(kcontext, mslsa_ccache, 0)) {
         com_err(argv[0], code, "while clearing flags");
-        krb5_cc_close(kcontext, mslsa_ccache);
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
     if ( !initial_ticket ) {
         fprintf(stderr, "%s: Initial Ticket Getting Tickets are not available from the MS LSA\n",
                 argv[0]);
-        krb5_cc_close(kcontext, mslsa_ccache);
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
     if (code = krb5_cc_get_principal(kcontext, mslsa_ccache, &princ)) {
         com_err(argv[0], code, "while obtaining MS LSA principal");
-        krb5_cc_close(kcontext, mslsa_ccache);
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
     if (ccachestr)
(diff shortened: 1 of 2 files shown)


greghudson:
The is_tgs_principal() test on the server principal will return true for cross-realm TGTs, which doesn't seem right.  It might be simpler to just keep the existing test for the INITIAL flag, which should be adequate in practice.

If there is only an uncopyable TGT in the LSA cache, we still want to set the default to MSLSA; the code in the PR will not do so.  I'm not sure if we can implement the desired behavior without looping over the MSLSA cache twice--once with TC_NOTICKET to find out if there's a TGT, and once without to determine if we can read the TGT.

krb5_free_principal() and krb5_free_context() are no-ops on null pointers (as are all krb5_free functions), so we do not need conditionals before calling them in the cleanup code, and our practice is not to conditionalize such calls.  We do need conditionals on the calls to krb5_cc_close(), although that should arguably change at some point.

SaharahSarah:
I know at one point @tlyu mentioned that TGT's aren't the only tickets that can have the INITIAL flag, for instance during a password change. At least we could filter out any tickets that could have an INITIAL flag but not be a TGT.

SaharahSarah:
OK, commit changed to set KRB5_TC_NOTICKET while checking if the MSLSA has any tickets. Also, a new commit is there for setting API as the default if no name is specifically specified to ms2mit and credentials were successfully copied from MSLSA. Still haven't decided if it should be a separate commit or added to the first commit. Thoughts are appreciated.

greghudson:
The existing breakdown of commits seems good to me.

In the first commit, the handling of error codes seems problematic.  A krb5_error_code value should not be directly returned from main(), although "return code ? 1 : 0;" would be reasonable.  has_tickets() should not use errno to communicate a krb5_error_code value, as errno should only ever contain values from <errno.h>.  Making has_tickets() return a krb5_error_code and set a boolean through an output parameter would be cleaner.

The check for is_tgs_principal() still returns true on cross-realm TGTs, which aren't interesting starting creds.  I suggest borrowing clients/klist/klist.c:is_local_tgt() instead of the KDC function.

tlyu:
I would remove wording in the commit message about a filtered PAC; it is probably too speculative. Further information (from Nico, I think) sent in IRC implies that the filtered access token that the MSDN documentation refers to is a kernel internal state structure (containing SIDs, etc.) rather than a Kerberos ticket, so it's probably not the PAC in the ticket that's getting filtered.

It's probably sufficient to say that giving full access to the TGT would somehow allow unprivileged software to bypass UAC. It's also probably more accurate to say that it's the TGT session keys that are inaccessible, rather than the entire LSA cache entry for the TGT. (The Windows native klist.exe can display a hexdump of the TGT and session key, but the key displays as all zeros in a UAC-restricted session.) I know this probably sounds overly pedantic, but information in this area is hard to come by, so increased accuracy will help people working on this code in the future. (I couldn't readily find anything authoritative about this from Microsoft.)

tlyu:
Also, I would find the first commit easier to review if the "goto cleanup" changes were split into a separate commit.

SaharahSarah:
When we set the default to API, should it be "API:" or "API:krb5"?

kaduk:
Just "API:".  It's possible to create separate sub-caches within the collection by using a trailing string, like "API:kadukroot" or "API:joeuser", but the default for leash should just be "API:" as it has been in the past.

SaharahSarah:
OK Cool. I was just asking because without the ms2mit changes it just automatically uses a cache set to API:krb5

src/windows/ms2mit/ms2mit.c
```diff
@@ -39,20 +41,61 @@ xusage(void)
...
-    char *argv[]
-    )
+/* Return true if princ is not a cross-realm principal. */
+krb5_boolean
+is_local_tgt(krb5_principal princ)
 {
-    krb5_context kcontext;
+    return princ->length == 2 &&
+    data_eq_string(princ->data[0], KRB5_TGS_NAME) &&
+    data_eq(princ->realm, princ->data[1]);
```
greghudson: The continuation lines should be indented four spaces, per the BSD style.

src/windows/ms2mit/ms2mit.c
```diff
@@ -39,20 +41,61 @@ xusage(void)
     exit(1);
 }
 
-void
-main(
-    int argc,
-    char *argv[]
-    )
+/* Return true if princ is not a cross-realm principal. */
```
greghudson: ... and is a TGT principal?  I suggest "Return true if princ is a local (not cross-realm) krbtgt principal."

src/windows/ms2mit/ms2mit.c
```diff
@@ -39,20 +41,61 @@ xusage(void)
...
+    if (code)
+        return code;
+
+    while (!has_tickets) {
+        code = krb5_cc_next_cred(kcontext, ccache, &cursor, &creds);
+        if (code)
+            break;
+
+        if ((!krb5_is_config_principal(kcontext, creds.server)) &&
+            (creds.times.endtime - time(0) > 0))
```
greghudson: This is over-parenthesized (you don't need the parens around either arm of the && operator).  I also suggest pulling the expression time(0) out of the loop, assigning its value to a krb5_timestamp variable, and using "creds.times.endtime > now" rather than subtraction.

src/windows/ms2mit/ms2mit.c
```diff
@@ -73,95 +116,85 @@ main(
...
         com_err (argv[0], code, "while copying MS LSA ccache to default ccache");
-        krb5_free_principal(kcontext, princ);
-        krb5_cc_close(kcontext, ccache);
-        krb5_cc_close(kcontext, mslsa_ccache);
-        krb5_free_context(kcontext);
-        exit(1);
+        goto cleanup;
     }
 
+    if (!ccachestr) {
```
greghudson: "if (ccachestr == NULL) {"

greghudson:
The code changes look good apart from the few style issues I noted.  I would lean towards not setting the default to LSA if ccachestr is provided--that is, keep ms2mit as a simple copy operation when the destination is specified.

I've changed my mind about the breakdown of commits; I think the second and third commits should be squished.  We've changed the contract of ms2lsa from "copy from LSA to the default ccache" to "set the registry default to either the LSA cache or a copied API cache", and it doesn't make sense to spread that change over two commits.

In the existing commit breakdown, the third commit message would need a shorter summary line and would need to talk about the change in default destination.  But that's moot if the second and third commits are combined.

tlyu:
Tag with JIRA issue.

tlyu:
Rebased, with fixup for the calling convention mismatch on master, at tlyu/krb5@729c0f551cc806329704dd1a0b85107fb1c2a83d.

SaharahSarah:
cc_has_tickets had a bug that I fixed. Everything else looks fine and it builds on Windows. Go ahead an push it when you would like. :+1:

tlyu:
Updated again at tlyu/krb5@117218c0a659598cef7b98e87deb41b1eb108923

tlyu:
Pushed to master as e033a81c891030741952e4743a0b5503bdbcea17..33b862799efa65b16e2acd1510c84d9f1ded2cbb.

PR #420, opened 2016-03-03
Add LDAP KDB authentication indicator handling

mrogers950:
This patch stores and retrieves authentication indicators in a new attribute "krbPrincipalAuthInd" during krb5_ldap_put_principal() and krb5_ldap_get_principal().

The OID for krbPrincipalAuthInd was just chosen for testing and will likely need to be replaced with an assigned OID

Usage scenario: http://www.freeipa.org/page/V4/Authentication_Indicators#Verify_Authentication_Indicators_During_Ticket_Issuance_.28TGSReq.29


greghudson:
Picking OIDs for new kerberos.schema attributes is unfortunately harder than it ought to be.  I will have to do some groundwork on that.  (Using an OID from an arc owned by Red Hat or the FreeIPA project is also an option.)

In the schema, add the new attribute to krbPrincipalAux where all of the existing principal attributes are, not to krbPrincipal.  Do not use a case-ignore match rule when the KDC treats the value as case-sensitive.

It's pretty much never right to add 20+ lines of open code to an oversized function like populate_krb5_db_entry(); use a helper function.  Also use the functions in k5-buf.h for this, instead of a count-allocate-fill pattern.  The additions to krb5_ldap_put_principal() should also use a helper function.

From libkdb5's perspective, all of a principal's string attributes are smushed together into a single tl-data item of type KRB5_TL_STRING_ATTRS, with alternating keys and values.  The code in the PR seems to assume that there is a separate KRB5_TL_STRING_ATTRS tl-data item for each string attribute, which isn't the case--we never repeat the same tl-data type multiple times in a tl-data list.

On read, a simpler way to make krbPrincipalAuthInd override krbExtraData on load would be to process krbPrincipalAuthInd after krbExtraData, and let krb5_dbe_set_string() overwrite the attribute.
I don't have a simple suggestion to handle writing, as the behavior is a little complicated.  Per the design here, a KRB5_TL_STRING_ATTRS tl-data item containing only a require_auth value should be dropped entirely, while an item containing require_auth and other attributes should be filtered.

We should also consider transition issues.  During KDC upgrades, the same LDAP database might be used by multiple versions of the KDC and kadmind.  If require_auth values are filtered out of krbExtraData on write, 1.14 KDCs won't see them, which could result in a privilege escalation.  Storing the require_auth values redundantly in both krbPrincipalAuthInd and krbExtraData, at least for now, would resolve that problem.  It's also easier to implement.

mrogers950:
Thanks for the review. I pushed an update that should address everything you mention. It also makes sense to have the delstr functionality to go along with this, so I will add that in to this PR.

If you would prefer to use a RH/IPA OID, then we can do that.

src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
```diff
@@ -1330,6 +1330,43 @@ free_princ_ent_contents(osa_princ_ent_t princ_ent)
...
+    char **auth_inds = NULL;
+    struct k5buf buf = EMPTY_K5BUF;
+
+    auth_inds = ldap_get_values(ld, ldap_ent, "krbPrincipalAuthInd");
+    if (auth_inds == NULL)
+        return 0;
+
+    k5_buf_init_dynamic(&buf);
+
+    /* space seperated list of indicators */
```
greghudson: Comments should usually be complete, punctuated sentences; in this case, I suggest "Make a space-separated list of indicators."  There are other comments in this commit which are complete sentences but lack punctuation; I won't comment on all of them but please fix them as well.

src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
```diff
@@ -1330,6 +1330,43 @@ free_princ_ent_contents(osa_princ_ent_t princ_ent)
...
+            k5_buf_add(&buf, " ");
+    }
+
+    ret = k5_buf_status(&buf);
+    if (ret != 0)
+        goto cleanup;
+
+    if (buf.data != NULL)
+        krb5_dbe_set_string(context, entry, KRB5_KDB_SK_REQUIRE_AUTH,
+                            (char *)buf.data);
```
greghudson: We require braces around any flow control statement body which doesn't fit on one line, even if it's only one statement.  buf.data does not need a cast, as C supports implicit casts from void *.  Finally, you are discarding the return value of krb5_dbe_set_string() here; it should be assigned to ret.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -618,6 +618,36 @@ tl_data2berval (krb5_tl_data *in, struct berval **out)
     return 0;
 }
 
+static krb5_error_code
```
greghudson: This helper function needs a brief descriptive comment.

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
```diff
@@ -618,6 +618,36 @@ tl_data2berval (krb5_tl_data *in, struct berval **out)
...
+    while (ai != NULL && i < sv_num) {
+        strval[i++] = ai;
+        ai = strtok_r(NULL, " ", &ai_save);
+    }
+
+    ret = krb5_add_str_mem_ldap_mod(mods, "krbPrincipalAuthInd",
+                                    LDAP_MOD_REPLACE, strval);
+
+cleanup:
+    free(auth_ind);
```
greghudson: Use krb5_dbe_free_string() instead of free() for this, per the contract of krb5_dbe_get_string().  (It doesn't matter in practice since the whole tree is compiled with the same malloc/free, but it's good hygiene.)

src/tests/t_kdb.py
```diff
@@ -117,6 +117,16 @@ def kldaputil(args, **kw):
 if not ldapmodify:
     skip_rest('some LDAP KDB tests', 'ldapmodify not found')
 
+ldapsearch = which('ldapsearch')
```
greghudson: Although it is very likely that ldapmodify and ldapsearch are either both present or both absent, we should explicitly check.  Put the ldapsearch assignment up next to the ldapmodify assignment, and change the conditional on skip_rest() to skip the rest of the script if either is not found.

src/tests/t_kdb.py
```diff
@@ -117,6 +117,16 @@ def kldaputil(args, **kw):
...
+
+def ldap_search(args):
+    proc = subprocess.Popen([ldapsearch, '-H', ldap_uri, '-b', top_dn,
+                             '-D', admin_dn, '-w', admin_pw, args],
+                            stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT)
+    (out, dummy) = proc.communicate('\n')
```
greghudson: Why send a newline to the process?

greghudson:
The code changes look good except for the few comments I made.

In the first commit message, I don't understand "Enable kadmin setstr require_auth/getstrs functionality for the ldap KDB plugin."  You're not enabling that functionality; you're just changing the representation.
Also, please use two spaces after periods in commit messages.

In the second commit message summary, "auth indicator" should not be capitalized, but "LDAP" should be in all caps.

Assigning an RH/IPA OID to the new attribute is the quickest path forward on this change, so let's do that.

mrogers950:
I've addressed your comments, and replaced the test OID with the assigned FreeIPA OID.

greghudson:
I was looking over the changes to the schema again, and I don't think the comment "This attribute holds authentication indicator strings" useful at the same level as other attribute descriptions.  I suggest "A list of authentication indicator strings, one of which must be satisfied to authenticate to the principal as a service."

greghudson:
Shouldn't krbPrincipalAuthInd be added to attributes_set in ldap_principal.c?  It's an attribute we set, so it seems like it should be on the list of attributes we remove when deleting a principal from a non-standalone LDAP entry.

Also, we've been gravitating towards "if (ret)" rather than "if (ret != 0)", as it's more concise krb5_error_code values can be used as boolean expressions when not checking for a specific error.  I changed the one place on the coding practices wiki page which disagreed with that.

greghudson:
Sorry, one more thing about the schema: in the comment for the new attribute, can you give a breakdown of the OID by arc, sort of like the ones at the top of the file?  It doesn't have to be written in a big cascading list like in that comment, just "joint-iso-ccitt(2) country(16) us(840) ...".

(Also, out of curiosity, how did FreeIPA's arc 2.16.840.1.113730.3.8 wind up under Netscape's arc?  I see it at http://www.alvestrand.no/objectid/2.16.840.1.113730.3.8.html but it seems odd.)

mrogers950:
> I was looking over the changes to the schema again, and I don't think the comment "This attribute holds authentication indicator strings" useful at the same level as other attribute descriptions. I suggest "A list of authentication indicator strings, one of which must be satisfied to authenticate to the principal as a service."

That's fine with me.

>Shouldn't krbPrincipalAuthInd be added to attributes_set in ldap_principal.c? It's an attribute we set, so it seems like it should be on the list of attributes we remove when deleting a principal from a non-standalone LDAP entry.
Thanks for pointing that out, I believe you are right. However, adding one more value to the mask (next in the set for KRB5_KDB values is 0x010000) that is used for this pushes it over the UINT16_MAX limit imposed in expand_tl_data() and store_tl_data(). Just bumping it to UINT32_MAX in these functions and changing the store_tl_data() calls of store_16_be() to store_32_be() doesn't appear to be enough so there is probably other code that uses a UINT16_MAX mask limit. Any ideas?

> Also, we've been gravitating towards "if (ret)" rather than "if (ret != 0)", as it's more concise to use krb5_error_code values as boolean expressions when not checking for a specific error. I changed the one place on the coding practices wiki page which disagreed with that.

No problem, I'll stick to that style from now on.

> Sorry, one more thing about the schema: in the comment for the new attribute, can you give a breakdown of the OID by arc, sort of like the ones at the top of the file? It doesn't have to be written in a big cascading list like in that comment, just "joint-iso-ccitt(2) country(16) us(840) ...".

Will do.

greghudson:
> Thanks for pointing that out, I believe you are right. However, adding one more value to the mask (next in the set for the KDB_*_ATTR values is 0x010000) that is used for this pushes it over the UINT16_MAX limit imposed in expand_tl_data() and store_tl_data(). Just bumping it to UINT32_MAX in these functions and changing the store_tl_data() calls of store_16_be() to store_32_be() doesn't appear to be enough so there is probably other code that uses a UINT16_MAX mask limit. Any ideas?

I see.  store_tl_data() and decode_tl_data() define a persistent format, so just changing them to store four-byte masks would create transition issues.  For this PR, you can steal the unused flag KDB_UP_FLAG_ATTR (just remove the unused #define and replace it with the new one), but we will have to deal with this problem before we add another normalized principal attribute.

My guess is that we will add a KDB_TL_MASK8, store both the two-byte KDB_TL_MASK and the eight-byte KDB_TL_MASK8, and check for both values on read.

greghudson:
It's not relevant to this PR, but I wanted to note that I might have been wrong about store_tl_data() and decode_tl_data() defining a persistent format.  The serialized format may only be used in memory and not be stored in the LDAP krbExtraData attribute, in which case we can change it without transition concerns.

greghudson:
Pushed to master as 0bdd3b8058ed4ec9acc050e316bea86f6830b15f 18252584fd8cbfb43144337833fc83d4d2d12da5

PR #422, opened 2016-03-04
Allow zero cksumtype in krb5_k_verify_checksum()

greghudson:
A checksum type of 0 means to use the mandatory checksum type in
krb5_k_make_checksum(), krb5_k_make_checksum_iov(), and
krb5_k_verify_checksum_iov().  Extend this meaning to
krb5_k_verify_checksum() for the checksum type in the krb5_checksum
argument.  This change also applies to krb5_c_verify_checksum().
    
Add code to t_cksums.c to test checksum verification, including with
checksum type 0 for applicable test cases.


tlyu:
This looks ok to me.

PR #423, opened 2016-03-08
Add send and receive hooks for sendto_kdc

gladiac:
This add hooks which will be called before sending or after receiving a packet from the KDC.

This is work in progress, I post here for early comments so we can discuss naming of functions. Also we can decide how much tests would be needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -361,6 +361,7 @@ local.properties
 /src/tests/etinfo
 /src/tests/gcred
 /src/tests/hist
+/src/tests/hooks
 /src/tests/hrealm
 /src/tests/icred
 /src/tests/kdbtest
--- a/doc/appdev/refs/api/index.rst
+++ b/doc/appdev/refs/api/index.rst
@@ -267,6 +267,8 @@ Rarely used public interfaces
    krb5_salttype_to_string.rst
    krb5_server_decrypt_ticket_keytab.rst
    krb5_set_default_tgs_enctypes.rst
+   krb5_set_kdc_send_hook.rst
+   krb5_set_kdc_recv_hook.rst
    krb5_set_error_message.rst
    krb5_set_real_time.rst
    krb5_string_to_cksumtype.rst
--- a/doc/appdev/refs/types/index.rst
+++ b/doc/appdev/refs/types/index.rst
@@ -57,7 +57,9 @@ Public
    krb5_pa_svr_referral_data.rst
    krb5_pa_data.rst
    krb5_pointer.rst
+   krb5_post_recv_fn.rst
    krb5_preauthtype.rst
+   krb5_pre_send_fn.rst
    krb5_principal.rst
    krb5_principal_data.rst
    krb5_const_principal.rst
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -1242,6 +1242,12 @@ struct _krb5_context {
     krb5_trace_callback trace_callback;
     void *trace_callback_data;
 
+    krb5_pre_send_fn kdc_send_hook;
+    void *kdc_send_hook_data;
+
+    krb5_post_recv_fn kdc_recv_hook;
+    void *kdc_recv_hook_data;
+
     struct plugin_interface plugins[PLUGIN_NUM_INTERFACES];
     char *plugin_base_dir;
 };
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -8292,6 +8292,114 @@ krb5_set_trace_callback(krb5_context context, krb5_trace_callback fn,
 krb5_error_code KRB5_CALLCONV
 krb5_set_trace_filename(krb5_context context, const char *filename);
 
+
+/**
+ * @brief A function which will be executed as a hook to inspect or modify the
+ *        message being sent to the KDC.
+ *
+ * If a new a message is being created and returned, it will override the message
+ * which is being sent to the KDC. It is also possible to create a direct reply.
+ * If a new reply is created, the message will not be transmitted to the KDC but
+ * the sendto_kdc function will return the new reply to the library.
+ *
+ * @param[in]                  The library context to set the hook.
+ *
+ * @param[in]  data            User data passed via the set function.
+ *
+ * @param[in]  realm           The realm the message will be sent to.
+ *
+ * @param[in]  message         The original message we plan to send to the KDC.
+ *
+ * @param[in]  new_message_out A newly allocated message which will be sent to
+ *                             the KDC. The message will be freed using
+ *                             krb5_free_data().
+ *
+ * @param[in]  new_reply_out   A newly allocated fake reply we will return
+ *                             without sending any data to the KDC. The new
+ *                             reply will be freed using krb5_free_data().
+ *
+ * @retval 0 Success
+ * @return An Kerberos error code
+ */
+typedef krb5_error_code
+(KRB5_CALLCONV * krb5_pre_send_fn)(krb5_context context,
+                                   void *data,
+                                   const krb5_data *realm,
+                                   const krb5_data *message,
+                                   krb5_data **new_message_out,
+                                   krb5_data **new_reply_out);
+
+/**
+ * @brief A function poninter used as a hook to inspect the KDC reply or override
+ *        with a new reply.
+ *
+ * If a new reply is created, the reply received from the KDC will be discarded.
+ *
+ * @param[in]             The library context to set the hook.
+ *
+ * @param[in]  data       The user data passed via the set function.
+ *
+ * @param[in]  kdc_retval The return value of the sendto_kdc function. This
+ *                        return value should be returned by the receive hook
+ *                        too unless you want to override the reply or a serious
+ *                        error occurred.
+ *
+ * @param[in]  realm      The realm the reply was received from.
+ *
+ * @param[in]  message    The message sent to the KDC of the specified realm.
+ *
+ * @param[in]  reply      The reply received from the KDC.
+ *
+ * @param[in]  new_reply_out A new reply which will override the reply received by the
+ *                           KDC.
+ *
+ * @return The kdc_retval or an appropriate error code which will override the
+ *         KDC return value.
+ */
+typedef krb5_error_code
+(KRB5_CALLCONV * krb5_post_recv_fn)(krb5_context context,
+                                    void *data,
+                                    krb5_error_code kdc_retval,
+                                    const krb5_data *realm,
+                                    const krb5_data *message,
+                                    const krb5_data *reply,
+                                    krb5_data **new_reply_out);
+
+/**
+ * @brief Set a KDC pre-send hook.
+ *
+ * The hook will be called before the message is sent to the KDC.
+ *
+ * @param[in]  context  The library context.
+ *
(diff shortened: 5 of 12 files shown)


greghudson:
Before we get into bikeshedding, there are a number of more serious problems with the hook APIs and the layer they are hooked in at.

1. The send hook needs to be able to modify the outgoing request or synthesize a reply; the receive hook needs to be able to modify the reply.  To do this, the hook may need to be able to allocate memory.  It is not okay to assume that hook functions have access to the same malloc/free/realloc functions as libkrb5, so any memory allocated by the hook vtable must be freed by the hook vtable.  You have a free function for the hook data argument, but that's not needed; the caller can manage the hook data memory itself.

I suggest that modified or synthesized requests/replies be returned from the hook functions in krb5_data ** output arguments (where NULL means no modification or synthesis).  The libkrb5 code should copy the result and then call a vtable function to free a krb5_data * value.  Using a pointer to krb5_data allows hook functions to synthesize 0-length requests and replies.

2. From IRC conversation, it sounds like you want this to work with password changes as well as KDC requests.  For password change requests, the request packet isn't constructed until the server address is run (see where start_connection() calls callback_info->pfn_callback).  So, contrary to my previous suggestion, you will need to hook in at a lower level, probably just before the call to set_transport_message().

Because we are hooking in at this low level, we could consider passing the socket address (and length) of the server to the hook function, along with the realm.  I wouldn't expect typical hook functions to use it.

gladiac:
The test calls krb5_get_init_creds_password() with a wrong password, and then checks that the first error returned from the KDC is KRB5KDC_ERR_PREAUTH_REQUIRED and the second packet has KRB5KDC_ERR_PREAUTH_FAILED as the error. So I think the position of the hooks are fine.

I will work on the other changes ...

greghudson:
I don't understand; I said the current position of the hooks won't work for password change requests.  What does that have to do with a test using krb5_get_init_creds_password()?

gladiac:
We do not intercept password change request in our test in the Samba source code. What we do is what I described above with krb5_get_init_creds_password() using a wrong password.

greghudson:
I was judging by this text from IRC:

 > gladiac: ghudson: we have also tests, where we modify the packet before it is sent to the KDC and check that the KDC returns the correct error
> gladiac: like detect a clock skew correctly and reply with the correct error code
> gladiac: or if we change try to change the password, modify the password and check the KDC reacts correctly

If Samba doesn't need hooks for password-change requests, then we should try to anticipate whether that functionality might be needed in the future and whether it should inform the design of these hooks.

gladiac:
I dived deeper into the Samba code to actually find out what we are doing, we are not going this deep. The problem with a deeper level is that it is triggered several times, as it tries different ways to connect the KDC. I would go with this variant now. And maybe have something deeper if needed.

We had this issue with heimdal and started to use sendto_realm instead of sendto_kdc, because the latter might be triggered several times ...

greghudson:
Okay, that's fair.

gladiac:
I've updated the patches. Could you please take a look?

greghudson:
When a new reply is given by the hook, you need to copy and free it.  What you're doing now will result in the new reply being allocated by the hook and freed by libkrb5.  I think you're also leaking the krb5_data containers.

I thought of another approach, which is to document that implementations of these hook functions must use krb5_copy_data() to construct krb5_data pointers for their outputs, so that libkrb5 allocates them.  Then we wouldn't need the caller to supply a free function, and we could ditch the vtable and just have two setter functions for the send and receive hooks.  Callers which ignore the requirement would get away with it on Unix and be buggy on Windows, but since these are debugging hooks I'm not sure how much we care.  What do you think?

Anyway, please move the "data" argument to be just after the krb5_context argument.  If you search for "callback" in krb5.hin you'll see that's more consistent.

gladiac:
I see what you mean, here is the next round of changes.

greghudson:
For the setter functions, I would go with krb5_set_kdc_send_hook() and krb5_set_kdc_recv_hook().  I would also make them return void, and allow NULL values of func (to disable the relevant hook after setting one).  The hook type names (with "pre_send" and "post_recv") can remain as they are, I think, but I would name the context fields "kdc_send_hook" and "kdc_recv_hook".

For the send hook type, I would rename "new_message" to "new_message_out" and "new_reply" to "reply_out".

For the recv hook type, I would rename "new_reply" to "new_reply_out".  I would also allow the hook to return an error code (for example, if it runs out of memory trying to construct a new reply).

If the send hook returns a non-null but empty reply, that should still be used as the reply; checking for non-zero length seems counterproductive.

Your code for copying hook replies isn't correct; perhaps what I said about using krb5_copy_data() was misunderstood?  I suggest using  "*reply = *hook_reply; free(hook_reply);" instead.  Of course you won't want to do this if reply is NULL, so I would just immediately jump to cleanup if the send hook returns nonzero, and then conditionalize on non-null reply.

Freeing the hook new_message output from the send hook should not be contingent on a recv hook being set; you can do it in the cleanup handler.  Also, if a new_message is given, we have no further use for the input message, so we can just set "message = new_message;" rather than conditionalizing every use of message.

Recurring style issues (don't worry too much about these, I can fix them when pushing): we don't put braces around control statement bodies if they fit on one line.  Comments should usually be complete, punctuated sentences.  We don't fall into the "one line per argument" style when calling functions; we always put multiple arguments on one line when they fit.  Calls through function pointers can be written as "ptr(args);" rather than "(*ptr)(args)" if ptr involves a dereference, so "retval = context->kdc_send_hook(...)" and so forth.

gladiac:
I've did the changes you requested.

About style:

"we don't put braces around control statement bodies if they fit on one line."

Not adding braces leaded to numerous  issues in the Samba code. The most prominent issue with not adding them is the "goto fail" security bug from Apple. So I would prefer to use them instead of deleting them.

With the "one line per argument" style it is easier to see what changed in a commit or patch. You directly see which argument changed.

I've changed to ptr(args)

Tests
=====

What about tests, do we want more tests here? Should I implement more tests we have in Samba? The one I currently implemented is one test we have in Samba.

gladiac:
I will add documentation and fix comments once you say the code is fine ... ;)

greghudson:
The _out suffix should only be used on out parameter names, not on the names of local variables used to hold those values in the caller (in this case, the caller being krb5_sendto_kdc()).

It seems like it should be possible for a recv hook to see that the real transmission resulted in an error, synthesize a reply, and return success.  In the current PR this is impossible, because a return code of zero from the hook is ignored.  On the flip side, the current PR allows the recv hook to cause krb5_sendto_kdc() to fill in reply when returning failure, which is nonsensical.

The current PR leaks memory when the recv hook creates a new reply.  I think this should be corrected as follows:

* In krb5_sendto_kdc(), rename the parameter "reply" to "reply_out" and introduce a local variable "krb5_data reply = empty_data();".
* Set *reply_out = empty_data() at the beginning of the function.
* When calling k5_sendto(), store the result in the local variable reply.
* At the end of the function's successful path, fill in *reply_out by stealing from either reply or the hook's reply (to steal from reply, do "*reply_out = reply; reply = empty_data();")
* In the cleanup handler, free the contents of reply with krb5_free_data_contents().

greghudson:
krb5_sendto_kdc() should not set contents in reply_out when it returns an error, so the "if (retval) goto cleanup" should precede the part where *reply_out is set.  (Since a recv hook function shouldn't set a new_reply and return an error, we probably don't need to worry about cleaning up new_reply in the cleanup handler.)

doxygen comments are needed for the two new API functions as well as the types.  entries also need to be added to doc/appdev/refs/types/index.rst and doc/appdev/refs/api/index.rst for the new types and functions.

I had a first look at the test program.  It doesn't seem to cover very many of the hook function edge cases--setting new messages or replies, or changing the retval.  The hook functions also arguably do too much, by checking for specific aspects of the decoded messages.  The test program can just include k5-int.h rather than copy-pasting its contents.  Finally, you should try to cut down on the verbosity of error handling by using a helper function, or by just using assert()/abort() on error.

gladiac:
For the test I would prefer to use cmocka ;)

greghudson:
That is an option, but it will delay PR integration while we work out how to conditionally integrate test programs using cmocka (and for this case, test programs using cmocka which are run from Python test scripts).

gladiac:
I've added docs and more tests, but I'm not finished yet. I have to leave now. More tomorrow or on Monday.

greghudson:
In krb5_sendto_kdc():

* Free reply in the cleanup label.  Right now it can leak if the recv hook returns an error status.
* Freeing new_message in the cleanup handler doesn't need to be conditionalized (or commented).
* I think hook_message and hook_reply would be clearer variable names than new_message and new_reply.

I'll hold off making detailed comments on the tests as you're not done, but I'll note that we pretty much never use leading underscores in identifiers, as you have in _assert_krb5_error_code().  There's no reason to risk intruding on reserved parts of the C namespace.

gladiac:
a) I've did the changes you've asked for krb5_sendto_kdc()
b) The name of the function is _assert_krb5_error_code() because there is a macro with the name assert_krb5_error_code() the macro calls the function to get the function name and line number, see line 92.
c) I did not include k5-int.h because I want the file also to be an example how to implement the hooks.

gladiac:
I've implemented the "corner" cases you wanted to have coverage for. Let me know if you want more tests. The test should also be useable as an example how to implement the hooks outside of the krb5 tree.

greghudson:
You want it to be an example, but it's using internal libkrb5 functions?  That doesn't seem like a good combination.  It's not generally a good idea for test programs to serve double duty as example programs in general.

I saw that you have a macro with the name assert_krb5_error_code(), but pick a different name for the function.  In C, identifiers with leading underscores are best treated as the property of the C implementation, not as a generic "internal" prefix.

greghudson:
More feedback on the test programs:

* You use isflagset() exactly once.  It would probably be more readable to just expand that out into the appropriate bitwise operation.

* In krb5_recv_as_rep_test(), you test that the ticket encryption enctype lies between ENCTYPE_DES3_CBC_SHA1 and ENCTYPE_UNKNOWN.  This check doesn't make sense; there are no guarantees about the ordering of enctype values.  You can use krb5_c_valid_enctype() to test that an enctype is valid.

* The test program always passes NULL as the data argument when setting hooks, and therefore doesn't test that krb5_sendto_kdc() is properly communicating data arguments to the hooks.

I am also wondering whether krb5_sendto_kdc() should pass NULL instead of &reply to the recv hook when it gets an error from k5_sendto().  It would add complexity to the krb5_sendto_kdc() code, but it's probably a cleaner interface.  Do you have an opinion?

gladiac:
I've fixed the tests and added tests for the user data pointer.

"I am also wondering whether krb5_sendto_kdc() should pass NULL instead of &reply to the recv hook when it gets an error from k5_sendto(). It would add complexity to the krb5_sendto_kdc() code, but it's probably a cleaner interface. Do you have an opinion?"

reply is always initialized, so data and length are set to 0. We could pass NULL in case of KRB5_KDC_UNREACH or ENOMEM. But the developer who implements the hook could implement the test for reply->length != 0 too or check the return value.

greghudson:
Pushed to master as fb4d426ddeb9d4802a53dfbd74189ef8eacbe65e 117f6e415238b507a42cbc3ccdb45ad8408bfdfd

PR #424, opened 2016-03-13
Cleaner handling of keyboard interrupts

greghudson:
I think I've seen people complain about printing com_err messages on keyboard interrupts in the past, but we don't seem to have an existing ticket for it, so maybe it was out of band.  Anyway, it's been annoying me for years, so the first commit fixes that.  (I thought we also did fallback-to-master on keyboard interrupts, therefore attempting to read the password twice, but I can't reproduce this and we explicitly avoid fallback on KRB5_LIBOS_PWDINTR in gic_pwd.c, so I must be misremembering.)

The second commit also fixes a personal annoyance in the test suite.  Displaying stack traces on keyboard exit is a universal problem for Python scripts which don't explicitly avoid it.  Fortunately, we can solve this problem in k5test.py by installing a signal handler, and avoid having to wrap every Python test script in a try/except.

### DIFF
--- a/src/lib/krb5/krb/gic_pwd.c
+++ b/src/lib/krb5/krb/gic_pwd.c
@@ -326,9 +326,8 @@ krb5_get_init_creds_password(krb5_context context,
     /* If all the kdc's are unavailable, or if the error was due to a
        user interrupt, fail */
 
-    if ((ret == KRB5_KDC_UNREACH) ||
-        (ret == KRB5_LIBOS_PWDINTR) ||
-        (ret == KRB5_REALM_CANT_RESOLVE))
+    if (ret == KRB5_KDC_UNREACH || ret == KRB5_REALM_CANT_RESOLVE ||
+        ret == KRB5_LIBOS_PWDINTR || ret == KRB5_LIBOS_CANTREADPWD)
         goto cleanup;
 
     /* if the reply did not come from the master kdc, try again with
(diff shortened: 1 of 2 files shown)


greghudson:
> (I thought we also did fallback-to-master on keyboard interrupts, therefore attempting to read the password twice, but I can't reproduce this and we explicitly avoid fallback on KRB5_LIBOS_PWDINTR in gic_pwd.c, so I must be misremembering.)

I figured out what I was remembering here, and pushed a third commit to fix it.  (Basically, don't fall back to master on KRB5_LIBOS_CANTREADPWD.)

tlyu:
I think having visible feedback about pressing ^C is useful. When reading a password, we turn off terminal echo, so if a user has echoctl set on the terminal (often the default?), they won't see an echo of "^C" if they press ^C during password read.

greghudson:
I couldn't think of many terminal programs which prompt for passwords, but surveying the ones I could think of:

* su ignores SIGINT entirely
* gpg displays a message and exits (like we currently do)
* sudo displays a newline and exits (like this PR's first commit does)
* ssh displays two newlines and exits

Personally, I think that if you hit an interrupt key and the program exits, that's feedback enough.  But I can remove the first commit from the PR if there's not consensus to change the behavior.

greghudson:
I have removed the first commit from the branch.

tlyu:
This looks ok to me.

PR #425, opened 2016-03-13
cc_file: always truncate, instead of relying on O_EXCL to guarantee creation

danc86:
The problem with unlink() followed by open() with O_CREAT|O_EXCL is that
there is a race window where another process initializing the same
ccache filename could have already re-created the file by the time we
reach open(). In that case open() will return EEXIST, which
fcc_initialize() makes no attempt to recover from and just returns
KRB5_FCC_INTERNAL.

We can instead drop O_EXCL, meaning that the file might already exist at
the time of the open() call, but then call ftruncate() on it *after*
acquiring the write lock, ensuring that the file is written safely even
when multiple concurrent processes are initializing the ccache.

### DIFF
--- a/src/lib/krb5/ccache/cc_file.c
+++ b/src/lib/krb5/ccache/cc_file.c
@@ -456,7 +456,7 @@ fcc_initialize(krb5_context context, krb5_ccache id, krb5_principal princ)
     k5_cc_mutex_lock(context, &data->lock);
 
     unlink(data->filename);
-    flags = O_CREAT | O_EXCL | O_RDWR | O_BINARY | O_CLOEXEC;
+    flags = O_CREAT | O_RDWR | O_BINARY | O_CLOEXEC;
     fd = open(data->filename, flags, 0600);
     if (fd == -1) {
         ret = interpret_errno(context, errno);
@@ -480,6 +480,12 @@ fcc_initialize(krb5_context context, krb5_ccache id, krb5_principal princ)
     if (ret)
         goto cleanup;
     file_locked = TRUE;
+
+    st = ftruncate(fd, 0);
+    if (st == -1) {
+        ret = interpret_errno(context, errno);
+        goto cleanup;
+    }
 
     /* Prepare the header and principal in buf. */
     k5_buf_init_dynamic(&buf);


greghudson:
If we truncate an existing file, we have to worry about the possibility that someone else is iterating over it.

There are also more difficult issues surrounding FILE ccache atomicity because of how the credential cache API works, as described in http://krbdev.mit.edu/rt/Ticket/Display.html?id=7707

danc86:
In this patch the truncation is only done after acquiring a write lock, so no other processes will be using the ccache file at the same time.

danc86:
As for http://krbdev.mit.edu/rt/Ticket/Display.html?id=7707 it seems like you could solve that by using this patch, plus also just removing the `unlink()` call, which would eliminate any window of time where another process sees either the ccache file missing, or with incomplete contents (assuming the other process is also using suitable file locks).

I could imagine there are some edge cases where the `unlink()` is important though (for example, a ccache file that exists but isn't writable) so maybe it needs some more thought.

danc86:
BTW there is a Fedora bug about the unlink()/open() race window, I meant to link it here but forgot when I was filing the PR:

https://bugzilla.redhat.com/show_bug.cgi?id=1316798

danc86:
Oh sorry, I re-read http://krbdev.mit.edu/rt/Ticket/Display.html?id=7707 more closely and I guess that is referring to the fact that the file lock is released and then reacquired between calls to `krb5_cc_initialize()` and `krb5_cc_store_cred()` or whatever API is used to save credentials. That ticket talks about other processes seeing an "empty" cache, which I first thought meant a 0-byte file, but actually it means a cache which is correctly initialized but has no credentials. So that is indeed an issue with the API itself and not something that can be solved here.

But this patch certainly *does* help to make `fcc_initialize()` more atomic from the point of view of other readers, which is kind of a necessary prerequisite for solving the above ticket anyway.

greghudson:
Iterators do not keep the ccache file locked during the whole iteration, as there is no limit on how long the whole iteration can take.  So even if you have a write lock, there could be another process (or thread) with the file handle open at some offset, prepared to read another record.  If, during that interval, the file is truncated and written to, the iterator's offset could wind up being in the middle of a record.

PR #426, opened 2016-03-15
Revisit inquire_attrs_for_mech on old mechs

greghudson:
[Tom: I'm putting this in ticket: 8358 along with the previous change to gss_inquire_attrs_for_mech(), since we haven't yet included ticket 8358 in any release.  Let me know if I should use a separate ticket instead.]

In gss_inquire_attrs_for_mech(), if the mech does not implement RFC
5587, return success with empty mech_attrs and known_mech_attrs sets
to indicate a lack of knowledge for all attributes.  The previous
behavior of returning an error caused gss_indicate_mechs_by_attr() to
fail out in the presence of an old mechanism, in turn causing
gss_acquire_cred() and SPNEGO to break.

### DIFF
--- a/src/lib/gssapi/mechglue/g_mechattr.c
+++ b/src/lib/gssapi/mechglue/g_mechattr.c
@@ -181,8 +181,12 @@ gss_inquire_attrs_for_mech(
     mech = gssint_get_mechanism(selected_mech);
     if (mech == NULL)
         return GSS_S_BAD_MECH;
-    else if (mech->gss_inquire_attrs_for_mech == NULL)
-        return GSS_S_UNAVAILABLE;
+
+    /* If the mech does not implement RFC 5587, return success with an empty
+     * mech_attrs and known_mech_attrs. */
+    if (mech->gss_inquire_attrs_for_mech == NULL)
+        return GSS_S_COMPLETE;
+
     public_mech = gssint_get_public_oid(selected_mech);
     status = mech->gss_inquire_attrs_for_mech(minor, public_mech, mech_attrs,
                                               known_mech_attrs);


frozencemetery:
_mea culpa_ on the return code.  I like that this fixes that without pretending to know about attributes.  Thank you.

simo5:
@dwmw2 this should fix the gss-ntlmssp issue with th current packages if you wwant to test it.

dwmw2:
It does. But that proposed fix you wanted in gss-ntlmssp, to provide a ```gss_inquire_attrs_for_mech ()``` which explicitly says that no attrs are known? Still appears broken since the krb5 code overrides that and "assumes" that gss-ntlmssp knows all attrs from RFC5587 despite having been explicitly told otherwise.

greghudson:
Yes; the SPI contract for this function is that you must provide a known attributes list or the mechglue will provide the base RFC 5587 attribute list.  I don't like that very much, but it's not necessarily safe to change.

dwmw2:
I did provide a known attributes list: The empty set ```GSS_C_NO_OID_SET```. I'm missing the part in RFC5587 where it says that's not valid.

greghudson:
The SPI contract for this function is not determined by RFC 5587 (which specifies the API contract), but by the mechglue implementation.  Luke's implementation in the mechglue "helpfully" substitutes the RFC 5587 known attributes list for an empty known attributes list, making it impossible to return an empty known attributes list to the caller.  I wouldn't have chosen to do this, but on the other hand, a mechanism which implements gss_inquire_attrs_for_mechs() should really know all of the RFC 5587 attributes.

dwmw2:
Hey, I only implemented gss_inquire_attrs_for_mechs() because you completely refused to use my mechanism otherwise. I don't know *any* of the RFC5587 attributes, and was trying to indicate that. 

I'm a little unsure about how implementation details become equivalent to "an SPI contract" — and why that *isn't* the case for the implementation detail that if mechanisms neglect to implement gss_inquire_attrs_for_mech() they don't get used at all. Either way, if RFC5587 isn't definitive about these functions and their behaviour then it at least needs an erratum to reflect reality.

greghudson:
> Hey, I only implemented gss_inquire_attrs_for_mechs() because you completely refused to use my mechanism otherwise.

In this case, "you" is the Fedora krb5 package, because Red Hat backported a commit (authored by a Red Hat developer) from the master branch which introduced a bug.  The bug is not present in any MIT krb5 release.

There is no formal specification about the contract between GSS-API mechglue and mech; it tends to use mostly the same function signatures as the GSS-API itself but sometimes has different edge-case behavior.  This isn't an ideal situation, but it's the one we live in.  RFC 5587 only defines the API for this function as seen by applications.

frozencemetery:
While I am responsible for both the commit in question and its backport to Fedora, I do think it's reasonable that master should be in a usable state.  It doesn't matter whether it's released or not; if it's broken, we should fix it, which is exactly what I think is the intent of this PR.

dwmw2:
To be clear: this PR (commit 67b1d4502) does fix the issue and lets gss-ntlmssp work again.

The alternative patch which I created at @simo5's prompting, to add gss_inquire_attrs_for_mech() in gss-ntlmssp, is *not* correct and should not be applied. Because as @greghudson notes, the interface between mechglue and mech is *similar* but not identical to that described in RFC5587, and there is additional fun to be had in the undocumented differences.

PR #427, opened 2016-03-18
Fix keytab file format description

greghudson:
The key length and count of principal components are 16-bit fields.


kaduk:
Looks good to me.

PR #428, opened 2016-03-22
Iprop improvements

mrogers950:
These are downstream patches from RHEL krb5-1.10.3-50.el6 that I've ported to master. The first three address problems with a configuration where the default_realm is different than the realm being propagated. This should (at least in part) resolve ticket 8277. The fourth patch fixes a kpropd issue where the configured iprop_port was not being passed to the spawned kprop command.

### DIFF
--- a/src/kadmin/server/ipropd_svc.c
+++ b/src/kadmin/server/ipropd_svc.c
@@ -335,12 +335,12 @@ ipropx_resync(uint32_t vers, struct svc_req *rqstp)
      * and timestamp are in the ulog (then the slaves can get the
      * subsequent updates very iprop).
      */
-    if (asprintf(&ubuf, "%s dump -i%d -c %s",
-		 kdb5_util, vers, dump_file) < 0) {
-	krb5_klog_syslog(LOG_ERR,
-			 _("%s: cannot construct kdb5 util dump string too long; out of memory"),
-			 whoami);
-	goto out;
+    if (asprintf(&ubuf, "%s -r %s dump -i%d -c %s", kdb5_util,
+                 handle->params.realm, vers, dump_file) < 0) {
+        krb5_klog_syslog(LOG_ERR,
+                         _("%s: cannot construct kdb5 util dump string too long;"
+                           " out of memory"), whoami);
+        goto out;
     }
 
     /*
@@ -390,14 +390,15 @@ ipropx_resync(uint32_t vers, struct svc_req *rqstp)
 	    _exit(1);
 	}
 
-	DPRINT("%s: exec `kprop -f %s %s' ...\n",
-		whoami, dump_file, clhost);
+	DPRINT("%s: exec `kprop -r %s -f %s %s' ...\n",
+	       handle->params.realm, whoami, dump_file, clhost);
 	/* XXX Yuck!  */
 	if (getenv("KPROP_PORT")) {
-	    pret = execl(kprop, "kprop", "-f", dump_file, "-P",
-			 getenv("KPROP_PORT"), clhost, NULL);
+	    pret = execl(kprop, "kprop", "-r", handle->params.realm, "-f",
+                     dump_file, "-P", getenv("KPROP_PORT"), clhost, NULL);
 	} else {
-	    pret = execl(kprop, "kprop", "-f", dump_file, clhost, NULL);
+	    pret = execl(kprop, "kprop", "-r", handle->params.realm, "-f",
+                     dump_file, clhost, NULL);
 	}
 	perror(whoami);
 	krb5_klog_syslog(LOG_ERR,
--- a/src/slave/kprop.c
+++ b/src/slave/kprop.c
@@ -52,6 +52,7 @@ static int debug = 0;
 static char *srvtab = NULL;
 static char *slave_host;
 static char *realm = NULL;
+static char *def_realm = NULL;
 static char *file = KPROP_DEFAULT_FILE;
 
 /* The Kerberos principal we'll be sending as, initialized in get_tickets. */
@@ -63,7 +64,7 @@ static krb5_address *receiver_addr;
 static const char *port = KPROP_SERVICE;
 static char *dbpathname;
 
-static void parse_args(int argc, char **argv);
+static void parse_args(krb5_context context, int argc, char **argv);
 static void get_tickets(krb5_context context);
 static void usage(void);
 static void open_connection(krb5_context context, char *host, int *fd_out);
@@ -101,7 +102,7 @@ main(int argc, char **argv)
         com_err(argv[0], retval, _("while initializing krb5"));
         exit(1);
     }
-    parse_args(argc, argv);
+    parse_args(context, argc, argv);
     get_tickets(context);
 
     database_fd = open_database(context, file, &database_size);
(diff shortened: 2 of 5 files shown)


SaharahSarah:
Well, I guess that will make my life easier. I had been looking into and slowly fixing 8277.

greghudson:
The first commit should error out on strdup() failure, not continue with incorrect behavior.

mrogers950:
Thanks, I've corrected that.

greghudson:
Okay, sorry for the slow review.

First of all, please move the last commit (related to port setting) to a different PR.  It solves a related but separate problem, the solution in that commit is possibly too simplistic (if I read correctly that it's using a server port configuration variable for a client port), and the realm problem is complex enough that we are better off avoiding side fixes in the same PR.

The following eight problem code areas are identified in RT issue 8277:

1. kprop.c:get_tickets() uses sn2princ
1. kpropd.c:do_iprop() uses sn2princ
1. kpropd.c:parse_args() uses sn2princ
1. kpropd.c:parse_args() doesn't set realm for kadm5_get_config_params()
1. kpropd.c:do_iprop() uses krb5_get_default_realm()
1. kadmind does not pass -r realm to kdb5_util dump
1. kadmind does not pass -r realm to kprop
1. kpropd.c does not set default realm in context before using ulog

The first three problems are not fixed in this PR.  They should be fixed by creating a helper function (which can be duplicated in kprop and kpropd for now) which calls krb5_sname_to_principal() and then unconditionally substitutes in the realm name.

The fifth problem is fixed in this PR, but the fix leaves behind technical debt.  The "def_realm" global variable should be renamed to "realm" and set in all cases by parse_args(), either to the configured realm or the default realm.  do_iprop() can then be simplified.

The PR appears to contain the beginnings of a fix for #6 and #7, but that commit doesn't seem to make any functional change beyond introducing a memory leak.  Also, kadmind code does not need to conditionalize on handle->params.realm being non-NULL; you can find numerous places in kadmind which rely on kadm5_init() always setting this field.

Please fix the remaining problems and squash everything into one commit.  Then, in a second commit, please add automated tests.  t_kprop.py and t_iprop.py should each create a setup where the default realm and the domain_realm mapping are both wrong for the realm being propagated, and perform and verify a simple propagation.

mrogers950:
I believe this should address all of the issues stated, please let me know if I had missed anything. The commit that only leaked memory I had left a chunk out on accident.

I just noticed that the CI build fails on the commit that only changes the testing files - I'm not sure what would have caused that.

tlyu:
kpropd.c:1176:13: error: void function 'parse_args' should not return a value
          [-Wreturn-type]
                return retval;
                ^      ~~~~~~

tlyu:
I think Travis only builds the merge commit, so any commit in the series could be responsible for a build failure. I've heard that some CI systems might have "auto-bisect" to help locate a failing commit but that could get expensive.

mrogers950:
Thanks, I've corrected that.

greghudson:
The realm argument for sn2princ_with_realm() should not be optional, as we never want to use the sn2princ result.

mrogers950:
I've changed sn2princ_with_realm() to require the rrealm argument, and updated kprop to use the default realm where -r is not provided.  I also added the t_iprop.py and t_kprop.py mismatch tests, refactored a bit and added comments, and updated the commit descriptions.

greghudson:
In ipropd_svc.c, you construct realmbuf, then use it conditionally in the construction of ubuf.  Why the condition, when realmbuf will never be NULL?  And why not just put -r %s directly in the construction of ubuf?  Also, please have a look at what the style checker says about your changes to this file after you fix this; it is a weird file because it still uses tabs (because it was originally rpcgen output; that's probably not a good enough reason and we should reformat it).

In kprop.c, the memory handling of the realm global variable is problematic because:

* It might be the sole owner of an allocated string or it might point into argv.
* In the former case, it is never freed; this is functionally harmless but looks slightly bad in memory leak checkers (only slightly bad because the memory is still accessible on process exit).
* It is memoized in get_tickets(); if the code ever changes such that it might be used before get_tickets, we could run into bugs where it is still NULL.

A better choice would be to (1) use a separate global variable default_realm to own the default realm if we fetch it (solely for the purpose of freeing it), so that realm is always an alias pointer; (2) ensure that realm is properly set on exit from parse_args(), and (3) free default_realm at the end of main().  This also applies in kpropd.c.

In get_tickets(), the krb5_is_referral_realm() check is no longer necessary because sn2princ_with_realm() always sets a concrete realm.  (I guess a user could specify -r "" on the command line, but we have no reason to paper over that.)  That code block should be removed.  This also applies in do_iprop() in kpropd.c.

In kpropd.c:parse_args(), "params.realm = strdup(realm);" should be "params.realm = realm;".  The current code leaks the strdup() result and doesn't check for failure.  Also, the params setting should happen immediately before calling kadm5_get_config_params() (without even a blank line); right now it is separated from the call by the entirely unrelated construction of temp_file_name.

greghudson:
In the tests, {'libdefaults': {..., 'domain_realm' : 'krbtest.com = BAR.$realm'}} won't do what you're trying to do, for multiple reasons.  You want a 'domain_realm' key at the top level of the conf dictionary, with the value {hostname: 'BAR.$realm'}.  (You could also simplify by not incorporating $realm into the realm names.  FOO and BAR are perfectly serviceable realm names by themselves.)

mrogers950:
- Removed the use of realmbuf and use the realm directly in ubuf creation.
- Removed referral realm code.
- Use and free def_realm as an alias pointer.
- Correct strdup() usage and move the param setting statements.
- Corrected domain_realm test config options.
I've begun converting ipropd_svc.c to use spaces with correct formatting in a new branch on top of these changes.  I can merge it in this PR or submit it as a new PR afterwards ,if you prefer.

greghudson:
def_realm was allocated with krb5_get_default_realm() and should be freed with krb5_free_default_realm() for form's sake.

In kprop.c, you are still setting up the realm variable at the beginning of get_tickets().  This should be done in parse_args().  (You will have to pass the krb5 context to parse_args() to do this.)

mrogers950:
Updated with those corrections.

greghudson:
The code changes look good, but reviewing the list of problems identified in RT ticket 8277 from my earlier comment, it appears that problems 7 and 8 still aren't fixed.  I would like to know why the tests succeed even though these problems remain, and I would like to see those problems fixed.

greghudson:
In the tests:

* The override "'libdefaults' : { 'default_realm' : '$realm' }" doesn't seem like it should change anything; the default realm should already be set to $realm in the parent configuration.  This override appears in t_kprop.py slave3 and t_iprop.py slave4/slave5.  slave5 doesn't seem to have any useful differences from slave2.

* Problems 6 and 7 manifest only on the master.  You're not setting up a master with a non-default realm, so you're not testing the problem 6 fix or revealing the lack of a problem 7 fix.

* Problem 8 would manifest with ulog_replay() failing to open the DB.  Since you are only testing the initial full propagation for slave3/slave4/slave5, I don't think your new tests ever reach ulog_replay(), so they don't reveal the lack of a problem 8 fix.

* PEP 8 prohibits spaces just inside braces (see https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements); some of your config declarations have them.  Also don't put spaces before the ':' in a dictionary map entry.

* Some continuation lines are not indented correctly.

mrogers950:
- Added in the fixes for 7 and 8 (with a minor declaration scope change).
- In t_iprop.py I changed the slave1m profile to use the FOO realm and set the slave1 kadmind to use -r, which verifies 6 and 7.
- Add an incremental update test for slave3 to verify fix 8, and update the ulog numbers for the tests that follow.
- Removed the redundant slave5 test, and adjusted dictionary formatting.

greghudson:
The context default realm only needs to be set once, not before every call to ulog_replay().  Do it in parse_args() like so:

    if (realm == NULL) {
        fetch the default realm into def_realm and set realm
    } else {
        set the default realm in kpropd_context to realm
    }

mrogers950:
That's fixed. I reverted the scope change since that would be unrelated now.

greghudson:
Pushed to master as 0e7b4c901bf475f61b74ca56363089906e324272 972ccb9205c28a1177853ad748aea1387ac15fde

PR #430, opened 2016-03-23
Amend KDC hook documentation

greghudson:
[Unfortunately I didn't catch these issues on my final pass over PR #423.]

In the Doxygen comments for the new APIs and types, include @version
tags indicating that they are new in 1.15, and put @param declarations
just after the brief message for consistency with other comments.


tlyu:
This looks ok to me.

PR #431, opened 2016-03-28
Improve errors when DB2 database cannot be opened

greghudson:


tlyu:
This looks ok to me. Manually verified that the reported error message adds the filename, resolving the issue from the original bug report.

PR #432, opened 2016-03-30
Integrate with appveyor for Windows CI

greghudson:
appveyor.com is a hosted continuous integration service for Windows.
Add an appveyor.yml file containing build instructions.  The appveyor
virtual machines do not include the MFC libraries, so change
util/wshelper/resource.rc to avoid including <afxres.h> (which it does
not need) and add a build conditional for leash.

Right now we do not build the installers; the appveyor VMs do not
appear to have the version of the WiX toolkit we need, and we would
also have problems with the missing leash executable.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,16 @@
+build_script:
+  - call "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /x86
+  - mkdir C:\kfw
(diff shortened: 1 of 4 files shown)


SaharahSarah:
Fails to build on windows: [Build Log](http://pastebin.com/VVHp6Xf8)

SaharahSarah:
Builds fine on my Windows VM now :+1:

PR #433, opened 2016-04-03
Add indentation checking to cstyle-file.py

greghudson:
For each file we check in cstyle-file.py, try to use emacs to
batch-reindent a copy of the file, and compare the resulting lines to
the input to detect indentation errors.

[PR commentary only:] Using emacs for this checking is kind of slow, because the emacs startup has to happen for every changed .c or .h file.  The layering of cstyle.py and cstyle-file.py doesn't make it easy to do multiple files in one emacs invocation.  It's still a lot faster to have a slow tool checking for these errors than to have a human notice it and give PR feedback.

We could alternatively use clang-format or indent, but I think we would have to change our coding practices to strictly conform to a configurable style in one tool or the other, or else they would generate a lot of false positives.  clang-format is also a bit of a moving target, and doesn't make it easy to write one style file for multiple versions.

### DIFF
--- a/src/util/cstyle-file.py
+++ b/src/util/cstyle-file.py
@@ -37,12 +37,12 @@
 #   Assignment at the beginning of an if conditional
 #   Use of prohibited string functions
 #   Lack of braces around 2+ line flow control body
+#   Incorrect indentation as determined by emacs c-mode (if possible)
 #
 # This program does not check for the following:
 #
 #   Anything outside of a function body except line length/whitespace
 #   Anything non-syntactic (proper cleanup flow control, naming, etc.)
-#   Indentation or alignment of continuation lines
 #   UTF-8 violations
 #   Implicit tests against NULL or '\0'
 #   Inner-scope variable declarations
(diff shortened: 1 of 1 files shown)


frozencemetery:
Would it perhaps be faster to start an emacs server (`emacs --daemon`) and run emacsclient to connect to it for each file?  Naively it looks like it should be possible with --eval.

tlyu:
That might not work well for those of us who keep emacs running all the time with an emacs server. Connecting to the currently running emacs could contaminate what should be a standardized process with our personal configuration quirks. Also I'm not sure if having multiple emacs server processes running simultaneously is a good idea.

tlyu:
I tested this change against a known-nonconforming file (via make reindent) and it seems to work as expected. I'm ok with this change as is, unless we want to pursue @frozencemetery's suggestion.

frozencemetery:
Interacting with existing emacs daemons is a good point.  I see flags for emacsclient to select which emacs daemon to connect to, but I don't see a way to start multiple emacs daemons.

I am therefore okay with this approach as written.

PR #434, opened 2016-04-05
Use APPVEYOR_BUILD_FOLDER

tlyu:
The directory that AppVeyor does checkouts into can vary.  Use the
documented APPVEYOR_BUILD_FOLDER environment variable instead of
hardcoding a value.


greghudson:
This change looks fine to me.

PR #435, opened 2016-04-06
Add German translation

greghudson:
We received a German translation by email to krbdev@mit.edu; this pull request adds it to the tree.

### DIFF
--- a/src/po/Makefile.in
+++ b/src/po/Makefile.in
@@ -18,7 +18,7 @@ ETSRCS=	$(BUILDTOP)/lib/gssapi/generic/gssapi_err_generic.c \
 	$(BUILDTOP)/lib/krb5/error_tables/kv5m_err.c \
 	$(BUILDTOP)/lib/krb5/error_tables/krb524_err.c
 # This is a placeholder until we have an actual translation.
-CATALOGS=en_US.mo
+CATALOGS=en_US.mo de.mo
 
 .SUFFIXES: .po .mo
 .po.mo:
--- a/src/po/de.po
+++ b/src/po/de.po
@@ -1,0 +1,9301 @@
+# German translation of mit-krb5.
+# This file is distributed under the same license as the mit-krb5 package.
+# Copyright (C) 1985-2013 by the Massachusetts Institute of Technology.
+# Copyright (C) of this file 2014-2016 Chris Leick <c.leick@vollbio.de>.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: mit-krb5 13.2\n"
+"Report-Msgid-Bugs-To: krbdev@mit.edu\n"
+"POT-Creation-Date: 2015-05-06 14:59-0400\n"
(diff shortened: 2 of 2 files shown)


SaharahSarah:
Looks good to me, though, I can't tell if the German translations are accurate.

tiran:
Nathaniel asked me to review the translation. I'm a native German speaker.

tiran:
I have skimmed over the translation. In general the translations are good and technically accurate. For my personal taste it's sometimes over the top, e.g. some technical terms are translated like "Anmeldedatenzwischenspeicher" for "credential cache". This could it make harder to search for error messages. On the other hand I'm not used to German translation on Linux. I have been using English since I started to use Linux 18 years ago.

I found two errors:

The translation of "ASN.1 structure is missing a required field" is not correct. "ASN.1-Struktur fehlt in einem erforderlichen Feld" -> "ein erforderliches Feld fehlt in der ASN.1-Struktur"

"falsche Magische Zahl" -> "falsche magische Zahl"

tlyu:
@tiran: Thanks for looking at this translation.  I do have a few questions, if you have time to look them over.  My German is extremely rusty (a couple of years of a language elective class in high school), so I'm sorry if any of this doesn't make sense.

There are many translations of com_err() strings that are sentence fragments with an implied "while" at the beginning in English, e.g.,
```
 #: ../../src/clients/kinit/kinit.c:488 ../../src/clients/kinit/kinit.c:644
 #, c-format
 msgid "resolving ccache %s"
 msgstr "Ccache %s wird ermittelt"
```
in an English locale would print something like "kinit: File not found resolving ccache foo" and an English reader would read as equivalent to "kinit: File not found while resolving ccache foo". So 
should "Ccache %s wird ermittelt" be "beim Ccache %s ermitteln"? Or would a German reader infer a similar transformation?  (Yes, the "right" answer is probably explicitly writing "while" in the English version.)

The "ASN.1 structure..." correction looks reasonable to me, but would it be better to delete "in" from the original translation so the word order could be closer to the original?

It looks like https://de.wikipedia.org/wiki/Magische_Zahl_(Informatik) doesn't consistently capitalize "Magische", but I guess in context the lower case "magische Zahl" is no more ambiguous than in English.

greghudson:
Updated with both suggested changes.

tlyu:
It looks like a suggested patch that was posted to krbdev wasn't applied; there was also some discussion that implies that the second hunk (for msgid "\t-T armor credential cache\n") might not be correct.  (I agree with the discussion; neither translation seems correct to me.)

greghudson:
Rebased and updated with the patch from
http://mailman.mit.edu/pipermail/krbdev/2016-April/012567.html
(by the initial author of the translation), except for the change to "-T armor credential cache".  Looking back at the discussion of that text, I think a translation suggesting "hardened credential cache" is slightly better than a translation that suggests "armor (verb) the credential cache".  What we really want is a translation suggesting "credential cache to use as armor", but I won't hold up the entire po file for that.

PR #436, opened 2016-04-12
SPNEGO: Add special case for NTLMSSP+MechListMIC

simo5:
This is an initial RFC PR, the code does work and operates as intented in a synthetic test performed using python-gssapi and the gss-ntlmssp mechanism, but hasn't been tested against a conformant implementation (Windows) yet.

### DIFF
--- a/src/lib/gssapi/spnego/spnego_mech.c
+++ b/src/lib/gssapi/spnego/spnego_mech.c
@@ -507,6 +507,48 @@ mech_requires_mechlistMIC(spnego_gss_ctx_id_t sc)
 	return result;
 }
 
+/* iso(1) org(3) dod(6) internet(1) private(4) enterprise(1) Microsoft(311)
+ * security(2) mechanisms(2) NTLM(10) */
+static const gss_OID_desc gss_mech_ntlmssp_oid =
+	{ 10, "\x2b\x06\x01\x04\x01\x82\x37\x02\x02\x0a" };
+
+/* iso(1) org(3) dod(6) internet(1) private(4) enterprises(1) samba(7165)
+ * gssntlmssp(655) controls(1) ntlmssp_reset_crypto(3) */
+static const gss_OID_desc ntlmssp_reset_crypto_oid =
+	{ 11, "\x2B\x06\x01\x04\x01\xB7\x7D\x85\x0F\x01\x03" };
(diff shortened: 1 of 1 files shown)


simo5:
This is the way I tested this so far:
create a file with some credentials:
```
$ echo F23:guest:guest > ~/ntlmcreds.txt
$ export NTLM_USER_FILE=~/ntlmcreds.txt
$ python
import gssapi

spnego = gssapi.raw.oids.OID.from_int_seq('1.3.6.1.5.5.2')
c = gssapi.creds.Credentials(mechs=[spnego], usage='initiate')
tname = gssapi.raw.names.import_name("F23/server", name_type=gssapi.raw.types.NameType.hostbased_service)
ac = gssapi.creds.Credentials(mechs=[spnego], usage='accept')

seci = gssapi.SecurityContext(creds=c, name=tname, mech=spnego, usage='initiate')
seca = gssapi.SecurityContext(creds=ac, usage='accept')
it = seci.step(token=None)
ot = seca.step(token=it)
it = seci.step(token=ot)
ot = seca.step(token=it)
it = seci.step(token=ot)

e = seci.wrap("Secrets", True)
o = seca.unwrap(e.message)

o.message
'Secrets'
```

So far I checked with GDB that the seq numbers were reset after the MEchListMIC operation happened and started again from 0.

greghudson:
Please factor out the export/import/import into a helper to minimize the expansion of process_mic().

simo5:
Ok moved context duplication in a separate helper function.

greghudson:
I think this looks okay apart from trivial stuff like comments and style issues.  I'll await confirmation that this interoperates with Windows.

greghudson:
It would be nice to trim down the call sites since there are two of them.  Moving the comment and the mechanism test into the helper function would be a start.  I don't know if there's much we can do about the error-handling parts.

simo5:
Ok moved the ntlmssp oid check in the reset state function as well as the MS-SPNG related comment, which has been expanded to explain the behavior of the mechanism.
Once this was done I was able to call the function w/o having to duplicate the error handling.

HTH.

greghudson:
Pushed to master as cb96ca52a3354e5a0ea52e12495ff375de54f9b7.

PR #437, opened 2016-04-15
Fix segfault on NULL key_data

mrogers950:
Fail gracefully in case "krbprincipalkey" cannot be read from LDAP.

### DIFF
--- a/src/lib/kdb/kdb_default.c
+++ b/src/lib/kdb/kdb_default.c
@@ -445,6 +445,11 @@ krb5_def_fetch_mkey_list(krb5_context        context,
     if (retval)
         return (retval);
 
+    if (master_entry->n_key_data == 0) {
+        retval = KRB5_KDB_NOMASTERKEY;
+        goto clean_n_exit;
+    }
+
     /*
      * Check if the input mkey is the latest key and if it isn't then find the
      * latest mkey.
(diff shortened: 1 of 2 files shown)


greghudson:
For a list defined by a pointer and count, if you want to check whether the list contains at least one member, it's always more correct to check the count than the pointer.  In this case, check n_key_data.

Also, I don't think this test is sufficient; there is a later unconditional use of key_data[0].  We should probably just return an error (perhaps KRB5_KDB_NOMASTERKEY) if there is no key data in K/M.

mrogers950:
Thanks, revision pushed.

greghudson:
The new commit introduces a memory leak when the conditional is true.

mrogers950:
Fixed.

greghudson:
The code change looks okay now.  Can you try to improve the commit message?  Neither the summary nor the explanatory sentence mention master keys at all, and a K/M entry could contain no key data with any KDB module, not just with LDAP.

greghudson:
Oh, also, I found a way to easily reproduce this bug:

    $ kadmin.local purgekeys -all K/M
    $ kadmin.local getprinc user
    Segmentation fault (core dumped)

(It doesn't really matter what the second command is.)  Can you add a regression test to the end of t_mkey.py?

mrogers950:
Thanks, I've updated the commit message and added the t_mkey.py test.

greghudson:
Pushed to master as 83494605b2dd594ab33f9b3cfa5abc82cf0f9e92 f2286a6493c7aab267cf0d9cca680edaf2968707 with minor wording and spacing changes.  (Remember that we use two spaces after sentence-ending periods.)

PR #438, opened 2016-04-19
Improve kadmin parsing of time intervals

greghudson:
When parsing time intervals in kadmin commands, try
krb5_string_to_deltat() first, then fall back to subtracting the
current time from get_date().  If we do fall back, treat "never" as a
zero interval, and error out rather than yield a huge interval if
get_date() returns a time in the past.

Notable behavior differences: bare numbers of seconds and suffixed
numbers (e.g. "5m" or "6h") are now supported for all intervals;
HH:MM:SS and HH:MM are now treated as intervals rather than absolute
times with the current time subtracted.

### DIFF
--- a/doc/admin/admin_commands/kadmin_local.rst
+++ b/doc/admin/admin_commands/kadmin_local.rst
@@ -260,11 +260,12 @@ Options:
     (:ref:`getdate` string) The password expiration date.
 
 **-maxlife** *maxlife*
-    (:ref:`getdate` string) The maximum ticket life for the principal.
+    (:ref:`duration` or :ref:`getdate` string) The maximum ticket life
+    for the principal.
 
 **-maxrenewlife** *maxrenewlife*
-    (:ref:`getdate` string) The maximum renewable life of tickets for
-    the principal.
+    (:ref:`duration` or :ref:`getdate` string) The maximum renewable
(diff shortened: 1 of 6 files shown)


SaharahSarah:
In kadmin_local.rst, maxlife and maxrenewlife have their description start with

> The maximum...

while later in the file the descriptions start with

> Sets the maximum/minimum/etc...

I think it would be wise to make maxlife and maxrenewlife match the rest of the file in that it describes the action it does, i.e. begins with Sets the maximum....

greghudson:
Changing the documentation for -maxlife and -maxrenewlife in that way would make those option descriptions inconsistent with the ones immediately before and after.  Using a consistent form for documenting options would be an improvement, but I would prefer to consider it out of scope for this PR.

tlyu:
I think you mean HH:MM:SS instead of HH:MM:DD in the commit message.

greghudson:
Thanks; I fixed it.

greghudson:
After discussion with Tom, I added three more commits to:

* Fix a minor race condition in the parsing of intervals
* Improve the output format for intervals and principal password expiry
* Add tests for many of the accpted deltat and relative getdate.y input formats

tlyu:
Thanks for the tests. Looks ok to me now.

PR #439, opened 2016-04-22
Add debug message filtering to krb5_klog_syslog

greghudson:
Two commits here: one to remove some unnecessary conditionalization in logger.c, and another to add the debug relation.


tlyu:
This looks ok to me.

PR #440, opened 2016-04-26
Skip password prompt when running ksu as root

mrogers950:
A change introduced in 5fd5a67 resulted in root always being prompted for
the target user password when running ksu.  Restore the previous behavior
which is to only prompt if the principal is provided with -n.


greghudson:
Pushed to master as 2f8b9ef90829b031448d4547190c72438a69aacd with one trivial change.

PR #441, opened 2016-04-27
python-ldap-27::modify() function errors when parsing kerberos.ldif

smemery:
When using the python-ldap-27::modify() function, an error message is generated:

("{'info': 'olcAttributeTypes: Missing closing parenthesis before end of input', 'desc': 'Other (e.g., implementation specific) error'}", 1)

parsing the kerberos.ldif file.  Most related RFCs, such as 2307, do specify class declarations with a space in between the closing right parenthesis.  Adding a space allows the python library to complete successfully.  The updated LDIF file has been tested with OpenLDAP 2.4.44.

### DIFF
--- a/src/plugins/kdb/ldap/libkdb_ldap/kerberos.ldif
+++ b/src/plugins/kdb/ldap/libkdb_ldap/kerberos.ldif
@@ -46,7 +46,7 @@ attributetypes: ( 2.16.840.1.113719.1.301.4.1.1
                 NAME 'krbPrincipalName'
                 EQUALITY caseExactIA5Match
                 SUBSTR caseExactSubstringsMatch
-                SYNTAX 1.3.6.1.4.1.1466.115.121.1.26)
+                SYNTAX 1.3.6.1.4.1.1466.115.121.1.26 )
 
 
 ##### If there are multiple krbPrincipalName values for an entry, this
@@ -62,7 +62,7 @@ attributetypes: ( 1.2.840.113554.1.4.1.6.1
                 EQUALITY caseExactIA5Match
                 SUBSTR caseExactSubstringsMatch
                 SYNTAX 1.3.6.1.4.1.1466.115.121.1.26
-                SINGLE-VALUE)
+                SINGLE-VALUE )
 
 ##### This specifies the type of the principal, the types could be any of
 ##### the types mentioned in section 6.2 of RFC 4120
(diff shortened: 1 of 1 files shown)


greghudson:
After looking at RFC 4512 I'm pretty sure this is a bug in python-ldap-27, but we can accept the change anyway to work around it.

greghudson:
When you say "python-ldap-27", what precisely are you referring to?  If I google for python-ldap, I can't figure out where the 27 comes from.

greeneg:
Normally, I’m a fly on the wall on the list, but, this is a parsing bug in the python library you’re using. The LDIF format specification does not strictly require what you’re requesting. The appropriate fix for this should be in Python-LDAP’s parser.

--
Gary L. Greene, Jr.
==============================================================================
Volunteer developer of the KDE F/OSS project and Project Lead for AltimatOS
    http://www.kde.org/ <http://www.kde.org/>                 http://www.altimatos.com/ <http://www.altimatos.com/>
Please refrain from sending me proprietary binary documents (Doc, Xls, Ppt)
  Use a free office suite with standards approved formats like LibreOffice.
    http://www.libreoffice.org/ <http://www.libreoffice.org/>
==============================================================================

smemery:
Sorry, I didn't mean to imply that the current LDIF syntax was in violation of RFC 2849.  I meant to assert that the format is inconsistent with similar RFCs, among the declaration itself, and between other declarations in the file.  Given the unexpected controversy of this change-set I will pursue a fix upstream with the python-ldap community.
In the mean time I will patch this file in our distribution given the following reasons:
1. The unfamiliarity of contributions with the upstream python-ldap community
2. The unknown scope and impact of changing the their parser
3. The low risk and impact of adding spaces to this LDIF file
4. The ability for our application to function in the time-line requested

greghudson:
I can still take this patch.  I just want to make sure the commit message correctly identifies which software's bugs we are working around, which is why I asked the question about what the "27" in python-ldap-27 refers to.  (Perhaps it's part of an OS package name and refers to the Python version?)

smemery:
Yes, this is the package name of the overall version of Python, 2.7.  The specific library can be found here:
Project URL: http://www.python-ldap.org/
Source URL: http://pypi.python.org/packages/source/p/python-ldap/python-ldap-2.4.10.tar.gz
We haven't tested with the latest version, 2.4.25.

greghudson:
Pushed to master as 1ae3175858a4b91458fef51108d8c75f5dc0d0dd

PR #442, opened 2016-04-27
Add support for new openssl-1.1.0 APIs

pixelb:
openssl-1.1.0 made HMAC_CTX, EVP_MD_CTX, and EVP_CIPHER_CTX
opaque types, therefore adjust how we use these interfaces accordingly,
while also being backwards compatible with the older interfaces.
Note EVP_CIPHER_CTX_{new,free} have been available since openssl-0.9.8
and since krb5 already enforces that version or later we just
use those new interfaces unconditionally.

### DIFF
--- a/src/configure.in
+++ b/src/configure.in
@@ -244,6 +244,10 @@ builtin)
   ;;
 openssl)
   AC_CHECK_LIB(crypto, PKCS7_get_signer_info)
+  AC_CHECK_TYPES([HMAC_CTX, EVP_MD_CTX], [], [], [[
+#include <openssl/hmac.h>
+#include <openssl/evp.h>
+  ]])
   ;;
 *)
   AC_MSG_ERROR([Unknown crypto implementation $withval])
@@ -280,6 +284,10 @@ builtin|openssl)
   AC_CHECK_LIB(crypto, CMS_get0_content,
                [AC_DEFINE([HAVE_OPENSSL_CMS], 1,
                           [Define if OpenSSL supports cms.])])
+  AC_CHECK_TYPES([HMAC_CTX, EVP_MD_CTX], [], [], [[
+#include <openssl/hmac.h>
+#include <openssl/evp.h>
+  ]])
   ;;
 nss)
   if test "${PKINIT_CRYPTO_IMPL_CFLAGS+set}" != set; then
--- a/src/include/k5-platform.h
+++ b/src/include/k5-platform.h
@@ -1099,4 +1099,17 @@ extern int k5_getopt_long(int nargc, char **nargv, char *options,
 #define getopt_long k5_getopt_long
 #endif /* HAVE_GETOPT_LONG */
 
+/*
+ * Compatibility for changed APIs in openssl-1.1.0.
+ * Pre openssl-1.1.0 will have these defined.
+ */
+#if HAVE_HMAC_CTX
+# define HMAC_CTX_free HMAC_CTX_cleanup
+#endif /* HAVE_HMAC_CTX */
+
+#if HAVE_EVP_MD_CTX
+# define EVP_MD_CTX_new() EVP_MD_CTX_create()
+# define EVP_MD_CTX_free(x) EVP_MD_CTX_destroy(x)
+#endif /* HAVE_EVP_MD_CTX */
+
 #endif /* K5_PLATFORM_H */
--- a/src/lib/crypto/openssl/enc_provider/aes.c
+++ b/src/lib/crypto/openssl/enc_provider/aes.c
@@ -65,22 +65,27 @@ cbc_enc(krb5_key key, const krb5_data *ivec, krb5_crypto_iov *data,
 {
     int             ret, olen = BLOCK_SIZE;
     unsigned char   iblock[BLOCK_SIZE], oblock[BLOCK_SIZE];
-    EVP_CIPHER_CTX  ciph_ctx;
+    EVP_CIPHER_CTX  *ciph_ctx;
     struct iov_cursor cursor;
 
-    EVP_CIPHER_CTX_init(&ciph_ctx);
-    ret = EVP_EncryptInit_ex(&ciph_ctx, map_mode(key->keyblock.length),
-                             NULL, key->keyblock.contents, (ivec) ? (unsigned char*)ivec->data : NULL);
-    if (ret == 0)
+    ciph_ctx = EVP_CIPHER_CTX_new();
+    if (!ciph_ctx)
         return KRB5_CRYPTO_INTERNAL;
+
+    ret = EVP_EncryptInit_ex(ciph_ctx, map_mode(key->keyblock.length),
+                             NULL, key->keyblock.contents, (ivec) ? (unsigned char*)ivec->data : NULL);
+    if (ret == 0) {
+        EVP_CIPHER_CTX_free(ciph_ctx);
+        return KRB5_CRYPTO_INTERNAL;
+    }
 
     k5_iov_cursor_init(&cursor, data, num_data, BLOCK_SIZE, FALSE);
     k5_iov_cursor_get(&cursor, iblock);
-    EVP_CIPHER_CTX_set_padding(&ciph_ctx,0);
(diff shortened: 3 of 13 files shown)


nicowilliams:
I've mentioned this to @vdkhovni, and he says that there should be compatibility macros in OpenSSL, and that this should be opened with OpenSSL as an issue/PR.  Mind you, I did something in Heimdal like what you propose here: work around the problem.

pixelb:
Since the interfaces are changing from stack to heap, it's necessary to change the each project accordingly. Now openssl could have provided HMAC_CTX_new() and EVP_MD_CTX_new() for a few versions like it had done for EVP_CIPHER_CTX_new(), but there's nothing openssl can do really to change already deployed openssl. Luckily the mappings done here are trivial enough.

greghudson:
Nico is saying that EVP_MD_CTX_create() and EVP_MD_CTX_destroy() should continue working in OpenSSL 1.1.0, I think.  Code which uses stack-allocated crypto contexts would still have to change.

My notes so far:

* It's simpler to key off "#if OPENSSL_VERSION_NUMBER < 0x10100000L" than to add autoconf tests.

* These conditionals don't belong in k5-platform.h.  We could add a k5-openssl-compat.h, or we could use libcrypto and PKINIT private headers and just duplicate the little bit we need.

* The handling of HMAC_CTX is very unsatisfactory, with preprocessor conditionals introduced inside functions and a trap-door "#define HMAC_CTX_free HMAC_CTX_cleanup" added to k5-platform.h.  Luckily we only use HMAC_CTX in one place.

* In the hash providers, why put the output length check after the call to EVP_MD_CTX_new()?  It just makes the flow control uglier.

* There are some little style issues in various parts of the change, such as putting the "*" with the type name instead of the variable name in declarations, or not having the declarations separated from the code by a blank line.

I can fix these issues if necessary; I just don't want to lose track of them.

pixelb:
Agreed on all points Greg.

version checks rather than feature checks breakdown in the presence of forward and backports of course, but that's probably not an issue here. I've just noticed now that we only depend on openssl >= 0.9.8 with pkinit enabled. I.E. the patch is a lot simpler when we can assume the presence of EVP_CIPHER_new(). Should we unconditionally check for openssl >= 0.9.8 now?

greghudson:
Requiring OpenSSL 0.9.8 or later for the libk5crypto OpenSSL implemenation should be fine.  However, krb5 should build (without PKINIT and with the builtin libk5crypto implenentation) with an older OpenSSL or with no OpenSSL at all.

nicowilliams:
I say that OpenSSL 1.1 means to be source backwards compatible.  Where they aren't that should result in an issue/PR being filed against OpenSSL.  OpenSSL 1.1 explicitly wants to break ABI backwards compatibility however.

pixelb:
@greghudson I fixed the last two points you raised.

I had too many questions on the options for your first two points, so it's probably quicker to let you adjust for those as you see fit, if that's Ok.

thanks.

vdukhovni:
> [ @nicowilliams ] I've mentioned this to @vdukhovni, and he says that there should be compatibility macros in OpenSSL, and that this should be opened with OpenSSL as an issue/PR. Mind you, I did something in Heimdal like what you propose here: work around the problem.

A few points:
* Wherever reasonably possible the OpenSSL 1.1.0 release should be source compatible with previous releases, but this is not always possible. Whenever I've noticed cases where compatibility could be retained, but wasn't I've asked for appropriate changes, or fixed it myself.
* At my instigation, OpenSSL 1.1.0 has an OPENSSL_API_COMPAT macro that users can define to a particular release level (0x10100000L == 1.1.0 see <openssl/opensslv.h>) which can be used to indicate the most recent API with which the caller's code is compatible.
* As OpenSSL makes API changes, we want users to adopt newer interfaces, and to no longer rely on old cruft. Therefore applications should, as opportunities arise, attempt to bump this macro to the latest release they're porting to, and use the preferred interface for that release.
* Once you've ported to a particular API, you can use that OPENSSL_API_COMPAT value and expect reasonable compatibility with future versions.
* The 1.1.0 API made many previously exposed data structures opaque so full source compatibility is not possible. However if it is obvious that we could have been compatible, but missed an opportunity, open a pull-request with openssl/openssl or post to openssl-dev, ...
* I expect a loss less by way of incompatible changes (if any) in 1.2.0.
* Finally, you'll typically also want to make sure that code ported to 1.1.0 still works with older releases.  Here unfortunately, it is not possible to retroactively add compatibility interfaces to all the deployed old software versions out in the field.  Therefore, you'll typically end-up implementing compatibility macros or functions along the lines of (example from upcoming Postfix 3.2 code):
```
  /* Backwards compatibility with OpenSSL < 1.1.0 */
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
 #define OpenSSL_version_num SSLeay
 #define OpenSSL_version SSLeay_version
 #define OPENSSL_VERSION SSLEAY_VERSION
 #define X509_up_ref(x) \
        CRYPTO_add(&((x)->references), 1, CRYPTO_LOCK_X509)
 #define EVP_PKEY_up_ref(k) \
        CRYPTO_add(&((k)->references), 1, CRYPTO_LOCK_EVP_PKEY)
#define X509_STORE_CTX_get0_cert(ctx) ((ctx)->cert)
#define X509_STORE_CTX_get0_untrusted(ctx) ((ctx)->untrusted)
#define X509_STORE_CTX_set0_untrusted X509_STORE_CTX_set_chain
#define X509_STORE_CTX_set0_trusted_stack X509_STORE_CTX_trusted_stack
 #endif
```

So by all means port to the new way, and test with ``-DOPENSSL_API_COMPAT=0x10100000L``.  Get the same code to work with older releases via version-conditional forward-compatibility.  Then set OPENSSL_API_COMPAT in the final builds shipped to users.  Repeat as new OpenSSL releases appear that introduce relevant API changes.

I hope this is useful

greghudson:
My edited version is at https://github.com/greghudson/krb5/commits/pr442

rc4.c and hmac.c have the most changes, and could use additional review from another person.

greghudson:
I tried building krb5 against the OpenSSL master with this PR (as well as my edited branch), and PKINIT doesn't seem very close to building, mainly because it looks like the DH type is now opaque.  Was this PR only tested against beta 1, perhaps?

pixelb:
Yes I tested this with openssl-1.1.0-pre4

greghudson:
Assuming the changes on OpenSSL master are going to make it into 1.1, it looks like a lot of works remains to ensure that PKINIT will build, since DH and X509_STORE_CTX are both opaque.  The current state of the OpenSSL master also seems too broken to test against; <openssl/asn1_mac.h> errors out with an undefined reference to ASN1_const_CTX if you include it, and X509_STORE_set_verify_cb_func() is still a macro which dereferences the now-opaque X509_STORE type.

pixelb:
Yes, it seems we should park this for a while until openssl stabilizes.
I'm surprised the interface is changing at this stage :(

greghudson:
Actually, this is probably our window to get any upstream issues fixed before OpenSSL 1.1 releases.  The asn1_mac.h issue was my fault (it was a stale header from a previous install), but the X509 verify callbacks issue is upstream's and I think will be fixed soon.  I will try to figure out if there are any other upstream issues affecting the PKINIT code.

vdukhovni:
The X509_STORE macro will be fixed shortly.

vdukhovni:
https://github.com/openssl/openssl/commit/7cafbb4bd373f024c4900dcaa71aaf626c2ddbae

greghudson:
Okay, I don't think there are any more upstream issues in the way.  I have pushed the work I have to my pr442 branch.  It builds against OpenSSL 1.1 and passes tests, but it's not very close to ready:

* It doesn't build against OpenSSL 1.0 because I haven't added the necessary conditionalization.  (If there's anything in 1.0 and not 0.9.8 which will make this easier, I will just bump up the required OpenSSL version for PKINIT, I think.)

* There are significant changes to prepare_enc_data(), which is only used in RSA mode, which is therefore not exercised by the test suite (I think).  So I will want to see if I can add RSA-mode PKINIT tests.

greghudson:
I am closing this PR and replacing it with PR #447.

PR #443, opened 2016-04-28
Krb5 1.10 dns_canonicalize_hostname flag support

laebshade:
This is a backport of the dns_canonicalize_hostname feature present in later versions.

I realize this is a stale branch; however, my company uses this version on CentOS 6.  A bug report was submitted to have a similar patch implemented: https://bugs.centos.org/view.php?id=10775, but I believe committing here is important, too.

### DIFF
--- a/doc/admin.texinfo
+++ b/doc/admin.texinfo
@@ -587,6 +587,13 @@ depends on configure-time options; if none were given, the default is to
 disable this option.  If the DNS support is not compiled in, this entry
 has no effect.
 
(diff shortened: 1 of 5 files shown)


greghudson:
Hi, sorry for the slow response.  We don't backport new features to existing release branches and we no longer support the 1.10 branch, so I'm going to close this PR without accepting the change.  It's of course fine with us if CentOS wants to backport the feature to their package, of course.

PR #444, opened 2016-04-28
Unconstify some krb5 GSS OIDs

tlyu:
gssapi_krb5.h declared some well-known OID constants as pointers to
const gss_OID_desc, which can't be assigned to application-declared
gss_OID variables or passed to GSSAPI functions without causing
warnings.

Declare these OID constants without the const qualifier on
gss_OID_desc, at the expense of some type safety.  (Fixing this
"correctly" probably requires some standards revision.)

ticket: 8399 (new)


greghudson:
Looks good to me.

PR #445, opened 2016-05-02
Fix cstyle-file.py when emacs is not installed

greghudson:
emacs_reindent() is intended to fail gracefully when emacs is not
installed, but instead subprocess.call() throws an OSError.  Check for
this error and return normally.


tlyu:
Should there be a warning that missing emacs means omitting some indentation checks?

tlyu:
I also manually tested both master and this pull request's cstyle-file.py with an artificially restrictive PATH setting and this change appears to work as advertised.

greghudson:
I think a warning is unnecessary; if we notice indentation issues from a contributor but no other style issues, we can just recommend that they install emacs.  Making a warning visible would be mildly inconvenient given the architecture of the style checker, and would be an annoyance to anyone who just wants to use the style checker without emacs indentation checking.

PR #446, opened 2016-05-03
Implement principal renaming in LDAP

SaharahSarah:
The previous way of renaming principals by adding a new entry and deleting the old one did not work in LDAP due to issues such as requiring renaming the DN in standalone principals, and modifying the DN in place for non-standalone principals. A new DAL function for renaming principals is added so that the plugins can properly rename the principal for their particular database. A function is also added to libkdb_ldap that implements the DAL rename_principal function that properly renames the DN and attributes when necessary.

[ticket: 8065](https://krbdev.mit.edu/rt/Ticket/Display.html?id=8065)

### DIFF
--- a/src/lib/kdb/kdb5.c
+++ b/src/lib/kdb/kdb5.c
@@ -298,6 +298,8 @@ kdb_setup_opt_functions(db_library lib)
         lib->vftabl.decrypt_key_data = krb5_dbe_def_decrypt_key_data;
     if (lib->vftabl.encrypt_key_data == NULL)
         lib->vftabl.encrypt_key_data = krb5_dbe_def_encrypt_key_data;
+    if (lib->vftabl.rename_principal == NULL)
+        lib->vftabl.rename_principal = krb5_db_def_rename_principal;
 }
 
 #ifdef STATIC_PLUGINS
@@ -949,6 +951,37 @@ krb5_db_delete_principal(krb5_context kcontext, krb5_principal search_for)
     status = ulog_add_update(kcontext, &upd);
     free(princ_name);
     return status;
+}
+
+krb5_error_code
+krb5_db_rename_principal(krb5_context kcontext, krb5_principal source,
+                         krb5_principal target)
+{
+    kdb_vftabl *v;
+    krb5_error_code status = 0;
+    krb5_db_entry *entry;
+
+    status = get_vftabl(kcontext, &v);
+    if (status)
+        return status;
+
+    /*
+     * If the default rename function isn't used and logging is enabled, iprop
+     * would fail since it doesn't formally support renaming.  In that case
+     * return KRB5_PLUGIN_OP_NOTSUPP.
+     */
+    if (v->rename_principal != krb5_db_def_rename_principal &&
+        logging(kcontext))
+        return KRB5_PLUGIN_OP_NOTSUPP;
+
+    status = krb5_db_get_principal(kcontext, target, KRB5_KDB_FLAG_ALIAS_OK,
+                                   &entry);
+    if (status == 0) {
+        krb5_db_free_principal(kcontext, entry);
+        return KRB5_KDB_INUSE;
+    }
+
+    return v->rename_principal(kcontext, source, target);
 }
 
 /*
--- a/src/lib/kdb/kdb5.c
+++ b/src/lib/kdb/kdb5.c
@@ -2260,6 +2260,48 @@ krb5_dbe_compute_salt(krb5_context context, const krb5_key_data *key,
     return 0;
 }
 
+krb5_error_code
+krb5_dbe_specialize_salt(krb5_context context, krb5_db_entry *entry)
+{
+    krb5_int16 stype, i;
+    krb5_data *salt = NULL;
+    krb5_error_code ret = 0;
+    uint8_t *data;
+
+    if (context == NULL || entry == NULL)
+        return EINVAL;
+
+    /*
+     * Store salt values explicitly so that they don't depend on the principal
+     * name.
+     */
+    for (i = 0; i < entry->n_key_data; i++) {
+        ret = krb5_dbe_compute_salt(context, &entry->key_data[i], entry->princ,
+                                    &stype, &salt);
+        if (ret)
+            goto cleanup;
+
+        data = krb5_db_alloc(context, NULL, salt->length);
+        if (data == NULL) {
+            ret = ENOMEM;
+            goto cleanup;
+        }
+        memcpy(data, salt->data, salt->length);
+
+        entry->key_data[i].key_data_type[1] = KRB5_KDB_SALTTYPE_SPECIAL;
+        krb5_db_free(context, entry->key_data[i].key_data_contents[1]);
+        entry->key_data[i].key_data_contents[1] = data;
+        entry->key_data[i].key_data_length[1] = salt->length;
+        entry->key_data[i].key_data_ver = 2;
+        krb5_free_data(context, salt);
+        salt = NULL;
+    }
+
+cleanup:
+    krb5_free_data(context, salt);
+    return ret;
+}
+
 /* change password functions */
 krb5_error_code
 krb5_dbe_cpw(krb5_context kcontext, krb5_keyblock *master_key,
--- a/src/lib/kdb/kdb_default.c
+++ b/src/lib/kdb/kdb_default.c
@@ -538,3 +538,47 @@ clean_n_exit:
         krb5_dbe_free_key_list(context, mkey_list_head);
     return retval;
 }
+
+krb5_error_code
+krb5_db_def_rename_principal(krb5_context kcontext,
+                             krb5_const_principal source,
+                             krb5_const_principal target)
+{
+    krb5_db_entry *kdb = NULL;
+    krb5_principal old;
+
+    krb5_error_code ret;
+
+    if (source == NULL || target == NULL)
+        return EINVAL;
+
+    ret = krb5_db_get_principal(kcontext, source,
+                                KRB5_KDB_FLAG_ALIAS_OK, &kdb);
+    if (ret)
+        goto cleanup;
+
+    /*
+     * Store salt values explicitly so that they don't depend on the principal
+     * name.
+     */
+    ret = krb5_dbe_specialize_salt(kcontext, kdb);
+    if (ret)
+        goto cleanup;
+
+    old = kdb->princ;
+    kdb->princ = (krb5_principal) target;
+
+    ret = krb5_db_put_principal(kcontext, kdb);
+
+    kdb->princ = old;
+
+    if (ret)
+        goto cleanup;
+
+    ret = krb5_db_delete_principal(kcontext, (krb5_principal)source);
+
+
+cleanup:
+    krb5_db_free_principal(kcontext, kdb);
+    return ret;
+}
--- a/src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
+++ b/src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
@@ -799,6 +799,54 @@ get_str_from_tl_data(krb5_context context, krb5_db_entry *entry, int type,
     return 0;
 }
 
+/*
+ * Replace the rdn in a dn with a new one.
+ */
+krb5_error_code
+replace_rdn(krb5_context context, const char *dn, const char *newrdn,
+            char **newdn)
+{
+    krb5_error_code ret;
+    LDAPDN ldn = NULL;
+    LDAPRDN lrdn = NULL;
+    char *newdn_out = NULL, *next;
+
+    *newdn = NULL;
+    ret = ldap_str2dn(dn, &ldn, LDAP_DN_FORMAT_LDAPV3);
+    if (ret != LDAP_SUCCESS || ldn[0] == NULL) {
+        ret = EINVAL;
+        goto cleanup;
+    }
+
+    ret = ldap_str2rdn(newrdn, &lrdn, &next, LDAP_DN_FORMAT_LDAPV3);
+    if (ret != LDAP_SUCCESS) {
+        ret = EINVAL;
+        goto cleanup;
+    }
+
+    ldap_rdnfree(ldn[0]);
+    ldn[0] = lrdn;
+    lrdn = NULL;
+
+    ret = ldap_dn2str(ldn, &newdn_out, LDAP_DN_FORMAT_LDAPV3);
+    if (ret != LDAP_SUCCESS) {
+        ret = KRB5_KDB_SERVER_INTERNAL_ERR;
+        goto cleanup;
+    }
+
+    *newdn = newdn_out;
+    newdn_out = NULL;
+    ret = 0;
+
+cleanup:
+    free(newdn_out);
+    if (ldn != NULL)
+        ldap_dnfree(ldn);
+    if (lrdn != NULL)
+        ldap_rdnfree(lrdn);
+    return ret;
+}
+
 krb5_error_code
 krb5_get_userdn(krb5_context context, krb5_db_entry *entry, char **userdn)
 {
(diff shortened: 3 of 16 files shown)


greghudson:
We'll have to think about how this new kdb5 function interacts with iprop.  At a minimum some internal documentation might be required.

Assuming we don't modify the iprop protocol (which is certainly easiest), then we don't need a krb5int_rename_principal_no_log() function, and I would prefer that its contents be folded into krb5_db_rename_principal().

Also, I think the second call to log_princ_update() should have FALSE as the third argument.  It would be good to add a rename test to t_iprop.py, probably around line 344 after the delete test and before the forced full resync.

(I haven't looked over the third commit yet.  The first commit seemed okay with the exception that some comments could be improved.)

tlyu:
krb5_dbe_specialize_salt() appears to unconditionally convert all key data entries to special salt. It should probably only do so for normal and norealm salt types. I'm not sure about the onlyrealm salt type; we probably don't expect to change realm names of a principal during a rename operation, but I could be missing something (maybe cross-realm principals?).

greghudson:
The current rename behavior converts all key data to the special salt type.  If we want to change that behavior I would make it a separate commit, but I don't see a need to change the behavior.

SaharahSarah:
Iprop is fixed. krb5_db_rename_principal fails if the plugin doesn't use krb5_db_def_rename_principal as the rename function.

greghudson:
I think krb5_dbe_specialize_salt() should copy the computed salt into a memory region allocated with krb5_db_alloc(), since that memory will later be freed inside the DB module.  (The existing salt transformation code in svr_principal.c is incorrect for not doing so.)  See kdb_cpw.c for examples of this being done in krb5_dbe_* functions.

The comment for krb5_dbe_specialize_salt() could be improved.  I suggest "Modify the key data of entry to explicitly store salt values using the KRB5_KDB_SALTTYPE_SPECIAL salt type."

greghudson:
In the second commit:

* We try to avoid using the word "plugin" as a noun when writing formally.  So in the commit message and in comments, refer to "KDB modules" rather than "plugins".

* Use imperatives rather than passive voice in the commit message: "Add a new DAL function to support renames in all KDB modules, with a default implementation that performs the old functionality..."

* In the rename_principal comment, use two spaces after periods.  "the call will fail if iprop is enabled" should say "the rename will fail if iprop logging is enabled."  (The distinction is that iprop logging happens on the master, whereas "iprop is enabled" could apply to a master or a slave.)

* log_princ_update() is unused and can be removed from the commit.

* The default function for renaming should be set in kdb_setup_opt_functions(), and the DB2 function table should just have NULL in this spot.  (This design dates from when we were avoiding circular dependencies between kdb5 and KDB modules, so it's a little weird, but we should follow it for consistency.)  The default implementation should also be in kdb_default.c rather than kdb5.c.  I don't think there should be any need to include it in the library export list.  (Some, but not all, of the current default functions are in the export list, but I think that's an oversight as they shouldn't need to be there.)

* In  krb5_db_rename_principal() where we check for custom rename handlers on an iprop master, there is no need to check for v->rename_principal == NULL, but there should be a comment explaining why we have this restriction.  The commit message should also briefly mention this limitation.

* In krb5_db_def_rename_principal():
** Why begin by clearing the error message?  krb5_clear_error_message() should generally only be called when an error code was received from a subsidiary function which might have set an extended message, but the error is being handled internally or ignored and not passed up the chain.
** The comment "Transform salts as necessary" isn't very communicative.  "Store salt values explicitly so that they don't depend on the principal name." would be better.
** Where you replace the principal name, it would be safer to set kdb->princ = NULL right after freeing it, instead of when you see an error from krb5_copy_principal().  (I don't understand your comment "so freeing the dbe doesn't lose target data"; what you are avoiding is a double free of the original principal name.)
** There's another violation of malloc hygiene here; krb5_copy_principal() allocates using the libkrb5 allocator but the name is freed inside the KDB module.  We could finesse this issue by setting kdb->princ just before calling krb5_db_put_principal() and then resetting it back to its old value immediately after.

* Adding a function to kdb_vftabl is an incompatible change, so KRB5_KDB_DAL_MAJOR_VERSION needs to be bumped in kdb.h.

src/lib/kdb/kdb5.c
```diff
@@ -2260,6 +2260,46 @@ krb5_dbe_compute_salt(krb5_context context, const krb5_key_data *key,
...
+        if (data == NULL) {
+            ret = ENOMEM;
+            goto cleanup;
+        }
+        memcpy(data, salt->data, salt->length);
+        free(salt->data);
+
+        entry->key_data[i].key_data_type[1] = KRB5_KDB_SALTTYPE_SPECIAL;
+        krb5_db_free(context, entry->key_data[i].key_data_contents[1]);
+        entry->key_data[i].key_data_contents[1] = (krb5_octet *)data;
```
greghudson: You can avoid this cast by making data of type uint8_t * (or krb5_octet *, but we are phasing out krb5-specific type names) instead of char *.

src/lib/kdb/kdb5.c
```diff
@@ -2260,6 +2260,46 @@ krb5_dbe_compute_salt(krb5_context context, const krb5_key_data *key,
...
+        if (ret)
+            goto cleanup;
+
+        data = krb5_db_alloc(context, NULL, salt->length);
+        if (data == NULL) {
+            ret = ENOMEM;
+            goto cleanup;
+        }
+        memcpy(data, salt->data, salt->length);
+        free(salt->data);
```
greghudson: It would make more sense to free salt using krb5_free_data() at line 2295.

src/plugins/kdb/ldap/libkdb_ldap/ldap_misc.c
```diff
@@ -1069,6 +1117,20 @@ krb5_add_int_mem_ldap_mod(LDAPMod ***list, char *attribute, int op, int value)
...
+{
+    krb5_error_code ret;
+
+    ret = ldap_modify_ext_s(ld, DN, mods, NULL, NULL);
+    if (ret != LDAP_SUCCESS) {
+        ret = set_ldap_error(context, ret, op);
+    }
+
+    return ret;
+}
```
greghudson: The DN argument should be named "dn", not "DN".  This function can be shortened and made more correct (using the correct type for the return valid of ldap_modify_ext_s()) as follows:

    int st;

    st = ldap_modify_ext_s(ld, dn, mods, NULL, NULL);
    return (st == LDAP_SUCCESS) ? 0 : set_ldap_error(context, st, op);

greghudson:
In the final commit:

* Please keep variable names to lowercase; "DN" and "RDN" components should be "dn" and "rdn".

* is_standalone_principal(), unparse_principal_name(), and rename_principal_dn() need comments above their definitions.  I would also suggest naming the third function rename_principal_rdn().

* For a standalone principal, you are currently renaming the entry with deleteoldrdn=0 and then removing the old krbPrincipalName value.  Why not just rename it with deleteoldrdn=1?

* Tests should try very hard to avoid sleeping.  I think we can test the last modified data change without sleeping by looking at the principal name rather than the timestamp.  kadmin.local has a -p flag which can be used to affect the principal used for last modified data; it can be set to anything (even a principal name which doesn't exist) because kadmin.local authenticates locally.

src/lib/kdb/kdb_default.c
```diff
@@ -538,3 +538,52 @@ krb5_def_fetch_mkey_list(krb5_context        context,
...
+
+    /*
+     * Store salt values explicitly so that they don't depend on the principal
+     * name.
+     */
+    ret = krb5_dbe_specialize_salt(kcontext, kdb);
+    if (ret)
+        goto cleanup;
+
+    ret = krb5_copy_principal(kcontext, target, &new);
```
greghudson: We don't need to make this copy; we can just assign kdb->princ = target temporarily.

SaharahSarah:
#### https://github.com/krb5/krb5/pull/446#issuecomment-218493289

> I think krb5_dbe_specialize_salt() should copy the computed salt into a memory region allocated with krb5_db_alloc(), since that memory will later be freed inside the DB module. (The existing salt transformation code in svr_principal.c is incorrect for not doing so.) See kdb_cpw.c for examples of this being done in krb5_dbe_* functions.

Copied the salt into a memory region allocated with krb5_db_alloc.

> The comment for krb5_dbe_specialize_salt() could be improved. I suggest "Modify the key data of entry to explicitly store salt values using the KRB5_KDB_SALTTYPE_SPECIAL salt type."

Modified the comment with the suggestion.

#### https://github.com/krb5/krb5/pull/446#issuecomment-218506644

> In the second commit:
>
> * We try to avoid using the word "plugin" as a noun when writing formally.  So in the commit message and in comments, refer to "KDB modules" rather than "plugins".

Replaced "plugin" with "KDB Module".

> * Use imperatives rather than passive voice in the commit message: "Add a new DAL function to support renames in all KDB modules, with a default implementation that performs the old functionality..."

Reworded the commit with imperatives instead of passive voice..

> * In the rename_principal comment, use two spaces after periods.  "the call will fail if iprop is enabled" should say "the rename will fail if iprop logging is enabled."  (The distinction is that iprop logging happens on the master, whereas "iprop is enabled" could apply to a master or a slave.)

Added two spaces and reworded the comment.

> * log_princ_update() is unused and can be removed from the commit.

Removed log_princ_update()

> * The default function for renaming should be set in kdb_setup_opt_functions(), and the DB2 function table should just have NULL in this spot.  (This design dates from when we were avoiding circular dependencies between kdb5 and KDB modules, so it's a little weird, but we should follow it for consistency.)  The default implementation should also be in kdb_default.c rather than kdb5.c.  I don't think there should be any need to include it in the library export list.  (Some, but not all, of the current default functions are in the export list, but I think that's an oversight as they shouldn't need to be there.)

Set the default function for reaname in `kdb_setup_opt_functions()` instead of individually in the modules. If a module doesn't want to support renaming then it will have to manually add a function that returns KRB5_PLUGIN_OP_NOTSUPP.

`krb5_db_def_rename_principal()` was already in kdb_default.c and never in kdb5.c, so nothing was to be done with this.

Removed `krb5_db_def_rename_principal()` from export list.

> * In  krb5_db_rename_principal() where we check for custom rename handlers on an iprop master, there is no need to check for v->rename_principal == NULL, but there should be a comment explaining why we have this restriction.  The commit message should also briefly mention this limitation.

Comment explaining the restriction added, and commit message modified to mention the limitation.

> * In krb5_db_def_rename_principal():
>   * Why begin by clearing the error message?  krb5_clear_error_message() should generally only be called when an error code was received from a subsidiary function which might have set an extended message, but the error is being handled internally or ignored and not passed up the chain.

A lot of the code in `krb5_db_def_rename_principal()` was copied from `kadm5_rename_principal()`, which did this already. I'm not sure why it did, but it has been removed from `krb5_db_def_rename_principal()`
 
>   * The comment "Transform salts as necessary" isn't very communicative.  "Store salt values explicitly so that they don't depend on the principal name." would be better.

The comment was copied from `kadm5_rename_principal()`. Modified the comment to be more descriptive.

>   * Where you replace the principal name, it would be safer to set kdb->princ = NULL right after freeing it, instead of when you see an error from krb5_copy_principal().  (I don't understand your comment "so freeing the dbe doesn't lose target data"; what you are avoiding is a double free of the original principal name.)

This was copied from `kadm5_rename_principal()`. I have absolutely no idea why they were doing that. It seems that the original author thought that `krb5_copy_principal()` wasn't a deep copy.

>   * There's another violation of malloc hygiene here; krb5_copy_principal() allocates using the libkrb5 allocator but the name is freed inside the KDB module.  We could finesse this issue by setting kdb->princ just before calling krb5_db_put_principal() and then resetting it back to its old value immediately after.

The entry principal now is set directly to the target principal before calling, then set back to the old principal right after the call.

> * Adding a function to kdb_vftabl is an incompatible change, so KRB5_KDB_DAL_MAJOR_VERSION needs to be bumped in kdb.h.

Incremented `KRB5_KDB_DAL_MAJOR_VERSION`

#### kdb5.c: 2291

> You can avoid this cast by making data of type uint8_t * (or krb5_octet *, but we are phasing out krb5-specific type names) instead of char *.

Changed the type of `data` to `uint8_t *`

#### kdb5.c: 2287

> It would make more sense to free salt using krb5_free_data() at line 2295.

The salt is now freed using `krb5_free_data()`

#### ldap_misc.c: `krb5_ldap_modify_ext()`

> The DN argument should be named "dn", not "DN".  This function can be shortened and made more correct (using the correct type for the return valid of ldap_modify_ext_s()) as follows:
>
>     int st;
>
>     st = ldap_modify_ext_s(ld, dn, mods, NULL, NULL);
>     return (st == LDAP_SUCCESS) ? 0 : set_ldap_error(context, st, op);

`DN` set to `dn`. `ldap_modify_ext_s()` shortened using suggested changes.

#### https://github.com/krb5/krb5/pull/446#issuecomment-218804442

> In the final commit:

> * Please keep variable names to lowercase; "DN" and "RDN" components should be "dn" and "rdn".

This was done this way based off the surrounding code style. Set all the uppercase variable names to lowercase.

> * is_standalone_principal(), unparse_principal_name(), and rename_principal_dn() need comments above their definitions.  I would also suggest naming the third function rename_principal_rdn().

Added comments above the defintions, and renamed the third function.

> * For a standalone principal, you are currently renaming the entry with deleteoldrdn=0 and then removing the old krbPrincipalName value.  Why not just rename it with deleteoldrdn=1?

I personally am testing this on 389ds. The call failes if deleteoldrdn=1. Not sure if OpenLDAP fails as well but since at least one popular LDAP server fails then it should stay the current way of deleting the old rdn later.

> * Tests should try very hard to avoid sleeping.  I think we can test the last modified data change without sleeping by looking at the principal name rather than the timestamp.  kadmin.local has a -p flag which can be used to affect the principal used for last modified data; it can be set to anything (even a principal name which doesn't exist) because kadmin.local authenticates locally.

The test now uses the -p flag to kadmin.local instead of sleeping. I didn't know if sleep was appropriate, but earlier in the file there's a point where sleep is used as well.

greghudson:
> I personally am testing this on 389ds. The call failes if deleteoldrdn=1. Not sure if OpenLDAP fails as well but since at least one popular LDAP server fails then it should stay the current way of deleting the old rdn later.

We should definitely try to work on 389ds, but I can't think of a reason why deleteoldrn=1 should cause that operation to fail, and I think we should investigate further.  Can you reproduce this error and say what the reported error is?

SaharahSarah:
The error occurs in both the code and if I run it from the command line.

```
dn: krbprincipalname=test@TEST.KRB5.SARAH,cn=TEST.KRB5.SARAH,cn=krbcontainer,dc=krb5,dc=sarah
changetype: modrdn
newrdn: krbprincipalname=test1@TEST.KRB5.SARAH
deleteoldrdn: 1

modifying rdn of entry "krbprincipalname=test@TEST.KRB5.SARAH,cn=TEST.KRB5.SARAH,cn=krbcontainer,dc=krb5,dc=sarah"
ldap_rename: Operations error (1)
```

greghudson:
I can reproduce this problem, and I think it's a bug in 389ds.  I will try to report it upstream, but we will want to work with broken versions of 389ds at least for the moment.  Please add a comment to the function explaining why we're modifying the RDN in an awkward way; it's very important to leave a clear trail when working around other people's bugs in our code.

greghudson:
I filed this upstream bug report: https://fedorahosted.org/389/ticket/48848

SaharahSarah:
> I can reproduce this problem, and I think it's a bug in 389ds. I will try to report it upstream, but we will want to work with broken versions of 389ds at least for the moment. Please add a comment to the function explaining why we're modifying the RDN in an awkward way; it's very important to leave a clear trail when working around other people's bugs in our code.

The comment was already added. Thank you for filing the bug report.

greghudson:
Pushed to master as 09cdb3079508f9d3fcc107a8ee8e7537f70c9d37 c38838be956ce72fcd7142f14bc374dc13dd8bb2 2ac75e548afadde4f87f20fcc1ee1472cdac3fed.  I noticed two naming issues in the third commit and want to call attention to them:

* replace_rdn() and rename_principal_rdn() had parameters named "newdn" and a local variable named "newdn_out".  This is reversed from our usual practices for new code; we use the suffix _out on output parameter names, not for local variables which are used to temporarily hold results.  In both of those functions, I changed the parameter name to use the _out suffix and removed the local variable (which we don't need if the last operation in the function can set the output parameter).

* In unparse_principal_name(), the variable "res" held the krb5_error_code result of subsidiary calls.  It's okay to go with the prevailing name in the file being modified, or with "ret" which is our usual practice for new code, but try to avoid other variants.  I changed the name of res to "st", which is the prevailing name in the file for this purpose, and renamed the user parameter to user_out.

There were also a lot of multi-sentence comments with only one space after the period; I fixed those.

PR #449, opened 2016-05-08
Link correct VS2015 C libraries for debug builds

greghudson:
When building with VS2015 and NODEBUG is not set, link with the
debugging versions of ucrt.lib and vcruntime.lib.  Based on work by
Chris Kingsley.


SaharahSarah:
This builds fine in my Visual Studio 2010 install on Windows 7.

PR #450, opened 2016-05-09
Fix unlikely pointer error in get_in_tkt.c

greghudson:
In add_padata(), reset the caller's pointer as soon as realloc()
succeeds, to ensure that the old pointer isn't left behind if a later
allocation fails.

(Note to self: after merging this, make a note in the RT ticket about the 1.14/1.13 backport, since it won't be a trivial merge.)


SaharahSarah:
Looks good :+1:

greghudson:
I found a bug in my fix, and pushed a corrected commit.

tlyu:
The revised patch looks ok to me.

PR #451, opened 2016-05-12
Don't indicate deprecated mechanisms

frozencemetery:
This addresses issues with gsstest.  As such, it is the successor to @simo5's [initial approach](https://github.com/simo5/krb5/commit/a180341bea83ebc699c9f46bf59922d264b7656c).  I expect this to need revision; at very least, there was discussion around the possibility of a tunable that I'm not sure concluded.

### DIFF
--- a/src/lib/gssapi/mechglue/g_initialize.c
+++ b/src/lib/gssapi/mechglue/g_initialize.c
@@ -202,12 +202,56 @@ gss_OID *oid;
 	return (generic_gss_release_oid(minor_status, oid));
 } /* gss_release_oid */
 
+/*
+ * Wrapper around inquire_attrs_for_mech to determine whether a mechanism has
+ * the deprecated attribute.  Must be called without g_mechSetLock since it
+ * will call into the mechglue.
+ */
+static int
+is_deprecated(gss_OID element)
+{
+	OM_uint32 major, minor;
+	gss_OID_set mech_attrs;
+	int deprecated = 0;
+
+
+	major = gss_inquire_attrs_for_mech(&minor, element, &mech_attrs, NULL);
+	if (major == GSS_S_COMPLETE) {
+		gss_test_oid_set_member(&minor, (gss_OID)GSS_C_MA_DEPRECATED,
+					mech_attrs, &deprecated);
+	}
+
+	if (mech_attrs != GSS_C_NO_OID_SET)
+		gss_release_oid_set(&minor, &mech_attrs);
+
+	return deprecated;
+}
+
+/*
+ * Removes mechs with the deprecated attribute from an OID set.  Must be
+ * called without g_mechSetLock held since it calls into the mechglue.
+ */
+static void
+prune_deprecated(gss_OID_set mech_set)
+{
+	OM_uint32 i, j;
+
+        j = 0;
+        for (i = 0; i < mech_set->count; i++) {
+            if (!is_deprecated(&mech_set->elements[i]))
+                mech_set->elements[j++] = mech_set->elements[i];
+            else
+                gssalloc_free(mech_set->elements[i].elements);
+        }
+        mech_set->count = j;
+}
 
 /*
  * this function will return an oid set indicating available mechanisms.
  * The set returned is based on configuration file entries and
  * NOT on the loaded mechanisms.  This function does not check if any
  * of these can actually be loaded.
+ * Mechanisms which are deprecated will not be returned.
  * This routine needs direct access to the mechanism list.
  * To avoid reading the configuration file each call, we will save a
  * a mech oid set, and only update it once the file has changed.
@@ -245,6 +289,10 @@ gss_OID_set *mechSet_out;
 	k5_mutex_lock(&g_mechSetLock);
 	status = generic_gss_copy_oid_set(minorStatus, &g_mechSet, mechSet_out);
 	k5_mutex_unlock(&g_mechSetLock);
+
+	if (*mechSet_out != GSS_C_NO_OID_SET)
+		prune_deprecated(*mechSet_out);
+
 	return (status);
 } /* gss_indicate_mechs */
 


greghudson:
We generally avoid hardcoding mechanism OIDs in the mechglue (with at least one exception related to SPNEGO).  It would be better to do this using mechanism attributes.

frozencemetery:
Makes sense.  However, I don't seem to be able to call gss_inquire_attrs_for_mech from here; it looks like it causes the init function to run, and we're being called from init in some cases already.  Do you have a suggestion on how to do this without reimplementing inquire_attrs_for_mech (or should I just do that)?

greghudson:
What is the call chain which leads from initialization to gss_indicate_mechs()?

If it weren't for interposition, you could call mech->gss_inquire_attrs_for_mech() directly, which would be very easy.  But with interposition, reimplementing gss_inquire_attrs_for_mech() is a fair amount of code, and if we're calling gss_indicate_mechs() during initialization, there's the concern that the relevant interposer mechs hasn't been loaded yet.

Taking a step back, perhaps the mechglue should have access to each mechanism's asserted and known mechanism attributes in struct gss_mech_info, and a mechanism's gss_inquire_attrs_for_mech() should only be called once at initialization time.  That would carry some performance and simplicity benefits, but it's not a trivial change, especially given interposition (unless we say that querying mech attributes can't be interposed, which I still think is a fair position).

frozencemetery:
Hmm, that suggests a different idea.  We're already skipping interposer mechs here anyway, so this doesn't need to work on them and we can in fact just call `mech->gss_inquire_attrs_for_mech()` directly.  Pushed new version.

As for what goes wrong, here's the trace (assuming is_depricated() is a function that calls `gss_inquire_attrs_for_mech()` on each non-interposer mech) from when kadmin startd in the test suite and tries to get the lock twice:

```
(gdb) bt
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007ffff61a4b05 in __GI___pthread_mutex_lock (mutex=0x7ffff777b9e0 <g_mechListLock>) at ../nptl/pthread_mutex_lock.c:81
#2  0x00007ffff753b7e6 in k5_mutex_lock (m=0x7ffff777b9e0 <g_mechListLock>) at ../../../include/k5-thread.h:379
#3  0x00007ffff7540dff in gssint_select_mech_type (minor=0x7fffffffdd48, oid=0x6180c0, selected_oid=0x7fffffffdce8) at g_initialize.c:979
#4  0x00007ffff75430eb in gss_inquire_attrs_for_mech (minor=0x7fffffffdd48, mech_oid=0x6180c0, mech_attrs=0x7fffffffdd40, known_mech_attrs=0x0) at g_mechattr.c:177
#5  0x00007ffff753bcde in is_deprecated (mech=0x6180c0) at g_initialize.c:274
#6  0x00007ffff753be24 in build_mechSet () at g_initialize.c:340
#7  0x00007ffff753bbf3 in gss_indicate_mechs (minorStatus=0x7fffffffdf74, mechSet_out=0x7fffffffde00) at g_initialize.c:238
#8  0x00007ffff7542ef2 in gss_indicate_mechs_by_attrs (minor=0x7fffffffdf74, desired_mech_attrs=0x0, except_mech_attrs=0x7fffffffde90, critical_mech_attrs=0x0, mechs=0x7fffffffdea8) at g_mechattr.c:113
#9  0x00007ffff753413b in gss_acquire_cred_from (minor_status=0x7fffffffdf74, desired_name=0x620700, time_req=0, desired_mechs=0x0, cred_usage=2, cred_store=0x0, output_cred_handle=0x6206c0, actual_mechs=0x0, time_rec=0x0) at g_acquire_cred.c:164
#10 0x00007ffff7534012 in gss_acquire_cred (minor_status=0x7fffffffdf74, desired_name=0x620700, time_req=0, desired_mechs=0x0, cred_usage=2, output_cred_handle=0x6206c0, actual_mechs=0x0, time_rec=0x0) at g_acquire_cred.c:107
#11 0x00007ffff778f050 in gssrpc_svcauth_gssapi_set_names (names=0x7fffffffe050, num=2) at svc_auth_gssapi.c:975
#12 0x000000000040bf8c in main (argc=0, argv=0x7fffffffe1e8) at ovsec_kadmd.c:479
(gdb) 
```

greghudson:
Now that I look at that stack trace, I realize something I should have caught immediately: you shouldn't be removing deprecated mechanisms from g_mechSet.  That would prevent those mechanisms from being used by applications at all.  If that were acceptable, we'd just remove those mechanism OIDs from the krb5 mech and be done.  You need to do the filtering in gss_indicate_mechs().

Calling mech->gss_inquire_attrs_for_mech() directly does not have the same semantics as calling gss_inquire_attrs_for_mech(), even though we aren't reporting interposer OIDs.  If there is an interposer mech corresponding to a public mech OID, gss_inquire_attrs_for_mech() will call the interposer, while mech->gss_inquire_attrs_for_mech() will not.

frozencemetery:
New version.  This does the filtering in `gss_indicate_mechs()` itself.  I would have liked it to filter and copy in a single operation (rather than pruning a copy), but this would require substantial refactoring due to handling of `g_mechSetLock` acquisition and holding.

greghudson:
I'm happy with pruning after the copy.

It's probably better for is_deprecated() to return int, rather than OM_uint32.  Also, you can pass NULL for known_mech_attrs to gss_inquire_attrs_for_mech() to save a little bit of code and runtime.

prune_deprecated() uses an O(n^2) loop, which isn't likely to be a performance issue, but it's less code to do it with an O(n) loop.  (The process remains O(n^2) because gss_inquire_attrs_for_mech() is O(n) in the current implementation.)  Selectively copying a vector to itself is a fairly common pattern in C:

    j = 0;
    for (i = 0; i < mech_set->count; i++) {
        if (!is_deprecated(&mech_set->elements[i]))
            mech_set->elements[j++] = mech_set->elements[i];
        else
            gssalloc_free(mech_set->elements[i].elements)
    }
    mech_set->count = j;

frozencemetery:
Thanks for the comments.  New version fixes a spelling error, the return type on `is_deprecated()`, passes NULL for `known_mech_attrs`, and uses your O(n) loop instead.

greghudson:
Pushed to master as 8cbfec302659812efee9b67031f5ae0c5400953f.  Nitpicks I fixed:

* The style-checker script noted the use of eight spaces instead of tabs in prune_deprecated().
* There were two blank lines after the declarations in is_deprecated().

I also opted to initialize mech_attrs in is_deprecated() as a defensive measure.  I think we usually code defensively when it comes to the possibility of subsidiary functions leaving their output parameters unchanged, even though GSS-API functions are rigorous about never doing so.

PR #452, opened 2016-05-12
Check princ length in krb5_sname_match()

greghudson:
krb5_sname_match() can read past the end of princ's component array in
some circumstances (typically when a keytab contains both "x" and
"x/y" principals).  Add a length check.  Reported by Spencer Jackson.

There is also a second commit to add unit tests.


tlyu:
This looks ok to me.

PR #453, opened 2016-05-17
Use library allocator, not module allocator, for principal/policy entries

greghudson:
See http://mailman.mit.edu/pipermail/krbdev/2016-May/012583.html

This will have some conflicts with PR #446, so whichever one is merged first, the other will need adjusting.


greghudson:
I rebased this on master after pushing the LDAP rename changes.

tlyu:
This looks ok to me.

PR #455, opened 2016-05-26
Clean t_sname_match in lib/krb5/krb

tlyu:
Add a missing "$" to t_sname_match$(EXEEXT) in the clean rule in
lib/krb5/krb/Makefile.in.


greghudson:
Looks good.

PR #456, opened 2016-05-26
Relocate cb->set_as_key() in otp preauth

npmccallum:
The previous position of this code caused caused the as_key to be
replaced even when later code in the fuction would fail. The effect of
this is that a responder callback cannot choose another preauth method
when the otp method is offered.


npmccallum:
I would also like to propose this fix for any future minor releases.

greghudson:
Pushed to master as 0712d0059d72ddeaf1764f8fa173a321e3bc072d

PR #457, opened 2016-05-27
Relax t_sn2princ.py reverse resolution test

tlyu:
Relax t_sn2princ.py check of the reverse resolution of the test
hostname.  The new requirement is that it be different from the
forward resolved hostname.  (There is also an existing implicit
requirement that it be in the mit.edu domain.)  This makes
t_sn2princ.py more robust against changes in the reverse resolution of
the test hostname.

ticket: 8422 (new)
tags: pullup
target_version: 1.14-next
target_version: 1.13-next


greghudson:
Looks okay to me.

PR #460, opened 2016-06-02
Fix leak in test_check_allowed_to_delegate()

greghudson:
sprinc and tprinc must be freed.  Reported by Will Fiveash.


tlyu:
This looks ok to me.

greghudson:
Yes.  I'll move the free statements up before the conditional return in another PR.

PR #461, opened 2016-06-02
Fix bugs in recent locate_kdc.c change

greghudson:
The most recent change to locate_srv_conf_1() introduced a possible
double-free bug (detected by Coverity), and also broke MS-KKDCP
support.  Separate the three uses of the "host" variable: the C string
copy of the realm name (now "realmstr"), the pointer to the hostname
or hostname:port specification in the profile values array (now
"hoststr"), and the hostname result of k5_parse_host_string() (still
"host").  Pass the correct pointer to k5_parse_host_string() if the
profile value is a URI.


tlyu:
I would prefer "hostspec" to "hoststr", but that unfortunately would no longer be parallel with the name k5_parse_host_string. Otherwise looks ok to me.

PR #462, opened 2016-06-02
Fix test_check_allowed_to_delegate() leak for real

greghudson:
Remove the conditional return from so that sprinc and tprinc are
always freed.  Reported by Will Fiveash.


wfiveash:
That looks right.

PR #464, opened 2016-06-08
Fix kadmin min_life check with nonexistent policy

greghudson:
In kadmind, self-service key changes require a check against the
policy's min_life field.  If the policy does not exist, this check
should succeed according to the semantics introduced by ticket #7385.
Fix check_min_life() to return 0 if kadm5_get_policy() returns
KADM5_UNK_POLICY.  Reported by John Devitofranceschi.


tlyu:
This looks ok to me.

PR #463, opened 2016-06-07
Do not segfault if we can't check if the KDC is a master

gladiac:
I've run into a segfault when the client tries to talk to the KDC which isn't running.

Look at server_used in frame 8, it is set to 22002. So have an out of bound array access and segfault dereferencing the content.

```
0x00007f7d5c2e00da in waitpid () from /lib64/libc.so.6
#0  0x00007f7d5c2e00da in waitpid () from /lib64/libc.so.6
No symbol table info available.
#1  0x00007f7d5c25b9ab in do_system () from /lib64/libc.so.6
No symbol table info available.
#2  0x00007f7d5e24aae3 in smb_panic_default (why=0x7f7d5e29b655 "internal error") at ../lib/util/fault.c:141
        result = 32637
        pidstr = "29720\000\000\000H\265)^}\177\000\000\325\000\000\001"
        cmdstring = "/home/asn/workspace/projects/samba/selftest/gdb_backtrace 29720\000[\000\000\000n", '\000' <repeats 11 times>, ">\266)^}\177\000\000X\267)^}\177\000\000\260\aL\201\362U\000\000t", '\000' <repeats 23 times>, "\365\252c\323\000\000\000\000\324\000\000\000\000\000\000\000`\247c\323\376\177\000\000\220\323*\201\362U\000\000\324\000\000\000\000\000\000\000\060\302A\201\362U\000\000\060\302A\201\362U\000\000\320\301A\201\362U\000\000`\247c\323\376\177\000"
        __FUNCTION__ = "smb_panic_default"
#3  0x00007f7d5e24ac6e in smb_panic (why=0x7f7d5e29b655 "internal error") at ../lib/util/fault.c:169
No locals.
#4  0x00007f7d5e24a936 in fault_report (sig=11) at ../lib/util/fault.c:83
        counter = 1
        __FUNCTION__ = "fault_report"
#5  0x00007f7d5e24a94b in sig_fault (sig=11) at ../lib/util/fault.c:94
No locals.
#6  <signal handler called>
No symbol table info available.
#7  k5_kdc_is_master (context=context@entry=0x55f2814932a0, realm=realm@entry=0x7ffed363aea0, server=0x55f281842380) at locate_kdc.c:685
        list = {servers = 0x6eda360500000000, nservers = 1859794437}
[...]

PR #465, opened 2016-06-09
Fix use_master handling with KDC hook reply

greghudson:
PR #463 resulted from a buggy KDC recv hook, but a valid hook function could also trigger the uninitialized use of server_used.  Here is a fix.


tlyu:
I'm a little uneasy about the assertion, because it could fail due to an error in code written by an application developer. On the other hand, it isn't as bad as an assertion failing due to invalid user or network input. Otherwise, this commit looks ok to me.

tlyu:
Ease of debugging is important, so I guess I agree with you. The documentation for krb5_post_recv_fn does imply a requirement for the behavior that the assertion is checking, but it could use a little clarification to make it more obvious that if the hook returns successfully after k5_sendto failed, the hook must synthesize a reply.

PR #466, opened 2016-06-16
Minimize timing leaks in PKINIT decryption

greghudson:
This branch steals the first testing commit from PR #447, and then simplifies and improves the PKCS7 EnvelopedData decoding function.


tlyu:
This looks ok to me.

PR #467, opened 2016-06-17
Fix Makefile for paths containing '+' character

veprbl:
include/Makefile uses a regex to perform variable substitution with '+'
used for delimeter. The '+' is also a valid character in a unix path, so
the paths need to be escaped in regex context.

I've tried to replace the sed thing with substitution by means of autoconf,
but it didn't work out for me because paths would get substituted as
"${exec_prefix}/sbin" which is not appropriate for C header file. So this
time I'm only providing hacky fix.

### DIFF
--- a/src/include/Makefile.in
+++ b/src/include/Makefile.in
@@ -57,19 +57,19 @@ SBINDIR = @sbindir@
 LIBDIR  = @libdir@
 SYSCONFCONF = @SYSCONFCONF@
 
-PROCESS_REPLACE = -e "s+@KRB5RCTMPDIR+$(KRB5RCTMPDIR)+" \
-		  -e "s+@PREFIX+$(INSTALL_PREFIX)+" \
-		  -e "s+@EXEC_PREFIX+$(INSTALL_EXEC_PREFIX)+" \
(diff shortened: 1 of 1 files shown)


greghudson:
We don't require gmake, so I don't think "$(subst ...)" is in our toolkit.

As this is just the sed delimiter, perhaps it would be reasonable to use "|", which is less likely to be included in paths because it has a meaning in the shell.

veprbl:
Good point! One other candidate would be ". It would be better, I think, because paths containing " already fail anyway. The drawback is that \" will look a bit uglier. What do you think?

greghudson:
" is no different from | in this regard.  Paths can contain either one, but rarely do since you have to quote them in the shell.

veprbl:
If you configure with ```--prefix=/tmp/\"``` make will fail, but if you do ```--prefix=/tmp/test/\|``` it won't.

greghudson:
I see.  Sure, we can use " as the sed delimiter.

veprbl:
Good. Thank you!

vcunat:
Well, OK, staged.

greghudson:
Pushed to master as cf872b9f25955ea2d51589a094df3a78fb582a6d.

PR #468, opened 2016-06-20
Fix argument order for kprop debug message

mrogers950:
Small correction of 0e7b4c9 - handle->params.realm needs to be the second argument.


greghudson:
Pushed to master as e76c5f0b08da7129746fc8549c9a01d0af8d9ffa

PR #469, opened 2016-06-21
Fix authdata framework CAMMAC extraction function

greghudson:
[Discovered by Coverity scan.]

Commit f4f619e7905d762204695dd66450e586c183c9fd used the wrong
sizeof() expressions for reallocating and copying into the list.  No
misbehavior manifested in tests because sizeof(list) == sizeof(*list)
on typical platforms.  Use the correct sizeof expressions.


mrogers950:
LGTM.

greghudson:
I will throw this fix into the commit for PR #474, so I'm closing this PR.

PR #470, opened 2016-06-21
Clarify directions when a Python test fails

greghudson:
In _onexit() in k5test.py, output an absolute path to the testlog
file, and output explicit directions on re-running the script with the
-v flag.


tlyu:
This looks ok to me.

PR #471, opened 2016-06-21
Fix incorrect recv() size calculation in libkrad

npmccallum:
Before this patch libkrad would always subtract the existing buffer
length from pktlen before passing it to recv(). However, in the case of
stream sockets, this is incorrect since krad_packet_bytes_needed()
already performs this calculation. This would lead to a case where the
existing buffer length was getting subtracted twice, causing integer
underflow on the len parameter to recv().

### DIFF
--- a/src/lib/krad/remote.c
+++ b/src/lib/krad/remote.c
@@ -315,7 +315,7 @@ on_io_read(krad_remote *rr)
     request *tmp, *r;
     int i;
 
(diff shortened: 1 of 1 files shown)


npmccallum:
In case it isn't obvious, I'd like this applied to all minor releases as well.

greghudson:
Pushed to master as c969e8a37617e9c7743a28177dd3808f7d08cee9

PR #472, opened 2016-06-21
Fix and simplify auth-indicator authdata module

greghudson:
Remove the authind_context count field.  The indicators list must be
null-terminated because it is freed with k5_free_data_ptr_list().
authind_internalize() didn't null-terminate the list, and the presence
of the count field made it appear that this wasn't a bug.  Use a
different scheme for setting *more in authind_get_attribute() to avoid
requiring an element count.

Also check more thoroughly for errors in authind_externalize() and
authind_internalize(), and remove some unnecessary pointer casts.

### DIFF
--- a/src/lib/krb5/krb/ai_authdata.c
+++ b/src/lib/krb5/krb/ai_authdata.c
@@ -37,7 +37,6 @@
 
 struct authind_context {
     krb5_data **indicators;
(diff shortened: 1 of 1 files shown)


mrogers950:
I've stepped through these changes and they look good to me.

PR #473, opened 2016-06-21
Fix recent memory leak in locate_kdc.c

greghudson:
Commit ce112dec844e4650b5ad174bd40f21c32aebe1d1 introduced a memory
leak in locate_srv_conf_1() by moving the free(realmstr) call to the
cleanup handler, because there was an early return after realmstr is
allocated.  Convert that early return to a goto.


greghudson:
I added a second memory leak fix to an unrelated test function.

tlyu:
This looks ok to me.

PR #474, opened 2016-06-22
krb5int_authdata_verify() CAMMAC fixes

mrogers950:
This adds the module->ad_type filtering of the resulting CAMMAC authdata that was missed in PR#410, and prevents the leak of cammac_authdata.


greghudson:
Pushed to master as 7df4aec92aebabaaf14de9de062f526228e65f48

PR #476, opened 2016-06-29
Miscellaneous leaks

greghudson:
Here are 16 fixes for miscellaneous memory leaks discovered using ASAN.  Most of these aren't likely to occur very commonly in practice, but could occur repeatedly in some cases.  I didn't mark the libev upgrade for pullup because I think it mostly manifests during loop shutdown, and because it isn't a terribly safe change.  We could of course make a conservative backport commit which just changes ev_realloc_emul().


tlyu:
I thought I had already commented on this, but it looks ok to me.

PR #477, opened 2016-06-29
Make the test suite run without detecting leaks in asan

greghudson:
Together with PR #476, this PR makes it possible to run make check without detecting any memory leaks with asan.


greghudson:
Updated with additional leak fixes in the second and third commits.  (The asan in Ubuntu 16.04 finds significantly more leaks than the one in 14.04.)

tlyu:
This looks ok to me. (It took a little work to rebase onto master due to deleted files.)

PR #478, opened 2016-06-30
Add asan build support

greghudson:
In an effort to add memory leak checking to our continuous integration, here is a PR which enables building with asan.  We need the following things to happen for an asan build:

1. Compile source files with -fsanitize=address.
2. Link programs with -fsanitize=address.
3. When building shared libraries, do not check for undefined symbols.

It would be easy to address (1) and (2) by adding the -fsanitize flag to CFLAGS and CXXFLAGS, but since asan and non-asan objects can't be mixed, it would probably go against the spirit of being able to override CFLAGS at make time.  Instead I defined a new make variable ASAN_FLAGS.  Getting ASAN_FLAGS into link options isn't easy at the moment; we have 60-odd definitions of CC_LINK_SHARED, CXX_LINK_SHARED, CC_LINK_STATIC, and CXX_LINK_STATIC in shlib.conf, most of them identical.  For the moment I am adding ASAN_FLAGS in pre.in where we substitute @CC_LINK@ and @CXX_LINK@, but I don't feel great about it.

I dealt with (3) by breaking out the Linux -Wl,--no-undefined into a separate variable which can be suppressed in configure.in after shlib.conf is included.  I couldn't find any evidence that any other asan-supported platforms will need attention, but I can't be sure.


SaharahSarah:
Is this supposed to pass a check? After building with --enable-asan I ran check and I'm getting detected memory leaks.

greghudson:
Only with PR #476 and PR #477 folded in.

SaharahSarah:
The latest seems to work well. I don't get any leaks etc. in either Fedora 24 or FreeBSD

SaharahSarah:
OK, FreeBSD, Fedora 24, and Gentoo all are passing the checks without reporting any problems!

tlyu:
I tried to get this to work on Mac OS 10.10, but even when making the appropriate UNDEF_CHECK change (to remove `-undefined error`), `gcc -dynamiclib` complains about undefined symbols. Otherwise, looks ok to me.

PR #479, opened 2016-07-01
Improve kdb5 error when default realm is unset

greghudson:
When the default realm name is unspecified, and none was set in the
krb5_context object, return KRB5_CONFIG_NODEFREALM from libkdb5
instead of the confusing KRB5_KDB_DBTYPE_NOTFOUND.  To accomplish
this, make kdb_get_library_name() return a krb5_error_code.


tlyu:
This looks ok to me.

PR #480, opened 2016-07-05
Do not return a decrypt error return code but KRB5KDC_ERR_PREAUTH_FAILED

gladiac:
This makes it consistent with other Kerberos implementations.

### DIFF
--- a/src/kdc/kdc_preauth_encts.c
+++ b/src/kdc/kdc_preauth_encts.c
@@ -59,7 +59,6 @@ enc_ts_verify(krb5_context context, krb5_data *req_pkt, krb5_kdc_req *request,
     krb5_key_data *             client_key;
     krb5_int32                  start;
     krb5_timestamp              timenow;
-    krb5_error_code             decrypt_err = 0;
 
     scratch.data = (char *)pa->contents;
     scratch.length = pa->length;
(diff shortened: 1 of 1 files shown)


gladiac:
Fixed, I'm checking for both return codes now. It is possible that a new kinit is used against an older KDC.

greghudson:
Hm, I'm not sure that's correct either.  If you're using PKINIT and it doesn't work, for example, you might not even have entered a password.  I'll make a note to discuss this next Tuesday.

greghudson:
From today's discussion:

* We could treat the kinit behavior separately from this change (although the test suite needs to know what message to expect from kinit), since KDCs and clients are independent and non-MIT KDCs already have the new behavior.  But I would like to have a plan for dealing with the problem in kinit before changing the KDC behavior.

* kinit could use its own prompter function and make note of whether it was prompted for a password, and translate KRB5_PREAUTH_FAILED to "Password incorrect" only if there was a password prompt.  I will look into this when time permits.

greghudson:
I opened PR #491 to deal with the kinit reporting issue; please go ahead and narrow this commit back down to the KDC behavior change.

gladiac:
Done ...

greghudson:
Pushed to master as 2653d69e0705a925597dff10083a24a77e2a20af.

PR #481, opened 2016-07-05
Enable KDC discovery through URI records

mrogers950:
This adds the service locator for URI records for https://tools.ietf.org/html/draft-mccallum-kitten-krb-service-discovery-02. The leading flags field is not mentioned in the draft but is carried over from the initial TXT record plans. Another item addressed is letting k5_kdc_is_master() check the master flag from the URI lookup before performing an additional search.


greghudson:
Just from a very brief glance:

* entry->master should be a tristate field (of type int), with value -1 for unknown, FALSE or TRUE for known.  Otherwise we will wind up looking up the URI record an extra time whenever we get a response for a KDC not tagged as a master.

* The structural changes to support known master KDC status should probably be in a separate commit for clarity.  Of course all of the existing lookup mechanisms will just pass -1.

* The URI DNS query code should be abstracted and split out into a separate file, similar to how the SRV code is in dnssrv.c.  Parsing of the URI string should remain in locate_kdc.c.

* Reusing struct srv_dns_entry for URI results is okay (with "host" holding the URI instead of a hostname), but there should be a comment in the structure definition indicating that.

* The comment beginning with "There are 3 bytes..." suggests a mystery that we need to solve.  Was this something you discovered experimentally a particular DNS server implementation?

greghudson:
Some more notes after a slightly closer look:

* There is too much copy-paste from krb5int_make_srv_query_realm().  Some refactoring is needed.  There are differences between the SRV and URI records which prevent using one function for both, but there should not be so much repeated code.

* I think we have a major miscommunication about how to translate the TXT record design into a URI record design.  My understanding, which I believe matches Nathaniel's and Nico's, was that in http://k5wiki.kerberos.org/wiki/Projects/KDC_Discovery 
everything after the priority and weight is notionally part of a new URI scheme which we have to make up a name for and register, so the URIs we see should look like "krb5srv:M:udp:my.host.name:123" or "krb5srv::kkdcp:https://my.host.name/KdcProxy".  locate_uri() appears to be going back to trying to shoehorn krb5 server transport descriptions into existing or made-up URI schemes, which I thought was a design we had abandoned.

* I also see a problem in the project page, in that it assumes that the residual for every transport type is host[:port][/path].  The residual format should depend on the transport type; for tcp and udp it should be host[:port], while for mskkdcp it should be the syntax of an https URL.

* MS-KKDCP only operates over https.  locate_uri() appears to be trying to introduce support for a non-TLS variant.  Don't do that as part of this PR.

* locate_uri() has a bunch of logic to detect IP addresses, making the erroneous assumption that [] can only surround an IPv6 address, and use add_addr_to_list() instead of add_host_to_list().  We don't do this in locate_srv_conf_1(); don't do it here.

I see a problem with the KDC discovery project page

mrogers950:
Thanks, I'm working through your suggestions. A couple comments:

> The comment beginning with "There are 3 bytes..." suggests a mystery that we need to solve. Was this something you discovered experimentally a particular DNS server implementation?

Yes, those results were from BIND 9.10.3-13. I've looked through the BIND URI RR code, but it's not clear to me if there are fields being added to the target string on purpose.

> I think we have a major miscommunication about how to translate the TXT record design into a URI record design. My understanding, which I believe matches Nathaniel's and Nico's, was that in http://k5wiki.kerberos.org/wiki/Projects/KDC_Discovery 
> everything after the priority and weight is notionally part of a new URI scheme which we have to make up a name for and register, so the URIs we see should look like "krb5srv:M:udp:my.host.name:123" or "krb5srv::kkdcp:https://my.host.name/KdcProxy". locate_uri() appears to be going back to trying to shoehorn krb5 server transport descriptions into existing or made-up URI schemes, which I thought was a design we had abandoned.

Ah, I assumed that moving forward with the URI we would be using the discovery draft text as it is, so I will make those adjustments to the wiki and we can finalize the scheme. Although, the leading service name in the examples you give here were not part of the TXT scheme, and would not be useful unless we intend to have one lookup to find all possible service URIs.

greghudson:
"The leading service name" is a URI scheme.  It should be the same for every record we process; there is no need to use separate schemes for kpasswd vs. the KDC protocol, for instance.  To use the new URI scheme, we have to register and use a new URI scheme, because there is no proper way to describe Kerberos transports with existing URI schemes--particularly when the master flag is included.  There was no need for this when the design used TXT records, because we were not constrained to using a URI with TXT records.

On an implementation note, feel free to put the URI request code in dnssrv.c, so that you can share static helpers with krb5int_make_srv_query_realm().  It's a slight abuse of the name, but if you're reusing struct srv_dns_entry it makes sense.

tlyu:
I thought we wanted to get all possible service URIs with a single lookup to minimize DNS roundtrips.

greghudson:
After reviewing http://mailman.mit.edu/pipermail/krbdev/2016-May/012588.html, I do not see anyone arguing that we should mix krb5, kpasswd, and kadmin transport descriptions together in the same RR set.

(We did,  however, want to include master and normal krb5 KDCs in the same RR set, because the same krb5 protocol operation generally needs to know both along the normal code path.)

greghudson:
> Yes, those results were from BIND 9.10.3-13. I've looked through the BIND URI RR code, but it's not clear to me if there are fields being added to the target string on purpose.

After some experimentation, I don't believe BIND inserts any extra bytes after the end of the URI.  A terminating 0 byte is not expected because the RR value is length-counted.  rdlen - 4 is the correct number of bytes to copy because rdlen is not decremented by SAFE_GETUINT16(), so you have to subtract out the four bytes for the priority and weight.

mrogers950:
Updated with the following changes.
- The master variable change, URI lookup code, and locate KDC code changes are in seperate commits.
- The master flag is now an int with -1 as the unset state.
- T_URI was defined the same under both HAVE_RES_NSEARCH and !HAVE_RES_NSEARCH. Also, under HAVE_RES_NSEARCH it doesn't seem that there is even a ns_t_uri type available, so it is now a single define.
- There was also a master variable added to srv_dns_entry, which is not necessary.
- The URI lookup function is in dnssrv.c and shares helper functions of the common code with krb5int_make_srv_query_realm().
- I rewrote the URI parsing for the correct format. It is split into parts first by parse_uri_fields() to confirm the "krb5srv" scheme, master status, and fetch the transport and target fields. The target of URL/address, port and path is then handled by parse_uri_if_https() and k5_parse_host_string(), and finally added to the server list with add_host_to_list().
- Revised comments and commits.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,6 +45,122 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
     }
 }
 
+static krb5_error_code
```
greghudson: This needs a brief descriptive comment.  The function body also needs a blank line after the declaration of buf.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,6 +45,122 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+    if (memchr(realm->data, 0, realm->length))
+        return EINVAL;
+
+    k5_buf_init_fixed(&buf, fixed_buf, bufsize);
+    k5_buf_add_fmt(&buf, "%s.", service);
+    if (protocol != NULL)
+        k5_buf_add_fmt(&buf, "%s.", protocol);
+    k5_buf_add_len(&buf, realm->data, realm->length);
+
+    /* Realm names don't (normally) end with ".", but if the query
```
greghudson: This is of course moved code, but please reformat the multi-line comment according to current practice.  Also, please drop the second paragraph of the comment.  Nothing in the code checks for "only if a name is specified," and we never make SRV or URI queries on a prefix alone.  I tracked down the history of this comment and we didn't check for "only if a name is specified" back then either, so I don't know why that paragraph was added.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,6 +45,122 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+       a search on the prefix alone then the intention is to allow
+       the local domain or domain search lists to be expanded.  */
+
+    if (buf.len > 0 && fixed_buf[buf.len - 1] != '.')
+        k5_buf_add(&buf, ".");
+
+    return k5_buf_status(&buf);
+}
+
+/* Insert into the list, ordering by priority. Weight is not currently
```
greghudson: Two spaces between sentences, please.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,6 +45,122 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+    /* Realm names don't (normally) end with ".", but if the query
+       doesn't end with "." and doesn't get an answer as is, the
+       resolv code will try appending the local domain.  Since the
+       realm names are absolutes, let's stop that.
+
+       But only if a name has been specified.  If we are performing
+       a search on the prefix alone then the intention is to allow
+       the local domain or domain search lists to be expanded.  */
+
+    if (buf.len > 0 && fixed_buf[buf.len - 1] != '.')
```
greghudson: Use buf.data here for clarity.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,6 +45,122 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+    if (*head == NULL || (*head)->priority > new->priority) {
+        new->next = *head;
+        *head = new;
+        return;
+    }
+    /* Insert an entry into this spot if the next entry has a higher
+     * priority (lower priorities are preferred), or if there is no
+     * next entry (we're at the end).  */
+    for (entry = *head; entry != NULL; entry = entry->next) {
+        if ((entry->next &&
```
greghudson: This should be "entry->next != NULL".  Also, please format the comment above according to current conventions.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,6 +45,122 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+        if (uri->host == NULL) {
+            ret = errno;
+            goto out;
+        }
+
+        place_srv_entry(&head, uri);
+    }
+
+out:
+    if (ds != NULL) {
```
greghudson: krb5int_dns_fini() handles null pointers, so don't conditionalize the call.  Also don't assign ds = NULL right before the end of the function.  (k5_try_realm_txt_rr() has the same issue.  But when adding new code which mirrors existing code, don't mirror the bad or outdated practices in the old code.)

greghudson:
The implementation of parse_uri_fields() has several serious problems:

* It takes a const char * input, and casts it to char * in order to pass that pointer to strtok_r(), which has the effect of modifying the string.

* It looks at the save pointer output of strtok_r().  There are no contractual guarantees as to what goes into that pointer.  (I'm not sure if this is even necessary.

* It uses a state machine and loop structure, but the only shared part of the loop is a call to strtok_r().  It could be simplified by just having three steps and no loop.

It would probably be better not to use strtok_r() in this function; using strchr() instead would avoid the need to copy or modify the input string.

You are also using 0 as a distinguished value for k5_transport.  0 is the value of TCP_OR_UDP; while URI records don't traffic in that transport type, it's not very hygienic.  It's not okay to use -1 (a compiler may choose to use an unsigned type for an enum).  I would say it's okay to initialize *transport_out to 0, but don't test for that result in the caller.

Also, the addition of the MSKKDCP transport type seems wrong, and makes me question whether this code works for the intended purpose.  All of the existing MS-KKDCP code uses the HTTPS transport type.

mrogers950:
I've made the above changes, and revised parse_uri_fields() to just progress through the string without strtok().  I also changed some variable names from "target" to "host" to be more explicit about the part of the URI being held and so it is not confused with the use of target in the DNS RR context.                                                          

> This code also uses 0 as a distinguished value for k5_transport. 0 is the value of TCP_OR_UDP; while URI records don't traffic in that transport type, it's not very hygienic. It's not okay to use -1 (a compiler may choose to use an unsigned type for an enum). I would say it's okay to initialize *transport_out to 0, but don't test for that result in the caller.

> Also, the addition of the MSKKDCP transport type seems wrong, and makes me question whether this code works for the intended purpose. All of the existing MS-KKDCP code uses the HTTPS transport type.

You're right, it would have needed a check later on to process the MSKKDCP type as HTTPS, so
instead I'm setting the transport to HTTPS when an MSKKDCP entry is parsed and that allowed me to get rid of the k5_transport addition.

tlyu:
The Travis failure (in clang) seems to be a case of casting away const; it should probably be fixed.

mrogers950:
I missed the krb5int_dns_fini() related changes in the last update, so I will fix that.

greghudson:
There is no reason to cast away const in parse_uri_fields().  Make host_out a const char ** and adjust the caller's variable type.  The description of parse_uri_fields() should note that host is an alias into uri.  Also, "(address/URL plus optional port and path)" is a little muddled; an HTTPS URL contains an optional port and a path, so host is either a URL or it's an address plus optional port.

In locate_uri(), if transport is HTTPS, the result of parse_uri_if_https() should be checked to make sure that it was actually a URL.  We don't want to accept records like krb5srv::kkdcp:just.a.hostname if it's not allowed in the format.  Unfortunately parse_uri_if_https() is molded for a locate_srv_conf_1() shaped hole, making this check slightly unnatural.  You can either redesign it slightly, or you can check the result by passing in an initialized transport variable and verifying that it changes to HTTPS.

mrogers950:
Sorry for the delay on this small update - I've made those corrections, and added const to parse_uri_if_https() parameters to avoid casting away const when its used in locate_uri().

greghudson:
This needs to be rebased to master after PR #506.  Make sure the string "htons" does not appear in locate_kdc.c after the rebase.

Please fix the problems reported by the style checker.  Omit the #ifdef TEST block from k5_make_uri_query().  (We should add a trace logging point here and in krb5int_make_srv_query_realm(), but that's out of scope for this PR.)  Also, please re-wrap the commit messages to a fill width of 70; right now they do not display well in an 80-column terminal with git's four-space indent.

parse_uri_if_https() should also have const added to the uri_path parameter.  You're only avoiding a warning here because strchr() allows you to get rid of const without a cast.  It looks like add_host_to_list() will also need its uri_path parameter constified to accomodate the necessary changes in the caller variables.  These constification changes should probably be a separate commit.

mrogers950:
Thanks again, I've rebased and updated with those changes.

greghudson:
Apologies for not raising this concern sooner, but we need to think about documentation and testing.

SRV records are documented in realm_config.rst, and referred to in install_clients.rst, install_kdc.rst, and krb5_conf.rst.  Correspondingly, URI records should be documented in realm_config.rst, and the references in the other files should be updated to refer to both.

Testing is a little trickier.  The existing test for SRV records is conditional on specific records existing in the public DNS (as determined by dig or nslookup).  It also isn't a great test since it doesn't seem to verify the results of the lookup.  We will have to create a URI record, probably in kerberos.org, to use as a fixture.  Unfortunately one of the kerberos.org DNS servers is currently too old to support URI records.  So we might have to defer automated testing for the moment; I will discuss that with the team.

I think the style fixes commit is new.  At a brief glance it looked fine, but please remove the malloc() result cast where you changed a line involving a call to malloc().

"Constify function parameters" should orient the reader given that there is space to do so: "Constify function parameters in locate_kdc.c".

npmccallum:
I would very strongly like to see tests. I think a better solution might be to use resolv_wrapper: http://jhrozek.livejournal.com/4301.html

greghudson:
Coincidentally, Sarah is currently working on optionally using cwrap for the test suite.  She hadn't planned on adding support for tests that can only run when cwrap is available, but we might wind up doing that instead of creating a public fixture in the DNS.

npmccallum:
FYI, https://lists.samba.org/archive/samba-technical/2016-August/115632.html

mrogers950:
Pushing to get feedback on the doc text and configuration option, while I work on getting resolv_wrapper to return all of its URIs for a query, and finish the testing.

frozencemetery:
In "Enable KDC discovery through DNS URI records", I would prefer that NUL checks not look like `if (*++uri == '\0')`, especially in the case of `if (*++uri == '\0' || *uri != ':')`.

doc/admin/conf_files/krb5_conf.rst
```diff
@@ -212,6 +212,12 @@ The libdefaults section may contain any of the following relations:
     data), and anything the fake KDC sends will not be trusted without
     verification using some secret that it won't know.
 
+**dns_uri_lookup**
```
greghudson: This text should indicate the default value, and should say "New in release 1.15." at the end.

doc/admin/realm_config.rst
```diff
@@ -193,6 +193,54 @@ using the **kdc**, **master_kdc**, **admin_server**, and
...
+  contacted using the specified transport, with an optional port
+  extension.  The MS-KKDCP transport type uses a HTTPS URL, and can
+  include a port and/or path extension.
+
+An example of URI records in a zone file::
+
+  _kerberos.EXAMPLE.COM   URI   10 1 krb5srv:m:tcp:kdc1.example.com
+                          URI   20 1 krb5srv:m:udp:kdc2.example.com:89
+                          URI   40 1 krb5srv::udp:10.10.0.23
+                          URI   30 1 krb5srv::kkdcp:https://proxy.example.com:8080/auth
```
greghudson: This line might be too long when formatted for PDF; perhaps use a shorter example URL.

greghudson:
It looks like the documentation commit still needs to update install_clients.rst.

mrogers950:
- Changed parse_uri_fields() pointer incrementing per Robbie's feedback.
- Changed krb5_conf.rst and realm_config.rst text and added DNS record references to install_clients.rst 
- Added the working resolv_wrapper commit. At the moment this requires an unreleased resolv_wrapper version. Until then you can use resolv_wrapper 1.1.4 from git with the following patches applied:
https://lists.samba.org/archive/samba-technical/2016-August/115839.html
https://lists.samba.org/archive/samba-technical/2016-August/115840.html

greghudson:
If libresolv_wrapper.so is installed but isn't new enough to have the necessary features and bugfixes, does make check fail?  We don't want that to happen.

(Given the overlap between this work and Sarah's work, I'm not sure it's worth the effort to get all the build goo right in this PR.  The correct path forward will become clearer when Sarah's cwrap work is ready for review, which I think should be soon.)

mrogers950:
Yes, I'll have to see about getting a version check for resolv_wrapper. There doesn't seem to be a function available for the autoconf check other than __res_search().

frozencemetery:
Much better, thanks.

gladiac:
FYI: I've released resolv_wrapper-1.1.5 with support for faking PTR and URI resource record types.

@mrogers950 You can use 'pkg-config --atleast-version=1.1.5 resolv_wrapper'

mrogers950:
@gladiac thanks!

greghudson:
I'd like to wrap this up soon so we can branch for 1.15.

@mrogers950, can you modify the testing commit to use pkg-config --atleast-version?  We have a couple of other configure.in fragments using pkg-config which you can use as a guide.

Also, has there been any progress on provisionally registering the URL scheme?

mrogers950:
Updated.

greghudson:
Why start with AC_CHECK_LIB?  Why not just run pkg-config?

mrogers950:
Good point, updated again.

src/lib/krb5/os/locate_kdc.c
```diff
@@ -506,10 +535,188 @@ prof_locate_server(krb5_context context, const krb5_data *realm,
...
+{
+    k5_transport transport;
+    int master = FALSE;
+
+    *transport_out = 0;
+    *host_out = NULL;
+    *master_out = -1;
+
+    /* Confirm the scheme name. */
+    if (uri == NULL || (strncasecmp(uri, "krb5srv", 7) != 0))
```
greghudson: uri can't be NULL here (looking at the caller), so don't check.  (API functions can be paranoid since they expect an unbounded number of callers; helper functions should be strict and concise so that contracts are clear.)

src/lib/krb5/os/locate_kdc.c
```diff
@@ -506,10 +535,188 @@ prof_locate_server(krb5_context context, const krb5_data *realm,
...
 
 #ifdef KRB5_DNS_LOOKUP
+
+/*
+ * Parse the initial part of the URI, first confirming the scheme name.  Get
+ * the transport, master status, and host.  The host is either an address or
+ * hostname with an optional port, or an HTTPS URL.
+ * The format is krb5srv:[M]:udp|tcp|kkdcp:host
+ *
+ * Return a NULL *host_out if there's any problems parsing the URI.
```
greghudson: "if there are"

src/lib/krb5/os/locate_kdc.c
```diff
@@ -506,10 +535,188 @@ prof_locate_server(krb5_context context, const krb5_data *realm,
...
+        transport = HTTPS;
+        uri += 5;
+    } else {
+        return;
+    }
+
+    if (*uri == '\0' || *uri != ':')
+        return;
+
+    /* The rest of the URI is the host and any optional parts. */
```
greghudson: "the host (with optional port) or URI".

src/lib/krb5/os/locate_kdc.c
```diff
@@ -506,10 +535,188 @@ prof_locate_server(krb5_context context, const krb5_data *realm,
...
+
+        def_port = default_port;
+        path = NULL;
+
+        parse_uri_fields(entry->host, &transport, &host_field, &master);
+        if (host_field == NULL)
+            continue;
+
+        /* TCP_OR_UDP allows entries of any transport type, otherwise
+         * we're asking for a match. */
```
greghudson: This is a run-on sentence; use a semicolon instead of a comma.

(This comment also exposes some technical debt with regard to the internal use of TCP_OR_UDP in locate_srv.c.  In turn, I think we have a minor behavior problem where we are unnecessarily excluding MS-KKDCP when falling back to TCP after a truncated UDP request or reply.  Those problems predate this PR and do not need to be addressed.)

src/lib/krb5/os/locate_kdc.c
```diff
@@ -506,10 +535,188 @@ prof_locate_server(krb5_context context, const krb5_data *realm,
...
+
+    ret = k5_make_uri_query(realm, req_service, &answers);
+    if (ret)
+        return ret;
+
+    if (answers == NULL)
+        return 0;
+
+    for (entry = answers; entry != NULL; entry = entry->next) {
+        if (entry->host == NULL)
```
greghudson: This can't happen, so don't check.  (If it did happen, there would be a bug in k5_make_uri_query(), and we wouldn't want to paper it over.)

src/lib/krb5/os/t_locate_kdc.c
```diff
@@ -58,8 +58,10 @@ print_addrs (void)
         char hostbuf[NI_MAXHOST], srvbuf[NI_MAXSERV];
 
         if (entry->hostname != NULL) {
-            printf("%2d: host %s\t%s\tport %d\n", (int)i, entry->hostname,
-                   ttypename(entry->transport), entry->port);
+            printf("%d: h:%s t:%s p:%d m:%d P:%s\n", (int)i,
+                   entry->hostname, ttypename(entry->transport),
+                   entry->port, entry->master,
+                   entry->uri_path ? entry->uri_path:"");
```
greghudson: There should be spaces around ":" here.

src/lib/krb5/os/locate_kdc.c
```diff
@@ -582,7 +789,10 @@ locate_server(krb5_context context, const krb5_data *realm,
 
 #ifdef KRB5_DNS_LOOKUP
     if (list.nservers == 0)
-        ret = dns_locate_server(context, realm, &list, svc, transport);
+        ret = dns_locate_server_uri(context, realm, &list, svc, transport);
```
greghudson: If ret is non-zero here, we should goto done.  Otherwise you will discard the fatal error (unless some servers were added before it happened).

src/lib/krb5/os/os-proto.h
```diff
@@ -58,6 +58,7 @@ struct server_entry {
     k5_transport transport;     /* May be 0 for UDP/TCP if hostname set */
     char *uri_path;             /* Used only if transport is HTTPS */
     int family;                 /* May be 0 (aka AF_UNSPEC) if hostname set */
+    int master;                 /* Set to true/false by URI RR. -1 when unset. */
```
greghudson: This line runs over 79 characters.  Also, there may be other information sources in the future which can set the flag, so the comment shouldn't talk about URIs.  I suggest "True, false, or -1 for unknown."

Similarly, the commit message should mention URI lookups only when talking about the motivation for the change.  After that, it should say "... not been determined by the information source" and "If the flag is set, use it...".

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,11 +45,128 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
     }
 }
 
-/* Do DNS SRV query, return results in *answers.
+/*
+ * Prepares a DNS label of service.protocol.realm using a caller-supplied fixed
+ * buffer.  Since the protocol is not relevant to all callers it can be set to
+ * NULL to be excluded from the label.
+ */
```
greghudson: This should use the imperative mood, and can be more concise:

    Construct a DNS label of the form "service.[protocol.]realm.", placing the
    result into fixed_buf.  protocol may be NULL.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,11 +45,128 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+     */
+
+    if (buf.len > 0 && ((char *)buf.data)[buf.len - 1] != '.')
+        k5_buf_add(&buf, ".");
+
+    return k5_buf_status(&buf);
+}
+
+/* Insert into the list, ordering by priority.  Weight is not currently
+ * used. */
```
greghudson: This should be more specific about the parameters: "Insert new into the list *head, ordering..."

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,11 +45,128 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+    if (*head == NULL || (*head)->priority > new->priority) {
+        new->next = *head;
+        *head = new;
+        return;
+    }
+    /*
+     * Insert an entry into this spot if the next entry has a higher priority
+     * (lower priorities are preferred), or if there is no next entry (we're at
+     * the end).
+     */
```
greghudson: This comment should be inside the for loop, and should say "into the next spot."  The two cases should also be reversed because the conditional should be reversed (see next comment).

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,11 +45,128 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+    for (entry = *head; entry != NULL; entry = entry->next) {
+        if ((entry->next != NULL &&
+             entry->next->priority > new->priority) ||
+            entry->next == NULL) {
```
greghudson: This reads better as "if (entry->next == NULL || entry->next->priority > new->priority)", which fits on one line.

src/lib/krb5/os/dnssrv.c
```diff
@@ -45,11 +45,128 @@ krb5int_free_srv_dns_data (struct srv_dns_entry *p)
...
+{
+    const unsigned char *p = NULL, *base = NULL;
+    char host[MAXDNAME];
+    int size, ret, rdlen;
+    unsigned short priority, weight;
+    struct krb5int_dns_state *ds = NULL;
+    struct srv_dns_entry *head = NULL, *uri = NULL;
+
+    *answers = NULL;
+     /* Construct service.realm. */
```
greghudson: This is indented by five spaces instead of four.

doc/admin/conf_files/krb5_conf.rst
```diff
@@ -212,6 +212,13 @@ The libdefaults section may contain any of the following relations:
...
+    Indicate whether DNS URI records should be used to locate the KDCs
+    and other servers for a realm, if they are not listed in the
+    krb5.conf information for the realm.  SRV records are used as a
+    fallback if no URI records were found. The default value is true.
```
greghudson: There should be two spaces after the period here.

doc/admin/realm_config.rst
```diff
@@ -193,6 +193,54 @@ using the **kdc**, **master_kdc**, **admin_server**, and
...
+The client performs a query for the following URI records:
+
+* ``_kerberos.REALM`` for fiding KDCs.
+* ``_kerberos-adm.REALM`` for finding kadmin services.
+* ``_kpasswd.REALM`` for finding password services.
+
+The URI record includes a priority, weight, and a URI string that
+consists of case-insensitive colon separated fields:
+
+``scheme:[flags]:transport:residual``
```
greghudson: This is currently formatted a paragraph, which doesn't make sense.  It could be formatted as an example (using ""::" above and indenting it by four spaces), but I would instead bring it up into the previous paragraph, saying "... separated fields, in the form ``...``".

doc/admin/realm_config.rst
```diff
@@ -193,6 +193,54 @@ using the **kdc**, **master_kdc**, **admin_server**, and
...
+
+* ``scheme`` defines the registered URI type. It should always be
```
greghudson: In these descriptions, use \*scheme\* (no backslashes) for the variable parts of the expression.  Continue to use \`\` for specific literal values.

npmccallum:
Do we want to use case to express criticality?

greghudson:
No, I don't think we need to worry about supporting critical flags here.  I certainly don't think it warrants breaking case-insensitivity.

mrogers950:
The flags are no longer critical and I've adjusted some of the test entries to reflect that.  The nits are corrected too.

greghudson:
Pushed to master as 26ae4495eededc9854f533b1cb2c0f23160290d8 and prior.

PR #482, opened 2016-07-07
Update LDAP docs for password lockout

greghudson:
The KDC now needs write access to the LDAP KDB, unless password
lockout and tracking of the last successful authentication time are
disabled.  Update the example LDAP access control text in
conf_ldap.rst to reflect this.  Reported by Will Fiveash.


greghudson:
I pushed an updated commit which notes in the explanatory text that only read access is required if account lockout is disabled.  At Will's suggestion, I also added a section to lockout.rst cross-referencing the write access requirement.

SaharahSarah:
This looks good to me.

PR #483, opened 2016-07-15
Add a kprop-port option to kadmind

mrogers950:
This adds '-k' to kadmind for setting the port that kprop is spawned with during full resyncs, and preserves the use of KPROP_PORT if the option is not used.

### DIFF
--- a/doc/admin/admin_commands/kadmind.rst
+++ b/doc/admin/admin_commands/kadmind.rst
@@ -16,6 +16,7 @@ SYNOPSIS
 [**-P** *pid_file*]
 [**-p** *kdb5_util_path*]
 [**-K** *kprop_path*]
+[**-k** *kprop_port*]
 [**-F** *dump_file*]
 
 DESCRIPTION
@@ -100,6 +101,11 @@ OPTIONS
 **-K** *kprop_path*
     specifies the path to the kprop command to use to send full dumps
     to slaves in response to full resync requests.
+
+**-k** *kprop_port*
+    specifies the port by which the kprop process that is spawned by kadmind
+    connects to the slave kpropd, in order to transfer the dump file during
+    a full resync request.
 
 **-F** *dump_file*
     specifies the file path to be used for dumping the KDB in response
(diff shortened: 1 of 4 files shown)


greghudson:
Can we avoid sending this through realm options?  Space in the realm option mask will be extremely tight after PR #380, and I am thinking it might be better if kadmind-specific options weren't part of that structure.

mrogers950:
I see, I suppose alternately it could be a part of the global_server_handle.

greghudson:
That would also involve changing libkadm5 structures.  Why not just make it a global variable?  There are already globals for the kprop path, the kdb5_util path, and the dump file.

greghudson:
The documentation for kadmind -k really doesn't convey what the option does.  In case background is needed, I'll summarize where this option fits into the iprop architecture (this isn't suggested documentation text):

When incremental propagation is enabled, the kadmind daemon on the master KDC serves a second RPC protocol on a different port (of the administrator's choosing, determined by the iprop_port option; there is no default port).  On the slave KDC, kpropd periodically makes requests for incremental updates to kadmind via this second RPC protocol.  In response to these requests, the KDC can say "you're up to date", "here are some incremental updates", or "you are too far behind and will need a full resync."  In the last case, kpropd will make another request to kadmind for a full resync.  kadmind will service this requests by dumping the database and spawning a kprop process to transfer the dump file to the slave using the traditional (non-incremental) kprop protocol.  This new -k option determines the port number for this dump file transfer.

mrogers950:
I've used a global variable instead, and revised the doc text.

greghudson:
You can avoid the allocation if you just keep the string representation of the port in the global variable, instead of converting it to an integer.  I would also suggest doing the getenv("KPROP_PORT") in main() in order to simplify ipropx_resync().

mrogers950:
I've made those changes and updated.

greghudson:
Pushed to master as 7ed18b1e5a11a514461be2281ff884e8173299b1 77c3f655f684761efe957722e500887f3199fc5f.

PR #484, opened 2016-07-15
Fix error handling in k5_expand_path()

greghudson:
In k5_expand_path_tokens_extra(), don't throw away non-zero return
statuses in the cleanup handler.


mrogers950:
LGTM.

PR #485, opened 2016-07-19
Fix S4U2Self KDC crash when anon is restricted

greghudson:
The first commit fixes a bug where S4U2Proxy can fail on the client side when the KDC is configured to support PKINIT.

The second commit fixes a KDC crash when restrict_anonymous_to_tgt is set and an S4U2Self request is received.  The test case added in the second commit relies on the first commit; if the first commit is not present, it will fail on the client side (causing the test script to fail) before exercising the KDC fix.


tlyu:
This looks ok to me.

PR #486, opened 2016-07-19
Fix recent change to kdcdefaults processing

greghudson:
When I integrated PR #380, I made a change to the last commit to ensure sure that kdc_ports and kdc_tcp_ports would still work inside [kdcdefaults].  Through a fascinating COPY_PASTE_ERROR heuristic, Coverity detected a bug in that change.  This commit fixes the bug.


SaharahSarah:
Simple change; I don't see any problems.

PR #487, opened 2016-07-20
Modify code to handle the case where cred->client_keytab is NULL

wfiveash:


greghudson:
It looks like we have to make a Windows build system change in order to be able to use trace macros from the GSS library.  The appveyor build failed with:

    Creating library ..\lib\obj\i386\dbg\gssapi32.lib and object ..\lib\obj\i386\dbg\gssapi32.exp
    LINK : warning LNK4199: /DELAYLOAD:secur32.dll ignored; no imports found from secur32.dll
    gssapi.lib(acquire_cred.obj) : error LNK2019: unresolved external symbol _krb5int_trace referenced in function _acquire_init_cred

src/lib/gssapi/krb5/acquire_cred.c
```diff
@@ -522,7 +525,11 @@ get_name_from_client_keytab(krb5_context context, krb5_gss_cred_id_rec *cred)
     krb5_principal princ;
 
     assert(cred->name == NULL);
-    code = k5_kt_get_principal(context, cred->client_keytab, &princ);
+    if (cred->client_keytab) {
```
greghudson: I would just add "if (cred->client_keytab == NULL) return KRB5_KT_NOTFOUND;" here and not change the existing code.

wfiveash:
I've updated the pull request as you suggested.

kaduk:
Is the windows change not just adding it to krb5_32.def?

greghudson:
Experimentally, yes, that's enough to fix the Windows build.  I'll do that in a separate commit.

I will fix it, but direct calls to TRACE from source files aren't allowed; you have to go through a macro in k5-trace.h.

There are also some implicit pointer nullity checks in this patch ("if (ptr)" rather than "if (ptr != NULL)"), which I will change.

wfiveash:
Why are direct TRACE() calls not allowed?  Having to add a new trace macro to k5-trace.h for one, unique trace point seems like an unnecessary hassle.

greghudson:
When we introduced trace logging, Nico and others argued that it was important to make it easy for downstream distributions to integrate trace logging with DTrace and similar OS-native tracing facilities.  Therefore, each trace point uses a unique macro with programmatic arguments, and all of the macros funnel through a central header file which can be edited as necessary to integrate with the system tracing facility.

I don't think anyone has taken advantage of this design to date, but it is still the design; abandoning it when new trace points are added is inconvenient.

wfiveash:
OK, I can buy that argument.  Certainly integrating the MIT trace facility with DTrace is of interest to us.

wfiveash:
BTW, adding more info in the comments in k5-trace.h explaining that TRACE() should not be called directly and why would be good.  I actually looked in k5-trace.h before I added the TRACE() call to acquire_init_cred() but I did not see anything there that was compelling enough for me not to take the easy route which was inappropriate.  Also, maybe adding a check to something like the cstyle checker would be good?

greghudson:
Pushed to master as de92d6c5af514b88dc4fd36768b4ff05273ed184 bd2c2a02e22c609b3c7e9f92d6634e151d14e478.

PR #488, opened 2016-07-20
Fix SPNEGO imported cred initialization

greghudson:
[I ran into this while running a test program under valgrind when looking into something else.]

In spnego_gss_import_cred(), use create_spnego_cred() to create the
SPNEGO credential structure.  Prior to this change, an imported SPNEGO
cred did not initialize the no_ask_integ field (added by commit
cf39ed349976908626cad3e05e17788f8334bce9, ticket #6938).


greghudson:
I will fix it, but direct calls to TRACE from source files aren't allowed; you have to go through a macro in k5-trace.h.

There are also some implicit pointer nullity checks in this patch ("if (ptr)" rather than "if (ptr != NULL)"), which I will change.

tlyu:
This looks ok to me.

PR #489, opened 2016-07-22
Fix trying compatible keys of different realms

ricardofandrade:
This fix prevent trying keys of different realms.
Without the fix, if they key was created with the same attributes (using the same password, etc.) it will be able to decrypt the ticket even when it belongs to a different realm.

The portion of RFC 1510 to base this fix is [3.2.3. Receipt of KRB_AP_REQ message](https://tools.ietf.org/html/rfc1510#page-21)
More specifically:
Since it is possible for the server to be registered in multiple realms, with different keys in each, the **srealm** field in the unencrypted portion of the ticket in the KRB_AP_REQ is **used to specify** which secret key the server should use to decrypt that ticket.

My understanding it that keys from different realms should not be attempted.

### DIFF
--- a/src/lib/krb5/krb/rd_req_dec.c
+++ b/src/lib/krb5/krb/rd_req_dec.c
@@ -363,7 +363,7 @@ decrypt_ticket(krb5_context context, const krb5_ap_req *req,
     krb5_enctype tkt_etype = req->ticket->enc_part.enctype;
     krb5_boolean similar_enctype;
     krb5_boolean tkt_server_mismatch = FALSE, found_server_match = FALSE;
-    krb5_boolean found_tkt_server = FALSE, found_enctype = FALSE;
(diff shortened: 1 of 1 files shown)


greghudson:
The code in this PR would try all keys after the first one which matches the realm, rather than trying only keys which match the realm.

Keys are normally either random, or salted with the principal name (including realm), so in most cases it is vanishingly unlikely for keys in two different realms to be identical.  I think a change like this would cause more practical issues (e.g. due to realm case mismatches) than it solves.

(RFC 1510 was obsoleted by RFC 4120, although it has the same text.  The "try all keys" algorithm results from the introduction of aliases in RFC 6806, which can cause the unencrypted part of the ticket to contain a different principal name from the one in the keytab.)

ricardofandrade:
@greghudson thank you for the additional explanation.
You're right about the likelihood of such event.
And reading the RFC it does make sense allowing such behavior.

Instead people should use more secure guidelines while creating keys.
Closing this.

PR #490, opened 2016-07-23
Comment why TRACE() is not used directly

kaduk:
Since it wasn't clear to Will, maybe we should add a comment.


greghudson:
Change "transition to" to "integration with" and I'm sold.

PR #491, opened 2016-07-25
Improve bad password inference in kinit

greghudson:
The first commit on this branch sets prompt types in OTP client preauth.  I noticed this omission while testing an earlier version of the second commit.  I also noticed that clpreauth modules can invoke the prompter but currently have no way of setting prompt types, so I think code which calls krb5_get_prompt_types() needs to expect a possible NULL result.

The second commit implements a change we talked about on last Tuesday's dev call, which is to display "Password incorrect" in kinit if we get a preauth-failed error from the KDC and witnessed a password prompt.  I didn't expand this change to other programs which perform AS requests; I think it's not as big of a deal to see a "Preauthentication failed" error from another utility like kadmin or kpasswd.

### DIFF
--- a/src/clients/kinit/kinit.c
+++ b/src/clients/kinit/kinit.c
@@ -684,9 +684,18 @@ kinit_prompter(
     krb5_prompt prompts[]
 )
 {
-    krb5_error_code rc =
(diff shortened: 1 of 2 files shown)


tlyu:
This looks ok to me.

PR #493, opened 2016-07-25
Add cmocka support

SaharahSarah:
check-cmocka support is added to the Makefiles in a fashion similar to check-pytest. The configure script checks for both the cmocka.h header and the cmocka library before setting HAVE_CMOCKA to yes.

I tried adding a skiptests log to src/Makefile.in when cmocka is skipped due to not being available, but the skiptests file is dumped in check-postrecurse which is called before the local checks in the top-level Makefile, which means that the line isn't added to skiptests till after it's been dumped.


greghudson:
I don't see any problems with the first commit, except for the lack of whitespace alignment in the pre.in change.  After PR #495 it should be possible to print the warning.

In the second and third commits, some of the Makefile.in changes add new lines which exceed 79 characters.  The style checker also reports several problems in the C source files.

In t_parse_host_string.c, what is the point of the handle structure?  It seems like you could just pass the hostname and port number as arguments to call_k5_parse_host_string() and verify the results in that function, without creating or destroying a handle.

In t_sort_key_data.c, I think the TEST_KEYS macro is an overuse of the preprocessor.  I would rather see the declarations and calls spelled out in each test function, so that it is easier to decipher what is being done and easier to step through the code if something goes wrong with a test.  The KEY macro is fine, but I believe the first structure field (key_data_ver) should be 1, not kvno.

tlyu:
Looks good at first glance (I haven't tested the changes). License for new files should be the two-clause BSD-style one from the templates in src/prototype.

tlyu:
I also think TEST_KEYS is overusing the preprocessor somewhat. I prefer the sizeof(a)/sizeof(a[0]) idiom for counting the number of members of an initialized array.

SaharahSarah:
OK, branch has been updated to include the single colon changes from master, and to address the concerns that were raised. Also, the parse_host_string test didn't actually work with the version that made it into master, so the test and the function itself have been updated.

SaharahSarah:
> Do we still need the UNIT_TEST macro in t_parse_host_string.c? It's just a trivial wrapper around cmocka_unit_test, and isn't used consistently.

I removed it.

> We talked on IRC about changing the behavior of k5_parse_host_string() when address is NULL or empty, but my conclusion was that there is no benefit to making that change. NULL is clearly an invalid input for a non-optional parameter, and an empty string is an invalid input according to the documented format. net-server.c does not pass NULL or empty strings to k5_parse_host_string(), so does not require this behavior.

My mistake, I didn't mean to include this change in my commit. It's back to the original behavior now.

> If we are not going to treat ":port" as an empty-string address together with a port, then I think we should treat it as an invalid input. The documentation implies that you can only use the colon separator when both pieces are present.

We DO treat it as an empty-string address together with a port. It's just that the function sets the host address to NULL when it's either an empty string or absent, instead of an empty string. In the documentation for the function:

> An output pointer for the parsed host, or NULL if no host was specified or an error occured

The previous function was returning an empty string instead of NULL in that case which goes against the documentation. I'd rather the function be more permissive since this is user input in a config file.

greghudson:
In the abstract, there is a distinction between "no host was specified" and "an empty-string host was specified."  When you run ``cat ""`` in the shell, you are specifying a filename, albeit an invalid one, and the resulting behavior is not the same as if you just run ``cat``.  That said, since we reject a completely empty address specification, I guess we have imposed a syntactic restriction that a host part must be non-empty, and we should apply that restriction when parsing the "a:b" form to be consistent.

As to whether to be permissive about ":123", I would prefer not to be.  We have the same addr-port format as httpd, and apr_parse_addr_port() does not accept strings beginning with ":" [1].  If we do elect to be permissive, I think that should be made clear in the function description; a parsing function which is more permissive than its contract is dangerous.

[1] https://svn.apache.org/repos/asf/apr/apr/trunk/network_io/unix/sockaddr.c -- search for "string starts with ':' "

greghudson:
In an earlier in-person conversation, you talked about the possibility of redirecting the stdout of cmocka test programs to /dev/null to reduce noise.  Would you mind changing the make check rules to do that?  (My understanding is that this will suppress the reports from successful tests, which have no diagnostic value if a test program fails, while reports of errors from unsuccessful tests will go to stderr and still be displayed.)

greghudson:
Pushed to master as b97888354c97803b4d1617d28b7b597489825693 and preceding.

PR #495, opened 2016-07-26
Use single-colon Makefile rules and add Python skiptest warning

greghudson:


SaharahSarah:
This looks good. Also, compiles fine on Fedora and FreeBSD. Also, the python skiptests entry is confirmed working on FreeBSD.

PR #496, opened 2016-07-27
Move CFLAGS and CPP_FLAGS after local includes

SaharahSarah:
The gss-kernel-lib Makefile overrides ALL_CFLAGS.  It was setting
the CFLAGS and CPPFLAGS to occure before local includes, which
causes some compilers to include system header files before the
local header files.  Moving the CPPFLAGS and CFLAGS to the end of
ALL_CFLAGS corrects this behavior.


tlyu:
Pushed as 5392327b479fd1b964fd72b3b3a0b5b3e04df3a6.

PR #497, opened 2016-07-27
Rename k5-queue.h macros

tlyu:
Some BSD-derived systems (e.g., FreeBSD and Mac OS X) inappropriately
include sys/queue.h from some non-kernel network-related headers that
we include (net/if.h is one example).  Because our k5-queue.h is a
copy from a BSD sys/queue.h, many compilers will warn about macro
redefinitions on those systems.  Rename the queue macros to have a K5_
prefix.

ticket: 8466 (new)

### DIFF
--- a/src/include/k5-queue.h
+++ b/src/include/k5-queue.h
@@ -90,15 +90,15 @@
 /*
  * List definitions.
  */
-#define	LIST_HEAD(name, type)						\
+#define	K5_LIST_HEAD(name, type)					\
 struct name {								\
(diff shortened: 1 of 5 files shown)


greghudson:
This seems reasonable.  I probably would have done K5_QUEUEDEBUG instead of QUEUEDEBUG_K5, but it doesn't really matter.

greghudson:
This still looks okay after the change to remove the queuedebug macros.

PR #498, opened 2016-07-28
Fix build system hygiene issues

greghudson:
Add some missing files to .gitignore, EXTRADEPSRCS, and to clean
rules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ lib*.so.*
 Makefile
 OBJS.*
 obj/
+skiptests
 testdir/
 testlog
 
@@ -182,6 +183,7 @@ local.properties
 /src/include/db.h
 /src/include/gssapi/
 /src/include/kadm5/
+/src/include/kdc_j_encode.h
 /src/include/krb5.stamp
 /src/include/osconf.h
 /src/include/private-and-public-decls
@@ -202,6 +204,7 @@ local.properties
 /src/kadmin/dbutil/import_err.c
 /src/kadmin/dbutil/import_err.h
 /src/kadmin/dbutil/kdb5_util
+/src/kadmin/dbutil/t_tdumputil
 
 /src/kadmin/ktutil/ktutil
 /src/kadmin/ktutil/ktutil_ct.c
@@ -296,7 +299,17 @@ local.properties
 /src/lib/kdb/t_ulog
 /src/lib/kdb/test.ulog
 
+/src/lib/krad/t_attr
+/src/lib/krad/t_attrset
+/src/lib/krad/t_client
+/src/lib/krad/t_code
+/src/lib/krad/t_packet
+/src/lib/krad/t_remote
+
+/src/lib/krb5/ccache/kcmrpc.c
+/src/lib/krb5/ccache/kcmrpc.h
 /src/lib/krb5/ccache/t_cc
+/src/lib/krb5/ccache/t_cccol
 /src/lib/krb5/ccache/t_cccursor
 /src/lib/krb5/ccache/t_marshal
 /src/lib/krb5/ccache/testcache
@@ -359,6 +372,7 @@ local.properties
 /src/slave/kproplog
 
 /src/tests/adata
+/src/tests/au.log
 /src/tests/etinfo
 /src/tests/gcred
 /src/tests/hist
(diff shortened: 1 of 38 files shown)


tlyu:
After `make distclean` on Mac OS 10.10.5, I have
```
$ git status --ignored --untracked=all --porcelain
?? src/lib/krb5/ccache/kcmrpc.c
?? src/lib/krb5/ccache/kcmrpc.h
!! src/autom4te.cache/output.0
!! src/autom4te.cache/requests
!! src/autom4te.cache/traces.0
!! src/build-tools/gssrpc.pc
!! src/build-tools/kadm-client.pc
!! src/build-tools/kadm-server.pc
!! src/build-tools/kdb.pc
!! src/build-tools/krb5-config
!! src/build-tools/krb5-gssapi.pc
!! src/build-tools/krb5.pc
!! src/build-tools/mit-krb5-gssapi.pc
!! src/build-tools/mit-krb5.pc
!! src/configure
!! src/include/autoconf.h.in
!! src/include/autoconf.stmp
!! src/plugins/locate/python/Makefile
!! src/plugins/preauth/cksum_body/Makefile
!! src/plugins/preauth/securid_sam2/Makefile
!! src/plugins/preauth/test.so
!! src/plugins/preauth/wpse/Makefile
!! src/plugins/tls/k5tls.so
!! src/tests/threads/Makefile
!! src/util/collected-client-lib/Makefile
```
Some of those should only be cleaned by an as-yet nonexistent `maintainer-clean` target, but the rest should probably get cleaned.

tlyu:
Some of those Makefiles are in directories that we deliberately omit from SUBDIRS in the top-level Makefile.in, but in that case we might not want to even generate those Makefiles. src/plugins/preauth/test/Makefile.in might be missing a clean-liblinks, etc.

greghudson:
Categorizing that list:

* kcmrpc.c and kcmrpc.h should be removed by clean, and added to .gitignore.  (I missed these because I don't build on OS X.)

* plugins/preauth/test.so and plugins/tls/k5tls.so should be fixed by adding clean-liblinks dependencies.  (I missed these because I was searching for files when I should have been searching for not-directories.)

* build-tools/*.pc and build-tools/krb5-config should be removed by distclean.

* The six Makefiles could be fixed by not generating them, or removing those directories entirely.  Not generating the Makefiles makes it significantly harder to use those directories productively.  We could conceivably remove those Makefiles in the top-level Makefile.in's distclean rule, but it wouldn't be very elegant.

* configure and include/autoconf.h.in should not be removed by distclean, since they are generated by autoreconf and are necessary to run configure.

* autom4te.cache is also generated by autoreconf, but can be removed without affecting configure.  I think we could remove it in distclean, and perhaps we should?

* include/autoconf.stmp is our creation, and tracks the last time include/Makefile generated autoconf.h.in.  I think it should be removed in distclean.

greghudson:
I pushed changes to address everything except the six Makefiles, I think.  I didn't remove autom4te.cache in distclean because I found an assertion from an autoconf maintainer (Akim Demaille) on a mailing list that it should only be removed in maintainer-clean, although he didn't say why.

greghudson:
I ran a similar test to what Tom ran and got:

    !! src/autom4te.cache/
    !! src/configure
    !! src/include/autoconf.h.in
    !! src/include/kadm5/
    !! src/plugins/locate/python/Makefile
    !! src/plugins/preauth/cksum_body/Makefile
    !! src/plugins/preauth/securid_sam2/Makefile
    !! src/plugins/preauth/wpse/Makefile
    !! src/tests/threads/Makefile
    !! src/util/collected-client-lib/Makefile
    !! src/util/k5test.pyc
    !! src/util/princflags.pyc

Most of those are expected, but there is also src/include/kadm5 and the two .pyc files.

include/kadm5 is created by the include target in lib/kadm5.  The clean rule there removes the copied headers but not the directory.  I guess I will replace the clean rule there with $(RM) -r $(HDRDIR).

The .pyc files should be removed by make clean.  I'll add a rule to util/Makefile.in.  (It's annoying that simply placing .py files in a directory adds responsibility to that directory's Makefile.in because of Python's caching, but that's the world we live in.)

tlyu:
One strategy we considered on IRC is to add the problem Makefile directories to the top-level SUBDIRS and make their all-unix targets not depend on anything. (Also we'd move the previous dependencies of all-unix in those Makefiles to a differently-named and well-commented target so people who want to build/run them could do so.)

tlyu:
The current commits look good to me, unless you want to tackle the remnant Makefile issue. I think we agreed on IRC that adding a maintainer-clean target as a separate pull request made the most sense.

PR #499, opened 2016-08-01
Remove all makefiles in make distclean

greghudson:
In PR #498 we noted that "make distclean" still leaves behind nine files, three of which distclean is supposed to leave alone and six of which are Makefiles generated by configure and not visited by recursive make.  We discussed possible solutions to the six Makefiles but didn't implement them in that PR.

Of the six orphaned Makefiles, one is for the securid_sam2 preauth plugin and the other five are for sort of questionable directories which we could maybe do without.  We clearly can't remove the securid_sam2 directory (at least until SPAKE is implemented and deployed at MIT, which is many years away), so I figured it would act as a model for the other five.  However, it turns out that securid_sam2 should be handled by conditionally generating the Makefile, as we do for PKINIT, LDAP, and other conditionally built directories.

The commit I've attached here conditionally generates the securid_sam2 Makefile and implements Tom's suggested solution for the other five directories.  But we might consider deleting the other five directories instead, or at least some of them.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -20,9 +20,12 @@ SUBDIRS=util include lib \
 	plugins/kdb/db2 \
 	@ldap_plugin_dir@ \
 	plugins/kdb/test \
+	plugins/locate/python \
+	plugins/preauth/cksum_body \
 	plugins/preauth/otp \
 	plugins/preauth/pkinit \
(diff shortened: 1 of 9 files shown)


tlyu:
This looks good to me.

PR #500, opened 2016-08-03
Remove unnecessary directories

greghudson:
Remove the plugin modules wpse, cksum_body, and locate/python, which
aren't used by the test suite or built by default.

Remove util/collected-client-lib, as we no longer have a need to
create a smaller client-only library.

Remove util/gss-kernel-lib, as it turned out not to be useful for
facilitating kernel integrations.


SaharahSarah:
I definitely approve this pull request :+1:

tlyu:
This looks ok to me.

PR #501, opened 2016-08-03
Properly escape quotes for otp set_string example

kaduk:
The user from debian bug 828976 noted that the documentation has an example that does not escape the quotes.

Maybe this should get an RT ticket for pullup...

### DIFF
--- a/doc/admin/admin_commands/kadmin_local.rst
+++ b/doc/admin/admin_commands/kadmin_local.rst
@@ -668,7 +668,7 @@ Alias: **setstr**
 Example::
 
     set_string host/foo.mit.edu session_enctypes aes128-cts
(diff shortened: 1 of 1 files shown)


greghudson:
I pre-allocated RT issue 8469 for this, so you can add "ticket: 8469 (new)" and "target_version: 1.14-next" and "tags: pullup".

The parenthetical in the commit message ends with "..., and <noun phrase>" which isn't grammar.

The resulting line after the quotes are added is a little long, which might create an issue for the PDF-formatted documentation (although it wouldn't be our only issue).  I suggest changing "custom" to "al" as a cheap way of shortening it slightly.

kaduk:
I updated the commit; sorry for the delay.

tlyu:
This looks ok to me.

greghudson:
This now looks okay to me as well.

PR #502, opened 2016-08-03
Warn about dump -recurse nonfunctionality

tlyu:
kdb5_util dump -recurse hasn't behaved as documented since krb5-1.5,
when the DAL was integrated.  Restoring it is a nontrivial amount of
work, so just document it for now.

ticket: 8470 (new)
target_version: 1.14-next
target_version: 1.13-next
tags: pullup


greghudson:
This looks fine to me, although I haven't checked how the documentation changes look after formatting.

PR #503, opened 2016-08-04
Decruft top-level Makefile.in

tlyu:
Remove unused stuff from the top-level Makefile.in, and also make Emacs tags file generation actually work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,6 +129,9 @@ local.properties
 /doc/html/
 /doc/pdf/
 
+# Emacs tags table
+/src/TAGS
+
 /src/config.log
 /src/config.status
 /src/configure
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -35,8 +35,6 @@ HDRS =
 # Why aren't these flags showing up in Windows builds?
 ##DOS##CPPFLAGS=$(CPPFLAGS) -D_X86_=1  -DWIN32 -D_WIN32 -W3 -D_WINNT
 
-DISTFILES = $(SRCS) $(HDRS) COPYING COPYING.LIB ChangeLog Makefile.in
-
 # Lots of things will start to depend on the thread support, which
 # needs autoconf.h, but building "all" in include requires that util/et
 # have been built first.  Until we can untangle this, let's just check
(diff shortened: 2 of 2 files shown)


greghudson:
How can I try this out in emacs?  So far I haven't made use of emacs tag support (instead I have a handy shell function to recursively grep the source tree), but it could save me some time.

Does it make sense to have a rule for TAGS in both the source tree and the build directory?  When would a person benefit from one or the other?

Can etags handle yacc/bison source?  If so it might be nice to pick up getdate.y.  It also might be nice to use x-deltat.y in preference to deltat.c, but that seems difficult.  This is a small issue and isn't blocking.

tlyu:
You can use M-. (meta-period) to search for a tag. The Emacs info file has a little more on how to use tags tables.

The build directory target is there for convenience when building in a separate build directory (`make TAGS` from the top of the build tree instead of `make $(top_srcdir)/TAGS`).

`etags --help` shows a list of languages that includes YACC, so we could add the .y files.

greghudson:
Also, should TAGS be cleaned up in distclean?  I'm not sure whether it qualifies for maintainer-clean instead.

tlyu:
A non-maintainer user who is doing minor editing of source code might find the tags file useful, which would suggest distclean.

greghudson:
Adding a dependency to distclean-postrecurse means it could run before, after, or in parallel with distclean-recurse.  It doesn't really matter if that happens, but why not just add the dependency to distclean, or put the rule directly in the distclean target?

Otherwise this now looks fine to me.  (If you want to add the FORCE dependency back that's also okay with me, or you could use .PHONY, or you could rename the phony target so it's not case-insensitively the same as "TAGS".)

I tried out emacs tag support.  While emacs isn't clever enough to walk the parent chain to find TAGS automatically, it seems convenient to use after the first time.

tlyu:
I made some changes in line with your suggestions. Should I add a maintainer-clean target while I'm at it, or make it a separate PR?

greghudson:
Seems okay.  I would make a separate PR for maintainer-clean; I think we would need to discuss exactly what that should get rid of.

tlyu:
Added a missing `--append` to the etags command line for YACC files.

PR #504, opened 2016-08-05
Use responder in non-preauth AS reqs

greghudson:
The first commit fixes RT issue 8454 in a minimal way and adds a test case.  It adds a noticeable amount of code to an already overly long and complicated function, so I included a second refactoring commit.


frozencemetery:
Verified that the responder fix resolves the issue we're seeing, and code for that commit looks good.

In the refactor commit, I personally dislike having a variable named `context` and another named `ctx`, though they have the different types which makes it less bad.  It looks like this was also present in the original code, though.  Otherwise the refactor looks good to me; thanks for making it a separate commit.

PR #505, opened 2016-08-06
Set default_port to host byte order before calling k5_parse_host_string

frenche:
In last refactoring of locate_srv_conf_1() the port number is converted
to network order when passed to add_host_to_list(htons(port_num)).
However, the default_port is already in network order so this caused
my configuration (kdc = 10.0.0.200) to fail (it tried port 22528).

At first, I thought to remove the above htons() call but then it fails
when I specify the port (10.0.0.200:88) because k5_parse_host_string()
doesn't call htons(), so instead I call ntohs(default_port) when
calling k5_parse_host_string() (also, this way the sanity check of
default_port in k5_parse_host_string() makes more sense).

### DIFF
--- a/src/lib/krb5/os/locate_kdc.c
+++ b/src/lib/krb5/os/locate_kdc.c
@@ -254,7 +254,7 @@ locate_srv_conf_1(krb5_context context, const krb5_data *realm,
         parse_uri_if_https(hostspec, &this_transport, &hostspec, &uri_path);
 
         default_port = (this_transport == HTTPS) ? htons(443) : udpport;
-        code = k5_parse_host_string(hostspec, default_port, &host, &port_num);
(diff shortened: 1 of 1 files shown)


greghudson:
Thanks for reporting this; I'm glad this bug didn't make it into testing releases.  While we have lots of automated testing of locate_kdc.c and sendto_kdc.c, none of it uses the default port because only root processes can bind to that port.

I have opened PR #506 to fix this bug with a more aggressive change that simplifies the port handling in locate_kdc.c and other consumers of struct serverlist.  Ordinarily I want to would commit a minimal bugfix first, but since this bug isn't present in any release and therefore doesn't need a backported fix, there is no need.

frenche:
OK, I'll look at the new fix and test my use case later today.
Closing this PR, I'll report back on #506 - thanks!

PR #506, opened 2016-08-07
Fix and simplify locate_kdc.c port byte order

greghudson:
[This PR is intended to supersede #505.]

Commit c89587b4476139f05f34aa2323bd7c7db348c44c introduced a bug in
the handling of KDC specifications using the default port, by passing
it to k5_parse_host_string() in network byte order.  Fix this by
removing all of the port byte order transformations in locate_kdc.c,
and storing the struct serverlist port field in host byte order.
Adjust other consumers of struct serverlist to expect the port to be
in host field to be in host byte order.

Reported by Isaac Boukris.


frenche:
So we never really needed it in network order :)
Looks good to me, and works for all the use cases I tried.

tlyu:
This looks ok to me.

SaharahSarah:
After looking it over a bit I don't see any problems. :+1:

PR #508, opened 2016-08-08
Add another AD-SIGNTICKET corner case test

greghudson:
Prior to the fix for #8139, forwarded TGTs obtained across a krbtgt
re-key could fail if the preferred krbtgt enctype changed, because
krb5_c_verify_checksum() returns an bad-enctype error due to the
mismatched checksum.  Add a test case for this scenario, using a new
test harness program which obtains a forwarded TGT.


tlyu:
This looks ok to me.

PR #509, opened 2016-08-08
Convert k5-thread macros to functions

SaharahSarah:
k5-thread.h has several pthread support calls defined as macros.
These macros may end up directly calling pthread functions.  An
issue can arise if a program is linked with the krb5 support
library and uses these macros.  If the program isn't compiled with
-fPIC, then the calling program can have a segfault if the
pthread functions weren't linked at compile time.  The fix is to
convert the macros which call the pthread functions to functions
within the krb5 support library.

### DIFF
--- a/src/include/k5-thread.h
+++ b/src/include/k5-thread.h
@@ -189,7 +189,6 @@ typedef unsigned char k5_os_nothread_once_t;
      : (assert(*(O) != 4), assert(*(O) == 2 || *(O) == 3), 0))
 
 
-
 #ifndef ENABLE_THREADS
 
 typedef k5_os_nothread_mutex k5_os_mutex;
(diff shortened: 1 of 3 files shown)


greghudson:
I take back the part about removing krb5int_pthread_loaded from the exports file; it could create a transitional ABI problem for downstream packages where krb5 libraries are split into multiple packages.  It should still be removed from k5-threads.h, though.

greghudson:
I've looked over the combined diff and it seems correct, but the commits still need to be squashed.

SaharahSarah:
OK they're squashed :)

greghudson:
Pushed to master as 17932091cc0d5981c5a78d389ffa4a5c7b532bd6.

PR #510, opened 2016-08-11
Restore recursive dump

tlyu:
There's no automated test case for a KDB, but there is one for the lower layer libdb. Some of the bugfix commits could be squashed after review.

### DIFF
--- a/doc/admin/admin_commands/kdb5_util.rst
+++ b/doc/admin/admin_commands/kdb5_util.rst
@@ -182,11 +182,13 @@ load_dump version 7".  If filename is not specified, or is the string
     corruption, this option will probably retrieve more principals
     than the **-rev** option will.
 
-    .. note::
-        The **-recurse** option currently doesn't modify the dump
-        functionality as described above; it does a normal dump.
+    .. versionchanged:: 1.15
+        Release 1.15 restored the functionality of the **-recurse**
+        option.
 
-    .. deprecated:: 1.5
+    .. versionchanged:: 1.5
+        The **-recurse** option ceased working until release 1.15,
+        doing a normal dump instead of a recursive traversal.
 
 .. _kdb5_util_dump_end:
 
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -135,6 +135,7 @@
 /* KDB iteration flags */
 #define KRB5_DB_ITER_WRITE      0x00000001
 #define KRB5_DB_ITER_REV        0x00000002
+#define KRB5_DB_ITER_RECURSE    0x00000004
 
 /* String attribute names recognized by krb5 */
 #define KRB5_KDB_SK_SESSION_ENCTYPES            "session_enctypes"
--- a/src/kadmin/dbutil/dump.c
+++ b/src/kadmin/dbutil/dump.c
@@ -1304,11 +1304,7 @@ dump_db(int argc, char **argv)
         } else if (!strcmp(argv[aindex], "-rev")) {
             iterflags |= KRB5_DB_ITER_REV;
         } else if (!strcmp(argv[aindex], "-recurse")) {
-            /* Accept this for compatibility, but do nothing since
-             * krb5_db_iterate doesn't support it. */
-            fprintf(stderr,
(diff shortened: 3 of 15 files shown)


greghudson:
Commit ~11: What is the motivation for better EINTR handling in bttest?  (What does bttest do?  It's not built or run by default, right?)

Commit ~8: I see that this is making mpool_stat() visible to the linker and in the header file without STATISTICS define, but how does that "make btree stats work"?

Commit ~6: What is MPOOL_IGNOREPIN supposed to do, and how was it broken?

Commits ~5 through ~2: I will look at these more carefully in another pass, but more explanation in the commit messages would help.

Commit ~1: no issues (you could push this while the rest of the PR is still in review).

Head commit: If R_RNEXT isn't defined, I don't think this will compile.  Also, checking dbc->hashfirst to detect a hash database after open should work, but it's unclear to a casual reader why that is, given the field's name and documentation.

tlyu:
I squashed a few commits, and expanded commit messages and code comments where appropriate. I haven't come up with a solution for making it build in the system libdb case, but am working on it.

I should also note that although I don't have an automated test for KDB corruption (as opposed to having automated tests for the underlying btree code), I have manually verified that the recursive dump recovers otherwise unreachable records after using bttest to unlink pages. This is partly due to difficulties with running bttest from the build tree from a test script (though this might also be feasible with additional work).

tlyu:
Fixed the final commit so it builds when configured `--with-system-db`. Tests still fail in that case because recent BDB (I tested db-5.3) seems to not work with `kdb5_util load` for some reason.

tlyu:
Add some further comment improvements and copyright bumps.

greghudson:
Commit ~9 (EINTR handling): it looks okay for what it does.  I still don't quite understand the problem being solved; I tried writing a simple test program using fgets and attaching to it with lldb, but I couldn't produce any bad behavior.

Commit ~8 (string printing): in the commit message, 'format specifiers such as "%.*s"' is unnecessarily wishy-washy, as "%.*s" is the one and only format specifier used in new lines of code.  The code changes look okay.

Commit ~7 (exports): in the commit message, "On platforms that have use" contains the extra word "have".  The code changes look okay.

Commit ~6 (void return): Looks good.

Commit ~5 (MPOOL_IGNOREPIN): in the commit message, "use the MPOOL_IGNOREPIN to" is missing the word "flag".  The commit message only describes the mpool.c change, not the bttest changes.  The mpool.c change looks okay.  I assume the bttest changes are okay too, but I don't have the context to know why setting MPOOL_IGNOREPIN is necessary and why not calling mpool_put() is desirable.

Commit ~4 (refactor recursive traversal): in the commit message, "flags R_RNEXT (...) and R_RPREV instead of ..." is awkward because "instead of" intuitively applies only to R_RPREV.  I would just remove the "instead of" clause since it's pretty clear from the above text.  In the last paragraph, I would note that the existing rlist command in bttest is fixed up to use the new interface.  The code changes look okay.

Commit ~3 (bttest unlink): in the commit message, the sentence "This new bttest command..." has a grammatical structure issue.  The final paragraph should use a comma instead of "and".  In the code changes, the comment "Find a candidate page to ulink" should say "unlink".  candidatepg() can be slightly simplified by returning directly from the loop (return NULL if mpool_get() returns NULL, return h intsead of breaking later, return NULL after the loop).  The code changes otherwise look okay.

Commit ~2 (test case): this looks okay.

Commit ~1 (curs_init errors): as before, this looks okay.

Head commit (restore recursive dump): this looks okay, but I would expect it to have an associated ticket.

tlyu:
The lldb fgets/EINTR issue might be specific to Mac OS X; I'm not sure.

tlyu:
A short test program that calls fgets on /dev/tty has the same SIGTTIN problem when run out of lldb on Mac OS. Changing the file handle to be stdin makes that work when run directly out of lldb, but the there's still a SIGSTOP when I try to attach.

greghudson:
It might be appropriate to associate more of the commits with ticket 8476 (particularly ~4), but it's not important.

Otherwise, I think all of my feedback has been addressed; these changes look okay.

PR #511, opened 2016-08-13
Fix build with -O3 on ppc64el

kaduk:
Ubuntu runs ppc64el builds with -O3, which elicited a few warnings
from gcc that were not generated elsewhere, as documented at
https://bugs.launchpad.net/ubuntu/+source/krb5/+bug/1592841 .

Initialize the output variable at the top of a couple helper functions
to silence the uninitialized-variable warnings.

### DIFF
--- a/src/lib/krb5/os/sendto_kdc.c
+++ b/src/lib/krb5/os/sendto_kdc.c
@@ -173,6 +173,8 @@ get_curtime_ms(time_ms *time_out)
 {
     struct timeval tv;
 
+    *time_out = 0;
+
     if (gettimeofday(&tv, 0))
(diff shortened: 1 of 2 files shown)


greghudson:
This looks okay to me.

Please add to the commit message:

    ticket: 8475 (new)
    target_version: 1.14-next
    target_version: 1.13-next
    tags: pullup

tlyu:
The commit message is missing "of" in "a couple helper functions", but the diff looks OK to me.

frozencemetery:
Looks good to me.  As for grammar, both the commit message as written and tlyu's suggestion are correct in my dialect.

PR #512, opened 2016-08-14
Work around glibc OFD lock bug on 32-bit Linux

greghudson:
A bug in Gnu libc causes OFD locking to fail unpredictably on 32-bit
Linux, typically leading to deadlocks.  Work around this bug by using
the fcntl64 system call and struct flock64.

See also: https://sourceware.org/bugzilla/show_bug.cgi?id=20251


frenche:
Looks ok and works in my environment.

tlyu:
This looks ok to me.

PR #513, opened 2016-08-14
Allow libev to use pthreads on old platforms

greghudson:
The upgrade to libev 4.22 introduced the use of "memory fences" for
more reliable signal processing.  Memory fences are usually
implemented using assembly or compiler primitives, but may be
implemented using pthreads as a last resort.  The unmodified libev
errors out at compile time if pthreads is used, but notes that this
error can be removed if relying on pthreads is okay.  Because the
project's nightly build infrastructure includes an old Solaris machine
whose toolchain is too old for any of the non-pthreads memory fence
implementations, remove the error to allow the build to succeed.  (A
dependency on pthreads functions on Solaris does not require linking
with libpthread.)

### DIFF
--- a/src/util/verto/ev.c
+++ b/src/util/verto/ev.c
@@ -1511,7 +1511,14 @@ ecb_binary32_to_binary16 (uint32_t x)
  * alternatively, you can remove this #error and link against libpthread,
  * which will then provide the memory fences.
  */
+/*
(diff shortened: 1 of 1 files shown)


tlyu:
This looks ok to me, but I haven't tested it on Solaris.

PR #514, opened 2016-08-15
Fix KDC to drop repeated in-progress requests

SaharahSarah:
The KDC's behavior when it receives a repeated request while the
original request is still processing is supposed to be to drop the
request.  A bug was causing the KDC to instead replay an empty
reply.  Setting kdc_check_lookaside in replay.c to return a NULL
reply_packet when the request hasn't finished returns the KDC back
to the expected behavior.


greghudson:
Pushed to master as 847fc7b3caa823c219c97cc307ccb8d7d519a20f.

PR #515, opened 2016-08-17
Output last command info on k5test failures

greghudson:
When a k5test failure occurs, display the last executed command, its
command index, and its output.  This will make it easier to understand
"make check" failures when it is not easy to run subsequent commands
or investigate the filesystem of the build host.

### DIFF
--- a/src/util/k5test.py
+++ b/src/util/k5test.py
@@ -373,6 +373,11 @@ def fail(msg):
     """Print a message and exit with failure."""
     global _current_pass
     print "*** Failure:", msg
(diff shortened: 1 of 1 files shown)


tlyu:
This looks ok to me.

greghudson:
Fortuitously, we got the sporadic Travis t_workers.py failure on this branch; unfortunately, my changes didn't cover _start_daemon() so they didn't help.  I updated the patch to include that; it needs review again.

tlyu:
The updated changes look ok to me.

PR #516, opened 2016-08-17
Remove instances of "XXX" in error descriptors

frozencemetery:
These are confusing to end users who may encounter them and serve no clear purpose.

This also makes the strings for `KRB5_RC_IO` consistent.


greghudson:
Hm, we should possibly get rid of the texinfo documentation, since we haven't been keeping it up to date.

For this PR, I'd prefer if the commit just changed krb5_err.et.  The pot file gets regenerated periodically; we don't try to keep it in sync with the code on every commit.

frozencemetery:
Removed the pot file.  If you'd like I can also add a commit to remove the texinfo documentation.

tlyu:
I think the LaTeX documentation deletion should be a separate pull request. The rest of this looks ok to me.

frozencemetery:
Removed TeX file modification from this commit since there is now a deletion PR.

tlyu:
Pushed as c4fd3dcfc9b09f83ca724260f648f1a2052e6b83.

PR #517, opened 2016-08-19
Add kdc lookaside cache tests

SaharahSarah:
The kdc lookaside cache was missing any tests which caused me a pretty major headache, so I went ahead and created some cmocka unit tests for it. Also, I fixed a couple comments in replay.c and fixed a bug which caused the number of entries to be incorrect.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -233,6 +233,7 @@ local.properties
 /src/kdc/kdc5_err.[ch]
 /src/kdc/krb5kdc
 /src/kdc/rtest
+/src/kdc/t_replay
 
 /src/lib/crypto/builtin/aes/aes-gen
 /src/lib/crypto/builtin/aes/kresults.out
--- a/src/kdc/Makefile.in
+++ b/src/kdc/Makefile.in
@@ -27,7 +27,8 @@ SRCS= \
 	$(srcdir)/kdc_audit.c \
 	$(srcdir)/kdc_transit.c \
 	$(srcdir)/tgs_policy.c \
-	$(srcdir)/kdc_log.c
+	$(srcdir)/kdc_log.c \
+	$(srcdir)/t_replay.c
 
 OBJS= \
 	kdc5_err.o \
@@ -73,6 +74,14 @@ check-unix: rtest
 	cmp test.out $(srcdir)/rtest.good
 	$(RM) test.out
 
+T_REPLAY_OBJS=t_replay.o
+
+t_replay: $(T_REPLAY_OBJS) replay.o $(KRB5_BASE_DEPLIBS)
+	$(CC_LINK) -o $@ $(T_REPLAY_OBJS) $(CMOCKA_LIBS) $(KRB5_BASE_LIBS)
+
+check-cmocka: t_replay
+	$(RUN_TEST) ./t_replay > /dev/null
+
 check-pytests:
 	$(RUNPYTEST) $(srcdir)/t_workers.py $(PYTESTFLAGS)
 	$(RUNPYTEST) $(srcdir)/t_emptytgt.py $(PYTESTFLAGS)
@@ -81,5 +90,5 @@ install:
 	$(INSTALL_PROGRAM) krb5kdc ${DESTDIR}$(SERVER_BINDIR)/krb5kdc
 
 clean:
-	$(RM) kdc5_err.h kdc5_err.c krb5kdc rtest.o rtest
+	$(RM) kdc5_err.h kdc5_err.c krb5kdc rtest.o rtest t_replay.o t_replay
 
--- a/src/kdc/replay.c
+++ b/src/kdc/replay.c
@@ -110,11 +110,51 @@ entry_size(const krb5_data *req, const krb5_data *rep)
         ((rep == NULL) ? 0 : rep->length);
 }
 
+/* Insert an entry into the cache using the given hash value. */
+static struct entry *
+insert_entry(krb5_context context, krb5_data *req, krb5_data *rep,
+             krb5_timestamp time)
+{
+    krb5_error_code ret;
+    struct entry *entry;
+    krb5_ui_4 req_hash = murmurhash3(req);
+    size_t esize = entry_size(req, rep);
+
+    entry = calloc(1, sizeof(*entry));
+    if (entry == NULL)
+        return NULL;
+    entry->timein = time;
+
+    ret = krb5int_copy_data_contents(context, req, &entry->req_packet);
+    if (ret) {
+        free(entry);
+        return NULL;
+    }
+
+    if (rep != NULL) {
+        ret = krb5int_copy_data_contents(context, rep, &entry->reply_packet);
+        if (ret) {
+            krb5_free_data_contents(context, &entry->req_packet);
+            free(entry);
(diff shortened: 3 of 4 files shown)


greghudson:
t_replay needs to be added to .gitignore.

If we're fixing the comments for kdc_check_lookaside( ) and kdc_insert_lookaside(), it would be nice if they talked about null reply_packet pointers.

time_return() should be a function, not a macro, unless there's some reason it can't be a function that I've missed.

Unless we actually need a separate krb5_context per test, it's better to reuse the same one for all of them.  Using a global krb5_context variable is fine for a test program.

We usually write conditions as "variable == constant", not "constant == variable" as is done in most of the test functions.

There are some blank lines in odd places, like in the middle of variable declarations in several functions.  Blank lines inside functions should be used to separate the declarations from the code, and to indicate related "paragraphs" in the code.

The rotl32 tests don't appear to need any setup, so shouldn't be using the setup and teardown functios.  For the murmurhash tests, I think the only required setup is setting seed, so it feels inelegant to be doing more than that, but I guess it's simpler not to have a separate setup function.  (I'm honestly not very fond of cmocka setup and teardown functions; I can explain why in a more appropriate context.  But it's okay if you want to use them here.)

All instances of "lookahead" should be changed to "lookaside", in the commit summary and the code.

SaharahSarah:
OK, code has been updated to address your concerns!

greghudson:
Sorry to have missed this stuff on the first go-round:

* The first commit description should talk about the "KDC lookaside cache"; as it stands, a reader would tend to assume that it's talking about lib/krb5/rcache before looking at the commit changes.

* hc_data1 and friends should be declared static.

* I think it reads better to define __wrap_krb5_timeofday() before including replay.c so that no declaration is needed.  Also, with this method of wrapping, __wrap_krb5_timeofday() can be declared static and does not need KRB5_CALLCONV.

* In new code, we don't put a space after cast operators.  But also, none of the casts of *state to type (krb5_context) are required; C's implicit casting from void * will do.

* Use uint32_t instead of krb5_ui_4, even when interoperating with code which uses krb5_ui_4.  They are interchangeable types and we are phasing out the krb5-specific names.

* It is clearer to use make_data() and string2data() than static krb5_data initializers.  In most (all?) of your sample strings, the length includes the terminating zero byte, which makes string2data() not work.  It would probably be better to use string2data() and fix up the hash values to match the strings without the zero terminator.  For test_murmurhash3_byte() and similar tests, make_data() can be used.

* t_insert_entry() seems problematic, as it duplicates part of kdc_insert_lookaside().  Looking at how it is used, it doesn't seem like it's really necessary, although a dozen or so tests would need to be changed (e.g. by computing the murmurhash of the req_packet).

* Where you have a block comment introducing a group of functions, put a blank line afterwards so that it doesn't look like a comment for just the first function.

SaharahSarah:
> Sorry to have missed this stuff on the first go-round:
> 
> * The first commit description should talk about the "KDC lookaside cache"; as it stands, a reader would tend to assume that it's talking about lib/krb5/rcache before looking at the commit changes.

Done

> * hc_data1 and friends should be declared static.

Done

> * I think it reads better to define __wrap_krb5_timeofday() before including replay.c so that no declaration is needed.  Also, with this method of wrapping, __wrap_krb5_timeofday() can be declared static and does not need KRB5_CALLCONV.

Done

> * In new code, we don't put a space after cast operators.  But also, none of the casts of *state to type (krb5_context) are required; C's implicit casting from void * will do.

Done

> * Use uint32_t instead of krb5_ui_4, even when interoperating with code which uses krb5_ui_4.  They are interchangeable types and we are phasing out the krb5-specific names.

Done

> * It is clearer to use make_data() and string2data() than static krb5_data initializers.  In most (all?) of your sample strings, the length includes the terminating zero byte, which makes string2data() not work.  It would probably be better to use string2data() and fix up the hash values to match the strings without the zero terminator.  For test_murmurhash3_byte() and similar tests, make_data() can be used.

Done.

> * t_insert_entry() seems problematic, as it duplicates part of kdc_insert_lookaside().  Looking at how it is used, it doesn't seem like it's really necessary, although a dozen or so tests would need to be changed (e.g. by computing the murmurhash of the req_packet).

These are unit tests, and the problem with this is that then all those tests will depend on kdc_insert_lookaside working correctly. The unit tests SHOULD be as isolated from the rest of the code as possible (though I decided that it was pointless in completely redoing some of the functions). I did the tests in the same order they show up in replay.c because I wanted to at least make sure that a test that used another function came after when the other function was tested. I personally still think it's best not to use something like kdc_insert_lookaside since it's a rather complex function that uses many other parts of replay.c, so I believe it shouldn't be used or tested till last.

> * Where you have a block comment introducing a group of functions, put a blank line afterwards so that it doesn't look like a comment for just the first function.

Done

tlyu:
I kind of wish there were a way to replace stuff like the set of test_rotl32 functions with a single function and an annotated table of inputs and outputs, but I don't know whether cmocka makes that easy.

I think kdc_insert_lookaside needing to be partially duplicated for testing purposes is a sign that refactoring it might be a good idea.

greghudson:
> These are unit tests, and the problem with this is that then all those tests will depend on kdc_insert_lookaside working correctly.

That's an interesting point.  But as data structures grow more complex and intricate, unit tests written to this philosophy will require more and more complicated, implementation-aware code in order to set up the preconditions required to test accessor and mutator interfaces.

I looked online and found a variety of opinions as to whether unit tests should be black-box or white-box and how they should set up preconditions.  I'm not sure if there is a settled best practice.  For the purposes of this PR, I would be content with Tom's implied suggestion: create a static helper function in replay.c which does the work of t_insert_entry() (without using asserts, of course), use it in kdc_insert_lookaside(), and also use it in the unit tests.

> I kind of wish there were a way to replace stuff like the set of test_rotl32 functions with a single function and an annotated table of inputs and outputs, but I don't know whether cmocka makes that easy.

A quick look suggests that cmocka unit test entries consist only of three function pointers, and can't be parameterized by data values.  So we could have a single unit test which iterates over a table of inputs and expected results, but I don't think we can make each input and result a separate cmocka unit test.  (I could be missing something.)

SaharahSarah:
kdc_insert_lookaside has been refactored with the insertion code moved to a new function. tests have been added for that function.

greghudson:
Sorry for the delay in getting back to this.

In the tests, there are a number of code blocks with repeated uses of complicated expressions which leads to awkward wrapping, like:

    assert_non_null(K5_LIST_FIRST(&hash_table[req_hash2]));
    assert_true(data_eq(K5_LIST_FIRST(&hash_table[req_hash2])->req_packet,
                        req2));
    assert_int_equal(K5_LIST_FIRST(&hash_table[req_hash2])->
                     reply_packet.length,
                     0);

Can you introduce intermediate variables (such as a pointer to hold the result of K5_LIST_FIRST(&hash_table[req_hash2])) to avoid the repetition and make the asserts fit on one line?

greghudson:
Pushed to master as 23762eae1103d429493919fb0cbafb3a93307cc8 f208d94e2b99945023ac4c3dfad35b85aba41760.

PR #518, opened 2016-08-19
Test the min char classes policy in ldap

SaharahSarah:
t_kdb didn't have a test for the minimum character classes password policy in ldap. I figured it'd be a good idea to add a test to make sure it works.


greghudson:
Pushed to master as 058a688335019709a9135f5c6377bb41f7eac5c9.

PR #520, opened 2016-08-19
Remove outdated TeX documentation

frozencemetery:
As suggested in #516, this removes the TeX docs in the top-level doc dir (api, kadm5, rpc).


tlyu:
I agree with removing doc/api because it is (probably) redundant with the Doxygen/Sphinx API documentation. As far as I know, doc/kadm5 and doc/rpc still contain content that isn't yet anywhere else.

frozencemetery:
Makes sense.  Dropped doc/kadm5 and doc/rpc from this PR.

greghudson:
Pushed to master as e394d42e8fbd6ae176ad042f9668ec252188239a.

PR #521, opened 2016-08-23
Fix GSSRPC server credential memory leak

greghudson:
In svc_auth_gss.c, stop using the global svcauth_gss_creds, and
instead keep a credential in struct svc_rpc_gss_data.  This change
ensures that the same credential is used for each accept_sec_context
call for a particular context, and ensures that the credential is
freed when the authentication data is destroyed.  Also, do not acquire
a credential when the default name is used (as it is in kadmind) as it
is not needed.

Leave the svcauth_gss_creds around for the backportable fix as it is
in the library export list.  It will be removed in a subsequent
commit.

### DIFF
--- a/src/include/gssrpc/rename.h
+++ b/src/include/gssrpc/rename.h
@@ -257,8 +257,6 @@
 
 #define svc_debug_gss		gssrpc_svc_debug_gss
 
-#define svcauth_gss_creds	gssrpc_svc_auth_gss_creds
-
 /* svc_auth_gssapi.c */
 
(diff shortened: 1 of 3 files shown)


frozencemetery:
Looks good to me.  The backportable fix is appreciated.

tlyu:
Looks ok to me, and I verified by manual testing with valgrind on amd64 Ubuntu 14.04.

greghudson:
Pushed to master as 670d9828086e979d5cdfd26f00ca88958a03754e 15e73c80420a818473a1289d09347b7a61f17cf2.

PR #522, opened 2016-08-23
Improve checking of decoded DB2 principal values

greghudson:
In krb5_decode_princ_entry(), verify the length of the principal name
before calling krb5_parse_name() or strlen(), to avoid a possible
buffer read overrun.  Check all length fields for negative values.  If
the value of key_data_ver is unexpected, return KRB5_KDB_BAD_VERSION
instead of aborting.


frozencemetery:
Line 391 - `if ((sizeleft -= 4) < 0) {` should probably converted for consistency.

Otherwise looks good to me.

greghudson:
Thanks; I've updated the commit with that change, and also noted in the commit message that the new code is deliberately avoiding arithmetic during bounds checks (to avoid having to think about possible overflow).

frozencemetery:
Looks good to me.

PR #524, opened 2016-08-25
Guess the mutual flag based on ap_options in krb5 gssapi mech

gladiac:
To work correctly with older Samba clients we should guess the mutual
flag based on the ap_options from the AP-REQ and not set it always.

Found by samba torture testsuite built with MIT Kerberos.


greghudson:
Pushed to master as 7919818c0eec534828521aed01b89aa72e5e7e81.

PR #525, opened 2016-08-25
Consolidate sn2princ_realm() in kprop and kpropd

greghudson:
In kprop and kpropd, factor out the duplicated implementation of
sn2princ_with_realm() into kprop_util.c.  Rename it to
sn2princ_realm(), remove the type parameter, and require the sname
parameter to be specified.  Rewrite the function to use
krb5_expand_hostname(), avoiding an unnecessary hostrealm lookup.

[There is also a prior commit to rename kprop_sock.c to kprop_util.c.]

### DIFF
--- a/src/slave/Makefile.in
+++ b/src/slave/Makefile.in
@@ -3,11 +3,11 @@ BUILDTOP=$(REL)..
 
 all:	kprop kpropd kproplog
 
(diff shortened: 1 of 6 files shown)


mrogers950:
Thanks for taking care of this. LGTM.

tlyu:
This looks ok to me.

PR #526, opened 2016-08-25
Clean up libdb2 warnings

tlyu:
Clean up many pointer alignment warnings by casting through (void *).

Clean up many signed-unsigned comparison warnings by casting to
unsigned types as appropriate.

### DIFF
--- a/src/plugins/kdb/db2/libdb2/btree/bt_debug.c
+++ b/src/plugins/kdb/db2/libdb2/btree/bt_debug.c
@@ -59,7 +59,6 @@ static void
 __bt_dinit()
 {
 	static int first = 1;
-	char buf[1024];
 
 	if (!first)
 		return;
@@ -96,7 +95,7 @@ __bt_dump(dbp)
 	(void)fprintf(tracefp, "%s: pgsz %d",
 	    F_ISSET(t, B_INMEM) ? "memory" : "disk", t->bt_psize);
 	if (F_ISSET(t, R_RECNO))
-		(void)fprintf(tracefp, " keys %lu", t->bt_nrecs);
+		(void)fprintf(tracefp, " keys %lu", (u_long)t->bt_nrecs);
 #undef X
 #define	X(flag, name) \
 	if (F_ISSET(t, flag)) { \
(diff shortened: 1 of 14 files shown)


greghudson:
Can this PR now be rebased so it's just the one commit?

I don't see any warning-related reason to change "level" to unsigned where that's done.  Is that another change to sync to the NetBSD version?  Is that also true for the changes to make many fields of struct hashhdr unsigned?  (I haven't tracked down what all of the struct field changes do, so those might be warning-related as well.)

tlyu:
Changing the type of "level" to unsigned works around a gcc warning about assuming signed overflow doesn't occur. For some reason the Travis clang didn't warn about that, but the Travis gcc did. Making many hashhdr fields unsigned clears many signed/unsigned warnings without having to make a huge number of explicit casts.

On further reflection, the P_BIGKEY memcpy changes might actually fix data corruption issues (possibly rare) on strict-alignment platforms (due to crashes in mid-split), so we might want to split them out into a separate commit for back porting.

tlyu:
Now rebased on top of #537.

greghudson:
The style checker notes a whitespace problem at btree.h line 139; there are seven unnecessary spaces preceding the "\".

The code changes look good.  My compiler (gcc on Ubuntu 14.04) still reports these warnings:

    hash.c:546:9: warning: ignoring return value of ‘tmpnam’, declared with attribute warn_unused_result [-Wunused-result]
    hash_page.c:308:22: warning: variable ‘next_key’ set but not used [-Wunused-but-set-variable]
    hash_page.c:1130:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    hash_page.c:1142:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    hash_page.c:1166:5: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    hash_page.c:1193:3: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    hash_page.c:1226:3: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]

I completely disagree with the warn_unused_result attribute (its semantics and its application to write()), but if we want to be warning-clean we will have to deal with these eventually.

The next_key warning manifests because of an assertion inside "#ifdef DEBUG".

PR #527, opened 2016-08-25
Avoid byte-swap corruption in libdb2

tlyu:
Apply a patch from NetBSD to restore the cached copy of a page to the
machine byte order after a write operation swapped it to the file byte
order.

[This is on top of #526 but is probably cherry-pickable to existing release branches.]


tlyu:
Updated to include multiple byte swapping bug fixes:

1. unaligned accesses when swapping big key or big data overflow page pointers (NetBSD patch)
2. swapping of incorrect targets in the small key, big data case
3. the cached page issue fixed in the previous push (NetBSD patch)

Also add test cases for all of these. Apparently (2) might be original to the btree code (20-25 years ago) and still unfixed in the BSDs.

Rebased on top of #528 because that's needed to for testing this branch under Solaris.

greghudson:
This is no longer on top of #526, and #526 is now on top of this PR, right?

I believe I understand the fix for bug 2, but it could be explained more clearly in the commit message.  I suggest keeping the current text, and adding:  "Fix this bug by keeping p unchanged when processing the P_BIGKEY flag, and always adding the full key size to p when processing the P_BIGDATA flag."

The test for bug 2 introduces an unconditional dependency on Python for make check (though it looks like ancient versions of Python should work).  I can't remember if we already have an unconditional dependency on Perl, or if there is any way to de-hex a file using other tools.  From out of band conversation, it sounds like you tried to use awk but ran into a bug in the Solaris awk when printing a 0 byte.

The commit structure of the PR branch is a little inconsistent: fix 1, test 1, fix 2, test 2, fix+test 3.

All of the bugfix code changes look okay.

tlyu:
Clarified commit message for bug 2 and changed the test case to use a very small Perl script.

greghudson:
This looks good to me now.

PR #528, opened 2016-08-26
Fix thread support for Solaris and simplify

greghudson:
threads.c failed to build on Solaris afer commit
17932091cc0d5981c5a78d389ffa4a5c7b532bd6 because k5-thread.h did not
define the conditional k5_once_t structure (because NO_WEAK_PTHREADS
is defined) but threads.c tried to build the conditional k5_once()
function.

Use a single preprocessor symbol, USE_CONDITIONAL_PTHREADS, to
determine whether to use and define pthreads functions which
conditionalize on whether pthreads is loaded.  In threads.c, move the
new k5_once() definitions into the USE_CONDITIONAL_PTHREADS block,
defining a stub function if other code will not refer to it.

Also move #pragma weak declarations from k5-threads.h into threads.c,
as we should no longer be conditionally referring to those symbols
outside of threads.c.

Also eliminate some missing-prototype warnings where we define
functions for linker-visibility but don't have corresponding
prototypes in k5-threads.h.

### DIFF
--- a/src/include/k5-thread.h
+++ b/src/include/k5-thread.h
@@ -243,14 +243,10 @@ typedef k5_os_nothread_mutex k5_os_mutex;
    symbol tables of the current process.  */
 
 #if defined(HAVE_PRAGMA_WEAK_REF) && !defined(NO_WEAK_PTHREADS)
-# pragma weak pthread_once
-# pragma weak pthread_mutex_lock
-# pragma weak pthread_mutex_unlock
-# pragma weak pthread_mutex_destroy
-# pragma weak pthread_mutex_init
-# pragma weak pthread_self
-# pragma weak pthread_equal
-# define USE_PTHREAD_LOCK_ONLY_IF_LOADED
+# define USE_CONDITIONAL_PTHREADS
+#endif
+
+#ifdef USE_CONDITIONAL_PTHREADS
 
 /* Can't rely on useful stubs -- see above regarding Solaris.  */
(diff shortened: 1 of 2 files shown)


SaharahSarah:
This looks good to me.

frozencemetery:
Looks good to me as well.

tlyu:
The commit message says "move #pragma weak declarations from k5-threads.h into threads.c", but they remain in k5-threads.h. Why are the `#pragma weak` directives, etc. in threads.c under `HAVE_PRAGMA_WEAK_REF` instead of `USE_CONDITIONAL_PTHREADS`?

greghudson:
There were a couple of iterations of this patch, and apparently I missed a change when reconstructing it in a later iteration, which is why k5-thread.h still had the "#pragma weak" declarations.  I've removed them.

Commit 151a777ff7f122fc97daf1ab748249d8de934452 (back in 2007) added NO_WEAK_PTHREADS to k5-thread.h, but left threads.c alone, so that libkrb5support still defines krb5int_pthread_loaded() on Solaris.  I don't know if that was deliberate; it is sometimes useful during upgrades for libraries to interoperate with older versions of libraries within the tree, although I think it's much less of a concern for Solaris than it is for Linux distributions.

tlyu:
Looks good now. I guess we can leave the NO_WEAK_PTHREADS discrepancy alone for now.

PR #529, opened 2016-08-27
Fix check-pytests-no

tlyu:
Commit 0db097ba8b605ea7a6e0364ad786da6528868179 accidentally removed
the $(SKIPTESTS) line from check-pytests-no, causing a syntax error
when running "make check" when python is missing or not sufficiently
new.


greghudson:
Looks good to me.

PR #531, opened 2016-08-31
Update config.guess, config.sub

tlyu:
ticket: 8489 (new)
target_version: 1.14-next
target_version: 1.13-next
tags: pullup


greghudson:
This looks fine.

It seems a little amiss that we have to maintain copies of config.guess and config.sub even though we're using autoreconf, but experimentally if I remove them, autoreconf doesn't copy them in.

tlyu:
autoreconf might pull in config.guess and config.sub (via automake) if libtool is in use, but then we might only get whatever was current when libtool was installed, instead of the latest from the upstream repository.

PR #532, opened 2016-09-02
Fix kadmin test scripts buildsystem on FreeBSD

frozencemetery:
We depend on the behavior of having a separate subshell for each line,
so force it where make (on FreeBSD 10.3) does not create one.

### DIFF
--- a/src/build-tools/Makefile.in
+++ b/src/build-tools/Makefile.in
@@ -14,7 +14,7 @@ PKGCONFIG_FILES = \
 all-unix: krb5-config $(PKGCONFIG_FILES)
 
 krb5-config $(PKGCONFIG_FILES): $(BUILDTOP)/config.status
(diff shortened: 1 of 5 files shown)


greghudson:
grepping the makefiles for "cd ", I see other possible instances of this problem in src/doc and src/include.

tlyu:
I think the Makefile comment can be omitted.

frozencemetery:
Eliminated comment, added in other locations, and updated commit message.

tlyu:
Looks good. If we're putting all instances of `cd dir && do_stuff` in subshells even when they're not followed by other command lines, we should probably also include the stuff in config/post.in.

greghudson:
Pushed to master as 69571141ccc2483b54e015a5d7c77d1e926f2822.

PR #533, opened 2016-09-05
Add a krb5_rd_req() function returing the decryption key

gladiac:
I've implemented a kpasswd service in Samba. The reason is that we need ACL checking against Samba database (DSDB).

For this we have an krb5 implementation to decrypt the tickets. To get the authenticated session info, we get the ticket parse the PAC and verify the PAC signature. For the validation of the PAC signature I need the key we used to decrypt the ticket. So I added a new function called krb5_rd_req_key() which adds an out pointer to store the allocated keyblock.

As I'm still a total KRB5 newbie, I'm not sure if this is the right way to get it, but it was the only I found.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2179,7 +2179,8 @@ void KRB5_CALLCONV krb5_free_config_files(char **filenames);
 
 krb5_error_code krb5_rd_req_decoded(krb5_context, krb5_auth_context *,
                                     const krb5_ap_req *, krb5_const_principal,
-                                    krb5_keytab, krb5_flags *, krb5_ticket **);
+                                    krb5_keytab, krb5_flags *, krb5_ticket **,
+                                    krb5_keyblock **);
 
 krb5_error_code krb5_rd_req_decoded_anyflag(krb5_context, krb5_auth_context *,
                                             const krb5_ap_req *,
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2185,7 +2185,8 @@ krb5_error_code krb5_rd_req_decoded(krb5_context, krb5_auth_context *,
 krb5_error_code krb5_rd_req_decoded_anyflag(krb5_context, krb5_auth_context *,
                                             const krb5_ap_req *,
                                             krb5_const_principal, krb5_keytab,
-                                            krb5_flags *, krb5_ticket **);
+                                            krb5_flags *, krb5_ticket **,
+                                            krb5_keyblock **);
 
 krb5_error_code KRB5_CALLCONV
 krb5_cc_register(krb5_context, const krb5_cc_ops *, krb5_boolean );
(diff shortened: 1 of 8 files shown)


greghudson:
In the long run, we would like to support scenarios where long-term keys are held in TPM storage and are not directly accessible to libkrb5.  Adding a new public API which exposes a long-term key to the application would make this harder.  Also, the rd_req API surface (public and internal) is already complicated, and I would prefer not to make it more complicated.

After rd_req, the decoded and verified PAC is sitting in the authdata context.  A GSS application could retrieve it using name attributes.  Adding a public libkrb5 API to retrieve name attributes would be a better approach, I think.

We have also agreed in principle to adding pluggable kadmin ACL checking.  I apologize that we haven't made forward progress on that item, but I don't want to commit to supporting a public API to expose the ticket decryption key as a workaround.

gladiac:
What about just extending the krb5_rd_req_decoded() the symbol is public, I can use that for now?

At one point we need to start discussing about libkdc. I'm currently trying to get most of the stuff cleaned up and working probably we have to implement the KDC in Samba so we can support RODC in future.

greghudson:
> What about just extending the krb5_rd_req_decoded() the symbol is public, I can use that for now?

I don't like that change either.  I would never want to encourage an application to use a symbol that isn't part of the public API.  (Symbols are often in the library export list for use by other parts of the tree or test programs.  That isn't in any way a public commitment to not changing their ABIs or removing them.)

greghudson:
Wait, wouldn't it be better to validate the PAC using the KDC signature, not the signature that uses the service key?

gladiac:
We get the PAC from the ticket and decode it. The PAC contains a server signature which can be verfied using the service key. In this case the service case is the long term key. The key is stored in the keytab (DSDB accessed via the Samba KDB module). This key is used to decrypt the ticket. So to verify the signature I need this key. As krb5_rd_req() loads it from the keytab (KDB) the simplest way is to get it from there. I think I could probably do the same and search for the principal in the keytab and load the service key using krb5_kt_* functions, like decrypt_ticket() is doing it ...

Or are you saying it can be verified using a different key?

gladiac:
What is the KDC signature?

greghudson:
The PAC contains two signatures, one for the KDC and one for the service.  In the MS-PAC document you linked, see section 2.8.2 titled "KDC Signature".  Since you are implementing a KDC service, to me it makes sense to verify the PAC using the KDC signature, although I guess it doesn't matter a whole lot since kadmin/* should have the same privilege level as krbtgt/REALM.

I will also reiterate that I would be happy to add a public libkrb5 API to retrieve name attributes from the auth context after a krb5_rd_req()--basically a wrapper around krb5_authdata_get_attribute().  That API could be used to get at the validated PAC.

kaduk:
(Obligatory note that tickets issued for a KDC service don't actually require both signatures, so the code should be able to handle that as a possibility.  I forget whether our KDC implementation takes advantage of that potential optimization at present, though.)

greghudson:
kaduk: that applies to CAMMACs, not PACs, right?

kaduk:
greghudson: er, yes.  Sorry for reading too quickly.

PR #534, opened 2016-09-05
Make krb5_kt_register and krb5_kt_kdb_ops public

gladiac:
In Samba I've implemented a kpasswd service because we need to do ACL checking for incoming kpasswd requests. For the decryption of the packet I need the decryption key from the keytab. We have implemented a KDB module for samba. I'm able to load everything with kadm5_init() from kadm5srv_mit but then I need to register the KDB key table operations that I can use 'KDB:' as the keytable name. This way the samba KDB module for the realm which kadm5_init() set gets loaded.

### DIFF
--- a/src/clients/kinit/kinit_kdb.c
+++ b/src/clients/kinit/kinit_kdb.c
@@ -33,7 +33,7 @@
 
 #include <k5-int.h>
 #include <kadm5/admin.h>
-#include <kdb_kt.h>
+#include <kdb.h>
 #include "extern.h"
 
 /** Server handle */
@@ -66,7 +66,7 @@ kinit_kdb_init(krb5_context *pcontext, char *realm)
                         &server_handle);
     if (retval)
         return retval;
-    retval = krb5_kt_register(*pcontext, &krb5_kt_kdb_ops);
+    retval = krb5_db_register_keytab(*pcontext);
     return retval;
 }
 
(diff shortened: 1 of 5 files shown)


greghudson:
Can we instead add a libkdb5 function to register the KDB keytab, and keep krb5_kt_register() a private interface?  It would be strange to have a public API in libkrb5 which can only be used with a public pointer constant in libkdb5.

gladiac:
Are you fine with krb5_kt_register_kdb() or do you prefer a differnt name?

greghudson:
I think the name should be krb5_db_register_keytab(), since it's a libkdb5 function.

gladiac:
I've changed it.

greghudson:
Pushed to master as 2e99582062d9d6a70f2adb00fd8fe58a1f95b9b7 78cf203c0193c9746cc2835d18d66d4b76b2afcf.

Just to note here: this should be sufficient for a libkrb5 program like a kpasswd server, but a GSS server like kadmind will also need krb5_ktkdb_set_context().  Since that function doesn't have very useful threading semantics (it just sets a global variable, and the context it sets is then used by all threads), I think it's best to defer that problem until a need arises.

gladiac:
FYI: That's the reason why we have a gensec_krb5 module in addition to a gensec_gssapi module in Samba. gssapi creates a new krb5_context for each calls, so we can't register the HDB/KDB ops. It looks like that krb5_ktkdb_set_context() is the workaround for that issue in MIT.

PR #535, opened 2016-09-06
aes-sha2 support

greghudson:
It looks like the aes-sha2 RFC will be published soon enough for us to implement it in 1.15.  IANA has not assigned enctype and checksum type numbers yet, but supposedly will very soon.  The implementation is a big diff, so I am putting up the PR now for review.  I will update it with the assigned numbers as soon as they become available.


frozencemetery:
The first commit removes the build support of hash_{md4,md5,sha1}.c and replaces them with hash_evp.c, but does not actually delete these files.

greghudson:
Good catch; fixed.

frozencemetery:
Looks good now, thanks.

greghudson:
Updated with the assigned numbers.  (They may not be 100% official, but they appear in the public IANA registries and are unlikely to change.)

src/lib/crypto/krb/etypes.c
```diff
@@ -167,6 +167,27 @@ const struct krb5_keytypes krb5int_enctypes_list[] = {
       krb5int_dk_cmac_prf,
       CKSUMTYPE_CMAC_CAMELLIA256,
       0 /*flags */ },
+
+    { ENCTYPE_AES128_CTS_HMAC_SHA256_128,
+      "aes128-cts-hmac-sha256-128", { "aes128-cts-sha256" },
```
tlyu: I'm still looking over the rest of the changes, but I would like us to use shorter aliases than "aes128-cts-sha256"; maybe "aes128-sha2" and "aes256-sha2"?

greghudson: Done.  I also added (via a new commit) aes128-sha1 and aes256-sha1 aliases.

greghudson:
I realized the documentation still referred to the old alias names, so I fixed that up.  The documentation commit also adds the aes256-sha1 and aes128-sha1 aliases to the table.

greghudson:
Pushed to master as 6fd74a89ac6c2444a347a357fac51b3490467284 and preceding.

PR #536, opened 2016-09-06
Remove meaningless checks decoding DB2 principals

greghudson:
Commit e3d9f03a658e247dbb43cb345aa93a28782fd995 (ticket 8481) added
several checks for negative length values when decoding DB2 principal
entries, including two unnecessary checks on unsigned values.  Remove
those checks as they can generate warnings.


tlyu:
This looks ok to me.

PR #537, opened 2016-09-07
Fix unaligned accesses in bt_split.c

tlyu:
In the libdb2 btree back end, splitting a page at an overflow key
could result in an unaligned access, causing a crash (and data
corruption) on platforms with strict alignment.  This probably occurs
only rarely in practice.

Also includes regression test and debugging improvements.


greghudson:
This looks okay to me.

PR #538, opened 2016-09-12
Add getrandom() support

frozencemetery:
- First commit eliminates a codepath where we can Ouroboros urandom.
- Second commit adds support for getrandom() to both OS and Fortuna PRNGs.

One thing I did not do here (but could) is move the urandom socket reuse code from prng_os.c to prng.c so that Fortuna could take advantage of it.  I don't know how much of an advantage this will be in practice.  (Reuse of the /dev/random socket seems over-aggressive.)

I plan to backport this to 1.14, but as it adds features it is probably inappropriate for upstream 1.14-next.

### DIFF
--- a/src/lib/crypto/krb/prng.c
+++ b/src/lib/crypto/krb/prng.c
@@ -36,10 +36,12 @@ krb5_c_random_seed(krb5_context context, krb5_data *data)
 #if defined(_WIN32)
 
 krb5_boolean
-k5_get_os_entropy(unsigned char *buf, size_t len)
+k5_get_os_entropy(unsigned char *buf, size_t len, int strong)
 {
     krb5_boolean result;
     HCRYPTPROV provider;
+
+    /* CryptGenRandom is always considered strong. */
 
     if (!CryptAcquireContext(&provider, NULL, NULL, PROV_RSA_FULL,
                              CRYPT_VERIFYCONTEXT))
@@ -47,22 +49,6 @@ k5_get_os_entropy(unsigned char *buf, size_t len)
     result = CryptGenRandom(provider, len, buf);
     (void)CryptReleaseContext(provider, 0);
     return result;
-}
-
-krb5_error_code KRB5_CALLCONV
-krb5_c_random_os_entropy(krb5_context context, int strong, int *success)
-{
-    int oursuccess = 0;
-    char buf[1024];
-    krb5_data data = make_data(buf, sizeof(buf));
-
-    if (k5_get_os_entropy(buf, sizeof(buf)) &&
-        krb5_c_random_add_entropy(context, KRB5_C_RANDSOURCE_OSRAND,
-                                  &data) == 0)
-        oursuccess = 1;
-    if (success != NULL)
-        *success = oursuccess;
-    return 0;
 }
 
 #else /* not Windows */
@@ -107,44 +93,11 @@ cleanup:
 }
 
 krb5_boolean
-k5_get_os_entropy(unsigned char *buf, size_t len)
+k5_get_os_entropy(unsigned char *buf, size_t len, int strong)
 {
+    if (strong != 0)
+        return read_entropy_from_device("/dev/random", buf, len);
     return read_entropy_from_device("/dev/urandom", buf, len);
-}
-
-/* Read entropy from device and contribute it to the PRNG.  Returns true on
- * success. */
-static krb5_boolean
-add_entropy_from_device(krb5_context context, const char *device)
-{
-    krb5_data data;
-    unsigned char buf[64];
-
-    if (!read_entropy_from_device(device, buf, sizeof(buf)))
-        return FALSE;
-    data = make_data(buf, sizeof(buf));
-    return (krb5_c_random_add_entropy(context, KRB5_C_RANDSOURCE_OSRAND,
-                                      &data) == 0);
-}
-
-krb5_error_code KRB5_CALLCONV
-krb5_c_random_os_entropy(krb5_context context, int strong, int *success)
-{
-    int unused;
-    int *oursuccess = (success != NULL) ? success : &unused;
-
-    *oursuccess = 0;
-    /* If we are getting strong data then try that first.  We are
-       guaranteed to cause a reseed of some kind if strong is true and
-       we have both /dev/random and /dev/urandom.  We want the strong
-       data included in the reseed so we get it first.*/
-    if (strong) {
-        if (add_entropy_from_device(context, "/dev/random"))
-            *oursuccess = 1;
-    }
-    if (add_entropy_from_device(context, "/dev/urandom"))
-        *oursuccess = 1;
-    return 0;
 }
 
 #endif /* not Windows */
--- a/src/lib/crypto/krb/prng.c
+++ b/src/lib/crypto/krb/prng.c
@@ -59,8 +59,14 @@ k5_get_os_entropy(unsigned char *buf, size_t len, int strong)
 #include <sys/stat.h>
 #endif
 
-/* Open device, ensure that it is not a regular file, and read entropy.  Return
- * true on success, false on failure. */
+#ifdef __linux__
+#include <sys/syscall.h>
+#endif /* __linux__ */
+
+/*
+ * Open device, ensure that it is not a regular file, and read entropy.  Return
+ * true on success, false on failure.
+ */
 static krb5_boolean
 read_entropy_from_device(const char *device, unsigned char *buf, size_t len)
 {
@@ -95,6 +101,34 @@ cleanup:
 krb5_boolean
 k5_get_os_entropy(unsigned char *buf, size_t len, int strong)
 {
+#if defined(__linux__) && defined(SYS_getrandom)
+    while (len > 0) {
+        int ret;
+
+        /*
+         * Pull from the /dev/urandom pool, but ensure that it has been fully
+         * seeded.  This ensures strong randomness while only blocking during
+         * system bringup.
+         *
+         * glibc does not currently provide a binding for getrandom:
+         * https://sourceware.org/bugzilla/show_bug.cgi?id=17252
+         */
+        errno = 0;
+        ret = syscall(SYS_getrandom, buf, len, 0);
+        if (ret <= 0) {
+            if (errno == EINTR)
+                continue;
+
+            /* ENOSYS or other unrecoverable failure */
+            break;
+        }
+        len -= ret;
+        buf += ret;
+    }
+    if (len == 0)
+        return TRUE;
+#endif /* __linux__ && SYS_getrandom */
+
     if (strong != 0)
         return read_entropy_from_device("/dev/random", buf, len);
     return read_entropy_from_device("/dev/urandom", buf, len);
--- a/src/lib/crypto/krb/prng_fortuna.c
+++ b/src/lib/crypto/krb/prng_fortuna.c
@@ -366,7 +366,7 @@ k5_prng_init(void)
 #else
     last_pid = getpid();
 #endif
-    if (k5_get_os_entropy(osbuf, sizeof(osbuf))) {
+    if (k5_get_os_entropy(osbuf, sizeof(osbuf), 0)) {
         generator_reseed(&main_state, osbuf, sizeof(osbuf));
         have_entropy = TRUE;
     }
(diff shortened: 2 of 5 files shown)


src/lib/crypto/krb/prng_os.c
```diff
@@ -91,3 +91,11 @@ krb5_c_random_make_octets(krb5_context context, krb5_data *outdata)
     }
     return 0;
 }
+
+krb5_error_code KRB5_CALLCONV
+krb5_c_random_os_entropy(krb5_context context, int strong, int *success)
+{
+    /* The OS RNGs that accept input from userspace do not increase the
+     * entropy estimate based on this new entropy. */
```
greghudson: This comment confuses me.  krb5_c_random_os_entropy() doesn't contribute any entropy from the application (except the timing of the call, I guess), so there's no reason to expect an increase in the estimated entropy of the OS PRNG.

frozencemetery: In prng.c, krb5_c_random_os_entropy() calls add_entropy_from_device(), which first gets entropy from device in a call to read_entropy_from_device() and then passes it to krb5_c_random_add_entropy() which, in the prng_os.c version, writes it into urandom.

greghudson: I understand that, but that's just a harmless accident of the design when the OS PRNG is treated both as the source of seed entropy and the source of PRNG data.

It's perfectly okay to short-circuit that, but the comment doesn't make any sense; we weren't trying to increase the operating system's entropy estimate by reading from /dev/urandom (or /dev/random in some cases) and writing it back.

frozencemetery:
Deleted comment.  Sorry for the confusion.

greghudson:
As discussed by voice, the auxfd part of the new design seems inelegant, as is the burning of a file descriptor in the case where we wind up using SYS_getrandom.  We could get rid of both of these if we stopped reusing a file descriptor for /dev/urandom.

I ran some timing tests.  On my Linux laptop, it takes about 1.6us to read 16 bytes from /dev/urandom (or using SYS_getrandom), and about 2.6us to open-read-close.  That's a noticeable difference, but probably not worth worrying about for such a fast operation which happens in the midst of a bunch of other work.  On my ancient Solaris machine,  the impact is more severe: it takes about 13us to read from /dev/urandom and about 93us to open-read-close.  Everything else is slower on that machine too, but a seven-fold increase in the time taken to read from the PRNG device might still start to make it noticeable.

So, while I think it would be fine not to have fd reuse in the case where we fall back from SYS_getrandom to /dev/urandom, I'd like to preserve the reuse on systems with no SYS_getrandom at all.  That would argue for separate PRNG modules for the two cases, where the new module does no initialization and does nothing on krb5_c_random_add_entropy(), and calls k5_get_os_entropy() to service krb5_c_random_make_octets().

If we go this route we have to pick a name for the new PRNG module.  Annoyingly, "os" is the only good fit, which argues for renaming the current module to prng_urandom.c or prng_device.c.  That would probably be okay, given the small number of people we expect to be configuring non-default PRNG modules.

Other concerns:

* How portable is <sys/syscall.h>?  Right now we only use it in lock_file.c inside a conditional on Linux.

* Is it thread-safe to have an unprotected global variable whose value begins at 1 and can only ever be set to 0, given that it's not harmful to erroneously read the old value for a while?

* There's an unnecessary change to a comment at line 30 of prng_os.c.

frozencemetery:
That sounds good to me as long as we're okay with the duplication of reading logic (which is probably fine; it's only a handful of lines and unlikely to change).

Other possible names for the new module: "kernel", "system" (this is a bit close to "os", though it might not be a bad idea to name them both something different than "os" to avoid confusion); I don't have a preference between those three.

Regarding other concerns:
- sys/syscall.h exists on both Linux and BSD (I tested FreeBSD 10.3 which does not have SYS_getrandom at all; syscall() claims to date back to 4BSD).  We don't need to worry about Windows because that doesn't go through this path.  [Oracle docs](https://docs.oracle.com/cd/E19620-01/805-3175/6j31empjd/index.html) suggest that it is present on Solaris as well (tangent: Solaris seems to have both getrandom() and getentropy() support, though we never want the latter).
- It is thread safe since there's only one check-site per thread.  The worst case is we have multiple unnecessary trips through syscall() that fail.  I expect this is similar in cost to a mutex.
- If prng_os.c can use the getrandom() path as it can in that version, the comment is technically incorrect without the change since the bytes do not come from the device entry.  That said, if I read the rest of your comment correctly, it sounds like this will not be a concern for the final version.

greghudson:
The 8-ish lines of reading logic are already duplicated.  I don't object to de-duplicating them via a utility function in prng.c, but that change doesn't need to be in the scope of this PR.

frozencemetery:
Added new OS PRNG.  I'm willing to squash the top three commits as desired, but am not sure what the preferred separation is.

greghudson:
It occurs to me that syscall() is not type-safe and a getrandom system call is not standardized.  If another operating system introduced (or already had) a getrandom() system call with different parameters, we would fail unsafely.  So I think we want to conditionalize this support on __linux__ being defined as well as SYS_getrandom.

frozencemetery:
All getrandom() handling is now conditionalized on \__linux__.

src/lib/crypto/krb/prng_fortuna.c
```diff
@@ -443,4 +443,36 @@ krb5_c_random_make_octets(krb5_context context, krb5_data *outdata)
     return 0;
 }
 
+#ifdef _WIN32
+#define BUFLEN 1024
+#else
+#define BUFLEN 64
+#endif
```
greghudson: I see how this preserves the old behavior, but I don't like it.  I will talk through the changes with other team members tomorrow and see if we want to change this behavior (e.g. using a 64-byte buffer on Windows) or if maybe there's a simpler design.

src/lib/crypto/krb/prng_fortuna.c
```diff
@@ -443,4 +443,36 @@ krb5_c_random_make_octets(krb5_context context, krb5_data *outdata)
...
+
+krb5_error_code KRB5_CALLCONV
+krb5_c_random_os_entropy(krb5_context context, int strong, int *success)
+{
+    krb5_data data;
+    unsigned char buf[BUFLEN];
```
greghudson: Use uint8_t here.

src/lib/crypto/krb/prng_os.c
```diff
@@ -26,68 +26,49 @@
  */
 
 /*
- * This file implements a PRNG module which relies on the system's /dev/urandom
+ * This file implements a PRNG module which relies on the system's kernel
+ * PRNG, acccessed through getrandom() with fallback to the /dev/urandom
```
greghudson: This should be vaguer, since k5_os_getrandom() doesn't necessarily do what this describes (e.g. on Windows).  The comment at the top should also stay the same, I think.

src/lib/crypto/krb/prng_os.c
```diff
@@ -26,68 +26,49 @@
...
-        return errno;
     return 0;
 }
 
 void
 k5_prng_cleanup(void)
 {
-    close(fd);
-    fd = -1;
+    return;
```
greghudson: A function can be empty, right?  So we don't need "return" here.

src/lib/crypto/krb/prng_os.c
```diff
@@ -26,68 +26,49 @@
...
-        if (count == -1)
-            return errno;
-        buf += count;
-        len -= count;
-    }
+    res = k5_get_os_entropy(buf, len, 0);
+
+    if (res == FALSE)
+        return KRB5_CRYPTO_INTERNAL;
+    return 0;
```
greghudson: Use a trinary operator to simplify this.  Also, the blank line here doesn't really separate a code paragraph and should be removed.  Also, buf and len can be removed and their uses expanded out.

src/lib/crypto/krb/prng_fortuna.c
frozencemetery: Sounds good.  I'll wait the rest of the changes until it is decided.

frozencemetery:
No, no buildsystem checks happen on the result of prng-alg.  We just only build and link prng_$(ALG).c.  The only check that currently happens is whether it's fortuna or not.

I'm open to adding sanity checks to `--with-prng-alg`, but there aren't any right now.

tlyu:
Ah, so it's not any more broken than it already was. There are issues with being able to usefully run `make depend` in this sort of setup, but they already existed.

src/lib/crypto/krb/prng_os.c
```diff
@@ -0,0 +1,74 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* lib/crypto/krb/prng_os.c - OS kernel PRNG implementation */
+/*
+ * Copyright (C) 2016 by the Massachusetts Institute of Technology.
+ * All rights reserved.
+ *
+ *
+ * Export of this software from the United States of America may require
```
tlyu: License should be the BSD-style one from src/prototype/prototype.c.

frozencemetery:
Hopefully I got everything requested.  As per earlier conversation, I've moved to a 64-byte buffer on all platforms.

tlyu:
Also if we don't anticipate SYS_getrandom transitioning from functional to nonfunctional within a process lifetime, we could use pthread_once() to test it and set a variable that's read-only thereafter. This is probably less expensive than a mutex.

greghudson:
I think we currently only use pthread_once behind the library initializer framework, because we didn't know how to do pthread_once on Windows back when we designed our thread support.  Testing getrandom() could be done in a library initialization function.  libk5crypto's initialization function is crypto_libinit.c:cryptoint_initialize_library().  Currently that function calls k5_prng_init() which is implemented by the module.  Since we now want to do something in prng.c, I think we need to:

* Rename k5_prng_init() and k5_prng_cleanup() to k5_prng_module_init() and k5_prng_module_cleanup().

* Add a call to k5_prng_init() in cryptoint_initialize_library().

* In prng.c, add k5_prng_init() which tests getrandom() (on Linux with SYS_getrandom defined).  There is no need to add a generic prng.c cleanup function at this time.

frozencemetery:
I don't see why we need to do any initialization testing at all.  It seems much simpler to leave the current, known working design since:
- `use_getrandom` can only change from TRUE to FALSE.
- `use_getrandom` is only checked in one place, and only set in one place.
- if a thread has a stale `use_getrandom` value, the only negative consequence is an extraneous `syscall()` invocation.

tlyu:
We should try to avoid undefined behavior, and we should also try to satisfy static checkers. I recall that Coverity will complain about unsynchronized access to a variable if another thread might be able to modify it at the same time. The statement "use_getrandom can only change from TRUE to FALSE", isnot necessarily true, especially if writes to use_getrandom aren't atomic on some platform. (It's probably safe here if we interpret the construct as unspecified as opposed to undefined, but it's not safe in general.)

POSIX says in http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_11 that "Applications shall ensure that access to any memory location by more than one thread of control (threads or processes) is restricted such that no thread of control can read or modify a memory location while another thread of control may be modifying it.". Whether violating that "shall" results in undefined or unspecified behavior is unclear to me. http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap01.html#tag_21_01_15 weighs in slightly favor of "unspecified", but C99 says that violating a "shall" requirement outside a constraint is undefined behavior.

src/lib/crypto/krb/prng.c
```diff
@@ -95,6 +105,35 @@ read_entropy_from_device(const char *device, unsigned char *buf, size_t len)
 krb5_boolean
 k5_get_os_entropy(unsigned char *buf, size_t len, int strong)
 {
+#if defined(__linux__) && defined(SYS_getrandom)
+    while (use_getrandom == TRUE && len > 0) {
+        int ret;
+
+        /*
+         * Pull from the /dev/urandom pool, but ensure that it has ben fully
```
tlyu: Typo in comment; this should be "has been fully".

frozencemetery:
There are two possible states for `use_getrandom`: TRUE and not TRUE.  It doesn't matter whether writes are atomic: `use_getrandom` will always have one of those two states.

src/lib/crypto/krb/prng.c
frozencemetery: Will fix with next update.

tlyu:
If the behavior is unspecified, I would agree with you that there are two possible states. If the behavior is undefined, any number of unpredictable and/or catastrophic things might occur. I don't like making best-case assumptions about what might happen in undefined behavior, because history shows that implementations will eventually find some novel and surprising way to cause problems for you if you do.

I would like simpler behavior whenever possible. If we don't care about this small but dubiously thread-safe optimization, we can unconditionally call SYS_getrandom when it is defined and take a small penalty if the kernel doesn't implement it. If this optimization is important, we should try to make it actually thread-safe, probably using @greghudson's approach.

frozencemetery:
Given the choice between those two, I'd prefer to take the penalty when the kernel doesn't implement it since going forward we would expect more (\__linux__) kernels to have it than not.

kaduk:
> There are two possible states for use_getrandom: TRUE and not TRUE. It doesn't matter whether writes are atomic: use_getrandom will always have one of those two states.

Even if we set aside for the sake of argument the question of undefined behavior, this statement is still not true -- at least, for sane values of "not TRUE".  In particular, the bits that make up the object being accessed could be a trap representation during some non-atomic modification, and the trap resulting from an access to that state would definitely not be the desired behavior.

greghudson:
Trying getrandom unconditionally is fine with me; an invalid system call on Linux should be a small penalty.  (Also, the penalty only applies when the runtime platform uses an older kernel than the build-time platform, which should be unusual.)

frozencemetery:
- Made trying getrandom() unconditional on Linux
- Fixed typo in comment

greghudson:
Pushed to master as 0bbbc2bd3a42cfbd9e6eb34c273da8aaa077c29f 5e54525fbe40d56f44368e216c92938403cad96d a9a48392c088b53d8dd86b8008b4059ab78a3679 0be7642b2b6f7b9e0acebb2c3d60aa6c3f7543aa.

PR #539, opened 2016-09-14
Configuration option to increase the TCP/RPC connection limit

mrogers950:
This adds "kdc_tcp_max_conn" to the kdcdefaults section, and bumps the default limit (which is arbitrary like the original value) .

### DIFF
--- a/doc/admin/conf_files/kdc_conf.rst
+++ b/doc/admin/conf_files/kdc_conf.rst
@@ -43,7 +43,7 @@ The kdc.conf file may contain the following sections:
 [kdcdefaults]
 ~~~~~~~~~~~~~
 
-With one exception, relations in the [kdcdefaults] section specify
+With two exceptions, relations in the [kdcdefaults] section specify
 default values for realm variables, to be used if the [realms]
 subsection does not contain a relation for the tag.  See the
 :ref:`kdc_realms` section for the definitions of these relations.
@@ -60,6 +60,10 @@ subsection does not contain a relation for the tag.  See the
     Specifies the maximum packet size that can be sent over UDP.  The
     default value is 4096 bytes.
 
+**kdc_tcp_listen_backlog**
+    (Integer.)  Set the size of the listen() queue length for the KDC
+    daemon.  The value may be limited by OS settings.  The default value
+    is 5.
 
 .. _kdc_realms:
 
--- a/src/include/net-server.h
+++ b/src/include/net-server.h
@@ -67,7 +67,8 @@ krb5_error_code loop_add_rpc_service(int default_port, const char *addresses,
                                      void (*dispatchfn)());
 
 krb5_error_code loop_setup_network(verto_ctx *ctx, void *handle,
-                                   const char *progname);
+                                   const char *progname,
+                                   int tcp_listen_backlog);
 krb5_error_code loop_setup_signals(verto_ctx *ctx, void *handle,
                                    void (*reset)());
 void loop_free(verto_ctx *ctx);
--- a/src/kdc/main.c
+++ b/src/kdc/main.c
@@ -54,7 +54,8 @@ static void usage (char *);
 
 static krb5_error_code setup_sam (void);
 
-static void initialize_realms (krb5_context, int, char **);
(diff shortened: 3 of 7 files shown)


greghudson:
My takeaway from the meeting yesterday was that it was probably more important to change the listen() backlog than the number of pending connections awaiting data, but this PR seems to affect the second.  Do I remember wrong?

(I do not want to change the default value of the listen() backlog from 5, as it acts as a load-shedding mechanism in multi-KDC realms.)

mrogers950:
I don't think we wanted to mess with the listen() backlog until some data came out of load tests. This option should help with the testing, and for anyone else that happens to hit the limit.

tlyu:
My recollection matches Greg's. It's possible to increase the effective number of open connections by running worker processes (though there are some confounding factors). The working hypothesis that I recall nobody objecting to was that we are likely limited on how fast the KDC can call accept(), so changing the listen() backlog would be most likely to make an observable difference.

mrogers950:
Ok, thanks for the clarification. I've updated the option to set the listen() backlog.

src/kdc/main.c
```diff
@@ -955,6 +955,21 @@ int main(int argc, char **argv)
 
     initialize_kdc5_error_table();
 
+    retval = krb5_get_profile(kcontext, &profile);
```
greghudson: Since this file includes k5-int.h, you could skip this step and just use kcontext->profile.  However, you should instead be reading this in initialize_realms() adjacent to where kdc_max_dgram_reply_size is read.

src/include/net-server.h
```diff
@@ -33,6 +33,7 @@
 
 /* The delimeter characters supported by the addresses string. */
 #define ADDRESSES_DELIM ",; "
+#define DEFAULT_TCP_LISTEN_BACKLOG 5
```
greghudson: I think this belongs in include/osconf.hin.

doc/admin/conf_files/kdc_conf.rst
```diff
@@ -312,6 +313,11 @@ The following tags may be specified in a [realms] subsection:
     release 1.15 and later, it has the same meaning as
     **kdc_tcp_listen** if that relation is not defined.
 
+**kdc_tcp_listen_backlog**
```
greghudson: You are documenting this as a realm variable, when it is only read for [kdcdefaults].  Document it adjacent to kdc_max_dgram_reply_size in the kdcdefaults section.

src/kdc/main.c
mrogers950: Does this mean I should be reading it from the alternate profile like max_dgram_reply_size instead, or still use profile_get_integer() with kcontext->profile?

greghudson: Same way as max_dgram_reply_size is read; there is no categorical difference between the two parameters.  (Functionally, it makes no difference.)

mrogers950:
Thanks, I've updated.

doc/admin/conf_files/kdc_conf.rst
```diff
@@ -60,6 +60,10 @@ subsection does not contain a relation for the tag.  See the
     Specifies the maximum packet size that can be sent over UDP.  The
     default value is 4096 bytes.
 
+**kdc_tcp_listen_backlog**
+    (Integer.)  Set the size of the listen() queue length for the KDC
+    daemon. The value may be limited by OS settings. The default value
+    is 5.
```
greghudson: Above, "With one exception" needs to be changed to "with two exceptions."  Also, two spaces after periods.

src/kdc/main.c
```diff
@@ -65,6 +65,7 @@ static const char *pid_file = NULL;
 static int rkey_init_done = 0;
 static volatile int signal_received = 0;
 static volatile int sighup_received = 0;
+static int tcp_listen_backlog;
```
greghudson: I believe this could be put into struct socksetup.  There is no need to make it a global since it is only used within a call to loop_setup_network().

mrogers950:
Updated.

greghudson:
Pushed to master as bf1a0aeb6d0423ebaf4c3dd21fe2c51b894ccb38.

PR #540, opened 2016-09-17
Remove more non-DFSG documentation

tlyu:
Delete an Internet-Draft that we missed on an earlier pass.  Also
remove ISOC/BCP 78 copyright language because the remaining extracts
from RFCs are small enough to be fair use.

ticket: 8497 (new)


greghudson:
Looks good to me.

PR #541, opened 2016-09-20
Silence warnings in lib/rpc

tlyu:
Cast through (void *) in xdr.c to silence pointer target width
warnings.  In xdr_sizeof.c, use x_private and x_base as proper
pointers as in xdr_mem.c instead of trying to store an integer in a
pointer.


tlyu:
Also fixed another instance of setting x_private.

PR #542, opened 2016-09-22
Fix unlikely leak in KDC AS-REQ error path

greghudson:
In prepare_error_as(), if krb5_us_timeofday() fails and error pa-data
was supplied, the FAST cookie and a shallow copy of the error padata
can be leaked.  Reported by Will Fiveash.


tlyu:
Just checking: this isn't needed in 1.13 because the code that allocates the leaked memory isn't there, right? Otherwise, it looks ok to me.

greghudson:
That wasn't my reasoning, but it does happen to be true.

PR #545, opened 2016-09-29
SELinux label support

mrogers950:
I'm opening this PR to gauge interest in upstream carrying the SELinux code that is included in RHEL/Fedora. The purpose of the code is to set the correct SELinux context on files that krb5 processes create, which is based on the configured policy label (queried with selabel_lookup()) and the process user (queried with context_user_get()). Without it, the files get the default label inherited from the parent directory. 

These files include
- Stash files
- Primary ccache files
- krb5kdc/kadmin log files
- DB dump files
- pid files

I included some additional refactoring and a WIP test, and there will most likely need to be some more cleanup if we decide to carry it.


greghudson:
I looked over the patch.  It seems like:

* For most created files, the goal is to do an selabel_lookup() on the filename, then a setfscreatecon() based on the result (after some context manipulation functions), then the open call, then a setfscreatecon() back to the previous context.  My immediate question is: why does selinux impose such a huge burden on every application or library which creates files?  Why can't it do that itself?

* When we create a temporary file and move it into place, we want to do the selabel_lookup() with the destination name.  That seems like something selinux perhaps can't do itself, but there could certainly be an mkstemp wrapper to do it.

* Despite the use of a mutex, this code doesn't seem to be even a little bit thread-safe.  If two threads create a file at roughly the same time, we could wind up restoring the wrong context or applying the wrong context to a file.

mrogers950:
> For most created files, the goal is to do an selabel_lookup() on the filename, then a setfscreatecon() based on the result (after some context manipulation functions), then the open call, then a setfscreatecon() back to the previous context. My immediate question is: why does selinux impose such a huge burden on every application or library which creates files? Why can't it do that itself?

That's just by design. SElinux provides the libraries for interfacing with the policy store and the kernel components, and it's up to the application (or an associated userspace "object manager") to manage file labels. 

> When we create a temporary file and move it into place, we want to do the selabel_lookup() with the destination name. That seems like something selinux perhaps can't do itself, but there could certainly be an mkstemp wrapper to do it.
 
You may be right. It would be a good idea to confirm if rename() preserves the label or not.

> Despite the use of a mutex, this code doesn't seem to be even a little bit thread-safe. If two threads create a file at roughly the same time, we could wind up restoring the wrong context or applying the wrong context to a file.

It appears to me the mutex covers the context as well as the associated creation, so I'm missing something. What would we need to do to correct this?

greghudson:
> That's just by design.

I'm not sure we want to bear the weight of that design in the upstream code.  If there were library functions (not ours) to create files with the right context as determined by a lookup of the filename, I'd be much more inclined to support it.

> It appears to me the mutex covers the context as well as the associated creation, so I'm missing something. 

The locking makes more sense on a second read.  I didn't realize that the global lock was held across the whole push-open-pop operation.

mrogers950:
Before I closed out this PR I just wanted to confirm that you still wouldn't want to accept it with its current design. Unfortunately at the moment there are no SELinux library functions available that would be able to replace these file wrappers.

greghudson:
That's correct.  We can reevaluate in the future based on changes in external conditions, but I don't think I want this much complexity in our code base at the current time.

frozencemetery:
Understood, thanks @greghudson.  Should it ever be useful, I'm [carrying this patch](https://github.com/frozencemetery/krb5_fedora/blob/master/krb5-1.15.1-selinux-label.patch) in [my fedora repo](https://github.com/frozencemetery/krb5_fedora) (also viewable as a tree [here](https://github.com/frozencemetery/krb5/tree/rawhide).)

PR #550, opened 2016-10-05
Set alg param correctly for PKCS1

tlyu:
When using a smart card and constructing a DigestInfo to pass to the
CKM_RSA_PKCS mechanism, make sure to set the AlgorithmIdentifier
parameters correctly.  This is typically an ASN.1 NULL value.

Reported to Ubuntu in Launchpad #1629370.

ticket: 8506
target_version: 1.14-next
target_version: 1.13-next
tags: pullup

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -1319,8 +1319,7 @@ cms_signeddata_create(krb5_context context,
             alg = X509_ALGOR_new();
             if (alg == NULL)
                 goto cleanup2;
-            alg->algorithm = OBJ_nid2obj(NID_sha1);
-            alg->parameter = NULL;
+            X509_ALGOR_set0(alg, OBJ_nid2obj(NID_sha1), V_ASN1_NULL, NULL);
             alg_len = i2d_X509_ALGOR(alg, NULL);
             alg_buf = malloc(alg_len);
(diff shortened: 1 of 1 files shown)


tlyu:
I don't have an easy way to test this, but `make check` works for me on a system with OpenSSL but no soft-pkcs11.so.

greghudson:
It looks like X509_ALGOR_set_md was added partway through OpenSSL 1.0.x; it's not in 1.0.0t, for instance.

frozencemetery:
I was testing with 1.0.2j; t is also appears present in 1.0.1e.

greghudson:
Sounds like it was added in 1.0.1.  For krb5 1.15, I don't mind depending on OpenSSL 1.0.1, but at least for the backported change we should probably take the more conservative approach of creating the null parameter value ourselves.

frozencemetery:
Agreed, openssl-1.0.1 is a fine place to draw the line for me, though we should bump the OPENSSL_VERSION_NUMBER check in configure.ac if that's the route we take.

tlyu:
It looks like we might be able to use `X509_ALGOR_set0()` on 1.0.0?

greghudson:
That function exists in 0.9 (at least 0.9.8zh), so it should be safe to use.

tlyu:
Updated to use `X509_ALGOR_set0()`. A separate pull request will refactor the DigestInfo encoding into a helper function of `pkinit_signdata_pkcs11()`.

PR #551, opened 2016-10-05
Fix "make depend" when cmocka.h not present

greghudson:
Add an intermediary header file k5-cmocka.h, which only includes
cmocka.h if we detected an appropriate version of cmocka at configure
time.  This allows "make depend" to successfully run over cmocka test
programs when cmocka.h isn't present on the build platform.

For convenience, k5-cmocka.h also includes stdarg.h, stddef.h, and
setjmp.h, which are required to include cmocka.h.

[Tested with and without cmocka.h present.  Something seems to have added gssrpc/types.h to the dependencies of a bunch of files recently, so there is more churn than I would have expected in the "make depend" commit.  I can't figure out what it was, but it's consistent across my 14.04 and 16.04 machines.]


greghudson:
I needed to include autoconf.h in k5-cmocka.h since it's using an autoconf symbol.  Fixed now.

PR #553, opened 2016-10-06
Fix execl DPRINTF to show kprop_port when present

frozencemetery:
Previously, if kprop_port was set, the flags logged would not match the
actual invocation of kprop.


greghudson:
Pushed to master as 29dc209a135bae4e15d328348c2dab73e1f2afce.

PR #555, opened 2016-10-07
Adjust case rules for Discovery URI parsing

mrogers950:
An updated version of draft-mccallum-kitten-krb-service-discovery will clarify matters of case-sensitivity and adds IANA name spaces for the flags and transport fields. Since the literal IANA values registered will be lower-case, we must match as such. 

BCP 35 recommends that the scheme name, while registered as lower-case it should be case-insensitive, so I added a comment there.


greghudson:
Uh, why are we making the flags and transport types case-sensitive?  I don't understand.

mrogers950:
From feedback I received on draft-mccallum-kitten-krb-service-discovery-03, regarding defining the case-sensitivity of fields separately, suggested that it would be best to specify the whole URI as case-sensitive instead (RFC 3986 also suggests generic syntax components are case-sensitive).  I take that as not normalizing the fields to when we match for known flags or transport types, and the initial flags and transport type values will be registered as lower case values.

mrogers950:
This jumps the gun a bit on the draft discussions, sorry about that. We can hold off for now and I'll bring this topic up on the WG.

mrogers950:
This can be closed, we're sticking with the current behavior after WG discussions.

PR #556, opened 2016-10-17
Update README for krb5-1.15

tlyu:


README
```diff
@@ -73,12 +73,123 @@ from using single-DES cryptosystems.  Among these is a configuration
...
-Major changes in 1.14
+Major changes in 1.15
 ---------------------
 
-krb5-1.14 changes by ticket ID
+Administrator experience:
+
+* Add support to kadmin for remote extraction of current keys without
+  changing them (requires a special kadmin permission), with the
+  exception of highly protected keys.
```
greghudson: I'd split this out into two items, since the lockdown_keys attribute fulfills a larger purpose than just disabling the new extraction feature.  For the second one, I suggest "Add a lockdown_keys principal attribute to prevent retrieval of the principal's keys (old or new) via the kadmin protocol.  In new databases, this attribute is set on the krbtgt and kadmin principals."

README
```diff
@@ -73,12 +73,123 @@ from using single-DES cryptosystems.  Among these is a configuration
...
+
+* Restore recursive dump capability for DB2 back end, so sites can
+  more easily recover from database corruption resulting from power
+  failure events.
+
+* Add DNS auto-discovery of HTTP proxy by URI records.
```
greghudson: I'd flesh this out: "Add DNS auto-discovery of KDC and kpasswd servers from URI records, in addition to SRV records.  URI records can convey TCP and UDP servers and master KDC status in a single DNS lookup, and can also point to HTTPS proxy servers."

PR #559, opened 2016-10-24
DNSSEC-secured host-to-realm lookups with libunbound

vanrein:
This patch implements draft-vanrein-dnstxt-krb1-09 by introducing a backend for the hostrealm API based on libunbound.  It was developed against krb5-1.15 and then rebased to the master branch where I compiled and tested it again.  If so desired, I can also produce a PR against 1.15.

The extra backend is on by default, and can be disabled with `--disable-unbound-for-realm` if so desired.  It can completely replace the DNS/libresolv variant because it also implements the fallback and default calls, opportunistically using DNSSEC where it is present.  The secure hostrealm lookup on the other hand, will reject any unsigned response.

I can see only one potential problem to this patch, and that is slow-down of KDC performance if it is asked to look in DNS; the additional processing time for DNSSEC can take up to hundreds of milliseconds.

I am happy to donate something concrete and useful to the incredible MIT krb5 project!


greghudson:
Can you remove the six 1.15 commits from Tom from the PR?

vanrein:
Hi Greg,

>
> Can you remove the six 1.15 commits from Tom from the PR?


Done, the diff was a bit confusing to me too.

I had to tear down the branch and create it anew, so this has ended up
in a new Pull Request.

By the way, if the ISE asks me for suggestions for a 2nd reviewer for
the _kerberos TXT Internet Draft, might I suggest your name?  Nico
Williams has already made a detailed review, and I have processed his
comments.

Thanks,
 -Rick

kaduk:
> > Can you remove the six 1.15 commits from Tom from the PR?
> Done, the diff was a bit confusing to me too.

> I had to tear down the branch and create it anew, so this has ended up
in a new Pull Request.

You should be able to use "git rebase -i origin/master" to selectively remove those commits without making a new branch, FWIW.

PR #560, opened 2016-10-24
DNSSEC-secured host-to-realm lookups with libunbound

vanrein:
This patch implements draft-vanrein-dnstxt-krb1-09 by introducing a backend for the hostrealm API based on libunbound. It was developed against krb5-1.15 and then rebased to the master branch where I compiled and tested it again. If so desired, I can also produce a PR against 1.15.

The extra backend is on by default, and can be disabled with `--disable-unbound-for-realm` if so desired. It can completely replace the DNS/libresolv variant because it also implements the fallback and default calls, opportunistically using DNSSEC where it is present. The secure hostrealm lookup on the other hand, will reject any unsigned response.

I can see only one potential problem to this patch, and that is slow-down of KDC performance if it is asked to look in DNS; the additional processing time for DNSSEC can take up to hundreds of milliseconds.

I am happy to donate something concrete and useful to the incredible MIT krb5 project!


frozencemetery:
Hi @vanrein, in order for krb5 to merge this patch, the CI builds (Travis and Appveyor) do need to pass.  It looks like they are not because unbound isn't being linked in properly (and I don't think is available on Windows).  I'm sure there will be more comments but CI needs to pass first.

Regarding style, MIT prefers four spaces for indentation with no hard tabs.  You can run the command `make reindent` (if you have emacs installed) to see what changes this would cause (though if editing an existing file, the existing file's style is to be preferred).  Please make changes in the form of force-pushes (with rebased history) to this branch rather than opening a new pull request.

Thanks!

vanrein:
I updated the Windows building code, and now AppVeyor is happy.

The reason Travis fails is that it assumes libunbound in its environment (which I setup as a default) and it doesn't find it.  Please let me know how you want to proceed:
 1. either you add libunbound to the environment (just like libresolv is, my model for making it a default)
 2. or I disable libunbound by default
 3. or I try to autodetect libunbound and don't use it when the environment doesn't offer it.

This is for your team to decide, I suppose.  I think 3 is the most confusing option, and would prefer 1 or 2.

greghudson:
When changes are requested, please do not add them in the form of additional commits.  Change the existing commits (or commit, in this case) and force-push the branch.  (Not all projects work like this, but ours does.)

We don't generally add hard dependencies from one krb5 release to the next.  Of the three options you presented, option 1 would be mostly unprecedented, while there is precedent for option 2 (e.g. the LDAP KDB module, which is disabled by default) and option 3 (e.g. PKINIT, which is enabled when OpenSSL is auto-detected).  I know that not everyone in the wider open-source community agrees with auto-detection of third-party libraries as a practice.

Another concern is that we have traditionally been guarded in what external libraries we make libkrb5 depend on.  I'm not sure there is a good general reason for that other than licensing, which shouldn't be an issue here (libunbound has a BSD-like license).  If we do think of a reason, an alternative approach is to make this a dynamically loaded module like k5tls.  Hold off on making that change unless we come up with a good reason for it.

Finally, there's the question of whether libunbound is the best way to make secure DNS queries.  I don't have enough expertise in the area to know the answer to that.  I will try to discuss this and the other two issues with some people later today.

greghudson:
Please omit the addition of khostrealm from this PR.  It can be considered in a separate PR.

(At a minimum, a change like that should be in a separate commit.  But in this case, there is enough to consider about the use of libunbound without also worrying about adding a new user-facing command.)

greghudson:
I talked about this PR with other team members and people who work on downstream distributions of krb5 in packages.  The feedback was:

* An alternate approach to using libunbound is to use the libresolv APIs and look at the AD bit.  This is discussed here: https://sourceware.org/glibc/wiki/DNSSEC .  I don't fully understand the considerations, but I am concerned about adding a user-visible dependency on libunbound via the default trust root.  If some operation system integrates DNSSEC system-wide using another implementation (like getdns), we will have difficulty integrating with that if we advertise /etc/unbound/root.key as the default trust root.

* If we do use libunbound, auto-detecting it by default is fine, as long as we have configure flags to force the behavior on or off.  (If the behavior is forced on and libunbound isn't detected, we should fail at configure time.)

* If we do use libunbound, two downstream distributors strongly preferred using it via a dynamically loaded module rather than making libkrb5 depend directly on libunbound.  This is due to considerations around making small images (such as install images) containing a minimum of packages.

vanrein:
Hello Robbie,

> Hi @vanrein, in order for krb5 to merge this patch, the CI builds (Travis and Appveyor) do need to pass.

OK.

> It looks like they are not because unbound isn't being linked in properly

Just let me know how you would like it done.  I'm not very fluent in
autoconf tooling though, so anything there would be problematic for me.

> (and I don't think is available on Windows).

It is:

https://unbound.net/documentation/unbound-windows-manual-02.pdf

In addition, we've got an MXE extension (soon to be donated back to the
MXE.cc project) that adds Unbound,

https://github.com/arpa2/mxe/blob/master/src/unbound.mk

Also, the dependency can be switched off with --disable-unbound-for-realm

> I'm sure there will be more comments but CI needs to pass first.

Hit me :)

> Regarding style, MIT prefers four spaces for indentation with no hard tabs.  You can run the command `make reindent` (if you have emacs installed) to see what changes this would cause (though if editing an existing file, the existing file's style is to be preferred).

I will change that.

> Please make changes in the form of force-pushes (with rebased history) to this branch rather than opening a new pull request.

The new PR was necessary because I had to move back in history, which
Git / GitHub wouldn't let me do, but usually I just append to the branch
that I set aside just for this pull request.


Thanks,
 -Rick

greghudson:
> The new PR was necessary because I had to move back in history, which
Git / GitHub wouldn't let me do

Sure it will.  We do it all the time.  Just make sure to do a force-push, like "git push myrepo +branch".

vanrein:
Hello Greg,

I picked libunbound with a fair amount of care; it is a mature and stable product, whereas GetDNS is much younger.  The only serious contender might be BIND, but I am not surprised that nobody suggested that.

Note that using libunbound is an additional choice for the hostrealm API, and libresolv can be used just as easily.  Disable Unbound it during configuration, and the dependency is gone.  If the feedback should be satisfied in the default configuration, then the Unbound backend can be disabled by default and nobody would notice the addition.

And of course I looked into libresolv with AD bit.  There is no API to make this work, let alone a portable one.  Also, the AD bit relies on another component for validation of DNSSEC information.  Unless the resolver is run on the KDC host, it introduces a lot of senstivity to security assaults; that the libunbound approach evades structurally; this is a deliberate decision made in this patch.

It's really easy to have another definition for the default trust root, do you have any suggestions to make it more flexible (than the current profile variable in libkrb5.conf)?  If we want to resurrect the _kerberos TXT mechanism, which is what this patch wants to do, then we will need some place for this root key.

As for making the smallest possible packages: when disabling Unbound during ./configure, it is reduced to an entry in the hostrealm sequence with an initvt that sets up nothing.  If you like, I can take even that out with a #ifdef (thus making it differ from the currently present modules) but it would not aid code readability -- so I'd advise against it.

I will split off the khostrealm program into a separate patch, and learn how to --force git-pushes.  NP, just never used it before.

I think this set would satisfy the concerns that you picked up; do you agree?

-Rick

greghudson:
> As for making the smallest possible packages: when disabling Unbound during ./configure, it is reduced to ...

These downstreams want to build krb5 with all the features, but then slice it up into packages with minimal dependencies.  A dynamically loaded hostrealm module fits this need better than having libkrb5 link against libunbound directly.  Take a look at how k5tls is built and loaded to see how to make this a dynamically loaded module.

> It's really easy to have another definition for the default trust root, ...

The question is whether the application should be making this choice or the OS.  I'm wary of entering into a world where each DNS-using application makes a separate choice about what library to use and therefore how the user configures the DNSSEC trust root.  I agree that the platforms are making glacial progress on this front, so it's a frustrating situation.

vanrein:
Ah yes, separate modules, that sounds like the RedHat's and Debians of this world -- it is a clear request and I will look into k5tls.

Glaciers actually move quite fast these days :'-( but remarks of GetDNS are indicative of the problem "there's a new one, let's first sit and wait what it is going to do", in other words waiting until everything is stable -- which will not happen IMHO.  So I believe choosing a mature and stable library is a relatively safe choice, and Unbound most certainly is not a fad library.  That way, we're at least moving forward.  In the end, most products lean towards zone file format anyway, so they ought to be reasonably compatible.

I will find me some time to make these changes.  Thanks for the input.

vanrein:
Thanks for the feedback on my initial patch submission.  I had to get used
to your style of using Git, but I do like the way you are using it.

I have made the modifications to this patch to accommodate the requests:

 1. I made it into a dynamically loaded plugin library
 2. The Unbound dependency is auto-detected at configuration time
 3. The khostrealm program has been separated out (later patch will follow)
 4. I updated the branch with all modifications in one force-pushed patch
 5. The library is generally named k5dnssec.so; it could also house getdns/other impl

Note that a few remarks can be made:

 1. The removed khostrealm program has been useful for testing this patch
 2. For realm crossover, the KDC would load k5dnssec.so through libkrb5.conf
 3. When KDC SRV lookup is made pluggable, libunbound can reuse its cache
 4. The setup always builds k5dnssec.so -- even when DNSSEC_IMPL=no
 5. Nature of config, I could not help but causing many changes to `deps` files

If you are happy with these modifications, then please include this patch.

greghudson:
What change caused all the deps changes?  (This would be easier to see if the "make depend" were in a separate commit.)

In DNSSEC, what are the provisions for rolling over the root key?

vanrein:
I'm really only applying the "deps" process as an outsider... but my guess was either the tree has fallen behind, or there might be something configuration-dependent in the deps files?  I can recover the old deps files (and only leave new ones) if you want; I'm not sure what's desired here.  I noticed that "gssrpc" occurs a lot in the differences, which I certainly didn't touch.

Root key rolling in DNSSEC is rare, but is arranged by RFC 5011.  Software using DNSSEC is either [open to such changes](http://jpmens.net/2015/01/21/opendnssec-rfc-5011-bind-and-unbound/) or relies on an external component -- such as [autotrust](https://www.nlnetlabs.nl/projects/autotrust/) or, seeing the slow pace of rolling this key, distro upgrades.  I have not included this yet, but might add a few lines and configuration options for it -- and make them specific to unbound, as I've done for the root key file already.  Is that your suggestion?

vanrein:
Yes indeed, it's the addition

    $(BUILDTOP)/include/gssrpc/types.h

that shows up in most of the unexpected `deps` files.  I'm not sure why this is, maybe the project tree is behind on that one in this patch's predecessor `krb5-1.15-beta1`?

greghudson:
Hm, I don't see any changes made by "make depend" in the current master.  It could be churn caused by a different toolchain, although we make an attempt to control for that.

I don't have a specific suggestion for root key rollover.  I'm just wondering what would be required of a krb5 administrator if they are using this module with a manually populated root key file.  It sounds like they would need to pick up the new key before the old one is rotated out.

vanrein:
Since we aren't really sure of the origin of the many ``deps`` files, I've removed the overzealous changes manually.  The patch should now be the minimalistic patch that you're looking for.

Please review for inclusion, and do let me know if there are any further questions or concerns.

greghudson:
My concern about visibly depending on libunbound remains.  One possible way forward is to require the admin to configure dnssec_trust_anchor (with no "unbound" in the variable name) and disable the feature if it is not configured.

The KDC is single-threaded, and doing lookups to arbitrary places in the DNS could cause it to halt processing of all queries for multiple seconds.  I am not sure that this feature makes sense on the KDC unless we amend the hostrealm interface to support asynchronous lookup (which would not be easy).  But it could still make sense on clients.

Smaller concerns:

* This is a plugin module for the hostrealm interface, not the "dnssec" interface.  It should live in plugins/hostrealm.

* The module's initvt function should not do anything but initialize the vtable.  Checking the profile variable should happen in the init method.

* I don't think we want the --with-dnssec-impl goop.  (I realize in hindsight that this was a consequence of suggesting the k5tls module as a reference.)  Until and unless we have multiple implementations of a dnssec hostrealm module, we can just make it a module that we build if libunbound is found, and don't build if it isn't.

* ubctx can't be a global variable; the init method is per-krb5_context, not per-process.  Use the moddata pointer.

* If the trust anchor isn't present, will ub_ctx_add_ta_autr() fail?  I wouldn't want the module to perform DNS lookups which can never succeed because we have no trust anchor.

* Is the fallback method redundant with the existing TXT record support?  Should turning on dnssec_lookup_realm imply doing insecure DNS fallback lookups?  I would think not.

PR #562, opened 2016-10-28
Add doxygen comments for RFC 8009, RFC 4757

greghudson:
The aes-sha2 specification has been published as RFC 8009.  Add
Doxygen comments to the #defines for its enctype and checksum type
comments.  Also add comments for the RC4 enctype and checksum type
constants referring to RFC 4757.


kaduk:
LGTM.

greghudson:
I don't think the RC4 annotations are important enough to backport by themselves, since 1.15 will become the -latest documentation in a month or so anyway.

PR #563, opened 2016-10-28
Allow lowercase realms in a UPN SAN

mrogers950:
These changes allow lowercase realms in the MS UPN subjectAltName, and verifies the UPN with krb5_db_get_principal() (with flags to 0; do we want to canonicalize or accept aliases?) before matching. 
The validation with krb5_db_get_principal() is perhaps something we would want to do for non-UPN SANs as well.

There's a new test and since I added a UPN cert to make-certs.sh, I ran the script and updated the testing certs.


greghudson:
Why is it necessary to convert the realm to uppercase?  Why isn't a DB lookup sufficient by itself?

Also, this code leaks memory if the comparison is succesful.

mrogers950:
I was seeing that the principal could not be found unless I converted the realm. Can you clarify about the leak? If the comparison is succesful, there is a goto out, where I added a krb5_db_free_principal().

greghudson:
I missed the krb5_db_free_principal() in the out label, but then isn't there a double-free if the comparison isn't succesful?

mrogers950:
Yes, I believe you're right. I just pushed an update that should prevent that.

greghudson:
I'd like to discuss the realm uppercasing on a dev call.

We should not make the PKINIT shared object dependent on libkdb5, because the PKINIT object is used on clients as well as KDCs, and clients should not need to have libkdb5 installed.  We will need to create a kdcpreauth callback for this change.  (That was one of the options discussed in the last dev call.  At the time, I didn't think about the dependency issues and advocated for the simpler approach of doing the DB lookup inside PKINIT; I was mistaken.)

frozencemetery:
> clients should not need to have libkdb5 installed.

Much as I'd like that to be the case, they currently do.  kinit links to
it directly.

greghudson:
That's a good point.  However, I can imagine ways to change that (creating a keytab pluggable interface and making the KDB keytab a dynamically loaded plugin), so I don't want to make the situation worse.

mrogers950:
Just pushing what I have at the moment prior to the meeting. The callback may be badly named, as it still has UPN specific behavior in it. That being matching the UPN value of 'admin@example.com' to an alias like 'admin@example.com@EXAMPLE.COM', which on lookup finds this alias and sets the principal's realm to 'example.com', then matches it with the casefold flag.  This is probably wrong in both the alias and the match. What alias should be allowed to match the UPN and what restrictions would need to be placed on it if any? I also revised the test with this scenario using the test kdb module as suggested.

greghudson:
After another voice conversation, I think there are a couple of problems with the existing PKINIT code which are getting in the way of this PR:

1. crypto_retrieve_X509_sans() parses the text of id_ms_san_upn fields as if they were regular principal strings; it does not use krb5_parse_name_flags() with KRB5_PRINCIPAL_PARSE_ENTERPRISE.  (This code predates krb5_parse_name_flags().)  It's hard to correctly do a canonicalizing DB lookup on this result.

2. verify_client_san() matches against request->client, which is uncanonicalized.  So if an AS request is made using any kind of alias, the current code expects the client cert to match the alias, not the canonical name.

I think we want to fix crypto_retrieve_X509_sans() to return enterprise principals for UPNs.  Doing this naively could run the risk of breaking compatibility with existing deployments, as people might have PKINIT client certs with UPN fields which textually match principals in their DB2 databases.  The following match_client callback behavior might be a reasonable way to maintain compatibility and save some DB lookups:

1. If princ directly matches rock->request->client, return true.  (This saves a DB lookup if the AS request uses a UPN and the client cert uses the same UPN.)
2. If princ directly matches rock->client->princ, return true.  (This saves a DB lookup.)
3. If princ is a UPN and rock->request->client is not, unparse princ and re-parse it as a regular principal.  If that matches rock->request->client, return true.  (This maintains backward compatibility.)
4. Otherwise look up princ in the DB and match the result against rock->client->princ.

We should no longer need a "upn" flag in the signature of match_client, because the name type of the principal should disambiguate.

mrogers950:
I've updated. After coding these changes and debugging with the added tests, I noticed krb5_principal_compare() doesn't compare the types of the two principals, so the backwards-compatible
match outlined in step 3 is accomplished by the initial UPN->request princ compare eliminating the need for a re-parse. 
I included the change to SAN parsing, as well as a change to the test kdb module to allow returning an alias for canonical requests.

greghudson:
>  I noticed krb5_principal_compare() doesn't compare the types of the two principals, so the backwards-compatible match outlined in step 3 is accomplished by the initial UPN->request princ compare eliminating the need for a re-parse.

That doesn't sound right to me.  A parsed UPN has one component containing "abc@def", and no realm.  The same string parsed as a principal has one component containing "abc", and "def" in the realm.  They should not compare equal with krb5_principal_compare().  In the test labeled "Backwards-compatible UPN match," match_client() only succeeds because of the DB alias entry (I verified this using gdb).

I do note that we have krb5_principal_compare_flags() and KRB5_PRINCIPAL_COMPARE_ENTERPRISE, which we can use instead of reparsing the UPN ourselves.

match_client() should pass KRB5_KDB_FLAG_ALIAS_OK, not KRB5_KDB_FLAG_CANONICALIZE, when doing the DB lookup.  See the documentation in kdb.h for get_principal and also note the code in ldap_principal2.c at line 192.  With this change, the test KDB module should not need to be altered for the tests.

In the tests:

* Won't "make check" fail after the second commit and before the third commit, because the new certs aren't present?  This is not okay.  Reorder things so that make-certs.sh change and the new certs are in the second commit, and the new t_pkinit.py tests are in the third.

* It's probably simpler to just use "realm" instead of "tdbrealm" for the new tests; that's the convention in other test scripts which serially creates realms.

* The tests should not assume that realm.host_princ is "host/localhost@KRBTEST.COM"; this is why the Travis tests are failing.  Either use hostname (a global variable set by k5test.py) in testprincs, or use a fixed server principal name for kvno in these tests.  I'm also okay with just skipping the klist and kvno parts in these tests.

* Why set +preauth on the server principal?

* Why set pkinit_indicator in aliaskdcconf?  Also, this should be alias_kdc_conf for consistency.

mrogers950:
>That doesn't sound right to me. A parsed UPN has one component containing "abc@def", and no realm. The same string parsed as a principal has one component containing "abc", and "def" in the realm. They should not compare equal with krb5_principal_compare(). In the test labeled "Backwards-compatible UPN match," match_client() only succeeds because of the DB alias entry (I verified this using gdb).

The existing UPN matching code can only successfully match a UPN string that does not contain '@def' because of the crypto_retrieve_X509_sans() parsing results, so I made the test as UPN 'user' vs. principal 'user'. I checked it again and it succeeds with the first krb5_principal_compare (3rd command in the verbose output - you may have been looking at the second test, which is UPN 'user@krbtest.com' vs. 'user' and matches after the db lookup). crypto_retrieve_X509_sans() now parses to the principal 'user' with realm 'KRBTEST.COM' and type KRB5_NT_ENTERPRISE_PRINCIPAL, and that ends up as a match against the KRB5_NT_PRINCIPAL req_princ.

If krb5_principal_compare was strict about the principal type, only then it seems would there be the situation where princ is UPN and request/client is not and happens to match after a re-parse back to a regular principal. I coded this in initially but the post-parse match never succeeded in any of the tests.

>match_client() should pass KRB5_KDB_FLAG_ALIAS_OK, not KRB5_KDB_FLAG_CANONICALIZE, when doing the DB lookup. See the documentation in kdb.h for get_principal and also note the code in ldap_principal2.c at line 192. With this change, the test KDB module should not need to be altered for the tests.

Changed, thanks for pointing this out, I misunderstood the differences between the two flags.

> In the tests:

I corrected these items as well.

greghudson:
> The existing UPN matching code can only successfully match a UPN string that does not contain '@def' because of the crypto_retrieve_X509_sans() parsing results

I don't understand.  If a cert has a UPN SAN containing "user@KRBTEST.COM", won't current code match that successfully against user@KRBTEST.COM in the request client?  That's the backward-compatibility case I was concerned about preserving.

(However, I was wrong on two other counts: KRB5_PRINCIPAL_PARSE_ENTERPRISE does typically include the default realm in the result, whereas I said a parsed UPN didn't have a realm; and I was indeed mistakenly stepping through execution of the second test case rather than the first one.)

greghudson:
The kdcpreauth_plugin.h changes don't have all the correct changes for adding the first new callbacks for 1.16.  See 7b12eb4757f8dd05b79c9b49d4289f0caf1f6eec for an example of what should change.

mrogers950:
> I don't understand. If a cert has a UPN SAN containing "user@KRBTEST.COM", won't current code match that successfully against user@KRBTEST.COM in the request client? That's the backward-compatibility case I was concerned about preserving.

Ok, I had missed that UPN form. I added the enterprise krb5_principal_compare_flags() call and a test to confirm, and updated version information for the callback.

mrogers950:
My change to the version number in the callbacks structure was left out of yesterday's update, so I've corrected that.

mrogers950:
I've condensed the compares against req_princ down to a single krb5_principal_compare_flags() call.

greghudson:
I referred Matt to commit 7b12eb4757f8dd05b79c9b49d4289f0caf1f6eec for the changes needed to introduce a new callback version, but that commit contains a small error; it should not have changed the documented minor version of the kdcpreauth interface (which remains at 2, per kdc_preauth.c:get_plugin_vtables()).  I will submit a fix.  In this PR, just remove the change to kdcpreauth_plugin.h line 37.

mrogers950:
> In this PR, just remove the change to kdcpreauth_plugin.h line 37.

Done.

greghudson:
I pushed some updates.  Please have a look at the third commit in particular.

* In the first commit, I use match_client for both kinds of SANs, as id-pkinit-san should work against the post-canonicalized user principal, and should work if the id-pkinit-san is an alias.  Also, the code in the PR was using retval (a krb5_error_code) to hold the boolean result of the match_client call; I fixed that by directly checking the function return value like the old code did.

* In the second commit, I added user-upn3.csr to the cleanup.

* In the third commit, I removed the ./responder invocations from the tests, as they were redundant at best (for testing the KDC matching logic) and mismatched to the kinit tests in several cases.  I added comments for each test.  I simplified some of the kinit invocations to use the -E flag instead of weird-looking principal strings.  I added a test for an id-pkinit-san match against a canonicalized principal.

The mismatch test still seems a little weird to me.  Why create a DB entry for user2\@krbtest.com for the mismatch test, instead of using just "user2"?

greghudson:
Side note: the first commit creates a slightly long line in crypto_retrieve_X509_sans(), which cannot easily be corrected with the current code structure.  I would like to de-indent that function by a level by returning from the "if ((l = ..." test if it doesn't produce a non-negative result.  I can file a PR for that change after merging this one.

mrogers950:
I'm good with those changes. And there's no specific reason that the mismatch test used user2@krbtest.com other than it was similar to the user alias, from when I was initially creating the tests. It can be simplified to user2 and the test remains the same.

PR #564, opened 2016-10-31
Add KRB5_TRACE calls for DNS lookups

mrogers950:
To help admins troubleshoot DNS records without resorting to strace/gdb.

### DIFF
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -155,6 +155,20 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
     TRACE(c, "ccselect choosing default cache {ccache} for server " \
           "principal {princ}", cache, server)
 
+#define TRACE_DNS_SRV_ANS(c, host, port, prio, weight)                \
+    TRACE(c, "SRV answer: {int} {int} {int} \"{str}\"", prio, weight, \
+          port, host)
+#define TRACE_DNS_SRV_NOTFOUND(c)               \
+    TRACE(c, "No SRV records found")
+#define TRACE_DNS_SRV_SEND(c, domain)                   \
+    TRACE(c, "Sending DNS SRV query for {str}", domain)
+#define TRACE_DNS_URI_ANS(c, uri, prio, weight)                         \
+    TRACE(c, "URI answer: {int} {int} \"{str}\"", prio, weight, uri)
+#define TRACE_DNS_URI_NOTFOUND(c)               \
+    TRACE(c, "No URI records found")
+#define TRACE_DNS_URI_SEND(c, domain)                   \
+    TRACE(c, "Sending DNS URI query for {str}", domain)
+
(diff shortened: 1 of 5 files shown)


greghudson:
This looks about right, but please name the tracing macros TRACE_DNS_* and put them in alphabetical order in k5-trace.h.  (It looks like a few macros were added to the end out of order in commit 372b3e2a4f3bd9d1b2e05abec4c04b99962e582f, which was a mistake, but by and large the organization is alphabetical.)

mrogers950:
No problem, I've updated.

greghudson:
It seems like it would be easy to trace-log the priority and weight along with the host (or URI) when logging the answers.

In TRACE_DNS_URI_ANS(), please call the second parameter "uri".

In TRACE_DNS_SRV_SEND() and TRACE_DNS_URI_SEND(), please call the second parameter "domain".  (I am not sure why the local variable is named "host"; it is certainly not a hostname.)

mrogers950:
> It seems like it would be easy to trace-log the priority and weight along with the host (or URI) when logging the answers.

I added these to the trace in a manner that mirrors the zone file format, including the port for SRV records.

> How hard would it be to log what kind of failure happens when we don't get an answer?

Are you thinking similar to the Tprint() calls in dns_locate_server_* functions? Should we log an incorrect URI format failure specifically?

> In TRACE_DNS_SRV_SEND() and TRACE_DNS_URI_SEND(), please call the second parameter "domain". (I am not sure why the local variable is named "host"; it is certainly not a hostname.)

To me, "record", or "name" or similar would be better; We'll be printing the full record name here, so it's not just the domain. I agree the variable names don't make much sense now. I think they would make more sense if the local "host" variables are changed to "name" and the srv_dns_entry "host" is changed to "target".

greghudson:
> Are you thinking similar to the Tprint() calls in dns_locate_server_* functions?

I wasn't thinking about anything specific.  I just thought it would be nice if we attempt a lookup and don't get an answer, it would be nice to log what happened instead, if it's not too hard.

> To me, "record", or "name" or similar would be better; We'll be printing the full record name here, so it's not just the domain.

Technically, when you look up anything in DNS, the key is a domain (or domain name).  See RFC 1035 section 2.3.1.

mrogers950:
I added logs for lookup failures and empty results, and replaced the equivalent Tprint calls (their output was not used by any test program).

greghudson:
I don't think TRACE_DNS_SRV_ERR() or TRACE_DNS_URI_ERR() will fire unless add_host_to_list() returns ENOMEM in the current state of the code.  Meaningful error tracing has to happen at lower layers since we currently suppress errors at lower layers (in particular, k5_make_uri_query() and krb5int_make_srv_query_realm() never return non-zero).  We can defer meaningful DNS error tracing for now, but I'd prefer that two _ERR() tracing macros be removed.

mrogers950:
Sorry, I forgot to mention that I've pushed an update with those changes.

PR #565, opened 2016-11-03
Memory scrubbing improvements

greghudson:
The first commit here improves the reliability of zap() with current and (we hope) future compilers, in particular by using memset_s() when we can.  There's a potential technique not used here, which is to make a call to memset through a volatile function pointer, but I don't know that it's any more likely to work than a shared library call to a function using a volatile char *.

The second commit uses zap() in places where we currently use memset().

I tried creating an automated test for zap() by creating a core file and looking through it for a sentinel string, but so far I haven't been able to get the test to fail with gcc 5.4.0-6ubuntu1~16.04.2 or clang 3.8.0-2ubuntu4, even using just free() with no zap or memset.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -652,30 +652,33 @@ k5_sha256(const krb5_data *in, uint8_t out[K5_SHA256_HASHLEN]);
  */
 #ifdef _WIN32
 # define zap(ptr, len) SecureZeroMemory(ptr, len)
-#elif defined(__GNUC__)
+#elif defined(__STDC_LIB_EXT1__)
+/*
+ * Use memset_s() which cannot be optimized out.  Avoid memset_s(NULL, 0, 0, 0)
+ * which would cause a runtime constraint violation.
+ */
 static inline void zap(void *ptr, size_t len)
 {
-    memset(ptr, 0, len);
-    /*
-     * Some versions of gcc have gotten clever enough to eliminate a
-     * memset call right before the block in question is released.
-     * This (empty) asm requires it to assume that we're doing
-     * something interesting with the stored (zero) value, so the
-     * memset can't be eliminated.
-     *
-     * An optimizer that looks at assembly or object code may not be
-     * fooled, and may still cause the memset to go away.  Address
-     * that problem if and when we encounter it.
-     *
-     * This also may not be enough if free() does something
-     * interesting like purge memory locations from a write-back cache
-     * that hasn't written back the zero bytes yet.  A memory barrier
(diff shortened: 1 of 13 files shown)


greghudson:
I decided to push the third commit with the test program to the PR branchj.  Even if free() on Ubuntu is scrubbing memory or something, it could reveal a problem in other environments.

greghudson:
I wasn't setting up the linker environment when running the test script.  That's fixed now.  (Still not sure this method of testing is viable, but I wanted to at least fix the obvious problems while we evaluate it.)

greghudson:
I changed the test commit so that the test program changes into zaptestdir, so that the runtime setup works with a relative LD_LIBRARY_PATH.  I also fixed a typo which was referencing "skiptestdir" instead of "zaptestdir".

frozencemetery:
Looks okay to me on Debian, but not on Fedora.  Fedora uses the coredump location /var/lib/systemd/coredump; I'm not sure how to even handle the way that directory works since only root can write there.

First two commits are fine; not sure what to do about the third.

greghudson:
I removed zap_check from make check, removed the skiptests stuff from the script, and changed zaptest.c to use a fixed buffer.  I still haven't been able to induce Ubuntu 16.04's gcc or clang to optimize out a call to memset(), so I am still inclined to omit the third commit from master on the grounds that it doesn't seem to be a very useful test.

src/include/k5-int.h
```diff
@@ -652,30 +652,33 @@ k5_sha256(const krb5_data *in, uint8_t out[K5_SHA256_HASHLEN]);
...
-     * that hasn't written back the zero bytes yet.  A memory barrier
-     * instruction would help in that case.
-     */
-    asm volatile ("" : : "g" (ptr), "g" (len));
+    if (len > 0)
+        memset_s(ptr, len, 0, len);
+}
+#elif defined(__GNUC__) || defined(__clang__)
+/*
+ * Use a memory barrier to force the memset to be carried out.  Avoid
```
tlyu: This should probably say "Use an asm statement that declares a memory clobber...".  Also, why the switch to the double-underscore asm and volatile keywords?

greghudson: The example header file from Zhaomo uses double-underscore keywords and "r" instead of "g" (which should be unimportant), so I thought I would switch over to being consistent with their approach.  Using double underscore keywords is more friendly to -ansi or -std=c99, which isn't really important to us, but could be if this code is lifted into another project.

I will change the comment.

PR #566, opened 2016-11-04
Modify shlib.conf to cross compile for Chrome OS.

illiminable:
Sets use_linker_init_option and use_linker_fini_option to yes
when compiling for Chrome OS.


frozencemetery:
This seems good to me.  You may wish to adjust your git/github email configuration so that github properly picks up you as the commit author.

greghudson:
This commit message says what is changed (which is obvious from the diff), but doesn't say why.  What problem is it solving?

PR #568, opened 2016-11-16
Remove redeclaration of ttyname.

illiminable:
The original declaration is in unistd.h with a different
signature.

This caused an issue building with clang fortify on Chrome OS.

Caused the following error;

main.c:858:15: error: redeclaration of 'ttyname' must have the 'overloadable' attribute
    char *p, *ttyname();
              ^
/build/samus/usr/include/unistd.h:784:14: note: previous overload of function is here
extern char *ttyname (int __fd) __THROW __CLANG_NO_MANGLE (ttyname);

### DIFF
--- a/src/clients/ksu/main.c
+++ b/src/clients/ksu/main.c
@@ -932,7 +932,7 @@ int standard_shell(sh)
 
 static char * ontty()
 {
-    char *p, *ttyname();
+    char *p;
     static char buf[MAXPATHLEN + 5];
     int result;
 


frozencemetery:
`char *ttyname(int fd)` is the declaration as per ttyname(3), so the signatures do in fact match.

vapier:
the code is doing `ttyname()` which is not the same.  yes, old style C prototypes allow you to pass an arbitrary number of args in that case so it would accept a single fd, but no, they do not match.

further, it's a pointless decl when you get it from the headers, and it breaks building with modern compilers as he pointed out.

frozencemetery:
Well so... it's only worth getting rid of if it's harmful, which would include breaking the build (as you say, it breaks chrome OS).  It doesn't break modern clang or gcc, for instance.  I don't like the uses of "match" and "same" you two suggest, but I don't disagree with what you're saying, just the terminology.

vapier:
it breaks users of clang who want to enable hardening features.  that's why CrOS noticed -- we're enabling optional hardening features that clang supports.  it's not specific to CrOS.

as i said, i disagree with your assessment that they match and/or the same.  the current prototype allows the code to call `ttyname();` or `ttyname("foo");` or `ttyname(1, &i, "aa");` and never issue a complaint even though they're all very very wrong.

tlyu:
If we want to be pedantic (at least according to C99) about terminology, my reading is the declarations have compatible types, but are not identical.  (The types are function returning pointer to character taking a single int parameter vs function returning pointer to character taking unspecified parameters; they are compatible but not identical types.)  POSIX says `char *ttyname(int fildes);` which is compatible with `char *ttyname();` As @vapier notes, the empty parameter list declaration allows calling with an arbitrary number of parameters, but I note that's only true if there's no prototype declaration of `ttyname` in scope.

I think it's poor style to declare a symbol that a standard header provides, without good reason.  One reason is there might be platforms that lack a declaration for `ttyname`, so deleting the declaration isn't completely harmless on such platforms.

We could delete the declaration and worry about adding it back (conditionally) if there's a platform we care about that doesn't declare it in `<unistd.h>`.

frozencemetery:
The (re)declaration of ttyname has been there since the code was contributed with no explanation.  Searching does not reveal any reported instances of ttyname missing on specific platforms.

I agree that we should probably delete it and add it back later on any platforms where it's needed.

greghudson:
I adjusted the commit message; does this look correct to people?

vapier:
lgtm, thanks!

PR #571, opened 2016-11-29
Relicense ccapi/common/win/OldCC/autolock.hxx

greghudson:
With permission from Danilo Almeida, change the license on
autolock.hxx to the 2-clause BSD license used by MIT krb5.

ticket: 8520 (new)
target_version: 1.15-next
target_version: 1.14-next
tags: pullup


greghudson:
Updated to add entries to notice.rst and NOTICE.

PR #574, opened 2016-12-04
Add AD-SIGNTICKET test for renewing tickets

greghudson:
Add another regression test for #8139 using renewed tickets.

[Although it's hard to back out the 8139 fix or directly backport this test case to 1.13, I created a variant of this test case in krb5 1.13's t_general.py and verified that it fails with "KDC returned error string: HANDLE_AUTHDATA" as expected.]


greghudson:
The comment for the previous test already mentions those things; do you want those references repeated?

PR #576, opened 2016-12-06
Add code to properly handle plugins for dlclose()

simo5:
In Fedora 25 we started experiencing issues with the way libgssapi_krb5 calls dlclose() in some situations, various reports have been coalesced in this bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1398370

This patch implements a possible way to avoid calling dlclose() on plugins multiple times on a program exit() (once by the dynamic linker and then again by the gss librray destructor). It does require plugins to implement a destructor too, but it can cope with older plugins that don't.

Comments appreciated.

### DIFF
--- a/src/lib/gssapi/mechglue/g_initialize.c
+++ b/src/lib/gssapi/mechglue/g_initialize.c
@@ -57,6 +57,7 @@
 
 #define MECH_SYM "gss_mech_initialize"
 #define MECH_INTERPOSER_SYM "gss_mech_interposer"
+#define MECH_DESTROY_HOOK "gss_mech_destroy_hook"
 
 #ifndef MECH_CONF
 #define	MECH_CONF "/etc/gss/mech"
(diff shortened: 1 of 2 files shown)


frozencemetery:
Simo and I have discussed this patch some, and it looks okay to me.

Of note, glibc will likely fix the `assert()` issue, but that won't help us on older platforms (and we're still buggy even with that changed).

greghudson:
Can you clarify what you mean by "we're buggy even with that changed?"  I am not convinced that it is a bug to dlclose() a handle returned by dlopen(), even if it refers to a handle for a library that has already been unloaded.

fweimer:
@greghudson Do you mean “a handle for a library whose destructors have run” instead of “ a handle for a library that has already been unloaded”?

If the library is unloaded, the handle is invalid.  I think the dynamic linker just can't unload the library (at least not completely) until the dlopen reference count reaches zero.

greghudson:
Sure.  If the library's destructors have been run, it is presumably no longer safe to call any function in it, but it may not technically have been unloaded.  Regardless, I think it should still be defined to dlclose() a handle to it.

simo5:
Greg, what happens here is that on exit() the dynamic linker can dlclose any libarray in any order, so hat I thin is happening here is that the dl is dlclos()in the plugins first, then the it dlclose() libgssapi and the gss destructor tries to dlclose() the plugins again.
My patch assures that if plugins have destroyers and they corrctly call the hook passed in, when the main libgssapi destroyer is called we do not try to double close the plugins.
I haven't tested this yet, but I plan to rebuild the gssproxy plugin tomorrow and verify that I do not get asserts() anymore.

greghudson:
> Greg, what happens here is that on exit() the dynamic linker can dlclose any libarray in any order, so hat I thin is happening here is that the dl is dlclos()in the plugins first, then the it dlclose() libgssapi and the gss destructor tries to dlclose() the plugins again.

I don't think that's precisely correct.  The plugin object's destructors are run, and the l_init_called field is cleared to indicate that the library is no longer initialized, but the dlopen() handle to the library is not a dangling pointer and the library has not been unloaded.  Calling any function in the library would be unsafe (because any state set up by the initializer has been destroyed), but it should still be safe to close the handle.

> My patch assures that if plugins have destroyers and they corrctly call the hook passed in, when the main libgssapi destroyer is called we do not try to double close the plugins.

I would not call what we are doing a "double close".  We received a handle from dlopen(), and we call dlclose() on it exactly once.  So I object to this patch because I don't think we should add this much complexity to work around what seems like a pretty shallow rtld bug.

I also object to this patch because it causes GSS plugin objects to call into libgssapi_krb5 code during destruction.  Although the Linux rtld currently runs destructors on GSS plugin objects before libgssapi_krb5 itself, it could reasonably decide to run them in the other due to the runtime circular dependency.  In that scenario, running code inside libgssapi_krb5 during destruction wouldn't be safe.

fweimer:
I agree with Greg here: If you called `dlopen`, you should be able to call `dlclose` on that handle even if destructors for this shared object have been run.  `dlsym` should probably fail with an error, but `dlclose` should succeed.

simo5:
Ok, then I retire this PR.

PR #577, opened 2016-12-12
Correctly initialize the out ccache in gss_krb5_copy_ccache()

cryptomilk:
In Samba we call gss_krb5_copy_ccache() to copy a gssapi_cred to a new credential cache. This fails because the cache is not initialized. The code works with Heimdal Kerberos. I think we also should correctly initialize the cache because we are not able to do it outside of Kerberos because we do not have access to the principal.

I think the right spot is to do it in krb5_cc_copy_creds() and use that function.

### DIFF
--- a/src/lib/gssapi/krb5/copy_ccache.c
+++ b/src/lib/gssapi/krb5/copy_ccache.c
@@ -8,8 +8,6 @@ gss_krb5int_copy_ccache(OM_uint32 *minor_status,
                         const gss_buffer_t value)
 {
     krb5_gss_cred_id_t k5creds;
-    krb5_cc_cursor cursor;
-    krb5_creds creds;
     krb5_error_code code;
     krb5_context context;
     krb5_ccache out_ccache;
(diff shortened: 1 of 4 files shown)


greghudson:
Can Samba use gss_store_cred_into()?

cryptomilk:
I could try to implement it using gss_store_cred_into() if the function is available. Our in-tree Heimdal doesn't support that. I'm need to check if newer version do that.

frozencemetery:
Agreed that `gss_store_cred_into` should be preferred if possible, but I do wonder if there's a reason not to initialize the ccache here.

greghudson:
Probably not, but these two functions have existed in public APIs with the current semantics for longer than I have been working on krb5, and there is a risk to changing their behavior (as can be seen by the test suite failing in an S4U2Proxy test).

cryptomilk:
I've implemented it using gss_store_cred_into() now. However if you don't want to fix it, shouldn't it be at least documented that the function does not correctly initialize the ccache with the principal and suggest to use gss_store_cred_into().

However the second patch is still correct :-)

greghudson:
I pushed the second commit.  I will file a PR to add documentation to gss_krb5_copy_ccache() recommending that gss_store_cred_into() be used instead.

greghudson:
I filed PR #578.  Closing this one.

PR #579, opened 2016-12-14
Some fixes for arguments which should be const

cryptomilk:
Some fixes for arguments which should be const.


greghudson:
krb5_fwd_tgt_creds() needs some memory handling cleanup for this change, as shown by the clang build failure.  Instead of a free_rhost variable, it needs a separate char *def_rhost variable to hold the allocated value created at line 134.  rhost can then be alised to def_rhost, and def_rhost can be freed in the cleanup handler.

Similar cleanup is warranted for close_cc and cc, so it might be worth doing at the same time.

greghudson:
I don't appear to have permission to push to this PR branch, but see the pr579 branch in my repository for commits containing the changes I outlined.

https://github.com/greghudson/krb5/tree/pr579

cryptomilk:
Sorry, I didn't have the time to update the changes. I had to prepare for some Security updates. Your PR looks fine, thanks!

PR #581, opened 2016-12-14
Clean up PKINIT decode_data functions

greghudson:
[The first commit is from PR #580, which I submitted separately for Travis builds.]

In pkinit_crypto_openssl.c, fold decrypt_data() into its caller
pkinit_decode_data_fs(), and simplify its error-handling logic.
Initialize output parameters in pkinit_decode_data_pkcs11() and
pkinit_decode_data().


frozencemetery:
Would this be cleaner if the initialization fixups were their own commit?  Otherwise I'm happy about this.

PR #582, opened 2016-12-14
Fix documented kcpreauth minor version

greghudson:
Commit 7b12eb4757f8dd05b79c9b49d4289f0caf1f6eec erroneously changed
the documented kdcpreauth minor version from 2 to 3.  Only the
callback version changed in that commit; the interface minor version
remains at 2 (from commit 83b4ecd20e50ad330cd761977d5dadefe30a785b
which added the loop method).

### DIFF
--- a/src/include/krb5/kdcpreauth_plugin.h
+++ b/src/include/krb5/kdcpreauth_plugin.h
@@ -34,7 +34,7 @@
  * Declarations for kdcpreauth plugin module implementors.
  *
  * The kdcpreauth interface has a single supported major version, which is 1.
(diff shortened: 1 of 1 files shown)


frozencemetery:
This made it into both 1.14 and 1.15 released versions; would it be better to just leave it?

(Also, really not a fan of what GitHub does to the length of commit hashes for wrapping purposes, but that's neither here nor there.)

tlyu:
@frozencemetery this is a (header file and not RST) documentation change to match the shipped interface, so I don't see much reason to not make this change.

frozencemetery:
That's a fair point.

PR #585, opened 2016-12-23
Use LDIF config and add mdb support in t_kdb.py

greghudson:
When setting up slapd, use slapadd with cn=config LDIF directives
instead of the deprecated config file.  By adding one cn=config object
at a time, we can detect specific specific configuration failures,
including unsupported database types.  Try the mdb and bdb database
types, to work with older and newer OpenLDAP versions.

[There is also a first commit on this branch, to add an OpenLDAP LDIF version of the krb5 schema.]


greghudson:
If slapd is configured without module support, attempts to load the back_mdb and back_bdb module will fail with "unrecognized objectClass 'olcModuleList'", before slapd's module_load() gets a chance to intercept the request to load an already existing back end.  I will need to adjust the DB selection code to take that into account.

greghudson:
Okay, that's fixed.  This is ready for review again.

greghudson:
I'm going to merge this since Will has had to back-burner testing it.  If it doesn't fix the Solaris issues we can make another change later.

PR #587, opened 2017-01-03
Fix KDC and kadmind network setup issues

greghudson:
Here are fixes for the two problems discussed in ticket 8530.  (I created ticket 8531 to track the IPv4-only startup issue.)

### DIFF
--- a/src/lib/apputils/net-server.c
+++ b/src/lib/apputils/net-server.c
@@ -105,6 +105,17 @@ paddr(struct sockaddr *sa)
     return buf;
 }
 
+/* Return true if sa is an IPv4 or IPv6 wildcard address. */
+static int
+is_wildcard(struct sockaddr *sa)
+{
+    if (sa->sa_family == AF_INET6)
+        return IN6_IS_ADDR_UNSPECIFIED(&sa2sin6(sa)->sin6_addr);
+    else if (sa->sa_family == AF_INET)
+        return sa2sin(sa)->sin_addr.s_addr == INADDR_ANY;
+    return 0;
+}
+
 /* KDC data.  */
 
 enum conn_type {
@@ -753,7 +764,7 @@ setup_socket(struct socksetup *data, struct bind_address *ba,
     }
 
     /* Try to turn on pktinfo for UDP wildcard sockets. */
-    if (ba->type == UDP && ba->address == NULL) {
+    if (ba->type == UDP && is_wildcard(sock_address)) {
         krb5_klog_syslog(LOG_DEBUG, _("Setting pktinfo on socket %s"),
                          paddr(sock_address));
         ret = set_pktinfo(sock, sock_address->sa_family);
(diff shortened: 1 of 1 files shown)


src/lib/apputils/net-server.c
```diff
@@ -889,9 +893,15 @@ setup_addresses(struct socksetup *data)
                                  _("Failed setting up a %s socket (for %s)"),
                                  bind_type_names[addr.type],
                                  paddr(ai->ai_addr));
-                goto cleanup;
+                if (ret != EAFNOSUPPORT)
```
frozencemetery: On IPv4 systems, when `ret == EAFNOSUPPORT`, won't this result in "Failed setting up" messages being logged?  This seems potentially confusing since they may not be fatal.

greghudson: Maybe?  I'm not sure it's wrong to log that since it points to dubious system behavior.  (I think what's going on is that the system decided not to load the IPv6 kernel module because there are no IPv6 kernel interfaces, but it doesn't have a mechanism to suppress returning IPv6 addresses from getaddrinfo(NULL, ...).)  The code gets more complicated if we try to log an EAFNOSUPPORT error only if we get it for all getaddrinfo results, but not if we get it for some.

frozencemetery: You're right, it would rather complicate things to not log at all.  One alternative could be to log a different message for EAFNOTSUPPORT when it is encountered. That said, though, `loop_setup_network()` takes care of logging that there was a fatal error in setup, so I could live with the code as written.

src/lib/apputils/net-server.c
```diff
@@ -753,7 +764,7 @@ setup_socket(struct socksetup *data, struct bind_address *ba,
     }
 
     /* Try to turn on pktinfo for UDP wildcard sockets. */
-    if (ba->type == UDP && ba->address == NULL) {
+    if (ba->type == UDP && is_wildcard(sock_address)) {
         krb5_klog_syslog(LOG_DEBUG, _("Setting pktinfo on socket %s"),
```
SaharahSarah: I can't recall the exact reason right now but there was definitely a reason why I didn't use pktinfo for all addresses, just wildcard. I'll have to look into it when I have access to my laptop.

rbasch: I think the following conditions might have issues: Point-to-Point, Loopback, and Multicast.

I do think if pktinfo fails, the code should use the default socket address (whether or not it does currently is something that should be examined)... I agree pktinfo may not always work but I think it presumptuous to assume "wildcard" is the only condition where it is required. I would be curious what you find.

src/lib/apputils/net-server.c
rbasch: It works for me.

src/lib/apputils/net-server.c
greghudson: Thanks for testing.

For the purpose of a backported fix, I would like to avoid using pktinfo for non-wildcard sockets, as we have never done that.  (Prior to 1.15, we either used pktinfo on a wildcard socket or iterated over the interfaces and bound to each interface address without pktinfo.)  We can consider using pktinfo everywhere for 1.16, although my memory agrees with Sarah's that there is a portability issue lurking there.

PR #588, opened 2017-01-03
net-server.c cleanups

greghudson:
The first two commits are PR #587.  The next two commits remove some redundancy introduced by PR #587 by introducing the sa_is_wildcard() helper and using it.  The final two commits clean up some constructs made unnecessary when we removed the workaround for non-pktinfo platforms.

### DIFF
--- a/src/lib/apputils/net-server.c
+++ b/src/lib/apputils/net-server.c
@@ -514,7 +514,7 @@ free_socket(verto_ctx *ctx, verto_ev *ev)
 
 static verto_ev *
 make_event(verto_ctx *ctx, verto_ev_flag flags, verto_callback callback,
-           int sock, struct connection *conn, int addevent)
+           int sock, struct connection *conn)
 {
     verto_ev *ev;
     void *tmp;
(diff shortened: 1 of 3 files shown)


src/lib/apputils/net-server.c
```diff
@@ -581,77 +574,61 @@ static void process_tcp_connection_write(verto_ctx *ctx, verto_ev *ev);
...
+    *fd_out = -1;
+
     sock = socket(addr->sa_family, type, 0);
     if (sock == -1) {
-        data->retval = errno;
-        com_err(data->prog, errno, _("Cannot create TCP server socket on %s"),
+        com_err(prog, errno, _("Cannot create TCP server socket on %s"),
                 paddr(addr));
-        return -1;
+        return errno;
```
frozencemetery: `errno` may be clobbered by `paddr()` which calls `getnameinfo(3)`.

src/lib/apputils/net-server.c
```diff
@@ -581,77 +574,61 @@ static void process_tcp_connection_write(verto_ctx *ctx, verto_ev *ev);
...
     }
 
     if (bind(sock, addr, sa_socklen(addr)) == -1) {
-        data->retval = errno;
-        com_err(data->prog, errno, _("Cannot bind server socket on %s"),
+        com_err(prog, errno, _("Cannot bind server socket on %s"),
                 paddr(addr));
         close(sock);
-        return -1;
+        return errno;
```
frozencemetery: Save `errno` here too.

src/lib/apputils/net-server.c
greghudson: Good catch.  Fixed.

PR #593, opened 2017-01-09
Fix per-request preauth scoping

greghudson:
It should be possible to successfully use multiple initial credentials
contexts with the same library context.  Create a new internal type
krb5_preauth_req_context containing per-request preauth state,
including the clpreauth modreq handles and the list of preauth types
already tried.  Remove this state from clpreauth_handle and
krb5_preauth_context.

[There is also a preliminary commit to make krb5_preauth_context a pointer type, and follow-up commits to add tests and to check that the same library context is used for each step of an initial credentials context.]

### DIFF
--- a/src/tests/icinterleave.c
+++ b/src/tests/icinterleave.c
@@ -1,0 +1,124 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* tests/icinterleave.c - interleaved init_creds_step test harness */
+/*
+ * Copyright (C) 2017 by the Massachusetts Institute of Technology.
+ * All rights reserved.
+ *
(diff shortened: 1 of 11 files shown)


src/tests/icinterleave.c
```diff
@@ -0,0 +1,124 @@
...
+    char **princstrs;
+    krb5_principal client;
+    krb5_init_creds_context *iccs;
+    krb5_data req, *reps, realm;
+    krb5_boolean any_left;
+    int i, nclients, master;
+    unsigned int flags;
+
+    if (argc < 3) {
+        fprintf(stderr, "Usage: icred password princ1 princ2 ...\n");
```
greghudson: Yes, I will fix this.

PR #597, opened 2017-01-13
Fix missing dependencies in gssrpc headers

frozencemetery:
This is intended to allow each header file to be included individually without missing definitions.

### DIFF
--- a/src/include/gssrpc/auth.h
+++ b/src/include/gssrpc/auth.h
@@ -42,6 +42,7 @@
 #ifndef GSSRPC_AUTH_H
 #define GSSRPC_AUTH_H
 
+#include <gssrpc/types.h>
 #include <gssrpc/xdr.h>
 
 GSSRPC__BEGIN_DECLS
--- a/src/include/gssrpc/auth_gssapi.h
+++ b/src/include/gssrpc/auth_gssapi.h
@@ -8,6 +8,10 @@
 
 GSSRPC__BEGIN_DECLS
 
+#include <gssapi/gssapi.h>
+#include <gssrpc/clnt.h>
+#include <gssrpc/svc.h>
+
 #define AUTH_GSSAPI_EXIT		0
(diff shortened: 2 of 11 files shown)


tlyu:
What sort of application (or library) needs a subset of the gssrpc headers and has problems with rpc.h?  I think the "supported" approach is to include gssrpc/rpc.h.

frozencemetery:
This was reported downstream to me as [a bug with rubygem-rkerberos](https://bugzilla.redhat.com/show_bug.cgi?id=1412274).  While I suspect the actual, underlying issue lies elsewhere in their build process, this seemed like an easy fix to me.

tlyu:
Hm, the log from the reported downstream bug says it's including kadm5/admin.h, which should be OK.  Somehow traditional Unix types like caddr_t and u_int are missing from the system headers at compile time for the ruby gem but not when the krb5 package was built.  This is perplexing.

tlyu:
More details: at configure time, krb5 defines GSSRPC__BSD_TYPEALIASES=1 if it's necessary to provide BSD type aliases like u_int, but not caddr_t, and inserts that into types.h.  If the types provided by sys/types.h vary between configuring krb5 and building the ruby gem, that seems to be a ruby gem build system problem to me.  Also we don't currently compensate for a complete lack of definition of caddr_t (as we weren't previously aware of a need to do that on any platform).  Can you give more details about the nature of the highly restrictive namespace that seems to be present here?

greghudson:
I just looked harder at the upstream bug, and I think the conflict is with -std=c99.  Using -std=gnu99 instead might be a good workaround, but we might want to see if we can make our headers work with the c99 namespace.

(I don't think the changes in this PR address the underlying issue.)

frozencemetery:
It sounds like this change isn't wanted unless it's going to fix the problem, and I think we're all pretty sure it isn't, so I'll close this.

Thanks Greg and Tom for looking into it further!

PR #599, opened 2017-01-21
Echo KDC cookies in preauth tryagain

greghudson:
[This will be tested in the preauth fallback work.  I want the change reviewed separately as it's a behavior change independent of improving preauth fallback.]

When trying again after a mechanism-specific error, we should send the
KDC cookie for conformance with RFC 6113.

ticket: 8539


greghudson:
I realized that it's easy to test this change individually by amending the PKINIT tryagain test, so I added that to the commit.

PR #600, opened 2017-01-25
Document default realm and login authorization

greghudson:
[In a recent admin/ops call, the topic of wanting multiple default realms came up, and I thought it might be helpful to document what the default realm does and how to override it in each case.  We are also missing prose documentation for local authorization, which is one of the chief behaviors affected by the default realm.  This PR contains documentation for both.]

Add documentation to host_config.rst describing what the default realm
does.  Also add documentation discussing login authorization
configuration, and give an example showing how to give login access to
principals from a realm other than the default realm.

### DIFF
--- a/doc/admin/host_config.rst
+++ b/doc/admin/host_config.rst
@@ -7,6 +7,119 @@ application servers, or KDCs, can be configured using
 you might want to make.
 
 
+Default realm
+-------------
+
+In the :ref:`libdefaults` section, the **default_realm** realm
+relation sets the default Kerberos realm.  For example::
+
+    [libdefaults]
+        default_realm = ATHENA.MIT.EDU
+
+The default realm affects Kerberos behavior in the following ways:
+
+* When a principal name is parsed from text, the default realm is used
+  if no ``@REALM`` component is specified.
+
+* The default realm affects login authorization as described below.
+
+* For programs which operate on a Kerberos database, the default realm
+  is used to determine which database to operate on, unless the **-r**
+  parameter is given to specify a realm.
+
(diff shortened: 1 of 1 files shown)


doc/admin/host_config.rst
```diff
@@ -7,6 +7,119 @@ application servers, or KDCs, can be configured using
...
+    [realms]
+        ATHENA.MIT.EDU = {
+            auth_to_local_names = {
+                # Careful, these match principals in any realm!
+                host/example.com = hostaccount
+                fred = localfred
+            }
+        }
+
+Local authorization behavior can be modified using plugin modules; see
```
kaduk: Maybe "can also"?

greghudson: Sure.

doc/admin/host_config.rst
```diff
@@ -7,6 +7,119 @@ application servers, or KDCs, can be configured using
...
+files located in account home directories, the **k5login_directory**
+relation in the :ref:`libdefaults` section can specify a directory
+containing one file per account uname.
+
+By default, if a ``.k5login`` file is present, it controls
+authorization both positively and negatively--any principal name
+contained in the file is granted access and any other principal name
+is denied access, even if it would have had access if the ``.k5login``
+file didn't exist.  The **k5login_authoritative** relation in the
+:ref:`libdefaults` section can be used to make ``.k5login`` files
```
mrogers950: Changing to "section can be set to false to make.." may make it clearer about the k5login_authoritative default value in this paragraph.

greghudson: Sure, I will make that change.

PR #601, opened 2017-01-26
Add OID to inquire impersonator status of a credential

simo5:
I coded this offline trying to remember what @greghudson adviced to do, hopefully I got it right.


src/lib/gssapi/krb5/gssapiP_krb5.h
```diff
@@ -1425,4 +1425,14 @@ iakerb_gss_pseudo_random(OM_uint32 *minor_status, gss_ctx_id_t context_handle,
  * the format changes. */
 #define CRED_EXPORT_MAGIC "K5C1"
 
+#define GSS_KRB5_GET_CRED_IS_IMPERSONATOR_OID_LENGTH 11
+#define GSS_KRB5_GET_CRED_IS_IMPERSONATOR_OID  \
+    "\x2a\x86\x48\x86\xf7\x12\x01\x02\x02\x05\x0e"
+
+OM_uint32 KRB5_CALLCONV
+gss_krb5int_cred_is_impersonator(OM_uint32 *minor_status,
```
greghudson: I don't think you want KRB5_CALLCONV here.  I believe this is causing the appveyor build failure.

src/lib/gssapi/krb5/inq_cred_oid.c
```diff
@@ -0,0 +1,62 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ * Copyright 2017 Red Hat, Inc.
+ * All Rights Reserved.
+ *
+ * Export of this software from the United States of America may
```
greghudson: If the intent is to use the MIT krb5 license here, please use the 2-clause BSD one from src/prototype/prototype.c.

src/tests/gssapi/t_s4u.c
```diff
@@ -305,6 +307,32 @@ main(int argc, char *argv[])
...
+    if (delegated_cred_handle != GSS_C_NO_CREDENTIAL) {
+        /* Inquire impersonator status */
+        major = gss_inquire_cred_by_oid(&minor, user_cred_handle, &inq_oid,
+                                        &bufset);
+        check_gsserr("gss_inquire_cred_by_oid", major, minor);
+
+        if (bufset->count != 1 || bufset->elements[0].length != sizeof(int))
+            errout("gss_inquire_cred_by_oid() returned invalid buffer set");
+
+        if (*((int *)bufset->elements[0].value) != 1)
```
greghudson: This could bus error on a platform with strict alignment requirements.  (Same for the similar condition below.)

src/lib/gssapi/krb5/gssapiP_krb5.h
```diff
@@ -1425,4 +1425,14 @@ iakerb_gss_pseudo_random(OM_uint32 *minor_status, gss_ctx_id_t context_handle,
  * the format changes. */
 #define CRED_EXPORT_MAGIC "K5C1"
 
+#define GSS_KRB5_GET_CRED_IS_IMPERSONATOR_OID_LENGTH 11
+#define GSS_KRB5_GET_CRED_IS_IMPERSONATOR_OID  \
+    "\x2a\x86\x48\x86\xf7\x12\x01\x02\x02\x05\x0e"
```
simo5: @greghudson I mean to ask if this OID is ok or if I should get another allocated by you.

src/lib/gssapi/krb5/gssapiP_krb5.h
simo5: ok

src/lib/gssapi/krb5/inq_cred_oid.c
simo5: Yeah I just copied the header from another file in the same directory, I'll change it.

src/tests/gssapi/t_s4u.c
simo5: Going to put a string in there with the name of the impersonator, so this shouldn't then be a problem anymore.

simo5:
Ok, ging to rework it to return a name, is it ok to return the result of gss_display_name ? Or do you think we should return a gss_name_t ?

simo5:
Ok rebased to return a string with the impersonator principal name unparsed, let me know if this works better.

src/lib/gssapi/krb5/gssapiP_krb5.h
greghudson: I'm not sure what this OID is, but it's probably okay.  This PR should probably also export an OID object and declare it in gssapi_krb5.h with an annotation for the OID like the other declarations there.

greghudson:
> Ok, ging to rework it to return a name, is it ok to return the result of gss_display_name ? Or do you think we should return a gss_name_t ?

Returning a gss_name_t via a gss_buffer seems questionable, so I think returning the display name is fine.

greghudson:
The inquire-context-by-oid implements are in inq_context.c, not a separate file.  Can we put the inquite-cred-by-oid function in inq_cred.c instead of in a new inq_cred_oid.c file?

(Sorry not to include this in the first round of feedback.)

src/lib/gssapi/krb5/gssapiP_krb5.h
simo5: I did not do this because the context oids are not there either.
But I can do it if you think I should.

simo5:
Removed inq_cred_oid and merged content into inq_cred

src/lib/gssapi/krb5/gssapiP_krb5.h
greghudson: Which context OIDs?  Some of those are only of interest to mechs, so it's less important to have an exported symbol.  This one is of importance to applications.

simo5:
Exposed OID in gssapi_krb5.h

src/lib/gssapi/krb5/gssapi_krb5.h
```diff
@@ -96,6 +96,15 @@ GSS_DLLIMP extern const gss_OID_desc krb5_gss_oid_array[];
...
 GSS_DLLIMP extern const gss_OID GSS_KRB5_CRED_NO_CI_FLAGS_X;
 
+/*
+ * This OID can be used with gss_inquire_cred_by_oid(0 to retrieve the
+ * impersonator name (if any).
+ *
+ * iso(1) member-body(2) United States(840) mit(113554) infosys(1) gssapi(2)
+ * krb5(2) krb5_gssapi_ext(5) get_cred_impersonator(14)
+ */
+GSS_DLLIMP extern const gss_OID GSS_KRB5_GET_CRED_IMPERSONATOR;
```
greghudson: Now that I see the expansion of this OID, it's in an arc I haven't familiarized myself with, and which I don't see a registry for.  How did you arrive at 14 for the final component?  (This OID is probably fine, I just want a better idea of what's going on with this arc.)

simo5: I just looked at the OIDs used for the context stuff, and searched for the first free one across the code. It was meant to be temporary until you could confirm it was ok to use it.
Any OID will be fine by me and will correct it as required.

tlyu: @simo5 it's best to get approval from the responsible party for an OID arc to use an OID from that arc.  I don't know who is responsible for that arc in particular; I suspect I will need to dig through my email archives to be sure.  It's also not valid to use underscores in OID symbolic identifiers, so it probably wasn't me who assigned that.  I'm looking at the file that's supposed to be authoritative for that arc and the closest assignment in the 1.2.840.113554.1.2.2 arc is 1.2.840.113554.1.2.2.3 (which is for the user-to-user gss-krb5 mech).  It seems that someone hasn't been updating the file like they should.  (I don't have access to, but I in theory can get the people who do have access to do it for me.)

simo5: @tlyu I asked guidance on the OID in a previous message, feel free to assign me any OID from a space you control. If you do not control any and want me to donate an OID from a space I control, feel free to ask.

greghudson: The registry has now been updated to include this OID as well as the old unregistered ones, so this particular issue is taken care of.

greghudson:
I pushed changes to the PR branch to:

* Remove some redundant parameter checking from gss_krb5int_get_cred_impersonator() that we don't do in other inquire-by-OID methods
* Use dashes instead of underscores in the expansion of the OID to match the registry entries and conform to ASN.1 syntax
* Don't add a const qualifier to the oid field of krb5_gss_inquire_cred_by_oid_ops as it isn't there in similar structures.
* Export GSS_KRB5_GET_CRED_IMPERSONATOR from libgssapi_krb5.
* Use GSS_KRB5_GET_CRED_IMPERSONATOR in t_s4u.
* Flesh out the commit message and make the summary fit within 50 characters.

Ordinarily I would want this PR to include documentation of the new feature, but we don't have any S4U documentation in appdev/gssapi.rst.  I intend to write S4U2Self and S4U2Proxy documentation, including documentation for this feature, and file it in a separate PR after this one is merged.

If a cred has no impersonator, we return GSS_S_UNAVAILABLE.  This is the same status we return if the mech has no gss_inquire_cred_by_oid method or if the krb5 mechanism doesn't recognize the requested OID, so an application can't tell if the cred really isn't an impersonator cred or if it's using code too old to tell.  Since we can't invent new major status codes and none of the old ones really apply, we could return success with an empty buffer set, perhaps?

simo5:
What about GSS_S_CRED_UNAVAIL ?
After all we are looking to see if we have an impersonator cred, and that one cred is unavailable.

greghudson:
From gssapi.hin:

    /* XXXX This is a necessary evil until the spec is fixed */
    #define GSS_S_CRED_UNAVAIL GSS_S_FAILURE

I don't understand the history of this error code, but it doesn't appear in RFC 2743 and I don't think it's necessarily a good idea to add more uses of it.

(Heimdal has the same #define for that error code but has no uses of it.)

simo5:
Would it make sense to return GSS_S_NO_CRED ?

simo5:
Or perhaps a minor that is different from 0 ?
The other option is to return GSS_S_COMPLETE and an empty buffer I guess.

greghudson:
GSS_S_NO_CRED doesn't seem right.  What do you think of returning GSS_S_COMPLETE and an empty buffer set?

simo5:
Ok pushed change to return GSS_S_COMPLETE and an empty buffer if no impersonator is found, tests adjusted.

greghudson:
An empty buffer set is not the same as a one-element buffer set containing an empty buffer.  Do you have a preference for the one-element buffer set?  An empty buffer set seems cleaner to me.

simo5:
I do not have a preference, will change it to be an empty buffer_set.

simo5:
Rebased to send an empty buffer_set

PR #606, opened 2017-02-10
Try a fallback realm for GSSAPI ccache selection

mrogers950:
This follows the discussion at http://mailman.mit.edu/pipermail/krbdev/2017-January/012708.html
Note that the test output changes seem to be only a result of the default domain appearing as a fallback, so we might need a more specific KDC discovery test.

### DIFF
--- a/src/lib/krb5/ccache/ccselect.c
+++ b/src/lib/krb5/ccache/ccselect.c
@@ -132,6 +132,8 @@ krb5_cc_select(krb5_context context, krb5_principal server,
     struct ccselect_module_handle **hp, *h;
     krb5_ccache cache;
     krb5_principal princ;
+    krb5_principal srvcp = NULL;
+    char **fbrealms = NULL;
 
     *cache_out = NULL;
     *princ_out = NULL;
@@ -139,7 +141,27 @@ krb5_cc_select(krb5_context context, krb5_principal server,
     if (context->ccselect_handles == NULL) {
         ret = load_modules(context);
         if (ret)
-            return ret;
+            goto cleanup;
+    }
+
+    /* Try to use the fallback host realm for the server if there is no
+     * authoritative realm. */
+    if (krb5_is_referral_realm(&server->realm) &&
+        server->type == KRB5_NT_SRV_HST && server->length == 2) {
+        ret = krb5_get_fallback_host_realm(context, &server->data[1],
+                                           &fbrealms);
+        if (ret)
+            goto cleanup;
+
+        /* Make a copy with the first fallback realm. */
+        ret = krb5_copy_principal(context, server, &srvcp);
+        if (ret)
+            goto cleanup;
+        ret = krb5_set_principal_realm(context, srvcp, fbrealms[0]);
+        if (ret)
+            goto cleanup;
+
+        server = srvcp;
     }
 
     /* Consult authoritative modules first, then heuristic ones. */
@@ -155,20 +177,25 @@ krb5_cc_select(krb5_context context, krb5_principal server,
                                          princ);
                 *cache_out = cache;
                 *princ_out = princ;
-                return 0;
+                goto cleanup;
             } else if (ret == KRB5_CC_NOTFOUND) {
                 TRACE_CCSELECT_MODNOTFOUND(context, h->vt.name, server, princ);
                 *princ_out = princ;
-                return ret;
+                goto cleanup;
             } else if (ret != KRB5_PLUGIN_NO_HANDLE) {
                 TRACE_CCSELECT_MODFAIL(context, h->vt.name, ret, server);
-                return ret;
+                goto cleanup;
             }
         }
     }
 
     TRACE_CCSELECT_NOTFOUND(context, server);
-    return KRB5_CC_NOTFOUND;
+    ret = KRB5_CC_NOTFOUND;
+
+cleanup:
+    krb5_free_principal(context, srvcp);
+    krb5_free_host_realm(context, fbrealms);
+    return ret;
 }
 
 void
--- a/src/tests/gssapi/t_ccselect.py
+++ b/src/tests/gssapi/t_ccselect.py
@@ -31,12 +31,18 @@ r2 = K5Realm(create_user=False, realm='KRBTEST2.COM', portbase=62000,
 
 host1 = 'p:' + r1.host_princ
 host2 = 'p:' + r2.host_princ
+foo = 'foo.krbtest.com'
+foo2 = 'foo.krbtest2.com'
 
-# gsserver specifies the target as a GSS name.  The resulting
-# principal will have the host-based type, but the realm won't be
-# known before the client cache is selected (since k5test realms have
-# no domain-realm mapping by default).
-gssserver = 'h:host@' + hostname
+# These strings specify the target as a GSS name.  The resulting
+# principal will have the host-based type, with the referral realm
+# (since k5test realms have no domain-realm mapping by default).
+# krb5_cc_select() will use the fallback realm, which is either the
+# uppercased parent domain, or the default realm if the hostname is a
+# single component.
+gssserver = 'h:host@' + foo
+gssserver2 = 'h:host@' + foo2
+gsslocal = 'h:host@localhost'
 
 # refserver specifies the target as a principal in the referral realm.
 # The principal won't be treated as a host principal by the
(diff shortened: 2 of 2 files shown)


src/lib/krb5/ccache/ccselect.c
```diff
@@ -142,32 +144,55 @@ krb5_cc_select(krb5_context context, krb5_principal server,
             return ret;
     }
 
+    ret = krb5_copy_principal(context, server, &srvcp);
```
greghudson: I would prefer to avoid the copy if it's not needed, and set server to point to the copy if we made one.

Also, this PR frees the allocated value in five different exit paths.  Our coding practices state "do not get to the point of freeing the same resource in several different places within a function depending on the exit path."  Introduce a cleanup label and use it.

src/lib/krb5/ccache/ccselect.c
```diff
@@ -142,32 +144,55 @@ krb5_cc_select(krb5_context context, krb5_principal server,
...
+    if (ret)
+        return ret;
+
+    /* Try to use the fallback host realm for the server if there's no
+     * authoritative realm. */
+    if (krb5_is_referral_realm(&srvcp->realm)) {
+        ret = krb5_get_fallback_host_realm(context, &srvcp->data[1], &fbrealms);
```
greghudson: This crashes if there aren't two components in the principal.  Check the number of components and the name type before assuming there's a hostname in the second component.

src/lib/krb5/ccache/ccselect.c
```diff
@@ -142,32 +144,55 @@ krb5_cc_select(krb5_context context, krb5_principal server,
...
+        if (ret) {
+            krb5_free_principal(context, srvcp);
+            return ret;
+        }
+        if (fbrealms != NULL && fbrealms[0] != NULL) {
+            krb5_free_data_contents(context, &srvcp->realm);
+            srvcp->realm = string2data(fbrealms[0]);
```
greghudson: Use krb5_set_principal_realm().

src/lib/krb5/ccache/ccselect.c
```diff
@@ -142,32 +144,55 @@ krb5_cc_select(krb5_context context, krb5_principal server,
...
+            free(fbrealms);
```
greghudson: This leaks memory if fbrealms has more than one value.  fbrealms is also leaked if it contains no values, although that shouldn't happen.

src/tests/gssapi/t_ccselect.py
```diff
@@ -114,9 +117,12 @@
...
 output = r1.run(['./t_ccselect', refserver])
-if output != (bob + '\n'):
-    fail('bob not chosen via primary cache when no .k5identity line matches.')
+if output != (alice + '\n'):
+    fail('alice not chosen via primary cache when no .k5identity line matches.')
 r1.run(['./t_ccselect', 'h:bogus@' + hostname], expected_code=1,
-       expected_msg="Can't find client principal noprinc")
+       expected_msg=('gss_init_sec_context: Unspecified GSS failure.  '
+                     'Minor code may provide more information\n'
+                     'gss_init_sec_context: Server bogus/localhost@KRBTEST.COM '
```
greghudson: The local hostname is included in this expected message, which will cause tests to fail whenever the canonicalized local hostname is not "localhost".

mrogers950:
Thanks, I made those changes and looked at the tests in some more detail.
The test [22] change appears to be correct to me, since adding the fallback realm to server now finds alice@KRBTEST.COM, where before the empty realm led realm_choose() to return KRB5_PLUGIN_NO_HANDLE.
 
Test [26] broke because of neglecting to check the server principal type (The comment for refserver points out that the target is not treated as a host type). With the update this is fixed.

Test [27] finds alice now, since the first line of .k5identity is “alice@KRBTEST.COM realm=KRBTEST.COM”, and with the patch the server realm is now set and matches the first line instead of falling through to third line.  So the error message changed to complain about the service after the ccache selection finds alice.  I’m not sure what change we would need here in order to be able to keep a test for a service constraint to a non-existent principal, along with the other test entries in the same .k5identity.

greghudson:
As these tests are written, the fallback realm will depend on the canonicalized local hostname.  If it has multiple components (e.g. my dev machine's "equal-rites.mit.edu")), then domain_fallback_realm() will chop off the first component and uppercase the rest; in my case the result is "MIT.EDU".  If it has only only component (e.g. your machine's "localhost"), then domain_fallback_realm() will give up and krb5_get_fallback_host_realm() will supply the default realm.

The simple answer is to use a fixed hostname for gssserver.  Better would be to create two or three different h: server strings (one for host@foo.krbtest1.com, for instance, and another for host@localhost) for different behaviors when the fallback realm is used.

src/lib/krb5/ccache/ccselect.c
```diff
@@ -132,14 +132,39 @@ krb5_cc_select(krb5_context context, krb5_principal server,
...
         if (ret)
-            return ret;
+            goto cleanup;
+    }
+
+    /* Try to use the fallback host realm for the server if there's no
+     * authoritative realm. */
+    if (krb5_is_referral_realm(&server->realm) &&
+        server->type == KRB5_NT_SRV_HST && server->length == 2 &&
+        server->data[1].length != 0) {
```
greghudson: I don't think we need to check the length of this component.  If someone manages to feed in host/@ then there's no harm in looking up the fallback realm for the empty string.

src/lib/krb5/ccache/ccselect.c
```diff
@@ -132,14 +132,39 @@ krb5_cc_select(krb5_context context, krb5_principal server,
...
+            goto cleanup;
+
+        server = srvcp;
+        ret = krb5_get_fallback_host_realm(context, &server->data[1],
+                                           &fbrealms);
+        if (ret) {
+            goto cleanup;
+        }
+
+        if (fbrealms != NULL && fbrealms[0] != NULL) {
```
greghudson: If this condition is false, we already created the copy and aliased server, which is harmless but inelegant.  We should make the modified copy after we have a fallback realm to set it to.

mrogers950:
Updated, the tests should now cover the fallback behavior for both multi and single component hostnames regardless of the test machine hostname.

src/lib/krb5/ccache/ccselect.c
greghudson: Actually, looking at the contract of krb5_get_fallback_host_realm() and how it is used in get_creds.c, I don't think we need this check at all.  If krb5_get_fallback_host_realm() succeeds it will return at least one entry.

mrogers950:
Updated again.

src/tests/gssapi/t_ccselect.py
```diff
@@ -31,12 +31,16 @@
...
+foo1 = 'foo.krbtest.com'
+foo2 = 'foo.krbtest2.com'
 
-# gsserver specifies the target as a GSS name.  The resulting
+# gssserver specifies the target as a GSS name.  The resulting
 # principal will have the host-based type, but the realm won't be
 # known before the client cache is selected (since k5test realms have
 # no domain-realm mapping by default).
-gssserver = 'h:host@' + hostname
+gssserver = 'h:host@' + foo1
```
greghudson: Perhaps "foo1" should be "foo" if we're going to use "gssserver" and "gssserver2".

Also, the comment above isn't really accurate after the fallback realm changes, and should be clarified.

src/tests/gssapi/t_ccselect.py
```diff
@@ -92,31 +106,41 @@
 if output != (zaphod + '\n'):
     fail('zaphod not chosen as default initiator name for server in r1')
 
-# Check that primary cache is used if server realm is unknown.
-output = r2.run(['./t_ccselect', gssserver])
+# Check ccache selection using a fallback realm.
+output = r1.run(['./t_ccselect', gssserver])
+if output != (alice + '\n'):
+    fail('alice not chosen via primary cache for unknown server realm')
```
greghudson: My understanding of what happens here doesn't match the failure message.  alice will be chosen via the realm module because of the fallback realm, not via the primary ccache.

The same goes for the later tests.  I don't think you have a remaining test where the realm module produces no result, so we no longer have coverage of the case where krb5_cc_select() returns no result when invoked by kg_cred_resolve().

src/lib/krb5/ccache/ccselect.c
```diff
@@ -132,14 +132,36 @@ krb5_cc_select(krb5_context context, krb5_principal server,
...
+                                           &fbrealms);
+        if (ret)
+            goto cleanup;
+
+        /* Make a copy and set the realm to the fallback realm. */
+        ret = krb5_copy_principal(context, server, &srvcp);
+        if (ret)
+            goto cleanup;
+
+        server = srvcp;
```
greghudson: Please move this assignment to server after calling krb5_set_principal_realm().  server is a read-only alias to a principal owned by someone else (either srvcp or the caller), and should not be used when modifying the principal.

mrogers950:
Updated, I was also able to restore the primary cache fallback test by using the refserver service.

PR #607, opened 2017-02-12
plugins/preauth/pkinit Fix build with LibreSSL

Sp1l:
- LibreSSL defines OPENSSL_VERSION_NUMBER 0x20000000L
  - Add check for LIBRESSL_VERSION_NUMBER for checks >= 1.1.0

See also HardenedBSD/hardenedbsd-ports@fc7b1529a02f3569f13e7a0d54ffbe952390361b
See also https://bugs.freebsd.org/217027


greghudson:
I've had some discussion with colleagues about whether applications should accomodate LibreSSL's choice to define OPENSSL_VERSION to correspond to release 2.0.  Several of them said that they shouldn't, in order to push back on a bad decision rather than bear the pain of that decision in application code.  I haven't come to a definite conclusion.

If we do accept this change, we probably want an intermediate symbol to indicate the OpenSSL 1.1 API, instead of a composite test in three different places.

PR #608, opened 2017-02-15
Some interposer test related fixes

frozencemetery:
This series is changes that I would like to have before proposing the interposer test plugin.  (That's in my interposer_test branch, but it doesn't all work yet; it gets through most of t_gssapi.py.)  gssproxy does not currently implement reentry for gss_import_name, so changes to this function will not break us.

Thanks!


doc/plugindev/gssapi.rst
```diff
@@ -90,10 +90,13 @@ formats are:
...
-  length in big-endian order, followed by the mechanism OID, followed
-  by the bytes 06, followed by the OID length as a single byte,
-  followed by the mechanism OID, followed by the mechanism token.
+* For **gss_import_name**, the bytes 04 01, followed by two plus the
+  interposed mechanism OID length in big-endian order, followed by the
+  byte 06, followed by the interposed OID length as a single byte,
+  followed by the interposed mechanism OID itself, followed by the
+  four byte big-endian length of the mechanism token, followed by the
+  mechanism token itself.  input_name_type must also be set to
+  GSS_C_NT_EXPORT_NAME.
```
greghudson: I think the two commits affecting this file would be better combined.  All of the documentation changes will probably be backported to 1.15 so that they show up in krb5-latest on the web.

Before these changes, the text above the bullet points says "using the concatenated OID" (which pretty clearly refers to the "special mechanism OID which is the concatenation of...") and the bullet points say "mechanism OID".  Note that the bullet points are introduced as describing the mechglue token formats, not directly describing what an interposer mechanism should put there.

After these changes, there is an inconsistent mix of "interposer mechanism OID", "interposed OID", and "interposed mechanism OID", none of which have been defined.  I don't think this is clearer.

I would accept changing "mechanism OID" to "concatenated OID" in the bullet points, and changing the text above the bullet points.

src/tests/gssapi/t_gssapi.py
```diff
@@ -143,22 +143,23 @@
 realm.run(['./t_inq_cred', '-k', '-b'], expected_msg=realm.host_princ)
 
 # Test gss_export_name behavior.
-out = realm.run(['./t_export_name', 'u:x'])
-if out != '0401000B06092A864886F7120102020000000D78404B5242544553542E434F4D\n':
+output = realm.run(['./t_export_name', 'u:x'])
```
greghudson: Can we use expected_msg for these?

(Using "output" as a variable name for the output of realm.run() is a historical botch, since there is a k5test.py function named "output".  So standardizing on "output" for these tests is worse than standardizing on "out", but using expected_msg would render the issue moot.)

src/lib/gssapi/mechglue/g_initialize.c
```diff
@@ -867,6 +867,10 @@ build_interMech(void *dl, const gss_OID mech_type)
 	RESOLVE_GSSI_SYMBOL(dl, mech, gssspi, _import_sec_context_by_mech);
 	RESOLVE_GSSI_SYMBOL(dl, mech, gssspi, _import_name_by_mech);
 	RESOLVE_GSSI_SYMBOL(dl, mech, gssspi, _import_cred_by_mech);
+	/* gss_get_mic_iov extensions (added 1.12, implementable 1.16) */
+	RESOLVE_GSSI_SYMBOL(dl, mech, gss, _get_mic_iov);
+	RESOLVE_GSSI_SYMBOL(dl, mech, gss, _verify_mic_iov);
+	RESOLVE_GSSI_SYMBOL(dl, mech, gss, _get_mic_iov_length);
```
greghudson: Don't these also need to be added to build_dynamicMech()?

We probably want to backport this to 1.15 if it's simply an omission from when I added the IOV mic functions.

src/lib/gssapi/mechglue/mglueP.h
```diff
@@ -82,16 +82,13 @@ int gssint_mechglue_initialize_library(void);
...
- * the routines in the glue module (gssd_mech_glue.c)
+ * This table is used to indirectly access mechanism specific versions of the
+ * gssapi routines through the routines in the glue module (g_glue.c)
  *
- * This contants all of the functions defined in gssapi.h except for
- * gss_release_buffer() and gss_release_oid_set(), which I am
- * assuming, for now, to be equal across mechanisms.
+ * This contains all of the functions defined in gssapi.h except for
+ * gss_release_buffer() and gss_release_oid_set(), which are assumed to be
+ * identical across mechanisms.
```
greghudson: I would combine these paragraphs, starting the second sentence with "It contains...".

doc/plugindev/gssapi.rst
```diff
@@ -48,16 +48,22 @@ This function is invoked with the OID of the interposer mechanism as
 specified in ``/etc/gss/mech`` or in a ``/etc/gss/mech.d/*.conf``
 file, and returns a set of mechanism OIDs to be interposed.  The
 returned OID set must have been created using the mechglue's
-gss_create_empty_oid_set and gss_add_oid_set_member functions.
+gss_create_empty_oid_set and gss_add_oid_set_member functions.  The
+function must be reentrant.
```
greghudson: Why must it be reentrant?

frozencemetery: Reading error; will remove.

src/tests/gssapi/t_gssapi.py
frozencemetery: Yes, but: we lose the explanation of what the tests are.  If we're fine with that I'll make the change.

frozencemetery:
Updated, and tagged the first two commits for 1.15.

PR #609, opened 2017-02-17
Add GSSAPI S4U documentation

greghudson:
Describe how a GSS application can perform S4U2Self and S4U2Proxy
requests using the MIT krb5 GSS library.  Also add a reference to RFC
7546 at the top, and fix a reference to gssapi_krb5.h.

### DIFF
--- a/doc/appdev/gssapi.rst
+++ b/doc/appdev/gssapi.rst
@@ -7,10 +7,13 @@ We recommend using the GSSAPI (or a higher-level framework which
 encompasses GSSAPI, such as SASL) for secure network communication
 over using the libkrb5 API directly.
 
-GSSAPIv2 is specified in :rfc:`2743` and :rfc:`2744`.  This
-documentation will describe how various ways of using GSSAPI will
-behave with the krb5 mechanism as implemented in MIT krb5, as well as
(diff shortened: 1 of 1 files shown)


doc/appdev/gssapi.rst
```diff
@@ -239,6 +242,73 @@ resulting token may be imported even if the original memory credential
...
+gss_cred_id_t object whose underlying credential cache contains the
+evidence ticket and a krbtgt ticket for the intermediate service.
+
+To acquire a proxy credential during client authentication, the
+service should first create an acceptor credential using the
+**GSS_C_BOTH** usage.  The application should then pass this
+credential as the *acceptor_cred_handle* to gss_accept_sec_context_,
+and also pass a *delegated_cred_handle* output parameter to receive a
+proxy credential containing the evidence ticket.
+
```
frenche: Since the gas_accept interface collides with plain delegetion, perhaps it should be noted that if the client has delegated its tgt then the delegated_cred_handle will contains that tgt and not proxy credentials which may more limited (afai recall from old testing).

greghudson: Sure, I will add text to indicate the possible resulting delegated_cred_handle values.

PR #610, opened 2017-02-17
PKINIT certificate authorization pluggable interface

mrogers950:
This PR implements the certauth plugin interface and builtin modules as outlined in https://k5wiki.kerberos.org/wiki/Projects/Certificate_authorization_pluggable_interface

### DIFF
--- a/src/include/krb5/certauth_plugin.h
+++ b/src/include/krb5/certauth_plugin.h
@@ -1,0 +1,103 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* include/krb5/certauth_plugin.h - certauth plugin header. */
+/*
+ * Copyright (C) 2017 by Red Hat, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * Certificate authorization plugin interface.  The PKINIT server module uses
+ * this interface to check client certificate attributes after the certificate
+ * signature has been verified.
+ */
+#ifndef KRB5_CERTAUTH_PLUGIN_H
+#define KRB5_CERTAUTH_PLUGIN_H
+
+#include <krb5/krb5.h>
+#include <krb5/plugin.h>
+
+/* Abstract module data type. */
+typedef struct krb5_certauth_moddata_st *krb5_certauth_moddata;
+
+typedef struct _krb5_db_entry_new krb5_db_entry;
+
+/*
+ * Optional: Initialize module data.
+ */
+typedef krb5_error_code
+(*krb5_certauth_init_fn)(krb5_context context,
+                         krb5_certauth_moddata *moddata_out);
+
+/*
+ * Optional: Clean up the module data.
+ */
+typedef void
+(*krb5_certauth_fini_fn)(krb5_context context, krb5_certauth_moddata moddata);
+
+/*
+ * Mandatory:
+ * Return 0 if the DER-encoded cert is authorized for PKINIT authentication by
+ * princ; otherwise return one of the following error codes:
+ * - KRB5KDC_ERR_CLIENT_NAME_MISMATCH - incorrect SAN value
+ * - KRB5KDC_ERR_INCONSISTENT_KEY_PURPOSE - incorrect EKU
+ * - KRB5KDC_ERR_CERTIFICATE_MISMATCH - other extension error
+ * - KRB5_PLUGIN_NO_HANDLE - the module has no opinion about cert
+ *
+ * - opts is used by built-in modules to receive internal data, and must be
+ *   ignored by other modules.
+ * - db_entry receives the client principal database entry, and can be ignored
+ *   by modules that do not link with libkdb5.
+ * - *authinds_out optionally returns a null-terminated list of authentication
+ *   indicator strings upon KRB5_PLUGIN_NO_HANDLE or accepted authorization.
+ */
+typedef krb5_error_code
+(*krb5_certauth_authorize_fn)(krb5_context context,
+                              krb5_certauth_moddata moddata,
+                              const uint8_t *cert, size_t cert_len,
+                              krb5_const_principal princ, const void *opts,
+                              const krb5_db_entry *db_entry,
+                              char ***authinds_out);
+
+/*
+ * Free indicators allocated by a module.  Mandatory if authorize returns
+ * authentication indicators.
+ */
+typedef void
+(*krb5_certauth_free_indicator_fn)(krb5_context context,
+                                   krb5_certauth_moddata moddata,
+                                   char **authinds);
+
+typedef struct krb5_certauth_vtable_st {
+    char *name;
+    krb5_certauth_init_fn init;
+    krb5_certauth_fini_fn fini;
+    krb5_certauth_authorize_fn authorize;
+    krb5_certauth_free_indicator_fn free_ind;
+} *krb5_certauth_vtable;
+
+#endif /* KRB5_CERTAUTH_PLUGIN_H */
--- a/src/plugins/certauth/test/main.c
+++ b/src/plugins/certauth/test/main.c
@@ -1,0 +1,209 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* plugins/certauth/main.c - certauth plugin test modules. */
+/*
+ * Copyright (C) 2017 by Red Hat, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <k5-int.h>
+#include "krb5/certauth_plugin.h"
+
+struct krb5_certauth_moddata_st {
+    int initialized;
+};
+
+/* Test module 1 returns OK with an indicator. */
+static krb5_error_code
+test1_authorize(krb5_context context, krb5_certauth_moddata moddata,
+                const uint8_t *cert, size_t cert_len,
+                krb5_const_principal princ, const void *opts,
+                const krb5_db_entry *db_entry, char ***authinds_out)
+{
+    char **ais = NULL;
+
+    ais = calloc(2, sizeof(*ais));
+    assert(ais != NULL);
+    ais[0] = strdup("test1");
+    assert(ais[0] != NULL);
+    *authinds_out = ais;
+    return KRB5_PLUGIN_NO_HANDLE;
+}
+
+static void
+test_free_ind(krb5_context context, krb5_certauth_moddata moddata,
+              char **authinds)
+{
+    size_t i;
+
+    if (authinds == NULL)
+        return;
+    for (i = 0; authinds[i] != NULL; i++)
+        free(authinds[i]);
+    free(authinds);
+}
+
+/* A basic moddata test. */
+static krb5_error_code
+test2_init(krb5_context context, krb5_certauth_moddata *moddata_out)
+{
+    krb5_certauth_moddata mod;
+
+    mod = calloc(1, sizeof(*mod));
+    assert(mod != NULL);
+    mod->initialized = 1;
+    *moddata_out = mod;
+    return 0;
+}
+
+static void
+test2_fini(krb5_context context, krb5_certauth_moddata moddata)
+{
+    free(moddata);
+}
+
+/* Return true if cert appears to contain the CN name, based on a search of the
+ * DER encoding. */
+static krb5_boolean
+has_cn(krb5_context context, const uint8_t *cert, size_t cert_len,
+       const char *name)
+{
+    krb5_boolean match = FALSE;
+    uint8_t name_len, cntag[5] = "\x06\x03\x55\x04\x03";
+    const uint8_t *c;
+    struct k5buf buf;
+    size_t c_left;
+
+    /* Construct a DER search string of the CN AttributeType encoding followed
+     * by a UTF8String encoding containing name as the AttributeValue. */
+    k5_buf_init_dynamic(&buf);
+    k5_buf_add_len(&buf, cntag, sizeof(cntag));
+    k5_buf_add(&buf, "\x0C");
+    assert(strlen(name) < 128);
+    name_len = strlen(name);
+    k5_buf_add_len(&buf, &name_len, 1);
+    k5_buf_add_len(&buf, name, name_len);
+    assert(k5_buf_status(&buf) == 0);
+
+    /* Check for the CN needle in the certificate haystack. */
+    c_left = cert_len;
+    c = memchr(cert, *cntag, c_left);
+    while (c != NULL) {
+        c_left = cert_len - (c - cert);
+        if (buf.len > c_left)
+            break;
+        if (memcmp(c, buf.data, buf.len) == 0) {
+            match = TRUE;
+            break;
+        }
+        assert(c_left >= 1);
+        c = memchr(c + 1, *cntag, c_left - 1);
+    }
+
+    k5_buf_free(&buf);
+    return match;
+}
+
+/*
+ * Test module 2 returns OK if princ matches the CN part of the subject name,
+ * and returns indicators of the module name and princ.
+ */
+static krb5_error_code
+test2_authorize(krb5_context context, krb5_certauth_moddata moddata,
+                const uint8_t *cert, size_t cert_len,
+                krb5_const_principal princ, const void *opts,
+                const krb5_db_entry *db_entry, char ***authinds_out)
+{
+    krb5_error_code ret;
+    char *name = NULL, **ais = NULL;
+
+    *authinds_out = NULL;
+
+    assert(moddata != NULL && moddata->initialized);
+
+    ret = krb5_unparse_name_flags(context, princ,
+                                  KRB5_PRINCIPAL_UNPARSE_NO_REALM, &name);
+    if (ret)
+        goto cleanup;
+
+    if (!has_cn(context, cert, cert_len, name)) {
+        ret = KRB5KDC_ERR_CERTIFICATE_MISMATCH;
+        goto cleanup;
+    }
+
+    /* Create an indicator list with the module name and CN. */
+    ais = calloc(3, sizeof(*ais));
+    assert(ais != NULL);
+    ais[0] = strdup("test2");
+    ais[1] = strdup(name);
+    assert(ais[0] != NULL && ais[1] != NULL);
+    *authinds_out = ais;
+
+    ais = NULL;
+
+cleanup:
+    krb5_free_unparsed_name(context, name);
+    return ret;
+}
+
+krb5_error_code
+certauth_test1_initvt(krb5_context context, int maj_ver, int min_ver,
+                      krb5_plugin_vtable vtable);
+krb5_error_code
+certauth_test1_initvt(krb5_context context, int maj_ver, int min_ver,
+                      krb5_plugin_vtable vtable)
+{
+    krb5_certauth_vtable vt;
+
+    if (maj_ver != 1)
+        return KRB5_PLUGIN_VER_NOTSUPP;
+    vt = (krb5_certauth_vtable)vtable;
+    vt->name = "test1";
+    vt->authorize = test1_authorize;
+    vt->free_ind = test_free_ind;
+    return 0;
+}
+
+krb5_error_code
+certauth_test2_initvt(krb5_context context, int maj_ver, int min_ver,
+                      krb5_plugin_vtable vtable);
+krb5_error_code
+certauth_test2_initvt(krb5_context context, int maj_ver, int min_ver,
+                      krb5_plugin_vtable vtable)
+{
+    krb5_certauth_vtable vt;
+
+    if (maj_ver != 1)
+        return KRB5_PLUGIN_VER_NOTSUPP;
+    vt = (krb5_certauth_vtable)vtable;
+    vt->name = "test2";
+    vt->authorize = test2_authorize;
+    vt->init = test2_init;
+    vt->fini = test2_fini;
+    vt->free_ind = test_free_ind;
+    return 0;
+}
--- a/src/plugins/preauth/pkinit/pkinit_crypto.h
+++ b/src/plugins/preauth/pkinit/pkinit_crypto.h
@@ -96,7 +96,6 @@ typedef struct _pkinit_cert_iter_info *pkinit_cert_iter_handle;
 #define PKINIT_ITER_NO_MORE	0x11111111  /* XXX */
 
 typedef struct _pkinit_cert_matching_data {
-    pkinit_cert_handle ch;  /* cert handle for this certificate */
     char *subject_dn;	    /* rfc2253-style subject name string */
     char *issuer_dn;	    /* rfc2253-style issuer name string */
     unsigned int ku_bits;   /* key usage information */
@@ -458,68 +457,38 @@ krb5_error_code crypto_free_cert_info
 
 
 /*
- * Get number of certificates available after crypto_load_certs()
+ * Get a null-terminated list of certificate matching data objects for the
+ * certificates loaded in id_cryptoctx.
  */
-krb5_error_code crypto_cert_get_count
-	(krb5_context context,				/* IN */
-	pkinit_plg_crypto_context plg_cryptoctx,	/* IN */
-	pkinit_req_crypto_context req_cryptoctx,	/* IN */
-	pkinit_identity_crypto_context id_cryptoctx,	/* IN */
-	int *cert_count);				/* OUT */
+krb5_error_code
+crypto_cert_get_matching_data(krb5_context context,
+			      pkinit_plg_crypto_context plg_cryptoctx,
+			      pkinit_req_crypto_context req_cryptoctx,
+			      pkinit_identity_crypto_context id_cryptoctx,
+			      pkinit_cert_matching_data ***md_out);
 
 /*
- * Begin iteration over the certs loaded in crypto_load_certs()
+ * Free a matching data object.
  */
-krb5_error_code crypto_cert_iteration_begin
-	(krb5_context context,				/* IN */
-	pkinit_plg_crypto_context plg_cryptoctx,	/* IN */
-	pkinit_req_crypto_context req_cryptoctx,	/* IN */
-	pkinit_identity_crypto_context id_cryptoctx,	/* IN */
-	pkinit_cert_iter_handle *iter_handle);		/* OUT */
+void
+crypto_cert_free_matching_data(krb5_context context,
+			       pkinit_cert_matching_data *md);
 
 /*
- * End iteration over the certs loaded in crypto_load_certs()
+ * Free a list of matching data objects.
  */
-krb5_error_code crypto_cert_iteration_end
-	(krb5_context context,				/* IN */
-	pkinit_cert_iter_handle iter_handle);		/* IN */
+void
+crypto_cert_free_matching_data_list(krb5_context context,
+				    pkinit_cert_matching_data **matchdata);
 
 /*
- * Get next certificate handle
+ * Choose one of the certificates loaded in idctx to use for PKINIT client
+ * operations.  cred_index must be an index into the array of matching objects
+ * returned by crypto_cert_get_matching_data().
  */
-krb5_error_code crypto_cert_iteration_next
-	(krb5_context context,				/* IN */
-	pkinit_cert_iter_handle iter_handle,		/* IN */
-	pkinit_cert_handle *cert_handle);		/* OUT */
-
-/*
- * Release cert handle
- */
-krb5_error_code crypto_cert_release
-	(krb5_context context,				/* IN */
-	pkinit_cert_handle cert_handle);		/* IN */
-
-/*
- * Get certificate matching information
- */
-krb5_error_code crypto_cert_get_matching_data
-	(krb5_context context,				/* IN */
-	pkinit_cert_handle cert_handle,			/* IN */
-	pkinit_cert_matching_data **ret_data);		/* OUT */
-
-/*
- * Free certificate information
- */
-krb5_error_code crypto_cert_free_matching_data
-	(krb5_context context,				/* IN */
-	pkinit_cert_matching_data *data);		/* IN */
-
-/*
- * Make the given certificate "the chosen one"
- */
-krb5_error_code crypto_cert_select
-	(krb5_context context,				/* IN */
-	pkinit_cert_matching_data *data);		/* IN */
+krb5_error_code
+crypto_cert_select(krb5_context context, pkinit_identity_crypto_context idctx,
+		   size_t cred_index);
 
 /*
  * Select the default certificate as "the chosen one"
--- a/src/plugins/preauth/pkinit/pkinit_crypto.h
+++ b/src/plugins/preauth/pkinit/pkinit_crypto.h
@@ -664,4 +664,8 @@ extern const size_t  krb5_pkinit_sha512_oid_len;
  */
 extern krb5_data const * const supported_kdf_alg_ids[];
 
+krb5_error_code
+crypto_encode_der_cert(krb5_context context, pkinit_req_crypto_context reqctx,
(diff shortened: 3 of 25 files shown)


greghudson:
I will have more later, but from a quick look:

* It looks like auth indicator strings are allocated by the certauth module and freed by the PKINIT module.  That won't work if they don't use the same allocator.

* I'd be happier if the test module didn't use the PKINIT object files.  It could use OpenSSL directly, or its own parsing, or it could do something zany like look for a binary substring of the DER.

* There is no point in having per-request moddata in this interface; a module gets no benefit out of receiving the same fixed sequence of req_init, authorize, and req_fini every time a cert comes in.  There should instead be a moddata object which lasts the lifetime of the KDC process, or perhaps one per realm.

* If there is a free method for the moddata object, it should accept the moddata object, not a pointer to it.

greghudson:
> There should instead be a moddata object which lasts the lifetime of the KDC process, or perhaps one per realm.

To provide a little more guidance here: certauth_modules should not be a global variable; it should be a linked from the PKINIT kdcpreauth moddata, and should be cleaned up when that moddata is finalized.  The PKINIT kdcpreauth moddata is currently an array of pkinit_kdc_context objects, one per realm.  So there are two options:

1. For each configured realm, load the certauth modules and create an array of certauth_handle objects with a long-term moddata for each module.  The module receives a moddata initialization for each configured realm.

2. Extend the PKINIT kdcpreauth moddata to be a structure containing the array of pkinit_kdc_context objects and also the array of certauth_handle objects.  The module receives a single moddata initialization for the whole KDC process.

Option 1 may seem inefficient, but the overwhelmingly common case is for there only to be one realm served by a KDC, and even when there are multiple realms, they are relatively heavyweight objects.

Option 2 means that any per-realm data (currently represented in "opts") needs to be passed to the authorize method rather than a moddata initialization method.  That might be more convenient for modules anyway.

mrogers950:
> It looks like auth indicator strings are allocated by the certauth module and freed by the PKINIT module. That won't work if they don't use the same allocator.

Would it be preferable to add a "free_indicator" method for this, or provide a buffer allocated by the PKINIT module?

> I'd be happier if the test module didn't use the PKINIT object files. It could use OpenSSL directly, or its own parsing, or it could do something zany like look for a binary substring of the DER.

That's fair. I don't think that we need to go too far into proper decoding and picking attributes out of the cert in the test module anyways, so checking the DER directly for the subject seems sufficient.

> Option 2 means that any per-realm data (currently represented in "opts") needs to be passed to the authorize method rather than a moddata initialization method. That might be more convenient for modules anyway.

This sounds like the better option to me as well.

greghudson:
> Would it be preferable to add a "free_indicator" method for this, or provide a buffer allocated by the PKINIT module?

Probably the former.

Another option is for the module to just call the add_auth_indicator callback in the kdcpreauth callbacks structure.  There is some potential for bad behavior there if one module adds an authentication indicator and another module says no.  At that point it's unlikely that the KDC would issue a ticket because PKINIT will fail, so I'm not sure it's worth worrying about.

On an unrelated note, I was looking at the PR's server_authorize_cert() and it doesn't seem like it provides a way for a module to say "pass" as described in the wiki page.

mrogers950:
> On an unrelated note, I was looking at the PR's server_authorize_cert() and it doesn't seem like it provides a way for a module to say "pass" as described in the wiki page.

If failure happens with any of the modules returning a FALSE status, is there a reason to make a distinction between  a TRUE status and "pass"? I understood "pass" as meaning the module did not perform any authorization checks, for which it can just return TRUE.

greghudson:
> I understood "pass" as meaning the module did not perform any authorization checks, for which it can just return TRUE.

The desired accumulator behavior (which we use for several existing authorization pluggable interfaces) is that at least one module must say yes, and zero modules must say no.  If a module can't give distinct "yes" and "pass" results, then there is no way to know whether any of the modules had a positive authorization result, or if they all just had nothing to say about the cert.

mrogers950:
Updated with a round of changes:
- Changed the kdcpreauth_moddata to hold realm_contexts and certauth_modules, and load certauth module handles into the moddata.  Some of the existing init functions were revised to accommodate this.
- Changed req_init and req_fini methods to load-and-unload time fini/init. The builtin specific void *opts for the realm and request data is passed to the authorize function.
- With regards to a module’s use of add_auth_indicator(), It seemed to be simpler to provide an indicator freeing method, rather than require a custom module to depend on kdcpreauth callbacks if it wants to add an indicator.  So I changed server_authorize_cert() to use cb->add_auth_indicator() after the authorize method call for gathering module indicators, and then call the free_ind method each time.
- Use the accumulator pattern for calling the authorize method: authorize no longer has *status and returns 0 for “yes”, KRB5_PLUGIN_NO_HANDLE for “pass”, and KRB5KDC_ERR_CERTIFICATE_MISMATCH/KRB5KDC_ERR_CLIENT_NAME_MISMATCH or other errors for “no”. 
- The test module now checks a certificate using a DER substring.
- Combined the pkinit_srv.c module commits (made change rebasing easier), and reworded doc text and commits.

mrogers950:
Just fixed a typo and made some alignment adjustments in plugins/certauth/test/main.c.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,293 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+    list = NULL;
+
+cleanup:
+    k5_plugin_free_modules(context, modules);
+    free_certauth_handles(context, list);
+    return ret;
+}
+
+static krb5_error_code
+certauth_authorize(krb5_context context, krb5_certauth_moddata moddata,
```
greghudson: This function name seems off; it's for the pkinitsrv certauth module but its name suggests it's associated with the certauth interface itself.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,293 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+    vt->fini = NULL;
+    vt->free_ind = NULL;
+    return 0;
+}
+
+/*
+ * Do certificate auth based on match expressions in the pkinit_cert_match
+ * attribute string.  An expression should be in the same form as those used for
+ * the pkinit_cert_match configuration option.  Multiple expressions are
+ * separated by ':'.  The certificate is authorized if all expressions are
```
greghudson: Do we need support for multiple expressions?  An expression can already contain multiple rules whose results can either be and-ed or or-ed together.  If we have a use case for an and-of-ors, that functionality is probably best moved into pkinit_matching.c.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,293 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+                                   "pkinit_cert_match", &match);
+    if (ret)
+        goto out;
+
+    /* Pass if there's no match expression. */
+    if (match == NULL) {
+        ret = KRB5_PLUGIN_NO_HANDLE;
+        goto out;
+    }
+    matchcpy = k5memdup0(match, strlen(match), &ret);
```
greghudson: Possibly moot, but this is more clearly written using strdup(), even though that doesn't set ret.  (I've considered adding a k5strdup helper for that reason, but haven't done so yet.)

src/include/krb5/certauth_plugin.h
```diff
@@ -0,0 +1,99 @@
...
+/*
+ * Return 0 if the DER encoded cert is authorized for PKINIT
+ * authentication by princ, otherwise return one of the following error codes:
+ * - KRB5KDC_ERR_CLIENT_NAME_MISMATCH  - Incorrect SAN value
+ * - KRB5KDC_ERR_INCONSISTENT_KEY_PURPOSE - Incorrect EKU
+ * - KRB5KDC_ERR_CERTIFICATE_MISMATCH - Other extension error
+ * - KRB5_PLUGIN_NO_HANDLE - The module answers "passed" and had nothing to say
+ *   about the cert.
+ *
+ * opts is used by the builtin kdc modules.  db_entry receives the client
```
greghudson: We need a little more verbiage here.  "opts is used by built-in modules and should be ignored by other modules."

src/include/krb5/certauth_plugin.h
```diff
@@ -0,0 +1,99 @@
...
+ * after the certificate signature has been verified.  A module can be
+ * implemented in the same shared object as a kdb module and define an
+ * authorization policy on top of the builtin module, or override the builtin
+ * module with enable_only.
+ */
+
+/* Abstract module data type. */
+typedef struct krb5_certauth_moddata_st *krb5_certauth_moddata;
+
+/*
```
greghudson: Every pluggable interface method description should say whether it is optional or mandatory to implement.  See other interface headers for examples.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,220 @@
...
+     * check for princ as CN. */
+    c = malloc(sizeof(cntag) +  strlen(name) + 1);
+    if (c == NULL)
+        return FALSE;
+
+    memcpy(c, cntag, sizeof(cntag));
+    c[sizeof(cntag)] = strlen(name);
+    memcpy(c + sizeof(cntag) + 1, name, strlen(name));
+
+    cc = memmem(cert.data, cert.length, c, sizeof(cntag) + strlen(name) + 1);
```
greghudson: I think you'll want to use memchr() and memcmp() instead.  There will be a fiddly length check involved, but it shouldn't be a lot of code.

mrogers950:
- Rewrote has_cn() with k5buf and replaced memmem() with memchr/memcmp.
- Changed declarations of certauth initvt functions to static.
- Renamed certauth_authorize to pkinitsrv_authorize.
- Changed the suggested k5memdup0 to strdup. A k5strdup() would be nice for clarity over k5memdup0().
- Adjustments to certauth_plugin.h comments.

Since the pkinit_cert_match option allowed multiple expressions I just tried to keep the same behavior for the string attribute in some way, so there is no explicit requirement. Does the client certificate matching require multiple statements in a way that would not fit for the server's checking purposes? I'm OK with changing it to just use one expression if multiple are unnecessary.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,240 @@
...
+    ret = k5_buf_status(&buf);
+    if (ret)
+        goto out;
+
+    b_start = *(char *)buf.data;
+    b_len = buf.len;
+    c_len = cert.length;
+
+    /* Check for the CN needle in the certificate haystack. */
+    c = memchr(cert.data, b_start, c_len);
```
greghudson: What happens in the loop below if memchr() finds b_start at the last character of the cert?

greghudson:
I think pkinit_cert_matching() only supports multiple rule sets because the profile library naturally produces multiple results.  Since you can have several components in a rule set, I would say that we should start with support for just a single rule set in the string attribute.

mrogers950:
- Looking closer at the case of b_start at the end of a cert, this led to a memcmp() past bounds and an overflowed value passed to memchr().  Changing this to figure out the amount of certificate left to check (variable renamed to c_left) and comparing with b_len before doing the memcmp() handles this situation cleaner.
- Changed the dbmatch module to use just a single expression, and updated the related t_pkinit.py tests to exercise the matching code a bit further. Updated docs.

greghudson:
More partial feedback (sorry): we will need documentation in pkinit.rst describing how to use pkinit_cert_match (it can reference pkinit_cert_match in krb5_conf.rst but should give a simple example using a subject match).  The pkinit_cert_match string attribute should also be briefly described in kadmin_local.rst under set_string.

mrogers950:
While working on the doc text I thought it might make sense for the dbmatch module to have an option for a default matching expression that it could apply to all principals in a realm. This way when using the dbmatch module solo as enable_only, you would not need to apply a match expression attribute to each principal. Thoughts on this?

greghudson:
Wouldn't you need some way to substitute the principal name into a matching pattern, for that to be useful?  That starts to get complicated.

On a related note, I think the dbmatch module, as well as its documentation and tests, should be moved into a second commit.  That can happen after we've finished review on the code changes to simplify development.

mrogers950:
>Wouldn't you need some way to substitute the principal name into a matching pattern, for that to be useful? That starts to get complicated.

Without that, it could be useful for some broader cases, like say if you wanted the only cert restriction to be that the subject mentions a certain realm.  I figure that using dbmatch alone would be used as a common way around the standard policy imposed by the pkinitsrv module, and those doing so may see this as a problem.

> On a related note, I think the dbmatch module, as well as its documentation and tests, should be moved into a second commit. That can happen after we've finished review on the code changes to simplify development.

That's fine with me.

mrogers950:
Updated with added doc text.

greghudson:
> Without that, it could be useful for some broader cases, like say if you wanted the only cert restriction to be that the subject mentions a certain realm. I figure that using dbmatch alone would be used as a common way around the standard policy imposed by the pkinitsrv module, and those doing so may see this as a problem.

If you were to set a default match expression of, say, "<SUBJECT>.*@MYREALM", then the dbmatch module would say "yes" for any cert with a subject containing MYREALM, even if the user in the cert bears no relationship to the user in the request, wouldn't it?  That doesn't seem like a useful policy.

I could see it being useful to be able to supply a default expression which is necessary to match but not sufficient, but then you would still need some criteria to judge a certificate as sufficient.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,239 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* main.c - PKINIT certificate authorization plugin test modules.
```
greghudson: The comment formatting here does not exactly match what's in src/prototype/prototype.c.  That's actually important for some automated processing tools used by Oracle.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,239 @@
...
+        return ENOMEM;
+    }
+
+    *authinds_out = ais;
+    return 0;
+}
+
+static void
+test_free_ind(krb5_context context, krb5_certauth_moddata moddata,
+                                    char **authinds)
```
greghudson: The style checker reports bad indentation here; are you running it with emacs installed?

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,239 @@
...
+        return ENOMEM;
+    mod->initialized = 1;
+    *moddata_out = mod;
+    return 0;
+}
+
+static void
+test2_fini(krb5_context context, krb5_certauth_moddata moddata)
+{
+   free(moddata);
```
greghudson: More bad indentation here.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,239 @@
...
+
+    *authinds_out = ais;
+    return 0;
+}
+
+static void
+test_free_ind(krb5_context context, krb5_certauth_moddata moddata,
+                                    char **authinds)
+{
+    size_t i;
```
greghudson: Add a blank line here.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,239 @@
...
+        free(authinds[i]);
+    }
+    free(authinds);
+}
+
+/* A basic moddata test. */
+static krb5_error_code
+test2_init(krb5_context context, krb5_certauth_moddata *moddata_out)
+{
+    krb5_certauth_moddata mod;
```
greghudson: Add a blank line here.

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,239 @@
...
+
+    b_start = *(char *)buf.data;
+    b_len = buf.len;
+    c_left = cert.length;
+
+    /* Check for the CN needle in the certificate haystack. */
+    c = memchr(cert.data, b_start, c_left);
+    while (c != NULL) {
+        c_left = cert.length - ((c + 1) - cert.data);
+        if (b_len > c_left)
```
greghudson: Does this work if the needle is present at the last possible position in the haystack?

src/plugins/preauth/pkinit/pkinit_crypto.h
```diff
@@ -664,4 +664,26 @@ extern const size_t  krb5_pkinit_sha512_oid_len;
...
+                              pkinit_plg_crypto_context plgctx,
+                              pkinit_req_crypto_context reqctx,
+                              pkinit_cert_matching_data **md_out);
+
+extern krb5_error_code
+pkinit_client_cert_match(krb5_context context,
+                         pkinit_plg_crypto_context plgctx,
+                         pkinit_req_crypto_context reqctx,
+                         const char *match_rule,
+                         krb5_boolean *matched);
```
greghudson: The use of blank lines in these prototypes is inconsistent.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,284 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
     return 0;
 }
+
+/*
+ * Run the received, verified certificate through the certauth plugin interface,
+ * to verify that client can use the certificate based on the extension policy
+ * defined by modules.
+ */
+static krb5_error_code
+server_authorize_cert(krb5_context context, certauth_handle *certauth_modules,
```
greghudson: Do we need the server_ prefix on this name?  We're in pkinit_srv.c and this is a static function.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,284 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+
+cleanup:
+    crypto_free_cert_data(context, cert);
+    return ret;
+}
+
+static krb5_error_code
+pkinitsrv_authorize(krb5_context context, krb5_certauth_moddata moddata,
+                   krb5_data cert, krb5_principal princ, void *opts,
+                   void *db_entry, char ***authinds_out)
```
greghudson: I don't think we're feeding the right things into the accumulator here, with the result that administrators will have to use module configuration to disable the pkinitsrv module when they shouldn't have to.

If the EKU is not found, I think we should say no (KRB5KDC_ERR_INCONSISTENT_KEY_PURPOSE).  An admin can disable EKU checking in configuration if desired.

If the SAN is not found, I think we should pass.  The certificate may still match the client principal according to another module's criteria, and there is no way to disable SAN checking in configuration.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,284 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+            if (hd->vt.free_ind != NULL)
+                hd->vt.free_ind(context, hd->moddata, ais);
+            else
+                free_ai_list(ais);
+            ais = NULL;
+        }
+    }
+
+    if (accepted)
+        ret = 0;
```
greghudson: And otherwise we return KRB5_PLUGIN_NO_HANDLE?  I don't think that's meaningful to the caller.  I think we should return KRB5KDC_ERR_CLIENT_NAME_MISMATCH, in that no module has asserted a match between the certificate and the client principal name.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,284 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+
+    /* Check the certificate against the match expression. */
+    ret = pkinit_client_cert_match(context, req_opts->plgctx->cryptoctx,
+                                   req_opts->reqctx->cryptoctx, matchcpy,
+                                   &matched);
+    if (ret)
+        goto out;
+
+    if (!matched)
+        ret = KRB5KDC_ERR_CERTIFICATE_MISMATCH;
```
greghudson: I'm not sure if this should be a "no", but leave it alone for now; there's a reasonable argument that if the string attribute is present, it should be necessary as well as sufficient.

mrogers950:
- In certauth_plugin.h, add the krb5_db_entry typedef, change method arguments and comments.
- Fixed certauth/test/Makefile.in
- Use assert() in test modules, fix top comment.
- Indentation (my own fault rather than vim) and space fixes.
- Renamed server_authorize_cert() to authorize_cert()
- Make authorize_cert() return KRB5KDC_ERR_CLIENT_NAME_MISMATCH on no module match.

> I'm not sure why one would want to put wildcards in this regexp. And actually, I think the regexp match isn't anchored (can you confirm whether or not that's the case?) so I think you would want \<SUBJECT\>^user@REALM$.

You might be confusing it with \<SAN\> which would make sense with an anchored match. For that example I was thinking of a cert subject that also contains OU, CN=, etc., the tests also run this same \<SUBJECT\> regexp.

> It isn't clear from this documentation what certificate criteria are necessary and sufficient when a pkinit_cert_match attribute is present.

I revised this part a bit. If it’s still not clear some suggestions would be welcome.

> Does this work if the needle is present at the last possible position in the haystack?

Actually no; the offset-by-one was wrong between the c_left assignment and the in-loop memchr(), so my update corrects this (although I don't think we would see this with our test certs).

> I don't think we're feeding the right things into the accumulator here, with the result that administrators will have to use module configuration to disable the pkinitsrv module when they shouldn't have to.
If the EKU is not found, I think we should say no (KRB5KDC_ERR_INCONSISTENT_KEY_PURPOSE). An admin can disable EKU checking in configuration if desired.
If the SAN is not found, I think we should pass. The certificate may still match the client principal according to another module's criteria, and there is no way to disable SAN checking in configuration.

In order for the SAN check to be able to return pass independently from the EKU check,  I split them into separate pkinitsrv modules, and made verify_client_san() return ENOENT when no SAN exists.

greghudson:
> You might be confusing it with <SAN> which would make sense with an anchored match. For that example I was thinking of a cert subject that also contains OU, CN=, etc., the tests also run this same <SUBJECT> regexp.

If I am right that the code does an unanchored regexp match, then the leading and trailing `.*` wildcards add nothing to the regexp.

More importantly, simply looking for the substring `user@realm` in the subject would also match `anotheruser@realm` which is a clear escalation of privilege issue.

mrogers950:
> If I am right that the code does an unanchored regexp match, then the leading and trailing .* wildcards add nothing to the regexp.
More importantly, simply looking for the substring user@realm in the subject would also match anotheruser@realm which is a clear escalation of privilege issue.

I see, so I've updated the doc text and the test regexps to what should be more sensible examples.

greghudson:
> I revised this part a bit. If it’s still not clear some suggestions would be welcome.

I suggest the following:

    By default, the KDC requires PKINIT client certificates to have the standard Extended Key
    Usage and Subject Alternative Name attributes for PKINIT.  Starting in release 1.16, it is
    possible to authorize client certificates based on the subject or other criteria instead of the
    standard PKINIT Subject Alternative Name, by setting the **pkinit_cert_match** string
    attribute on each client principal entry.  For example::

        kadmin set_string user@REALM pkinit_cert_match "<SUBJECT>CN=user@REALM$"

    The **pkinit_cert_match** string attribute follows the syntax used by the :ref:`krb5.conf(5)`
    **pkinit_cert_match** relation.  To allow the use of non-PKINIT client certificates, it will also
    be necessary to disable key usage checking using the **pkinit_eku_checking** relation; for
    example::

        [kdcdefaults]
            pkinit_eku_checking = none

src/include/krb5/certauth_plugin.h
```diff
@@ -0,0 +1,103 @@
...
+ * - opts is used by built-in modules to receive internal data, and must be
+ *   ignored by other modules.
+ * - db_entry receives the client principal database entry and can be ignored by
+ *   modules that do not link with libkdb5.
+ * - *authinds_out optionally returns a NULL terminated list of authentication
+ *   indicator strings upon "passed" or accepted authorization.
+ */
+typedef krb5_error_code
+(*krb5_certauth_authorize_fn)(krb5_context context,
+                              krb5_certauth_moddata moddata, krb5_data cert,
```
greghudson: We don't like passing structure types as function parameters or return values (as stated in http://k5wiki.kerberos.org/wiki/Coding_style/Practices).  I recommend using separate "const uint8_t *cert" and "size_t len" parameters; I think that will be easier for most consumers to deal with than a krb5_data object anyway.

src/include/krb5/certauth_plugin.h
```diff
@@ -0,0 +1,103 @@
...
+                              krb5_principal princ, void *opts,
+                              krb5_db_entry *db_entry,
```
greghudson: db_entry should probably be of type `const krb5_db_entry *`.  princ should also be of type krb5_const_principal, and opts should be of type const void * (which means that in the built-in consumers, it should be cast to a const pointer to the options structure).

src/plugins/certauth/test/main.c
```diff
@@ -0,0 +1,227 @@
...
+
+static void
+test_free_ind(krb5_context context, krb5_certauth_moddata moddata,
+              char **authinds)
+{
+    size_t i;
+
+    if (authinds == NULL)
+        return;
+    for (i = 0; authinds[i] != NULL; i++) {
```
greghudson: No braces here.

mrogers950:
Updated:
- Constify method arguments. Use separate *cert and cert_len arguments and make related adjustments in pkinit_srv.c.
- Make buf and data type adjustments to has_cn(), add name length assert().
- Fix testing asserts.
- Correct the test1 module to return pass now that the framework has the accumulator behavior, since test1 is intended to test the behavior of a passed module asserting indicators.
- Use suggested text for pkinit.rst.
- Fix manpage ref in kadmin_local.rst.
- Remove referring to pointer as opaque, changed "KDB" to say "libkdb5" and adjusted the builtin module text in certauth.rst.

> But if the cert we're authorizing is in received_cert, why decode it again in each built-in callback? Why not just use received_cert as is?

Good point, thinking about it now this was only due to my initial idea of how the interface was to be designed (and that initially I used the pkinit objects during testing). Now that it has changed, I see the decoding part is unnecessary. I've removed the decoding step from the builtins and assuming we do not care to export any OpenSSL decoding for use by custom modules, I removed the decoding functions.

greghudson:
> On a related note, I think the dbmatch module, as well as its documentation and tests, should be moved into a second commit. That can happen after we've finished review on the code changes to simplify development.

What I see now is a breakdown into five commits, and no separation of dbmatch from the certauth interface.  The second commit contains support functions used only by dbmatch; the third commit contains the dbmatch module along with the certauth interface consuming code; the fourth commit contains dbmatch tests along with certauth tests; the fifth commit contains dbmatch documentation mixed in with certauth documentation.

What I would like to see is a first commit containing the certauth interface, with its tests and its documentation, as if we had never conceived of dbmatch as a clever piece of added functionality, and then a second commit containing the dbmatch feature, with its tests and its documentation.

mrogers950:
Updated with revised commits.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+     * an error code other than KRB5_PLUGIN_NO_HANDLE (pass).  Add indicators
+     * from modules that return 0 or pass.
+     */
+    ret = KRB5_PLUGIN_NO_HANDLE;
+    for (i = 0; certauth_modules != NULL && certauth_modules[i] != NULL; i++) {
+        hd = certauth_modules[i];
+        if (hd->vt.authorize == NULL)
+            continue;
+
+        ret = hd->vt.authorize(context, hd->moddata, (const uint8_t *)cert.data,
```
greghudson: Likely moot due to other changes, but it's only necessary to cast to (uint8_t *) here.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+
+        if (ais != NULL) {
+            /* Assert authentication indicators from certificate authorization. */
+            for (ai = ais; *ai != NULL; ai++) {
+                ret = cb->add_auth_indicator(context, rock, *ai);
+                if (ret)
+                    goto cleanup;
+            }
+
+            if (hd->vt.free_ind != NULL)
```
greghudson: This method is documented as mandatory if indicators are returned; providing a fallback doesn't seem like a good idea.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -51,6 +79,47 @@ pkinit_find_realm_context(krb5_context context,
                           krb5_kdcpreauth_moddata moddata,
                           krb5_principal princ);
 
+static void
+free_ai_list(char **ai_total)
+{
+    size_t i;
+    if (ai_total == NULL)
+        return;
+    for (i = 0; ai_total[i] != NULL; i++) {
```
greghudson: No braces here.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -51,6 +79,47 @@ pkinit_find_realm_context(krb5_context context,
                           krb5_kdcpreauth_moddata moddata,
                           krb5_principal princ);
 
+static void
+free_ai_list(char **ai_total)
+{
+    size_t i;
```
greghudson: Add a blank line after declarations.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -51,6 +79,47 @@ pkinit_find_realm_context(krb5_context context,
                           krb5_kdcpreauth_moddata moddata,
                           krb5_principal princ);
 
+static void
+free_ai_list(char **ai_total)
```
greghudson: "ai_total" seems like a relic from when this was part of another piece of code.  "list" would be more natural.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -51,6 +79,47 @@ pkinit_find_realm_context(krb5_context context,
...
+    for (i = 0; ai_total[i] != NULL; i++) {
+        free(ai_total[i]);
+    }
+    free(ai_total);
+}
+
+static void
+free_realm_contexts(krb5_context context, pkinit_kdc_context *realm_contexts)
+{
+    int i;
```
greghudson: Add a blank line after declarations.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1405,17 +1475,25 @@ static void
...
+    if (moddata == NULL)
         return;
 
-    for (i = 0; realm_contexts[i] != NULL; i++) {
-        pkinit_server_plugin_fini_realm(context, realm_contexts[i]);
+    realm_contexts = moddata->realm_contexts;
+    certauth_modules = moddata->certauth_modules;
+
+    if (realm_contexts != NULL) {
+        for (i = 0; realm_contexts[i] != NULL; i++) {
```
greghudson: No braces here.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+pkinitsrv_san_authorize(krb5_context context, krb5_certauth_moddata moddata,
+                        const uint8_t *cert, size_t cert_len,
+                        krb5_const_principal princ, const void *opts,
+                        const krb5_db_entry *db_entry, char ***authinds_out)
+{
+    krb5_error_code ret;
+    int valid_san;
+    const struct certauth_req_opts *req_opts = opts;
+
+    if (authinds_out != NULL)
```
greghudson: No need to test for nullity here; it's not an optional return parameter.  (Also applies to eku module.)

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+    if (ret == ENOENT) {
+        ret = KRB5_PLUGIN_NO_HANDLE;
+        goto out;
+    }
+
+    if (ret)
+        goto out;
+
+    if (!valid_san) {
+        pkiDebug("%s: did not find an acceptable SAN in user "
```
greghudson: This string constant now fits on one line.  (Also applies to EKU module.)

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+                 "certificate\n", __FUNCTION__);
+        ret = KRB5KDC_ERR_CLIENT_NAME_MISMATCH;
+    }
+
+out:
```
greghudson: If there's nothing to put in the cleanup handler, don't use a cleanup label; just have earlier return points.  (Also applies to EKU module.)

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+    krb5_certauth_vtable vt;
+
+    if (maj_ver != 1)
+        return KRB5_PLUGIN_VER_NOTSUPP;
+    vt = (krb5_certauth_vtable)vtable;
+    vt->name = "pkinitsrv";
+    vt->authorize = pkinitsrv_san_authorize;
+    vt->init = NULL;
+    vt->fini = NULL;
+    vt->free_ind = NULL;
```
greghudson: It's not our practice to explicitly set methods to NULL in initvt functions; not having to do so is a nice benefit of using a zero-filled vtable provided by the caller.  (Also applies to the EKU module, the test1 module, and the dbmatch module.)

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+    return 0;
+}
+
+/*
+ * Do certificate auth based on match expressions in the pkinit_cert_match
+ * attribute string.  An expression should be in the same form as those used for
+ * the pkinit_cert_match configuration option.  Multiple expressions are
+ * separated by ':'.  The certificate is authorized if all expressions are
```
greghudson: This comment still needs to be updated now that the code supports only one expression.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -1489,3 +1567,316 @@ kdcpreauth_pkinit_initvt(krb5_context context, int maj_ver, int min_ver,
...
+
+        ret = (*mod)(context, 1, 1, (krb5_plugin_vtable)&h->vt);
+        if (ret) {
+            free(h);
+            continue;
+        }
+        h->moddata = NULL;
+        if (h->vt.init != NULL) {
+            ret = h->vt.init(context, &h->moddata);
+            if (ret) {
```
greghudson: I'd like to see a trace log here; see the localauth, hostrealm, or ccselect consumers for examples.  (Note that PKINIT trace log definitions are in pkinit_trace.h, though.)

mrogers950:
- Update pkinit_crypto.h as a whole to conform to current style with spaced tabs. I made this a separate commit following the first two since it's a large formatting-only change, if that's OK to throw in here.
- Rename and revise SAN and EKU modules as suggested.
- Changed crypto_encode_der_cert() output arguments. 
- Remove the free_ai_list() fallback and function.
- Remove initvt NULL assignments.
- Update dbmatch comment.
- Fix free_realm_contexts() declarations.
- Add trace functions and use them during module loading.

greghudson:
Why make a copy of the get_string() result in dbmatch_authorize()?

greghudson:
I'm going to push some adjusted commits to the PR branch to, among various miscellaneous edits:

* change has_cn() to return a boolean value
* move documentation for the builtin modules to krb5_conf.rst for consistency with other interfaces
* use expected_msg in t_certauth.py
* reorder the new pkinit_srv.c code to avoid needing forward prototypes
* allocate our own memory in crypto_encode_der_cert() so we can free it with free(), as we do in all the other places where we use OpenSSL i2d functions
* eliminate a copy in dbmatch_authorize() which is no longer needed
* prototype pkinit_client_cert_match() in pkinit.h instead of pkinit_crypto.h
* avoid using k5-int.h in pkinit_crypto_openssl.c

I will also be removing the third reformatting commit.  It changes the license statement, which is not necessary okay, and this PR is too involved to include out-of-scope commits.

greghudson:
I still haven't delved fully into the new matching code, but I am concerned about the memory management in crypto_req_cert_matching_data().  It allocates cd, fills it in, passes it to crypto_cert_get_matching_data() which makes an alias within the resulting md, and then frees it while md is still live.  Am I missing something or does this create a dangling pointer?

mrogers950:
> I still haven't delved fully into the new matching code, but I am concerned about the memory management in crypto_req_cert_matching_data(). It allocates cd, fills it in, passes it to crypto_cert_get_matching_data() which makes an alias within the resulting md, and then frees it while md is still live. Am I missing something or does this create a dangling pointer?

Yes, so technically there is a dangling pointer in md (and this would also apply for the ci pointer) however those two aliases of md->ch and md->cred are not used during the component_match() calls in pkinit_client_cert_match(), and md lasts until crypto_cert_free_matching_data(). Note that crypto_cert_free_matching_data() does not clean up md->ch or md->cred.  There may be a cleaner way to handle the md here, it might need changing crypto_cert_free_matching_data() to be able to (optionally) release md->ch/md->cred.

greghudson:
I see how that pointer is only used by crypto_cert_select() which we aren't calling.  I think the way this code uses the existing matching interfaces creates too much technical debt.  To make the matching interface properly usable for the server code, I think we need to:

1. Remove the ch field from pkinit_cert_matching_data and make the existing caller track the cert handle separately.
2. Change crypto_cert_select() to take a pkinit_cert_handle.
3. From crypto_cert_get_matching_data(), factor out a function which produces matching data from a cert (given a plgctx and reqctx object as separate parameters) and use that interface in crypto_req_cert_matching_data().

For step 1, there are a couple of approaches.  We could create an array of cert_handle/matching_data pairs in obtain_all_cert_matching_data().  Another approach might be to replace crypto_cert_iteration_begin()/crypto_cert_iteration_next() with a function that returns an array of cert handles (this would be substantially less code) and maintain two parallel arrays with the same indexes.

These internal changes should be in a separate commit that doesn't introduce any new functionality or change any existing behavior.

greghudson:
> For step 1, there are a couple of approaches. We could create an array of cert_handle/matching_data pairs in obtain_all_cert_matching_data(). Another approach might be to replace crypto_cert_iteration_begin()/crypto_cert_iteration_next() with a function that returns an array of cert handles (this would be substantially less code) and maintain two parallel arrays with the same indexes.

Actually, I think we can go simpler still: get rid of cert handles entirely, and just ask the crypto code for an array of matching data objects for an id_cryptoctx (with a plgctx and reqctx as additional parameters for for the SAN/EKU functions).  Then select a cert by id_cryptoctx and index.

mrogers950:
I've updated with a new commit implementing those changes.

greghudson:
I flipped the second and third commit, and made some minor edits.  I will make a final review pass probably tomorrow, and hopefully merge this.

PR #612, opened 2017-02-28
Fix udp_preference_limit with SRV records

greghudson:
In sendto_kdc:resolve_server() when resolving a server entry with a
specified transport, defer the resulting addresses if the strategy
dictates that the specified transport is not preferred.  Reported by
Jochen Hein.

### DIFF
--- a/src/lib/krb5/os/sendto_kdc.c
+++ b/src/lib/krb5/os/sendto_kdc.c
@@ -791,7 +791,7 @@ resolve_server(krb5_context context, const krb5_data *realm,
     struct server_entry *entry = &servers->servers[ind];
     k5_transport transport;
     struct addrinfo *addrs, *a, hint, ai;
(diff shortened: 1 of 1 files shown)


greghudson:
I considered two other possible ways to express the new code which might be clearer (opinions welcome):

    transport = entry->transport;
    defer = ((entry->transport == TCP && strategy == UDP_FIRST) ||
                  (entry->transport == UDP && strategy == UDP_LAST));

which just adds a set of parentheses to align the clauses, also consistent with the previous assignment to defer which puts parentheses around a boolean expression for clarity.  Or:

    if (entry->transport == TCP || entry->transport == UDP)
        defer = (entry->transport != transport);
    transport = entry->transport;

This way uses the exact same assignment as we do above for module entries, and maybe makes it clearer that deferral is something we only do for TCP and UDP transports.

PR #613, opened 2017-03-04
Fix segfault in gss_get_name_attribute() call

frenche:
In krb5_authdata_get_attribute(), check value and
display_value arguments before dereferencing them.

Signed-off-by: Isaac Boukris <iboukris@gmail.com>

### DIFF
--- a/src/lib/gssapi/krb5/naming_exts.c
+++ b/src/lib/gssapi/krb5/naming_exts.c
@@ -319,8 +319,8 @@ krb5_gss_get_name_attribute(OM_uint32 *minor_status,
     krb5_data kattr;
     krb5_boolean kauthenticated;
     krb5_boolean kcomplete;
-    krb5_data kvalue;
-    krb5_data kdisplay_value;
+    krb5_data kvalue = empty_data();
+    krb5_data kdisplay_value = empty_data();
 
     if (minor_status != NULL)
         *minor_status = 0;
@@ -355,8 +355,8 @@ krb5_gss_get_name_attribute(OM_uint32 *minor_status,
                                        &kattr,
                                        &kauthenticated,
                                        &kcomplete,
-                                       value ? &kvalue : NULL,
-                                       display_value ? &kdisplay_value : NULL,
+                                       &kvalue,
+                                       &kdisplay_value,
                                        more);
     if (code == 0) {
         if (value != NULL)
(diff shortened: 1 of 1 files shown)


frenche:
To reproduce, just call gss_get_name_attribute() with either value or dispaly_value as GSS_C_NO_BUFFER.
Thanks.

greghudson:
I can't find any indication in RFC 6680 that it's allowed to pass GSS_C_NO_BUFFER for value or display_value to gss_get_name_attribute().  It does seem like it might be convenient to be able to do so, as the caller probably only cares about one or the other.

There are also some unconditional dereferences of value in authdata module get_attribute methods (authind_get_attribute, s4u2proxy_get_attribute), although some dereferences are conditionalized.

frenche:
> I can't find any indication in RFC 6680 that it's allowed to pass GSS_C_NO_BUFFER for value or display_value to gss_get_name_attribute().

Right, man pages are usually more explicit about these details but I didn't find one, and the RFC is silent.

> It does seem like it might be convenient to be able to do so, as the caller probably only cares about one or the other. 

+1

> There are also some unconditional dereferences of value in authdata module get_attribute methods (authind_get_attribute, s4u2proxy_get_attribute), although some dereferences are conditionalized.

I'll take a look, my run-flow reached mspac_get_attribute() which does check for null.

frenche:
Right, so authind_get_attribute() and s4u2proxy_get_attribute() would need fixing as well (and greet_get_attribute() too I guess).
Or perhaps, we can make a simpler fix at the gss level, something like:
https://github.com/frenche/krb5/commit/702d4540d0248225a2224eefdff35990b5d0f94c

Thoughts?

greghudson:
data_to_gss() always clears its input, so that commit can be simplified by just freeing the data fields at the end.

frenche:
Done, and pushed to this branch along side.

greghudson:
In I don't think you should need to change any of the code inside the first (code == 0) conditional, except to get rid of the call to free() when display_value != NULL and code isn't 0.  (And then it would make sense to combine the conditions for calling data_to_gss(&kdisplay_value, display_value); into a single if statement.)

frenche:
Ok fixed, hope I got it correct now, thanks.

greghudson:
That looks right.  I will merge the V2 commit tomorrow, I hope.

We should decide whether to treat this as a bugfix or an added convenience.  Since the spec doesn't label the output parameters as optional, it's not strictly a bugfix, but we can be flexible about it.  If we do treat it as a bugfix I will mark it for pullup to 1.15 and 1.14.

PR #614, opened 2017-03-12
Memory leak fixes, checking

greghudson:
I did another round of asan leak checking after a report of a memory leak in kvno.  (That leak turns out to be quite old, and wasn't caught by asan because we never run kvno with multiple principal arguments in the test suite.)  This PR has five commits:

* Fix two leaks that crept into test programs since the last round of fixes.
* Fix a leak in gss_inquire_cred_by_oid(), which is old but wasn't caught by asan until 750cd27317d8e8af506c3b9e412a3ccbeef299b0.
* Fix the reported leak in kvno, minimally but somewhat awkwardly.
* Refactor the kvno loop into a helper function so its memory management isn't as awkward.
* Enable asan for Travis builds.

That last change will have a significant impact on the time it takes to run checks on PRs, but it's clear that we need it in order for the test suite to remain leak-clean.  I had originally intended on adding asan checking to a buildbot worker instead, but that would lead to a lot of follow-on commits after we merge a PR without realizing that it introduces a leak (or other memory error).

### DIFF
--- a/src/clients/kvno/kvno.c
+++ b/src/clients/kvno/kvno.c
@@ -326,9 +326,7 @@ static void do_v5_kvno (int count, char *names[],
                 printf(_("%s: kvno = %d\n"), princ, ticket->enc_part.kvno);
         }
 
-        continue;
-
-    error:
+    cleanup:
         if (server != NULL)
             krb5_free_principal(context, server);
         if (ticket != NULL)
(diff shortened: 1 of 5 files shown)


greghudson:
I reverted the last commit as it doesn't pass Travis checks.  The language needs to be set to C++ to avoid a clang/g++ mismatch (easy) and in the Travis build environment asan reports a leak in Cyrus SASL during t_kdb.py which I can't reproduce locally or get a full stack trace for (harder).

greghudson:
Pushed to master as commit 6d82212636eec154bb7322ce7a6e02fd6d1aa596 and preceding.

PR #616, opened 2017-03-16
Port pkinit debug code to OpenSSL 1.1.0 API

Jakuje:
there are still missing bits using old OpenSSL 1.0.x API in the DEBUG ifdefs. I ran into this one while building `pkinit` with `DEBUG` so I am providing a fix, but there might be others in other parts of the project.


greghudson:
I added some more variables to make the code more readable, and tested with all of the PKINIT DEBUG options set.  I think this is the only necessary change.

PR #621, opened 2017-03-27
Coverity

makittel:
Hi,

as discussed on the dev mailing list last week this is a pull request for some of the Coverity issues we found and we think might need fixing. Please let me know in case of questions or if something needs to be changed in the patches.
Would be great to hear back from you.

Best wishes,

Martin.

### DIFF
--- a/src/lib/krb5/krb/preauth2.c
+++ b/src/lib/krb5/krb/preauth2.c
@@ -497,7 +497,7 @@ responder_get_answer(krb5_context context, krb5_clpreauth_rock rock,
     krb5_init_creds_context ctx = (krb5_init_creds_context)rock;
 
     /* Don't let plugins get the raw password. */
-    if (question && strcmp(KRB5_RESPONDER_QUESTION_PASSWORD, question) == 0)
+    if (strcmp(KRB5_RESPONDER_QUESTION_PASSWORD, question) == 0)
         return NULL;
     return k5_response_items_get_answer(ctx->rctx.items, question);
 }
--- a/src/lib/krb5/os/trace.c
+++ b/src/lib/krb5/os/trace.c
@@ -173,7 +173,7 @@ trace_format(krb5_context context, const char *fmt, va_list ap)
             p = va_arg(ap, const char *);
             if (p == NULL && len != 0)
                 k5_buf_add(&buf, "(null)");
-            else
+            else if (p != NULL)
                 buf_add_printable_len(&buf, p, len);
         } else if (strcmp(tmpbuf, "hexlenstr") == 0) {
             len = va_arg(ap, size_t);
--- a/src/lib/krb5/rcache/ser_rc.c
+++ b/src/lib/krb5/rcache/ser_rc.c
@@ -72,7 +72,7 @@ krb5_rcache_size(krb5_context kcontext, krb5_pointer arg, size_t *sizep)
          *      krb5_int32      for KV5M_RCACHE
          */
         required = sizeof(krb5_int32) * 3;
-        if (rcache->ops && rcache->ops->type)
+        if (rcache->ops)
             required += (strlen(rcache->ops->type)+1);
 
         /*
--- a/src/util/support/plugins.c
+++ b/src/util/support/plugins.c
@@ -592,9 +592,10 @@ krb5int_open_plugin_dirs (const char * const *dirnames,
                     }
                 }
 
-                if (krb5int_open_plugin (filepath, &handle, ep) == 0) {
+                if (!err && krb5int_open_plugin(filepath, &handle, ep) == 0) {
                     err = krb5int_plugin_file_handle_array_add (&h, &count, handle);
-                    if (!err) { handle = NULL; }  /* h takes ownership */
+                    if (!err)
+                        handle = NULL; /* h takes ownership */
                 }
 
                 free(filepath);
--- a/src/lib/gssapi/krb5/naming_exts.c
+++ b/src/lib/gssapi/krb5/naming_exts.c
@@ -261,8 +261,7 @@ krb5_gss_inquire_name(OM_uint32 *minor_status,
     krb5_gss_name_t kname;
     krb5_data *kattrs = NULL;
 
-    if (minor_status != NULL)
-        *minor_status = 0;
+    *minor_status = 0;
 
     if (attrs != NULL)
         *attrs = GSS_C_NO_BUFFER_SET;
@@ -322,8 +321,7 @@ krb5_gss_get_name_attribute(OM_uint32 *minor_status,
     krb5_data kvalue = empty_data();
     krb5_data kdisplay_value = empty_data();
 
-    if (minor_status != NULL)
-        *minor_status = 0;
+    *minor_status = 0;
 
     code = krb5_gss_init_context(&context);
     if (code != 0) {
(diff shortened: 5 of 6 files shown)


src/lib/krb5/krb/preauth2.c
```diff
@@ -499,6 +499,9 @@ responder_get_answer(krb5_context context, krb5_clpreauth_rock rock,
     /* Don't let plugins get the raw password. */
     if (question && strcmp(KRB5_RESPONDER_QUESTION_PASSWORD, question) == 0)
         return NULL;
+
+    if (question == NULL)
```
frozencemetery: Would prefer changing the previous condition to `!question || strcmp() == 0` instead.

src/lib/krb5/os/trace.c
```diff
@@ -173,7 +173,7 @@ trace_format(krb5_context context, const char *fmt, va_list ap)
             p = va_arg(ap, const char *);
             if (p == NULL && len != 0)
                 k5_buf_add(&buf, "(null)");
-            else
+            else if (p != NULL)
```
frozencemetery: I think we should probably fail here, not keep going.

src/lib/krb5/rcache/ser_rc.c
```diff
@@ -64,15 +64,15 @@ krb5_rcache_size(krb5_context kcontext, krb5_pointer arg, size_t *sizep)
     size_t              required;
 
     kret = EINVAL;
-    if ((rcache = (krb5_rcache) arg)) {
+    if ((rcache = (krb5_rcache)arg) && (rcache->ops != NULL)) {
```
frozencemetery: Why is this check moving?

src/lib/crypto/krb/string_to_key.c
```diff
@@ -51,8 +51,11 @@ krb5_c_string_to_key_with_params(krb5_context context, krb5_enctype enctype,
         return KRB5_BAD_ENCTYPE;
     keylength = ktp->enc->keylength;
 
+    if (salt == NULL)
```
frozencemetery: Maybe collapse this into the next check?  `if (!salt || salt->length == SALT_TYPE_AFS_LENGTH)` or so?

mrogers950:
lib/kadm5/unit-test/setkey-test.c calls krb5_string_to_key() with a NULL salt (this is what appears to be failing the CI). Looking at some of the str2key functions there are still some that use salt optionally and some that don't, so that change doesn't seem right to me.  Maybe it would be better to evaluate if the SALT_TYPE_AFS_LENGTH hack is still needed, although I don't know the backstory on it.

src/lib/crypto/krb/string_to_key.c
greghudson: The s2k implementations are inconsistent about whether to allow a null salt.  The DES string-to-key allows it, but the AFS variant does not.  The AES string-to-key allows it, unless pepper is set (as for Camellia and aes-sha2), in which case it does not.  And of course the RC4 string-to-key ignores salt altogether.

make check is failing on this change because lib/kadm5/unit-test/setkey-test.c passes a null salt to krb5_string_to_key() for some DES keys.  I don't think it would be completely safe to error out on null salt in the front end, but it should be safe to substitute a pointer to an empty krb5_data.  That would in turn render all of the null salt conditionals in s2k_des.c and s2k_pbkdf2.c unnecessary.

src/lib/krb5/rcache/ser_rc.c
makittel: Further down krb5_rc_get_name is passed rcache as an argument and dereferences rcache->ops in its body

makittel:
Just as a side note: we don't have any expertise in the krb5 code base and our efforts were triggered by the Coverity findings. So please don't be surprised if I cannot participate in the in-depth discussions. If you tell me what you would like to have changed in the commits just let me know. You can also take the patches as input for your own patches if you prefer that. From our point of view the main goal is to get any real issue fixed.

Best wishes,

Martin.

src/lib/krb5/rcache/ser_rc.c
frozencemetery: Understood.  This is fine then.

src/lib/crypto/krb/string_to_key.c
greghudson: I opened PR #622 to address this particular check as discussed above, so we can just eliminate this hunk from this commit.

greghudson:
For the first commit, I think the best answer is to remove the minor_status conditionals, not add more.  minor_status is not an optional parameter; also, the mechglue does a ton of argument validation and will not pass a null minor_status down to the krb5 mech.  The functions in krb5_gss_glue.c are a bit of an exception because they aren't called via the mechglue, but it's also less important that we be nice to buggy calling programs for those extensions.

(@makittel: don't worry about making any of the changes we mention if they are unclear or if you don't have time.  We can make them ourselves assuming you have left the default permission open to let upstream project committers push to the PR branch.)

makittel:
Please go ahead and update the branch as you see fit. Also please let me know if I should update some of patches so we don't run into conflicts.
Just for the way forward: would you rather have new commits with further corrections or updates to the existing commits and forcing the branch (which probably makes more sense but requires more coordination)?.

greghudson:
We prefer changes to the existing commits with force pushes, so that the version history looks clean when it is merged to master.

src/lib/krb5/os/trace.c
greghudson: No, if p is NULL and len is 0, that's not a failure case; we don't want to display "(null)" in this case, and functionally it doesn't matter if we do nothing or call buf_add_printable_len(&buf, NULL, 0) as that is a no-op.

It's sort of annoying that Coverity trips over this code, but adding the conditional isn't too bad as it just looks like an optimization for the p == NULL && len == 0 case.

src/lib/krb5/rcache/ser_rc.c
greghudson: We should remove the conditional on rcache->ops here rather than moving it.  We do not expect a null ops pointer, the code would already crash if we saw one, and there is no similar conditional in krb5_rcache_externalize().

src/util/support/plugins.c
```diff
@@ -592,9 +592,13 @@ krb5int_open_plugin_dirs (const char * const *dirnames,
                     }
                 }
 
-                if (krb5int_open_plugin (filepath, &handle, ep) == 0) {
-                    err = krb5int_plugin_file_handle_array_add (&h, &count, handle);
-                    if (!err) { handle = NULL; }  /* h takes ownership */
+                if (!err) {
+                    if (krb5int_open_plugin(filepath, &handle, ep) == 0) {
```
greghudson: I would prefer to combine these: ``if (!err && krb5int_open_plugin(...``

(This code should ideally be rewritten to use the currently favored ``goto cleanup`` style rather than an if-ladder.  But we should fix the bug minimally, and combining ``if (!err)`` conditionals with other checks appears to be normal in this file.)

makittel:
If you want to, I can try and incorporate your feedback tomorrow and push the changes, although I am not certain you would still want the commits grouped as they currently are. In the meantime feel free to update the PR as you see fit. I will be off for the remainder of the evening (German time) so there shouldn't be any competing changes.

makittel:
I uploaded new patches and also restructured the commits in the process. I removed the commit regarding the salt handling as this is now being discussed in #622.

PR #622, opened 2017-03-27
Simplify null salt handling in string-to-key

greghudson:
The per-enctype string_to_key implementations are inconsistent about
whether a null salt is treated as empty or results in a null
dereference.  Since the original DES string-to-key allowed a null
salt, substitute an empty salt in krb5_c_string_to_key_with_params().
Eliminate conditionals on accessing salt in the per-enctype
implementations as they are on longer needed.  Based on a patch by
Martin Kittel.


mrogers950:
Small typo in the commit message: "as they are on longer needed."

PR #624, opened 2017-03-29
Convert some pkiDebug messages to TRACE macros

mrogers950:
There may be some other pkiDebug() calls that would also be good for TRACE output, but these seemed to be the most useful for now.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -2308,7 +2308,6 @@ crypto_check_cert_eku(krb5_context context,
 
     X509_NAME_oneline(X509_get_subject_name(reqctx->received_cert),
                       buf, sizeof(buf));
-    pkiDebug("%s: looking for EKUs in cert = %s\n", __FUNCTION__, buf);
 
     if ((i = X509_get_ext_by_NID(reqctx->received_cert,
                                  NID_ext_key_usage, -1)) >= 0) {
@@ -2342,7 +2341,6 @@ crypto_check_cert_eku(krb5_context context,
 
         if (found_eku) {
             ASN1_BIT_STRING *usage = NULL;
-            pkiDebug("%s: found acceptable EKU, checking for digitalSignature\n", __FUNCTION__);
 
             /* check that digitalSignature KeyUsage is present */
             X509_check_ca(reqctx->received_cert);
@@ -2351,12 +2349,10 @@ crypto_check_cert_eku(krb5_context context,
 
                 if (!ku_reject(reqctx->received_cert,
                                X509v3_KU_DIGITAL_SIGNATURE)) {
-                    pkiDebug("%s: found digitalSignature KU\n",
-                             __FUNCTION__);
+                    TRACE_PKINIT_EKU(context);
                     *valid_eku = 1;
                 } else
-                    pkiDebug("%s: didn't find digitalSignature KU\n",
-                             __FUNCTION__);
+                    TRACE_PKINIT_EKU_NO_KU(context);
             }
             ASN1_BIT_STRING_free(usage);
         }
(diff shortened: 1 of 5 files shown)


src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -4287,28 +4286,26 @@ pkinit_get_certs_pkcs12(krb5_context context,
     EVP_PKEY *y = NULL;
 
     if (idopts->cert_filename == NULL) {
-        pkiDebug("%s: failed to get user's cert location\n", __FUNCTION__);
+        TRACE_PKINIT_NO_CERT(context);
```
greghudson: Looking at the code, I think this and some of the similar trace messages are unreachable.  pkinit_get_certs_pkcs12() is only called from crypto_load_certs() when idopts->idtype is IDTYPE_PKCS12, and in that case pkinit_identity.c used parse_pkcs12_options() and set cert_filename and key_filename.

There's an opportunity for code cleanup here (perhaps eliminating the pkinit_identity_opts type and using separate entry points into the crypto file for different ID types), but in the meantime I don't think we want to add tracing calls that can't be triggered.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -4957,7 +4949,7 @@ crypto_retrieve_X509_key_usage(krb5_context context,
     if (ret_eku_bits)
         *ret_eku_bits = 0;
     else {
-        pkiDebug("%s: EKUs not requested, not checking\n", __FUNCTION__);
+        TRACE_PKINIT_NO_EKU_CHECK(context);
```
greghudson: Never happens (only one call site which never passes NULL).  Same for the next message.  The conditionals should be removed instead, although that's not necessarily in scope for this PR.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -5498,8 +5490,7 @@ load_cas_and_crls_dir(krb5_context context,
 
     while ((dentry = readdir(d))) {
         if (strlen(dirname) + strlen(dentry->d_name) + 2 > sizeof(filename)) {
-            pkiDebug("%s: Path too long -- directory '%s' and file '%s'\n",
-                     __FUNCTION__, dirname, dentry->d_name);
+            TRACE_PKINIT_PATH_TOO_LONG(context, dirname, dentry->d_name);
```
greghudson: This code should use asprintf(); in the meantime, I wouldn't add a trace message.

src/plugins/preauth/pkinit/pkinit_identity.c
```diff
@@ -610,8 +610,7 @@ pkinit_identity_prompt(krb5_context context,
             retval = crypto_cert_select_default(context, plg_cryptoctx,
                                                 req_cryptoctx, id_cryptoctx);
             if (retval) {
-                pkiDebug("%s: Failed while selecting default certificate\n",
-                         __FUNCTION__);
+                TRACE_PKINIT_NO_DEFAULT_CERT(context);
```
greghudson: This can only fail in the OpenSSL implementation because there is more than one cert to choose from.  The trace message in crypto_cert_select_default() seems more informative in that case, and we don't need two messages.

For parallelism we might also want pkinit_cert_matching() to be responsible for tracing when there is no matching cert.

src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -449,7 +451,7 @@ pkinit_server_verify_padata(krb5_context context,
 
     switch ((int)data->pa_type) {
     case KRB5_PADATA_PK_AS_REQ:
-        pkiDebug("processing KRB5_PADATA_PK_AS_REQ\n");
+        TRACE_PKINIT_SERVER_PADATA_VERIFY(context, "KRB5_PADATA_PK_AS_REQ");
```
greghudson: We shouldn't call the same TRACE macro twice; they are supposed to be usable as DTrace tracing points if the macros are changed.  So use two different macros here, not the same one with a string literal parameter.

mrogers950:
Thanks, I've updated with those changes.

greghudson:
TRACE_PKINIT_NO_CERT is called twice.

TRACE_PKINIT_PATH_TOO_LONG still exists, and shouldn't; the relevant code should change to use asprintf instead (not necessarily right now).

TRACE_PKINIT_PKCS_DECODE_FAIL and TRACE_PKINIT_PKCS_OPEN_FAIL use split string literals but don't need to.

mrogers950:
Updated

PR #625, opened 2017-03-30
Add support to query the SSF of a context

simo5:
This patch adds a way to return the SSF number for a gssapi security context.

Implementation hasn't been discussed a lot yet so I expect changes will be needed, this is more of a PoC to start a discussion.


greghudson:
(Summarizing and expanding upon IRC conversation:)

The strength of a Kerberos session, to the extent that it can be represented by a single number at all, is dependent on the weakest link in the authentication chain.  The links are:

1. AS exchange (perhaps a weak password)
2. TGS exchange (if one was used; may be multiple exchanges for cross-realm)
3. Service ticket encryption key
4. Service ticket session key
5. Initiator subkey
6. Acceptor subkey

In the GSS code we do not have visibility into all of that, and as we start looking further backwards it gets more inconvenient.  Looking back to step 4 requires keeping new state in the context (as is done in the current PR).  Looking back to step 3 requires decoding the ticket on the client, and the client also might not understand the ticket encryption key enctype, leading to a disagreement in SSF value between the client and server.  Looking back to step 2 is impossible on the server, and may require combing the ccache on the client for a TGT.  Looking back to step 1 would require a bunch of new infrastructure on the client and in the AS exchange process, and is impossible on the server.

Since the whole SSF concept is kind of bogus, and we can't give a properly conservative answer in all cases (in fact, one could argue that in the common case, the user's password is the weakest link), perhaps we should take the easiest route and just look at the two keys that are already handy in the context.  Apologies for reversing course here.

simo5:
No apologies necessary and I actually concur, I have pondered for a while.
There is an ideal SSF value and a praactical one.
The ideal value would go all the way back to the set up of the server account (how was the keytab obtained and stored ?) and even further back to the admins credentials strenght and how the KDC was installed and so on and so forth. Clearly there is no ideal SSF.

So back to being practical, what is that the SASL connection wants to measure ? It wants to measure the strenght of the cipher used to actually encrypt data on the wire, and it assumes the admin did due diligence in properly securing all the comes before that.

So yeah in the end I think justtaking in account the actual subkey used only is the way to go.
I'll change the PR back

simo5:
Ok changed back to query just the resuting key.
If you feel strongly I could check both subkey and acceptor_subkey, but I am going to assume that the strength of the actual key is what matters for now.

greghudson:
Pushing changes to:

* Rename krb5_enctype_to_ssf to k5_enctype_to_ssf.  (Miscommunication in voice discussion.)
* Not use KRB5_CALLCONV for k5_enctype_to_ssf.
* Edit some comments and the commit message.
* Export GSS_C_SEC_CONTEXT_SASL_SSF on Windows.

greghudson:
I took a stab at writing docs for our inquire_context extensions, but got bogged down in the details of some functions having wrappers and others not.  I will probably merge this without documentation as we only expect a single consumer anyway.

PR #626, opened 2017-03-31
Coverity

makittel:
Hi,

as discussed on the mailing list this is the next set of commits based on Coverity findings. Based on the mailing list feedback I am not quite certain which ones are actually considered to matter, so to be on the safe side I included all of them. Feel free to clean up as you please.

Best wishes,

Martin.

### DIFF
--- a/src/kadmin/dbutil/dump.c
+++ b/src/kadmin/dbutil/dump.c
@@ -688,6 +688,10 @@ process_tl_data(const char *fname, FILE *filep, int lineno,
                      _("cannot read tagged data type and length"));
             return EINVAL;
         }
+        if (i1 < INT16_MIN || i1 > INT16_MAX || u1 > UINT16_MAX) {
+            load_err(fname, lineno, _("data type or length overflowed"));
+            return EINVAL;
+        }
         tl->tl_data_type = i1;
         tl->tl_data_length = u1;
         if (read_octets_or_minus1(filep, tl->tl_data_length,
@@ -735,6 +739,10 @@ process_k5beta7_princ(krb5_context context, const char *fname, FILE *filep,
         goto fail;
 
     /* Get memory for and form tagged data linked list */
+    if (u3 > UINT16_MAX) {
+        load_err(fname, *linenop, _("cannot allocate tl_data (too large)"));
+        goto fail;
+    }
     if (alloc_tl_data(u3, &dbentry->tl_data))
         goto fail;
     dbentry->n_tl_data = u3;
@@ -823,13 +831,17 @@ process_k5beta7_princ(krb5_context context, const char *fname, FILE *filep,
             load_err(fname, *linenop, _("cannot read key size and version"));
             goto fail;
         }
+        if (t1 > KRB5_KDB_V1_KEY_DATA_ARRAY) {
+            load_err(fname, *linenop, _("unsupported key_data_ver version"));
+            goto fail;
+        }
 
         kd->key_data_ver = t1;
         kd->key_data_kvno = t2;
 
         for (j = 0; j < t1; j++) {
             nread = fscanf(filep, "%d\t%d\t", &t3, &t4);
-            if (nread != 2) {
+            if (nread != 2 || t4 < 0) {
                 load_err(fname, *linenop,
                          _("cannot read key type and length"));
                 goto fail;
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -841,6 +841,8 @@ prepare_error_as(struct kdc_request_state *rstate, krb5_kdc_req *request,
     kdc_realm_t *kdc_active_realm = rstate->realm_data;
     size_t count;
 
+    errpkt.magic = KV5M_ERROR;
+
     if (e_data_in != NULL) {
         /* Add a PA-FX-COOKIE to e_data_in.  e_data is a shallow copy
          * containing aliases. */
--- a/src/lib/gssapi/krb5/k5sealv3.c
+++ b/src/lib/gssapi/krb5/k5sealv3.c
@@ -301,7 +301,7 @@ gss_krb5int_unseal_token_v3(krb5_context *contextptr,
                             int *conf_state, gss_qop_t *qop_state, int toktype)
 {
     krb5_context context = *contextptr;
-    krb5_data plain;
+    krb5_data plain = empty_data();
     uint64_t seqnum;
     size_t ec, rrc;
     int key_usage;
--- a/src/lib/gssapi/krb5/k5sealv3.c
+++ b/src/lib/gssapi/krb5/k5sealv3.c
@@ -110,6 +110,7 @@ gss_krb5int_make_seal_token_v3 (krb5_context context,
         krb5_data plain;
         krb5_enc_data cipher;
         size_t ec_max;
+        size_t encrypt_size;
 
         /* 300: Adds some slop.  */
         if (SIZE_MAX - 300 < message->length)
@@ -128,7 +129,12 @@ gss_krb5int_make_seal_token_v3 (krb5_context context,
             return err;
 
         /* Get size of ciphertext.  */
-        bufsize = 16 + krb5_encrypt_size (plain.length, key->keyblock.enctype);
+        encrypt_size = krb5_encrypt_size(plain.length, key->keyblock.enctype);
+        if (encrypt_size > SIZE_MAX / 2) {
+            err = ENOMEM;
+            goto error;
+        }
+        bufsize = 16 + encrypt_size;
         /* Allocate space for header plus encrypted data.  */
         outbuf = gssalloc_malloc(bufsize);
         if (outbuf == NULL) {
--- a/src/lib/kdb/kdb_default.c
+++ b/src/lib/kdb/kdb_default.c
@@ -282,7 +282,7 @@ krb5_db_def_fetch_mkey_stash(krb5_context   context,
     key->length = keylength;
 #endif
 
-    if (!key->length || ((int) key->length) < 0) {
+    if (!key->length || key->length > 1024) {
         retval = KRB5_KDB_BADSTORED_MKEY;
         goto errout;
     }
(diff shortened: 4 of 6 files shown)


src/lib/kdb/kdb_default.c
```diff
@@ -287,7 +287,8 @@ krb5_db_def_fetch_mkey_stash(krb5_context   context,
         goto errout;
     }
 
-    if (!(key->contents = (krb5_octet *)malloc(key->length))) {
+    if (!(key->contents = (krb5_octet *)malloc(
+              sizeof(key->contents[0])*key->length))) {
```
frozencemetery: Needs whitespace around `*`.

src/kdc/do_as_req.c
```diff
@@ -841,6 +841,8 @@ prepare_error_as(struct kdc_request_state *rstate, krb5_kdc_req *request,
     kdc_realm_t *kdc_active_realm = rstate->realm_data;
     size_t count;
 
+    memset(&errpkt, 0, sizeof(krb5_error));
```
frozencemetery: A slightly less invasive change on these would be to set `errpkt.magic`.

src/kadmin/dbutil/dump.c
```diff
@@ -688,6 +688,10 @@ process_tl_data(const char *fname, FILE *filep, int lineno,
                      _("cannot read tagged data type and length"));
             return EINVAL;
         }
+        if (i1 > KRB5_INT16_MAX || u1 > KRB5_INT16_MAX) {
+            load_err(fname, lineno, ("overflow"));
```
frozencemetery: Perhaps "data type or length overflowed" instead?

src/kadmin/dbutil/dump.c
```diff
@@ -735,6 +739,10 @@ process_k5beta7_princ(krb5_context context, const char *fname, FILE *filep,
         goto fail;
 
     /* Get memory for and form tagged data linked list */
+    if (u3 > KRB5_INT16_MAX) {
+        load_err(fname, *linenop, _("u3 overflow"));
```
frozencemetery: "cannot allocate tl_data (too large)"?

src/lib/gssapi/krb5/k5sealv3.c
```diff
@@ -128,7 +130,14 @@ gss_krb5int_make_seal_token_v3 (krb5_context context,
             return err;
 
         /* Get size of ciphertext.  */
-        bufsize = 16 + krb5_encrypt_size (plain.length, key->keyblock.enctype);
+        encrypt_size = krb5_encrypt_size(plain.length, key->keyblock.enctype);
+        if ((encrypt_size == (size_t) - 1) ||
+            (encrypt_size > SIZE_MAX - 300)) {
```
frozencemetery: Why 300?  Needs a comment if it's not arbitrary.

src/plugins/kdb/db2/libdb2/btree/bt_open.c
```diff
@@ -285,6 +285,10 @@ __bt_open(fname, flags, mode, openinfo, dflags)
 	if (b.cachesize < b.psize * MINCACHE)
 		b.cachesize = b.psize * MINCACHE;
 
+	/* prevent overflow in next line */
+	if ((b.cachesize > 4000000000) || ((4000000000 - b.cachesize) < t->bt_psize))
+		goto err;
```
frozencemetery: Should set errno here.

src/lib/gssapi/krb5/k5sealv3.c
```diff
@@ -312,6 +312,8 @@ gss_krb5int_unseal_token_v3(krb5_context *contextptr,
     krb5_key key;
     krb5_cksumtype cksumtype;
 
+    memset(&plain, 0, sizeof(krb5_data));
```
greghudson: We can initialize plain to empty_data() where it is declared, which sets the magic field to the nominally correct value for a krb5_data object.

src/kdc/do_as_req.c
greghudson: Right, it should nominally be set to KV5M_ERROR anyway, not to 0 (although it doesn't matter in practice).

src/kadmin/dbutil/dump.c
greghudson: i1 should be compared against INT16_MIN and INT16_MAX.  u1 should be compared against UINT16_MAX.  KRB5_INT16_MAX is defined to a bogus value (old technical debt) and shouldn't be used.  (Same applies to the next few hunks.)

src/lib/gssapi/krb5/k5sealv3.c
greghudson: Also, (size_t)-1 is > SIZE_MAX - 300, so these clauses are redundant.

Using SIZE_MAX / 2 would be reasonable.  The caller will never have an object that large, and it would give us lots of margin to avoid overflows since we don't multiply here.

src/plugins/kdb/db2/libdb2/btree/bt_open.c
greghudson: This code makes assumptions about the range of an int.

src/util/verto/ev.c
```diff
@@ -4346,6 +4346,11 @@ infy_cb (EV_P_ ev_io *w, int revents)
     {
       struct inotify_event *ev = (struct inotify_event *)(buf + ofs);
       infy_wd (EV_A_ ev->wd, ev->wd, ev);
+      if (ev->len > NAME_MAX) {
```
greghudson: This is third-party code, and we don't use it.  After PR #623 we won't compile it.  So we should leave this hunk out.

src/lib/kdb/kdb_default.c
greghudson: The more I think about this, the more I'm confused about why Coverity triggers on it for Martin (and not on any other code that allocates a uint8_t buffer).  key->contents is of type krb5_octet *, and krb5_octet is a typedef for uint8_t, and uint8_t is a typedef for unsigned char with glibc (looking at stdint.h).  So why does Coverity think we need to multiply?

makittel: My apologies. This fix was provided by a colleague. I rechecked the original Coverity message and it turns out that Coverity was not complaining about the missing multiplication but about the fact that key->length is being read from a file on disk and used without any upper bounds checking.
I had only looked at the fix and from there concluded that the problem was the missing multiplication.

src/kdc/do_as_req.c
makittel: Should I replace all memset calls by an initialization of errpkt.magic = KV5M_ERROR? This would look a lot nicer to me.

greghudson: Yes, please.

src/lib/kdb/kdb_default.c
greghudson: That makes more sense.

A bounds check could be good here in principle.  The data is being read from a trusted source, but the file could be corrupt.  The approach we use in cc_file.c is to stat the file and bound all allocations by the file size (itself bounded by SIZE_MAX).  That feels a little cumbersome for a one-off like this, though.  Since keys are relatively small (currently limited to 32 bytes for 256-bit ciphers), we could also just apply a limit like 1024 bytes, which should be good enough for the forseeable future.

makittel:
I uploaded a new version of the patches that hopefully address all comments that were made so far.
I also changed the overflow check in ure.c to match the one from OpenLDAP based on previous comments on the mailing list.

Best wishes,

Martin.

src/util/verto/ev.c
frozencemetery: This comment has not been addressed.

src/kadmin/dbutil/dump.c
```diff
@@ -823,13 +831,17 @@ process_k5beta7_princ(krb5_context context, const char *fname, FILE *filep,
             load_err(fname, *linenop, _("cannot read key size and version"));
             goto fail;
         }
+        if (t1 > sizeof(kd->key_data_type)/sizeof(kd->key_data_type[0])) {
+            load_err(fname, *linenop, _("not supported key_data_ver version"));
```
frozencemetery: "unsupported key_data_ver version" is more correct

makittel:
Hopefully now really all comments have been addressed :-) Thanks for pointing this out.
I think Greg still wanted to look at the DB2 stuff more closely. If need be I can also move these to a separate commit.

src/plugins/kdb/db2/libdb2/btree/bt_seq.c
```diff
@@ -239,13 +240,17 @@ __bt_seqset(t, ep, key, flags)
 				return (RET_ERROR);
 
 			/* Check for an empty tree. */
-			if (NEXTINDEX(h) == 0) {
+			if (NEXTINDEX(h) <= 0) {
```
greghudson: NEXTINDEX(h) expands to (h->lower - BTDATAOFF) / sizeof(indx_t).  h->lower is of type indx_t, which is u_int16_t.   I don't think this value can be less than zero.

src/plugins/kdb/db2/libdb2/btree/bt_seq.c
```diff
@@ -239,13 +240,17 @@ __bt_seqset(t, ep, key, flags)
...
 				mpool_put(t->bt_mp, h, 0);
 				return (RET_SPECIAL);
 			}
 
 			if (h->flags & (P_BLEAF | P_RLEAF))
 				break;
+			if (UNDERFLOW_NEXTINDEX(h) || NEXTINDEX(h) <= 1) {
```
greghudson: Again I think I would like to defer this change, until one of the BSDs introduces something similar.  (I don't have an analysis showing that an overflow is unlikely for this case as I do for cachesize, however.)

makittel:
I had updated the patches on April 6 and thought I had left a comment stating that I'll be out for one week - obviously I must have forgotten to press the button to actually post it... Sorry.
Anyway, I have removed all the db2 changes from the patch set, so hopefully we are good to go.

greghudson:
Yeah, I noticed a few days ago that the patches had been updated, but got distracted by Robbie's warning-clean build work and a request for UTF-16 support.  I don't have any outstanding concerns about the changes at this time, and will give them a final review and merge them soon.

PR #627, opened 2017-04-03
Add FAST encrypted challenge auth indicator

mrogers950:
Here's my attempt at adding AI capabilities to the FAST ec module. The configuration option may be inadequate as only a kdcdefaults section option, but currently the module does not use any realm-specific moddata in order to handle separate FAST AIs.

### DIFF
--- a/src/kdc/kdc_preauth_ec.c
+++ b/src/kdc/kdc_preauth_ec.c
@@ -66,6 +66,8 @@ ec_verify(krb5_context context, krb5_data *req_pkt, krb5_kdc_req *request,
     krb5_keyblock *kdc_challenge_key;
     krb5_kdcpreauth_modreq modreq = NULL;
     int i = 0;
+    char *ai = NULL, *realmstr = NULL;
+    krb5_data realm = request->server->realm;
 
     plain.data = NULL;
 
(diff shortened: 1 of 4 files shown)


greghudson:
I'd rather this use a per-realm variable; is that difficult to implement?

mrogers950:
> I'd rather this use a per-realm variable; is that difficult to implement?

It should just be a matter of adding moddata that keeps a list of realms and their related indicators, and consulting that list based on the server realm in the request (I'm thinking similar to the handling of the PKINIT module per-realm contexts), so I don't think that would be hard to do here.

greghudson:
I'm not sure you need to do all that.  PKINIT looks up its realm context using request->server->realm (case-insensitively for some reason), so I think you can just use that as a lookup key in ec_verify().

mrogers950:
This update makes it a per-realm variable.

src/kdc/kdc_preauth_ec.c
```diff
@@ -84,6 +86,19 @@ ec_verify(krb5_context context, krb5_data *req_pkt, krb5_kdc_req *request,
         if (plain.data == NULL)
             retval = ENOMEM;
     }
+
+    /* Check for a configured FAST ec auth indicator. */
+    realmstr = malloc(realm.length + 1);
```
greghudson: You can simplify this using k5memdup0(), I think.

mrogers950:
I've simplified it with k5memdup0.

greghudson:
Let's go with "encrypted_challenge_indicator" as the profile variable.  (Barely fits on one line in k5-int.h, doesn't quite fit the tabulation but it's only off by one.)  People see "ec" and naturally think "elliptic curve".  I got that feedback on the code when creating preauth_ec.c and kdc_preauth_ec.c but unfortunately didn't address it at the time.

mrogers950:
OK, updated.

PR #628, opened 2017-04-04
Use the client DB principal name for OTP

mrogers950:
Add the kdcpreauth callback client_name() to return the client DB
principal and use it when calling otp_state_verify().  This allows
canonicalization of the name before passing it along as the username
for RADIUS.


greghudson:
Looks good, but please separate the new callback from its use by OTP into two commits.   They will both be small commits, but this allows the new callback to be cherry-picked separately from the OTP behavior change in case something else depends on it.

mrogers950:
That's done.

PR #630, opened 2017-04-06
Fix all warnings with clang

frozencemetery:
If I've done this right, as promised, the clang build on Travis should now pass without any warnings.

* I thought about turning on -Werror for clang (on Travis only, not in general) but could not find a good way to do this.
* Second-last commit is again to db2; I'm not sure what we want to do about this.  I think in the past we submitted it to NetBSD?
* Last commit touches the SWIG-generated TCL profiling code committed into the tree.  I assume there's a better way to fix this issue, and would appreciate suggestions.  (I do not know SWIG.)
* How would people feel about our CI running `make -s`?  It would make poring over these logs easier, I think.

Thanks!

### DIFF
--- a/src/lib/crypto/builtin/sha2/sha256.c
+++ b/src/lib/crypto/builtin/sha2/sha256.c
@@ -211,14 +211,14 @@ k5_sha256_update(SHA256_CTX *m, const void *v, size_t len)
 #if !defined(WORDS_BIGENDIAN) || defined(_CRAY)
 	    int i;
 	    uint32_t current[16];
-	    struct x32 *u = (struct x32*)m->save;
+	    struct x32 *u = (struct x32*)(void*)m->save;
 	    for(i = 0; i < 8; i++){
 		current[2*i+0] = swap_uint32_t(u[i].a);
 		current[2*i+1] = swap_uint32_t(u[i].b);
 	    }
 	    calc(m, current);
 #else
-	    calc(m, (uint32_t*)m->save);
+	    calc(m, (uint32_t*)(void*)m->save);
 #endif
 	    offset = 0;
 	}
--- a/src/lib/crypto/krb/t_fortuna.c
+++ b/src/lib/crypto/krb/t_fortuna.c
@@ -85,7 +85,7 @@ head_tail_test(struct fortuna_state *st)
 {
     static unsigned char buffer[1024 * 1024];
     unsigned char c;
-    size_t i, len = sizeof(buffer);
+    int i, len = sizeof(buffer);
     int bit, bits[8] = { 0, 0, 0, 0, 0, 0, 0, 0 };
     double res;
 
--- a/src/lib/krb5/krb/sendauth.c
+++ b/src/lib/krb5/krb/sendauth.c
@@ -131,22 +131,21 @@ krb5_sendauth(krb5_context context, krb5_auth_context *auth_context,
            This isn't strong cryptographically; the point here is
            not to guarantee randomness, but to make it less likely
            that multiple sessions could pick the same subkey.  */
-        char rnd_data[1024];
+        struct sockaddr_storage rnd_data;
         GETPEERNAME_ARG3_TYPE len2;
-        krb5_data d;
-        d.length = sizeof (rnd_data);
-        d.data = rnd_data;
-        len2 = sizeof (rnd_data);
-        if (getpeername (*(int*)fd, (GETPEERNAME_ARG2_TYPE *) rnd_data,
-                         &len2) == 0) {
+        krb5_data d = make_data(&rnd_data, sizeof(rnd_data));
+
+        len2 = sizeof(rnd_data);
+        if (getpeername(*(int *)fd, ss2sa(&rnd_data), &len2) == 0) {
             d.length = len2;
-            (void) krb5_c_random_add_entropy (context, KRB5_C_RANDSOURCE_EXTERNAL_PROTOCOL, &d);
+            (void)krb5_c_random_add_entropy(
+                context, KRB5_C_RANDSOURCE_EXTERNAL_PROTOCOL, &d);
         }
-        len2 = sizeof (rnd_data);
-        if (getsockname (*(int*)fd, (GETSOCKNAME_ARG2_TYPE *) rnd_data,
-                         &len2) == 0) {
+        len2 = sizeof(rnd_data);
+        if (getsockname(*(int *)fd, ss2sa(&rnd_data), &len2) == 0) {
             d.length = len2;
-            (void) krb5_c_random_add_entropy (context, KRB5_C_RANDSOURCE_EXTERNAL_PROTOCOL, &d);
+            (void)krb5_c_random_add_entropy(
+                context, KRB5_C_RANDSOURCE_EXTERNAL_PROTOCOL, &d);
         }
     }
 
--- a/src/lib/krb5/os/accessor.c
+++ b/src/lib/krb5/os/accessor.c
@@ -30,11 +30,14 @@
 
 /* If this trick gets used elsewhere, move it to k5-platform.h.  */
 #ifndef DESIGNATED_INITIALIZERS
-#define DESIGNATED_INITIALIZERS                         \
-    /* ANSI/ISO C 1999 supports this...  */             \
-    (__STDC_VERSION__ >= 199901L                        \
-     /* ...as does GCC, since version 2.something.  */  \
-     || (!defined __cplusplus && __GNUC__ >= 3))
+/* ANSI/ISO C 1999 supports this...  */
+#if __STDC_VERSION__ >= 199901L                       \
+    /* ...as does GCC, since version 2.something.  */ \
+    || (!defined __cplusplus && __GNUC__ >= 3)
+#define DESIGNATED_INITIALIZERS 1
+#else
+#define DESIGNATED_INITIALIZERS 0
+#endif
 #endif
 
 krb5_error_code KRB5_CALLCONV
--- a/src/util/support/threads.c
+++ b/src/util/support/threads.c
@@ -237,7 +237,6 @@ void *k5_getspecific (k5_key_t keynum)
     if (err)
         return NULL;
 
-    assert(keynum >= 0 && keynum < K5_KEY_MAX);
     assert(destructors_set[keynum] == 1);
 
 #ifndef ENABLE_THREADS
@@ -271,7 +270,6 @@ int k5_setspecific (k5_key_t keynum, void *value)
     if (err)
         return err;
 
-    assert(keynum >= 0 && keynum < K5_KEY_MAX);
     assert(destructors_set[keynum] == 1);
 
 #ifndef ENABLE_THREADS
@@ -334,8 +332,6 @@ int k5_key_register (k5_key_t keynum, void (*destructor)(void *))
     if (err)
         return err;
 
-    assert(keynum >= 0 && keynum < K5_KEY_MAX);
-
 #ifndef ENABLE_THREADS
 
     assert(destructors_set[keynum] == 0);
@@ -365,8 +361,6 @@ int k5_key_register (k5_key_t keynum, void (*destructor)(void *))
 
 int k5_key_delete (k5_key_t keynum)
 {
-    assert(keynum >= 0 && keynum < K5_KEY_MAX);
-
 #ifndef ENABLE_THREADS
 
     assert(destructors_set[keynum] == 1);
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -1367,7 +1367,6 @@ dnl =============================================================
 dnl Internal function for testing for getpeername prototype
 dnl
 AC_DEFUN([KRB5_GETPEERNAME_ARGS],[
-AC_DEFINE([GETPEERNAME_ARG2_TYPE],GETSOCKNAME_ARG2_TYPE,[Type of getpeername second argument.])
 AC_DEFINE([GETPEERNAME_ARG3_TYPE],GETSOCKNAME_ARG3_TYPE,[Type of getpeername second argument.])
 ])
 dnl
@@ -1412,7 +1411,6 @@ if test "$sock_set" = no; then
 fi
 res1=`echo "$res1" | tr -d '*' | sed -e 's/ *$//'`
 res2=`echo "$res2" | tr -d '*' | sed -e 's/ *$//'`
-AC_DEFINE_UNQUOTED([GETSOCKNAME_ARG2_TYPE],$res1,[Type of pointer target for argument 2 to getsockname])
 AC_DEFINE_UNQUOTED([GETSOCKNAME_ARG3_TYPE],$res2,[Type of pointer target for argument 3 to getsockname])
 ])
 dnl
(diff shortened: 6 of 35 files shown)


frozencemetery:
Well, it's clean with clang-3.4, which is what asking for clang on Trusty gives you.  I'll fix it to be clang-3.5 in a bit.

frozencemetery:
Okay, good with clang-3.5 now.  There remain warnings in the bundled libev code; how do we want to handle those?

kaduk:
> SWIG-generated TCL profiling code committed into the tree

I really want the answer to be "remove it all", though I did not look at what it's used for or the feasibility of doing so.

src/lib/crypto/builtin/sha2/sha256.c
```diff
@@ -211,7 +211,7 @@ k5_sha256_update(SHA256_CTX *m, const void *v, size_t len)
 #if !defined(WORDS_BIGENDIAN) || defined(_CRAY)
 	    int i;
 	    uint32_t current[16];
-	    struct x32 *u = (struct x32*)m->save;
+	    struct x32 *u = (struct x32*)(void*)m->save;
```
greghudson: There's another branch down below which needs similar treatment.  (Also for the next hunk.)

There might be a more correct way to express this code (using a different type for save), but I'm reluctant to diverge materially from the upstream Heimdal code.  I suppose there might be different upstream sha256/sha512 code we could think about using instead.

src/util/profile/prof_file.c
```diff
@@ -60,6 +59,10 @@ int profile_library_initializer(void)
 
     return k5_mutex_finish_init(&g_shared_trees_mutex);
 }
+
+#ifdef _WIN32
```
greghudson: How is this Windows-specific?

src/lib/crypto/krb/t_fortuna.c
```diff
@@ -102,7 +102,7 @@ head_tail_test(struct fortuna_state *st)
     }
 
     for (bit = 0; bit < 8; bit++) {
-        res = ((double)abs(len - bits[bit] * 2)) / (double)len;
+        res = ((double)(len - bits[bit] * 2)) / (double)len;
```
greghudson: I think the code is broken here and should be fixed rather than papered over.  I'm not sure why the test isn't failing; you'd expect at least one of the eight bits to be biased slightly high such that len - bits[bit] * 2 is a huge number.

greghudson:
I looked at the db2 sources in the various *BSD libcs.  None of them appear to use any kind of temporary filename at all; they just don't open a file if none is specified.  I will need to examine in further detail how they handle the backing store in this case.

src/util/support/threads.c
```diff
@@ -237,7 +237,7 @@ void *k5_getspecific (k5_key_t keynum)
     if (err)
         return NULL;
 
-    assert(keynum >= 0 && keynum < K5_KEY_MAX);
+    assert(keynum < K5_KEY_MAX);
```
frozencemetery: Hmm, I wonder why clang warns so aggressively there then.  If you're find removing the assert entirely, I'll go ahead and do that.

src/lib/crypto/builtin/sha2/sha256.c
frozencemetery: Weird, no warning from the compiler about it.  Will fix.

src/util/support/threads.c
greghudson: clang presumably chooses an unsigned type for the enumeration, and doesn't remember enough state to relax the warning on the grounds that the type might be signed.  We can just get rid of the assert; it's not likely that code would pass an invalid key value.

src/lib/crypto/builtin/sha2/sha256.c
greghudson: Building on an Intel platform, WORDS_BIGENDIAN would be false, so only the first branch gets compiled.

src/lib/crypto/krb/t_fortuna.c
greghudson: Okay, I ran this code through a debugger.  For bit 1 (the second iteration of the loop), bits[bit] * 2 is indeed slightly larger than len, so the expression (len - bits[bit] * 2) evaluates to a large size_t value.  But since abs takes an int argument, the value is coerced to -1090 and the result of abs() is 1090.  What is the clang warning here?

The change in the PR causes t_fortuna to print an error message and exit with status 1.  However, because of the way Makefile.in runs t_fortuna, it doesn't notice (it only notices that stdout matched the expected output) so the Travis tests pass.

Simply making i and len have type ``int`` appears to be a better fix.

frozencemetery: clang warns that `abs()` is being called on something that is unsigned.  I'll change the types instead in update.

src/util/support/utf8_conv.c
```diff
@@ -430,7 +433,7 @@ krb5int_ucs2lecs_to_utf8s(const unsigned char *ucs2les,
     if (ucs2leslen > SSIZE_MAX)
         return ERANGE;
 
-    len = k5_ucs2s_to_utf8s(NULL, (krb5_ucs2 *)ucs2les, 0,
+    len = k5_ucs2s_to_utf8s(NULL, ucs2_of_utf8(ucs2les), 0,
```
frozencemetery: `krb5int_ucs2lecs_to_utf8s` is used in pac.c

src/util/profile/prof_file.c
frozencemetery: This isn't quite the right fix.  Most of the time, `MAKE_FINI_FUNCTION` creates a static function - Windows is one of (I misread as "only") the cases where it doesn't.  I think it's more correct to just not warn about this function being potentially unused.  If it were more than one function doing this, I'd suggest modifying the definitions of `MAKE_FINI_FUNCTION`, but it's not.

src/util/support/utf8_conv.c
greghudson: My bad; it looks like krb5int_ucs2lecs_to_utf8s is the only entry point into k5_ucs2s_to_utf8s whihc is used.

This conversion is unsafe because we don't know the alignment of the input pointer.  The byte swapping insid ek5_ucs2s_to_utf8s is also broken; if we ever asked for non-little-endian input on a little-endian platform, we would get the wrong values.  The upstream OpenLDAP code (in utf-8-conv.c) operates on a wchar_t array and does not do byte swapping, so this is all stuff introduced by lukeh when the code was adapted for krb5.

For what we want, I think the right answer is to:

* Remove the unused entry points, so that we are only trying to work on byte arrays in little-endian.

* Change k5_ucs2s_to_utf8s() to operate on a byte array and use load_16_le() to pull UCS-2 values out.  Probably also to use k5buf instead of operating in two passes and worrying about the output buffer size.

That's a fair amount of code surgery; I can take care of it if need be.

src/lib/krb5/os/accessor.c
```diff
@@ -41,7 +41,7 @@ krb5_error_code KRB5_CALLCONV
 krb5int_accessor(krb5int_access *internals, krb5_int32 version)
 {
     if (version == KRB5INT_ACCESS_VERSION) {
-#if DESIGNATED_INITIALIZERS
+#ifdef DESIGNATED_INITIALIZERS
```
greghudson: This change would break the intent of the code, which is to use designated initializers only for C99 compilers or suitable versions of gcc.  DESIGNATED_INITIALIZERS is always defined by isn't always true.  What is the warning related to this change?

src/util/profile/prof_file.c
greghudson: I don't understand how this code is different from other uses of MAKE_FINI_FUNCTION such that there is a warning here but not elsewhere.  (For example, crypto_libinit.c.)

Unconditionally using __attribute__(unused) breaks on non-gcc/clang compilers, as witnessed by the appveyor build failure.

frozencemetery:
* Rebased on top of #631 and dropped corresponding commit.
* Two new commits (ifdef and kdc_j_encode.c) added to bring the clean clang set to: 3.4, 3.5, 3.8, 3.9, 4.0
* I've opted to yank the PRNG games, which I've separated to #633.
* Addressed most other comments (still working on some)

src/lib/crypto/builtin/sha2/sha256.c
frozencemetery: (This has been addressed)

src/lib/krb5/os/accessor.c
frozencemetery: Oops, I've done the wrong thing.  I'll fix it in a moment, but for posterity:

```
accessor.c:44:5: warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
#if DESIGNATED_INITIALIZERS
    ^
```
and
```
accessor.c:108:5: warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
#if DESIGNATED_INITIALIZERS
    ^
```

frozencemetery:
Regarding db2: hsearch.c explicitly passes NULL to `__hash_open()`, which is an alias for `__kdb2_hash_open()`, which is the function that uses `tmpnam()` currently.

src/util/profile/prof_file.c
frozencemetery: I do not know what makes it different.  Appveyor build has been fixed, but the fix is admittedly not great.

greghudson:
hsearch.c appears to implement leaf functionality which we don't use.  If we break it nothing should notice.

frozencemetery:
Alright.  I couldn't find another way to get NULL in there, so I'm going forward with the EINVAL approach.

src/util/support/utf8_conv.c
frozencemetery: Test suite calls into `k5_utf8s_to_ucs2s()` eight times, and `k5_ucs2s_to_utf8s()` once.  I have also not tested this code on a big endian machine.  I did have a version with k5buf, but it's longer and more involved; the ucs2 converter function really wants to operate on the raw string, so the abstraction doesn't gain much.

src/lib/krb5/krb/sendauth.c
```diff
@@ -137,13 +137,13 @@ krb5_sendauth(krb5_context context, krb5_auth_context *auth_context,
         d.length = sizeof (rnd_data);
         d.data = rnd_data;
         len2 = sizeof (rnd_data);
-        if (getpeername (*(int*)fd, (GETPEERNAME_ARG2_TYPE *) rnd_data,
+        if (getpeername (*(int*)fd, (GETPEERNAME_ARG2_TYPE *)(void *)rnd_data,
```
frozencemetery: Added a new fix removing the _ARG2_TYPE macros.  I think if for some reason we want to keep them we could also just declare `GETPEERNAME_ARG2_TYPE rnd_data;` and it would work.

frozencemetery:
(I believe all outstanding comments have been addressed for the moment.)

src/util/support/utf8_conv.c
```diff
@@ -99,24 +98,16 @@ k5_utf8s_to_ucs2s(krb5_ucs2 *ucs2str,
...
-#ifdef K5_BE
-#ifndef SWAP16
-#define SWAP16(X)       ((((X) << 8) | ((X) >> 8)) & 0xFFFF)
-#endif
-            if (little_endian)
-                ucs2str[ucs2len] = SWAP16(ch);
-            else
-#endif
-                ucs2str[ucs2len] = ch;
+            ucs2str[ucs2len] = load_16_le(&ch);
```
greghudson: Something is wrong here (and in the old code), although I will have to look more closely to see how it ought to work.  load_16_le() must only be applied to a pointer to two bytes, not to a 16-bit value.  We should not be constructing a krb5_ucs2 array filled with byte-swaped values; we should produce a byte array and use store_16_le() to write the 16-bit values into it.

greghudson:
Why don't we pull the utf8_conv.c changes out into a separate PR, and I can hopefully merge the rest?

frozencemetery:
Sounds good; see #635.

greghudson:
I want to investigate the prof_file.c warning further, but I unfortunately can't reproduce it with clang 3.4 or 3.8.0 (on Ubuntu 14.04 and Ubuntu 16.04 respectively).

frozencemetery:
It shows up during `make check` for me on all tested versions (Ubuntu Trusty with 3.4 and 3.5 to mirror what Travis runs, Debian testing/sid with 3.8, 3.9, and 4.0 to futureproof).

greghudson:
Aha.  This is because we build a non-shared version of prof-file.o for the tests.  When that is build, SHARED is not defined, so k5-platform.h chooses the static function prototype with no uses of it.  The other source files which use finalizers do not get non-shared object files built.

I think it would make the most sense to put the __attribute__(unused) in the first MAKE_FINI_FUNCTION definition in k5-platform.h.  The other definitions shouldn't require it.

frozencemetery:
Sounds good, updated.

greghudson:
It still needs to be conditional on gcc.

frozencemetery:
So it does.  Updated.

src/lib/krb5/krb/sendauth.c
```diff
@@ -127,26 +127,27 @@ krb5_sendauth(krb5_context context, krb5_auth_context *auth_context,
...
-           This isn't strong cryptographically; the point here is
-           not to guarantee randomness, but to make it less likely
-           that multiple sessions could pick the same subkey.  */
-        char rnd_data[1024];
+        /* Provide some more fodder for random number code.  This isn't strong
+         * cryptographically; the point here is not to guarantee randomness,
+         * but to make it less likely that multiple sessions could pick the
+         * same subkey.
+         */
+        struct sockaddr rnd_data;
```
greghudson: ``struct sockaddr`` won't necessarily be large enough to hold the result of getpeername() or getsockname().  rnd_data should be of type ``struct sockaddr_storage`` and then its address should be casted with ss2sa() when passing it to getpeername/getsockname.

frozencemetery:
Fixed.

greghudson:
I pushed some minor changes to the PR branch; I expect to merge this tomorrow.

> I thought about turning on -Werror for clang (on Travis only, not in general) but could not find a good way to do this.

I will look into this.  One option is to turn on -Werror in configure when --enable-maintainer-mode is on and the compiler is clang, and then specify --enable-maintainer-mode for Travis.

> How would people feel about our CI running make -s? It would make poring over these logs easier, I think.

I don't think we want to do this.  When something fails in Travis we can't easily get at the build environment to debug it, so we shouldn't throw away information from the log.

> There remain warnings in the bundled libev code; how do we want to handle those?

Maybe try (in a separate PR) adding pragmas to verto-k5ev.c to disable the warning that pop up, before including ev.c.  I will also look at the current libev sources and see if they might fix any of the warnings, but I suspect that won't be enough by itself.

greghudson:
Confirmed: none of the ev.c warnings is fixed in the latest libev sources.  I think they're all of just three types (most of them are unused-value due to the string annotations in assertions), so it should only require a few pragmas to suppress them.

kaduk:
Yeah, the libev maintainer was not at all interested in taking patches just to appease compiler warnings, the last time I tried.
(Even one that was a legitimately unclear order of operations thing, where the intent of the code could not be reliably determined from just the code.)

PR #633, opened 2017-04-10
Remove all calls to krb5_c_random_add_entropy()

frozencemetery:
Feeding PRNG output back into the PRNG has never been helpful, and in
almost all non-Fortuna cases it will be ignored.  Seed Fortuna from the OS
if asked for octets and there isn't entropy.

I don't enable Fortuna, so this is a minor performance/convenience improvement for me.  I believe Debian does though; @kaduk, do you want to take a look at this?

### DIFF
--- a/src/kdc/dispatch.c
+++ b/src/kdc/dispatch.c
@@ -33,7 +33,7 @@
 #include <arpa/inet.h>
 #include <string.h>
 
-static krb5_int32 last_usec = 0, last_os_random = 0;
+static krb5_int32 last_os_random = 0;
 
 static krb5_error_code make_too_big_error(kdc_realm_t *kdc_active_realm,
                                           krb5_data **out);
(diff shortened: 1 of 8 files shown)


frozencemetery:
As a further rationale for this approach: if one applies the PR *without* the addition of `krb5_c_random_get_os_entropy()` to prng_fortuna.c, various parts of the test suite will fail out because the test suite doesn't get seeded.  This seems problematic to me because it means we rely on these reaches for entropy in normal-ish operation.

frozencemetery:
More concisely: I think it's probably incorrect to read from Fortuna without having seeded from the OS at least once.  If it is not desirable to remove these other seeds as well, I would like to at least fix it so we don't do that.

greghudson:
To be clear: what we rely on for entropy in normal operation is the call to krb5_c_random_os_entropy() from krb5_init_context().

frozencemetery:
(For clarity: I was mistaken about initialization because I accidentally removed the call to get OS entropy.  Thanks for your review.)

PR #635, opened 2017-04-12
Modernize utf8_conv.c

frozencemetery:
This is an outgrowth of #630 (and is based on those commits).  So far:

* Removed several unused functions, and thereby moved to little endian only
* Changed `k5_ucs2s_to_utf8s()` to operate on a byte array
* Using `load_16_le()` to load bytes
* Changed `k5_ucs2s_to_utf8s()` to operate in a single pass

### DIFF
--- a/src/include/k5-utf8.h
+++ b/src/include/k5-utf8.h
@@ -81,49 +81,22 @@ size_t krb5int_ucs2_to_utf8(krb5_ucs2 c, char *buf);
 int krb5int_utf8_to_ucs4(const char *p, krb5_ucs4 *out);
 size_t krb5int_ucs4_to_utf8(krb5_ucs4 c, char *buf);
 
-int
-krb5int_ucs2s_to_utf8s(const krb5_ucs2 *ucs2s,
-                       char **utf8s,
-                       size_t *utf8slen);
+/*
+ * Convert a little-endian UCS-2 string to an allocated null-terminated UTF-8
+ * string.  nbytes is the length of ucs2bytes in bytes, and must be an even
+ * number.  Return EINVAL on invalid input, ENOMEM on out of memory, or 0 on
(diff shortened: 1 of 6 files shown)


greghudson:
I pushed a second commit to show how I'd like the code to look, but I also want to tamper with the function names and contracts a bit, and maybe split this into a couple of commits.

greghudson:
My new UTF8->UCS2 code ignores the utf8 string length (that's the reason for the CI failure).  While investigating, I noticed that the old code isn't perfect about respecting the length either, even after commit eb52da21d72faa3d00b1205a5a0fdbabc45c9e6d.  krb5int_utf8cs_to_ucs2s() calls krb5int_utf8c_chars() to get a count of the UTF-8 characters in utf8s/utf8slen, and that function just keeps skipping over utf8 characters with KRB5_UTF8_INCR() until it hits or runs past the end.  But KRB5_UTF8_INCR() doesn't take a length parameter, so if the counted string contains an invalid UTF-8 sequence at the end, we could overrun the count by a few bytes.

The security impact seems minimal, but we'll want to fix this bug in a separate commit prior to any modernization of the code for easy backporting.  Since we don't currently have a length-bounded primitive to get the length of a UTF-8 character, I am thinking the most robust fix is to abandon the length-counted-UTF8 variant of UTF8->UCS2 and just k5memdup0 the string in RC4 string-to-key before calling the null-terminated variant.

greghudson:
I rebased onto PR #636, fixed up my work a bit, and squashed it in with the existing one (I decided that splitting it into multiple commits wasn't needed).  Please have a look when convenient.

frozencemetery:
Ah, I had been hoping to avoid allocation there, but I think you're right that this is better.  I'm okay with this approach.

PR #638, opened 2017-04-18
Replace references to OS X

mmattioli:
Replace references to Apple OS X with the updated OS name "macOS" where
applicable.


greghudson:
Thanks for this.

config.guess and ev.c both have upstream sources, and we shouldn't change them.

A few places would now say "Apple macOS", including a couple that didn't say "Apple OS X" before (as well as a couple that did).  Is there a particular reason for these?

I'm not sure what I think about saying "macOS Panther" or otherwise referring to old versions using the new branding, although I guess Apple considers their rebranding to be retroactive.

mmattioli:
@greghudson I removed the changes to `config.guess` and `ev.c` as well as changed everything to simply "macOS" rather than "Apple macOS" in some places. In regards to the Panther reference, I modified it to refer to the OS version number (10.3).

PR #640, opened 2017-04-19
Update Travis CI distribution to Xenial

mmattioli:
Update the Travis CI OS distribution to Ubuntu 16.04 LTS (Xenial Xerus). The current OS distribution, Ubuntu 14.04 LTS (Trusty Tahr), was chosen before 16.04 LTS was released and, presumably, the update was simply overlooked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: c++
 
 sudo: required
 
-dist: trusty
+dist: xenial
 
 matrix:
   include:


frozencemetery:
Travis is doing something weird here.  Xenial should be clang-3.8 by default, but it shows as having 3.5.  It looks like Xenial does ship both; maybe we should ask it to build with the newer one?

greghudson:
Building with clang-3.8 seems like a good idea, but I'm not sure how.  I couldn't find a declarative way to ask for a specific clang version.  Adding CC and CXX to the "env" line for that entry in the build matrix didn't work--the "compiler: clang" environment variables are added after the "env" variables, overriding them.  I tried adding a ``$CONFVARS`` substitution to the script's configure command and setting ``CONFVARS='CC=clang-3.8 CXX=clang++-3.8'`` in the env line, but configure blows out with "C compiler cannot create executables".

frozencemetery:
Weird, I can't find any information on how to do this either.  I guess their support is better for other languages.

My guess is that they're explicitly installing the 3.5 package and not the default package.  We already require sudo, so we could adjust what's installed (and any symlinks, though I don't think we'd need to).  On the other hand, 3.5 is what we were testing with before, so not doing that isn't making anything worse.

greghudson:
After spinning up an inordinate number of Travis builds, I think I've determined that Travis doesn't install any clang packages, but instead puts a build of clang 3.5 in /usr/local/clang-3.5.0 which takes precedence in the path over anything in /usr/bin.  So to use clang 3.8, we'll need to install a clang package and ``sudo rm -rf /usr/local/clang*`` (or do some path manipulation).

mmattioli:
According to the [Travis build for this commit](https://travis-ci.org/krb5/krb5/builds/223391390), the `.travis.yml` that currently exists in the `master` branch is being used which is causing the [Trusty distribution to be used for this particular job](https://travis-ci.org/krb5/krb5/jobs/223391391#L476).

@greghudson, I imagine in the various Travis builds you've done you've used a `.travis.yml` file which specifically chooses Xenial before checking the clang version?

greghudson:
> @greghudson, I imagine in the various Travis builds you've done you've used a .travis.yml file which specifically chooses Xenial before checking the clang version?

Yes, I used a modified version of your commit for testing, on a branch in my fork (which is also configured for Travis).

greghudson:
> So to use clang 3.8, we'll need to install a clang package and sudo rm -rf /usr/local/clang* (or do some path manipulation).
Hm, nope.  After nuking /usr/local/clang*, the next command fails with "/home/travis/.travis/job_stages: line 53: /usr/local/clang-3.5.0/bin/clang: No such file or directory".  Not sure what magic is involved there.

I will probably just take this PR as-is, as there doesn't seem to be a good path to using clang 3.8.

kaduk:
FWIW, https://github.com/openssl/openssl/blob/master/.travis.yml has stanzas that claim to enable clang-3.9, though I have not personally verified that they actually do so.

greghudson:
> FWIW, https://github.com/openssl/openssl/blob/master/.travis.yml has stanzas that claim to enable clang-3.9, though I have not personally verified that they actually do so.

That does seem to work for OpenSSL, but after some additional experimentation I'm not sure quite what the secret sauce is.  If I say ``compiler: clang-3.9`` in the build matrix and add an addon directive to install clang-3.9, I see in the build log:

    $ export CXX=clang++
    $ export CC=clang
    $ clang-3.9 --version
    clang-3.9: command not found

The build then continues on successfully, although I'm guessing with clang 3.5.  In the OpenSSL build logs, I see instead:

    $ export CC=clang-3.9
    [some stuff relating to ccache, which is a compiler caching tool, not a credential cache]
    $ clang-3.9 --version
    clang version 3.9.1-svn288847-1~exp1 (branches/release_39)
    [...]

There are some differences which might be relevant, but I can't figure out whicih: OpenSSL is using ``language: c`` while we recently switched to ``language: c++``; OpenSSL is using trusty and we want to use xenial; OpenSSL is adding an apt source for the clang-3.9 package since Trusty doesn't have one natively.

frozencemetery:

> I will probably just take this PR as-is, as there doesn't seem to be a
> good path to using clang 3.8.

Agreed, we shouldn't hold this PR back if it's going to be nontrivial.

mmattioli:
I did some more digging on Travis and Xenial support w/clang 3.8 and it appears as though [as of November 2016 support for Xenial (clang 3.8 as the default version) is still on their roadmap](https://twitter.com/travisci/status/803596876620693504?lang=en). I noticed that even when specifying `xenial` as the distribution, Travis still uses Trusty (Ubuntu 14.04 LTS). There is also an open [issue about this on GitHub](https://github.com/travis-ci/travis-ci/issues/5821) and several people from Travis are more or less saying that they recognize people want it and they're working to support it but, as of right now, the only way to use Xenial (Ubuntu 16.04 LTS) is through Docker.

PR #641, opened 2017-04-19
Add UTF-16 conversions in place of UCS-2

greghudson:
[This will be kind of dense to review, but it isn't a lot of code, and I tried to comment and structure it well enough that it isn't an impenetrable wall of magic numbers.  The wikipedia article on UTF-16 contains a nice summary of the encoding.]

Where we convert between UTF-8 and UCS-2 (RC4 string-to-key and PAC
client info), use UTF-16 instead of UCS-2.  Add a test program for
the conversion functions.

### DIFF
--- a/src/util/support/utf8_conv.c
+++ b/src/util/support/utf8_conv.c
@@ -1,7 +1,7 @@
 /* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 /* util/support/utf8_conv.c */
 /*
- * Copyright 2008 by the Massachusetts Institute of Technology.
+ * Copyright 2008, 2017 by the Massachusetts Institute of Technology.
  * All Rights Reserved.
  *
  * Export of this software from the United States of America may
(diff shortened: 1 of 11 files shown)


src/util/support/utf8_conv.c
```diff
@@ -47,34 +47,56 @@
  * THE PERPETRATOR TO CRIMINAL AND CIVIL LIABILITY.
  */
 
-/* This work is part of OpenLDAP Software <http://www.openldap.org/>. */
+/* This work is based on OpenLDAP Software <http://www.openldap.org/>. */
```
frozencemetery: Should copyright update here as well?

frozencemetery:
I was also confused by the commit message initially - maybe it would be better to say something like "Replace UCS-2 conversions with UTF-16"?

greghudson:
Comments addressed.

I think there is a fair amount of unused stuff in utf8.c beyond just those two functions, and also a lot of stuff in k5-utf8.h which should move to a private header in util/support (or we could just combine utf8.c and utf8_conv.c).  I would like to defer that cleanup.

PR #643, opened 2017-04-20
Apply TCP timeouts to HTTPS (KKDCP) transport

greghudson:
We apply (as of ticket #7604) a ten-second minimum delay after a TCP
connection is accepted before creating new connections or sending UDP
packets.  Apply this timeout to HTTPS connections as well, by removing
the transport check in get_endtime().  As the endtime field is only
set by service_tcp_connect(), it will always have the value 0 for UDP
connection state objects, so there is no need to check the transport
type.


greghudson:
I have confirmation from the reporter that this works for him (on a FAST-OTP deployment that doesn't tolerate replays, with a KKDCP transport), so I will likely merge this tomorrow.

PR #647, opened 2017-04-25
Update cmocka to 1.1.1

mmattioli:
Update cmocka used in Travis builds to version 1.1.1; this version
was released on April 4, 2017. As of release 1.1.0, new features like
support to verify call ordering, to pass initial data to test cases,
catch multiple exceptions and improved XML output have been added.
Release 1.1.1 fixes several bugs in 1.1.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,10 @@ before_install:
   - sudo apt-get install -y bison dejagnu gettext keyutils ldap-utils libldap2-dev libkeyutils-dev libssl-dev python-cjson python-paste python-pyrad slapd tcl-dev tcsh
   - mkdir -p cmocka/build
   - cd cmocka
-  - wget https://cmocka.org/files/1.0/cmocka-1.0.1.tar.xz
-  - tar -xvf cmocka-1.0.1.tar.xz
+  - wget https://cmocka.org/files/1.1/cmocka-1.1.1.tar.xz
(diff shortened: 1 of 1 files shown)


frozencemetery:
Sorry, what's the reason for this change?  (Should be mentioned in commit message.)

mmattioli:
@frozencemetery I just modified the commit message

frozencemetery:
Okay, so since there isn't anything we know we need in this version, I'd prefer to wait until the distros have the 1.1 series available before we bump our expected version.  (Currently it's present in Fedora but missing in Debian/family, so I filed [a bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861266), but it'll probably not get fixed until after the Stretch release.)

greghudson:
I agree; there's no need to keep the Travis build script in sync with cmocka unless we think cmocka releases will get pickier about its callers, which isn't common for libraries.  (By contrast, C toolchains do seem to get pickier about source code over time, so it's best to do CI with the most recent toolchain available, if you're not going to do multiple versions.)

mmattioli:
While I am in agreement that the features added aren't necessarily a concern, there are bug fixes and performance improvements (although Travis is doing the builds so we don't necessarily care about performance) that I figured would be of benefit. I suppose you can close this if neither of you are keen on bumping the version.

PR #648, opened 2017-04-25
Refactor kvno for uniformity

mmattioli:
Several `if` and `if/else` statements were not enclosed with curly
braces similarly to the rest of the statements in the file; in some
situations it made reading through the code somewhat confusing.

The `quiet` flag is declared as static since it's use is local to kvno..

### DIFF
--- a/src/clients/kvno/kvno.c
+++ b/src/clients/kvno/kvno.c
@@ -36,6 +36,7 @@ extern int optind;
 extern char *optarg;
 
 static char *prog;
+static int quiet = 0;
 
 static void xusage()
 {
@@ -44,8 +45,6 @@ static void xusage()
     fprintf(stderr, _("\tservice1 service2 ...\n"));
     exit(1);
 }
-
-int quiet = 0;
 
 static void do_v5_kvno (int argc, char *argv[],
                         char *ccachestr, char *etypestr, char *keytab_name,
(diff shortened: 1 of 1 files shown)


frozencemetery:
Your concerns about readability are appreciated.

We tend to avoid this kind of code churn in order to keep backporting simpler.  Also, while we do try to match existing style within files, krb5 tends toward not bracing single-line `if` statements; your changes make this file less consistent on this, not more.  As someone who frequently does backports, NACK.

greghudson:
We do sometimes revise old code for consistency with the current style, but Robbie is right that we don't generally put braces around single-line control statement bodies.

greghudson:
We had some discussion about this in our weekly meeting.

Changes to improve coding style conformance carries a benefit in readability, but a cost in churn, which makes back-porting or forward-porting changes harder.  To mitigate this cost, we try to make readability improvements in somewhat chunks, so that the odds of a change crossing a churn barrier are reduced.

Normally, modernizing a whole program would be a reasonably-sized chunk.  But kvno is a small program (one 300-line file) and is closely associated with kinit, kdestroy, and klist.  So those four programs might make a reasonable unit.  They also have some residual complexity left behind from the days of supporting krb4, which could be cleaned up at the same time.

mmattioli:
@greghudson would you like me to modify this PR to include those other clients?

greghudson:
Sure, if you're interested in doing that work.  If you haven't seen it yet, our coding style is documented at http://k5wiki.kerberos.org/wiki/Coding_style

mmattioli:
I just opened a separate PR (#651); I'll close this one since the changes here are incorporated there.

PR #650, opened 2017-04-25
Use a random nonce in TGS requests

greghudson:
[KDC-REQ nonces equal to the current timestamp won't work reliably after 2038, because of a disagreement between RFC 4120 and MIT krb5/Heimdal on whether the nonce field is signed.  Of course we could mask off the high bit from the timestamp just as easily as from random bytes, but this is better.]

Commit ae0fee058ad883b2e82fa2b34f4e5f059e827a1b (ticket #5425) changed
the AS client code to use a random nonce, but left the TGS client code
using the current timestamp.  Use a random nonce for TGS requests as
well.

### DIFF
--- a/src/lib/krb5/krb/send_tgs.c
+++ b/src/lib/krb5/krb/send_tgs.c
@@ -28,6 +28,25 @@
 #include "int-proto.h"
 #include "fast.h"
 
+/* Choose a random nonce for an AS or TGS request. */
+krb5_error_code
+k5_generate_nonce(krb5_context context, int32_t *out)
+{
+    krb5_error_code ret;
+    unsigned char random_buf[4];
+    krb5_data random_data = make_data(random_buf, 4);
+
+    *out = 0;
+
+    /* We and Heimdal incorrectly encode nonces as signed, so make sure we use
+     * a non-negative value to avoid interoperability issues. */
+    ret = krb5_c_random_make_octets(context, &random_data);
+    if (ret)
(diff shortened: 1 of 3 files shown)


src/lib/krb5/krb/send_tgs.c
```diff
@@ -28,6 +28,25 @@
...
+    krb5_data random_data = make_data(random_buf, 4);
+
+    *out = 0;
+
+    /* We and Heimdal incorrectly encode nonces as signed, so make sure we use
+     * a non-negative value to avoid interoperability issues. */
+    ret = krb5_c_random_make_octets(context, &random_data);
+    if (ret)
+        return ret;
+    *out = 0x7FFFFFFF & load_32_n(random_buf);
```
kaduk: It feels a little silly to load_32_n a buffer of random bytes, but it seems to make sense to preserve the abstractions.

greghudson:
I was copying that code from get_in_tkt.c.  Ken actually created load_32_n for the commit which introduced that pattern, but I'm not sure what his reasoning was.  It seems like it would work to just put the random bytes into the output variable (via ``make_data(&var, sizeof(var)``) and then mask off the high bit.

src/lib/krb5/krb/send_tgs.c
frozencemetery: It's a little weird, but I don't think it's worth another commit since right now it's just code movement.

PR #651, opened 2017-04-27
Refactor clients to conform to the coding style

mmattioli:
Clients were refactored to more closely conform to the style and
formatting guide as specified in
http://k5wiki.kerberos.org/wiki/Coding_style

### DIFF
--- a/src/clients/kdeltkt/kdeltkt.c
+++ b/src/clients/kdeltkt/kdeltkt.c
@@ -7,26 +7,28 @@
 #include "k5-platform.h"
 
 static char *prog;
+static int quiet = 0;
 
-static void xusage()
+static void
+xusage()
 {
-    fprintf(stderr, "xusage: %s [-c ccache] [-e etype] [-f flags] service1 service2 ...\n", prog);
+    fprintf(stderr, "xusage: %s [-c ccache] [-e etype] [-f flags] service1 "
+            "service2 ...\n", prog);
     exit(1);
 }
 
-int quiet = 0;
+static void
+do_kdeltkt(int argc, char *argv[], char *ccachestr, char *etypestr, int flags);
 
-static void do_kdeltkt (int argc, char *argv[], char *ccachestr, char *etypestr, int flags);
-
-int main(int argc, char *argv[])
+int
+main(int argc, char *argv[])
 {
     int option;
-    char *etypestr = 0;
-    char *ccachestr = 0;
+    char *etypestr = NULL, *ccachestr = NULL;
     int flags = 0;
 
     prog = strrchr(argv[0], '/');
-    prog = prog ? (prog + 1) : argv[0];
+    prog = (prog != NULL) ? prog + 1 : argv[0];
 
     while ((option = getopt(argc, argv, "c:e:f:hq")) != -1) {
         switch (option) {
@@ -49,15 +51,16 @@ int main(int argc, char *argv[])
         }
     }
 
-    if ((argc - optind) < 1)
+    if (argc - optind < 1)
         xusage();
 
     do_kdeltkt(argc - optind, argv + optind, ccachestr, etypestr, flags);
     return 0;
 }
 
-static void do_kdeltkt (int count, char *names[],
-                        char *ccachestr, char *etypestr, int flags)
+static void
+do_kdeltkt(int count, char *names[], const char *ccachestr, char *etypestr,
+           int flags)
 {
     krb5_context context;
     krb5_error_code ret;
@@ -75,7 +78,7 @@ static void do_kdeltkt (int count, char *names[],
         exit(1);
     }
 
-    if (etypestr) {
+    if (etypestr != NULL) {
         ret = krb5_string_to_enctype(etypestr, &etype);
         if (ret) {
             com_err(prog, ret, "while converting etype");
@@ -111,9 +114,10 @@ static void do_kdeltkt (int count, char *names[],
 
         ret = krb5_parse_name(context, names[i], &in_creds.server);
         if (ret) {
-            if (!quiet)
+            if (!quiet) {
                 fprintf(stderr, "%s: %s while parsing principal name\n",
                         names[i], error_message(ret));
+            }
             errors++;
             continue;
         }
@@ -133,9 +137,7 @@ static void do_kdeltkt (int count, char *names[],
         if (ret) {
             fprintf(stderr, "%s: %s while retrieving credentials\n",
                     princ, error_message(ret));
-
             krb5_free_unparsed_name(context, princ);
-
             errors++;
             continue;
         }
@@ -147,14 +149,11 @@ static void do_kdeltkt (int count, char *names[],
         if (ret) {
             fprintf(stderr, "%s: %s while removing credentials\n",
                     princ, error_message(ret));
-
             krb5_free_cred_contents(context, &out_creds);
             krb5_free_unparsed_name(context, princ);
-
             errors++;
             continue;
         }
-
         krb5_free_unparsed_name(context, princ);
         krb5_free_cred_contents(context, &out_creds);
     }
--- a/src/clients/kdestroy/kdestroy.c
+++ b/src/clients/kdestroy/kdestroy.c
@@ -41,7 +41,7 @@ extern int optind;
 extern char *optarg;
 
 #ifndef _WIN32
-#define GET_PROGNAME(x) (strrchr((x), '/') ? strrchr((x), '/')+1 : (x))
+#define GET_PROGNAME(x) (strrchr((x), '/') ? strrchr((x), '/') + 1 : (x))
 #else
 #define GET_PROGNAME(x) max(max(strrchr((x), '/'), strrchr((x), '\\')) + 1,(x))
 #endif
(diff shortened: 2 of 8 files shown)


mmattioli:
@greghudson is there anything else you'd like to see in this PR?

greghudson:
Sorry for the delay.  We have a style checker described here: http://k5wiki.kerberos.org/wiki/Coding_style/Style_checker

It detects 105 problems with the changed lines in this commit, and might detect further problems with unchanged code.  (It's not perfect; if it seems to report something wrong, ignore it.)

mmattioli:
I ran that script before committing and noted that it reports false positives; I reviewed what it generated and didn't seem to find anything valid (unless you found otherwise?).

greghudson:
The kinds of messages I see are:

* Length exceeds 79 characters: this is pretty much never a false positive in a modernized source file.  String literals can be broken up using C's string literal concatenation ("a" "b" is the same as "ab").

* Body is 2+ lines but has no braces: if the body is a single line, we don't ordinarily use braces (unless the other arm of an if-else needs braces), but if it's a single statement that takes multiple lines, we do.

* Indentation may be incorrect: these arise for various reasons, sometimes whitespace on a blank line, sometimes misalignment.

* No space after flow control keyword: "for(;;)" should be "for (;;)" for instance.

* Assignment in if conditional: break these up using "retval = fn(...);" followed by "if (retval)".

* One arm of if/else statement braced but not the other: we put braces around single-line bodies of if-else statements if the other arm needs braces.

* Space before close parenthesis: "strdup( password )" should be "strdup(password)", for instance.

* Trailing whitespace: this should always be eliminated when changing a line of code.

* No space before/after binary operator: "found_count +1" should be "found_count + 1", for instance.

mmattioli:
I just squashed all of the messages generated from the previous commit and re-pushed. These are the only messages it generates and they're all confirmed false positives:
```
dbd935a36 Refactor clients to conform to the coding style
  clients/klist/klist.c:
      225  Preprocessor statement in function body
  clients/ksu/authorization.c:
       62  Indentation may be incorrect
      566  Indentation may be incorrect
  clients/ksu/ccache.c:
       52  Indentation may be incorrect
  clients/ksu/ksu.h:
      231  Indentation may be incorrect
      232  Indentation may be incorrect
      234  Indentation may be incorrect
      235  Indentation may be incorrect
      238  Indentation may be incorrect
  clients/ksu/main.c:
      181  No space before or after binary operator
  clients/kvno/kvno.c:
      347  Indentation may be incorrect
      347  Trailing whitespace
```

mmattioli:
@greghudson let me know if you still require further modifications to this PR. Thanks.

mmattioli:
@greghudson just checking in on this. Thanks.

mmattioli:
@greghudson 😁

greghudson:
Apologies for the slow review cycle on this PR; I had the flu for most of last week.

In the previous PR I had suggested kinit/kvno/klist/kdestroy.  Adding kcpytkt and kdeltkt to that set isn't a problem because they're small, but ksu is a beast of a different size, with a lot of technical debt.  Since I expect to need to make a pass over all of the affected files to look for additional style changes that need to be made, I would like to defer the ksu changes for now if that's okay.

The style checker's report of "trailing whitespace" at kvno.c:347 is not a false positive; the PR commit adds eight spaces to a blank line there.

mmattioli:
@greghudson I just fixed the whitespace on `kvno.c:347`. I don't mind if it takes a while for you to review the whole PR?

greghudson:
> I don't mind if it takes a while for you to review the whole PR? 

A brief look suggests that it would take a lot of my time to go over the ksu stuff; there are a ton of unfixed style issues there, including many pre-ANSI function definitions.  Red Hat also has a PAM patch for ksu which they don't really want to forward-port across a bunch of code churn.  So I will definitely be deferring the ksu changes from this PR.

greghudson:
I pushed a revised commit.  The original commit (with the ksu changes) is stashed at https://github.com/greghudson/krb5/commits/pr651.orig .

src/clients/kvno/kvno.c
```diff
@@ -36,35 +36,35 @@ extern int optind;
...
-                        char *sname, int canon, int unknown,
-                        char *for_user, int proxy);
+static void do_v5_kvno(int argc, char *argv[], char *ccachestr, char *etypestr,
+                       char *keytab_name, char *sname, int canon, int unknown,
+                       char *for_user, int proxy);
 
 #include <com_err.h>
-static void extended_com_err_fn (const char *, errcode_t, const char *,
-                                 va_list);
+static void extended_com_err_fn(const char *, errcode_t, const char *,
```
frozencemetery: Could we name the parameters, while we're here?

src/clients/kdestroy/kdestroy.c
```diff
@@ -135,62 +128,61 @@ main(argc, argv)
...
 
     if (all) {
-        code = krb5_cccol_cursor_new(kcontext, &cursor);
+        code = krb5_cccol_cursor_new(context, &cursor);
         if (code) {
             com_err(progname, code, _("while listing credential caches"));
             exit(1);
         }
-        while ((code = krb5_cccol_cursor_next(kcontext, cursor,
+        while ((code = krb5_cccol_cursor_next(context, cursor,
```
frozencemetery: The value of `code` isn't used here, so we don't need to do the assignment in the loop conditional.

src/clients/klist/klist.c
```diff
@@ -408,120 +404,126 @@ list_ccache(krb5_ccache cache)
...
         return 1;
     }
 
     printf(_("Ticket cache: %s:%s\nDefault principal: %s\n\n"),
-           krb5_cc_get_type(kcontext, cache),
-           krb5_cc_get_name(kcontext, cache), defname);
-    /* XXX Translating would disturb table alignment; skip for now. */
+           krb5_cc_get_type(context, cache), krb5_cc_get_name(context, cache),
+           defname);
+    /* Translating would disturb table alignment; skip for now. */
```
frozencemetery: Also here.

src/clients/kdeltkt/kdeltkt.c
```diff
@@ -7,26 +7,28 @@
...
 
 static char *prog;
+static int quiet = 0;
 
-static void xusage()
+static void
+xusage()
 {
-    fprintf(stderr, "xusage: %s [-c ccache] [-e etype] [-f flags] service1 service2 ...\n", prog);
+    fprintf(stderr, "xusage: %s [-c ccache] [-e etype] [-f flags] service1 "
```
greghudson: The string doesn't fit on one line even if we break before and after it.

frozencemetery: Correct, ignore that.

src/clients/kinit/kinit.c
```diff
@@ -199,34 +196,31 @@ usage()
...
+    vfprintf(stderr, fmt, args);
+    fprintf(stderr, "\n");
 }
 
 static int
 add_preauth_opt(struct k_opts *opts, char *av)
 {
     char *sep, *v;
     krb5_gic_opt_pa_data *p, *x;
+    size_t newsize = (opts->num_pa_opts + 1) * sizeof(*opts->pa_opts);
```
greghudson: realloc(NULL, n) for n>0 is equivalent to malloc(n).  We know that opts->pa_opts is NULL in this case since opts is allocated zero-filled.

src/clients/kinit/kinit.c
```diff
@@ -738,63 +697,61 @@ k5_kinit(opts, k5)
...
-    if ((opts->action == INIT_KT) && opts->keytab_name)
-    {
+    if (opts->action == INIT_KT && opts->keytab_name != NULL) {
 #ifndef _WIN32
         if (strncmp(opts->keytab_name, "KDB:", 4) == 0) {
-            code = kinit_kdb_init(&k5->ctx,
-                                  krb5_princ_realm(k5->ctx, k5->me)->data);
-            if (code != 0) {
-                com_err(progname, code,
+            ret = kinit_kdb_init(&k5->ctx, k5->me->realm.data);
```
greghudson: krb5_princ_realm() is a macro, and I've always found that those macros obfuscate code with no benefit, especially since they take an unused context argument.  We've removed uses of those macros in the past (e.g. caaf72893a5be61822763eb471f4d573992479ed ).

src/clients/klist/klist.c
```diff
@@ -560,32 +562,28 @@ check_ccache(krb5_ccache cache)
...
-    krb5_enctype enctype;
+static char *
+etype_string(krb5_enctype enctype)
 {
     static char buf[100];
-    krb5_error_code retval;
+    krb5_error_code ret;
 
-    if ((retval = krb5_enctype_to_name(enctype, FALSE, buf, sizeof(buf)))) {
-        /* XXX if there's an error != EINVAL, I should probably report it */
```
greghudson: I'm not sure it's enough of a concern to justify the comment.  krb5_enctype_to_name() can't return an error other than EINVAL unless you pass it an overly short buffer (and 100 bytes is plenty for any enctype name), and there's no reason to expect that to change.

greghudson:
I pushed a revised commit with changes to address all of the comments I didn't respond to.

PR #653, opened 2017-05-04
Year 2038 changes

greghudson:
Here are changes to make in-tree code work after y2038 and document what third-party code should do.

Closer to the year 2038, we might consider changing krb5_timestamp to uint32_t.  Doing that could break third-party code which subtracts timestamps and therefore does not seem advisable in 2017, but might be a reasonable tradeoff in, say, 2033.  The helper functions and documented practice for using krb5_timestamp values should continue to be safe if we make that change.

Project page: https://k5wiki.kerberos.org/wiki/Projects/Timestamps_after_2038

### DIFF
--- a/src/kdc/fast_util.c
+++ b/src/kdc/fast_util.c
@@ -607,7 +607,7 @@ kdc_fast_read_cookie(krb5_context context, struct kdc_request_state *state,
     ret = krb5_timeofday(context, &now);
     if (ret)
         goto cleanup;
-    if (now - COOKIE_LIFETIME > cookie->time) {
+    if (ts2tt(now) > cookie->time + COOKIE_LIFETIME) {
         /* Don't accept the cookie contents.  Only return an error if the
          * cookie is relevant to the request. */
         if (is_relevant(cookie->data, req->padata))
@@ -700,7 +700,7 @@ kdc_fast_make_cookie(krb5_context context, struct kdc_request_state *state,
     ret = krb5_timeofday(context, &now);
     if (ret)
         goto cleanup;
-    cookie.time = now;
+    cookie.time = ts2tt(now);
     cookie.data = contents;
     ret = encode_krb5_secure_cookie(&cookie, &der_cookie);
     if (ret)
--- a/src/lib/krb5/krb/get_in_tkt.c
+++ b/src/lib/krb5/krb/get_in_tkt.c
@@ -40,24 +40,6 @@ static krb5_error_code sort_krb5_padata_sequence(krb5_context context,
                                                  krb5_pa_data **padata);
 
 /*
- * This function performs 32 bit bounded addition so we can generate
- * lifetimes without overflowing krb5_int32
- */
-static krb5_int32
-krb5int_addint32 (krb5_int32 x, krb5_int32 y)
-{
-    if ((x > 0) && (y > (KRB5_INT32_MAX - x))) {
-        /* sum will be be greater than KRB5_INT32_MAX */
-        return KRB5_INT32_MAX;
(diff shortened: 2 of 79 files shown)


frozencemetery:
The test suite does not pass after year 2038; is this expected?

greghudson:
> The test suite does not pass after year 2038; is this expected?

No; where does it fail?

src/lib/krb5/krb/get_in_tkt.c
```diff
@@ -1881,7 +1863,9 @@ k5_populate_gic_opt(krb5_context context, krb5_get_init_creds_opt **out,
         if (retval)
             goto cleanup;
         if (creds->times.starttime) starttime = creds->times.starttime;
-        krb5_get_init_creds_opt_set_tkt_life(opt, creds->times.endtime - starttime);
+        krb5_get_init_creds_opt_set_tkt_life(opt,
```
greghudson: I agree this is awkward.  I don't like putting a newline right after the open paren of a function call, but it might make sense to introduce a local to hold ``ts_delta(cred->times.endtime, starttime)``.

(Fundamentally I wish the krb5_get_init_creds_opt_ functions had used krb5_gic_opt as a prefix so they weren't so long, but that ship sailed long ago.)

src/tests/t_y2038.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python
+from k5test import *
+
+# These tests will become much less important after the y2038 boundary
+# has elapsed, and may start exhibiting problems around the year 2075.
```
greghudson: I don't think we need a warning for y2038 elapsing.  After that point, testing positive-to-negative timestamp transition scenarios will be moot.

I also don't think we need a specific warning for the early y2106 failure in the test suite.  If the project still has enough life in it in 2075 to successfully do y2106 remediation, people will probably have started thinking it by then, or at least have revisited this test script.

src/kdc/fast_util.c
```diff
@@ -607,7 +607,7 @@ kdc_fast_read_cookie(krb5_context context, struct kdc_request_state *state,
     ret = krb5_timeofday(context, &now);
     if (ret)
         goto cleanup;
-    if (now - COOKIE_LIFETIME > cookie->time) {
+    if (ts_after(ts_incr(now, -COOKIE_LIFETIME), cookie->time)) {
```
greghudson: Sure.

greghudson:
I pushed changes to the second commit affecting:

* kadmin/dbutil/dump.c conversion of timestamps to ASCII decimal
* lib/kadm5/srv/svr_principal.c addition of current time and pw_expiration
* lib/kadm5/srv/server_acl.c addition and comparison of timestamps to enforce max lifetime restrictions
* kadmin/testing/util/tcl_kadm5.c conversion of timestamps to ASCII decimal

The last bullet point was the first reason why the test suite fails after y2038; the dejagnu tests were receiving timestamps as negative values, but ignoring the minus sign (because of the regexp they used).

The second reason for failure is that tests/dejagnu/config/default.exp sets ``default_principal_expiration = 2037.12.31.23.59.59`` so nothing can authenticate.  I haven't removed those lines yet because that's our only exercise of default_principal_expiration in the test suite, but I might anyway as it's only an exercise, not a test.

The third reason (not sure this is the final one) is that make-certs.sh makes certificates that expire after about ten years from the last time the script is run.  This isn't a y2038 issue.  We could consider setting  alonger expiration time, making certs that don't expire (not sure that this is possible), or making certs dynamically and skipping the appropriate tests if there's no openssl command in the path.  Regardless, this issue is out of scope for the current PR.

PR #654, opened 2017-05-11
Pass the address from the connecting client to audit_as_req

cryptomilk:
In Samba we've added an audit logging system. To be able to call that from the Samba KDB module I need the client address. We have that information but it is not passed down to the KDB module.

As this is an incompatible change to the API also raise the DAL version.

### DIFF
--- a/src/lib/kdb/kdb5.c
+++ b/src/lib/kdb/kdb5.c
@@ -322,12 +322,7 @@ copy_vtable(const kdb_vftabl *in, kdb_vftabl *out)
     out->audit_as_req = in->audit_as_req;
     out->refresh_config = in->refresh_config;
     out->check_allowed_to_delegate = in->check_allowed_to_delegate;
-
-    /* Copy fields for minor version 1 (major version 6). */
-    assert(KRB5_KDB_DAL_MAJOR_VERSION == 6);
-    out->free_principal_e_data = NULL;
-    if (in->min_ver >= 1)
-        out->free_principal_e_data = in->free_principal_e_data;
+    out->free_principal_e_data = in->free_principal_e_data;
 
     /* Set defaults for optional fields. */
     if (out->fetch_master_key == NULL)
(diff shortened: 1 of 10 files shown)


src/lib/kdb/kdb5.c
```diff
@@ -323,10 +323,8 @@ copy_vtable(const kdb_vftabl *in, kdb_vftabl *out)
     out->refresh_config = in->refresh_config;
     out->check_allowed_to_delegate = in->check_allowed_to_delegate;
 
-    /* Copy fields for minor version 1 (major version 6). */
-    assert(KRB5_KDB_DAL_MAJOR_VERSION == 6);
     out->free_principal_e_data = NULL;
-    if (in->min_ver >= 1)
+    if (in->maj_ver > 6 || (in->maj_ver == 6 && in->min_ver >= 1))
```
greghudson: You can just remove this conditional; all loaded DB modules will have maj_ver > 6.

src/tests/kdbtest.c
```diff
@@ -244,7 +244,7 @@ static void
 sim_preauth(krb5_timestamp authtime, krb5_boolean ok, krb5_db_entry **entp)
 {
     /* Both back ends ignore the request parameter for now. */
-    krb5_db_audit_as_req(ctx, NULL, *entp, *entp, authtime,
+    krb5_db_audit_as_req(ctx, NULL, NULL, *entp, *entp, authtime,
```
greghudson: The comment above this line explains why it's okay to pass a null pointer for the request; it should be adjusted to also mention from.

cryptomilk:
I've did the changes you requested. As the port number is random from the client I do not really see any use for that.

Thanks for the review!

greghudson:
This is also an incompatible change to the libkdb API, so you need to bump KRB5_KDB_API_VERSION in kdb.h and LIBMAJOR in lib/kdb/Makefile.in

cryptomilk:
Ok, I've change the KDB version too and it build with new version number: ./lib/kdb/libkdb5.so.9.0

I've adapted the commit message to mention KDB too.

PR #655, opened 2017-05-17
y2038 followup work

greghudson:
While doing the y2038 work I noted that some code used krb5_int32 directly instead of krb5_timestamp to hold timestamp values, and figured that would be simplest to treat as a cleanup opportunity for later.  It turns out I should have done it as part of the original PR, because in searching for those uses I found four areas of timestamp manipulation code that I had missed.  This PR has three commits:

1. Fix up the untouched timestamp manipulations.
2. krb5_int32 -> krb5_timestamp where needed
3. Clean up some really ancient libkadm5 code that should have been deleted rather than stuck inside ``#if 0``.

### DIFF
--- a/src/lib/krb5/os/c_ustime.c
+++ b/src/lib/krb5/os/c_ustime.c
@@ -29,7 +29,10 @@
 
 k5_mutex_t krb5int_us_time_mutex = K5_MUTEX_PARTIAL_INITIALIZER;
 
-struct time_now { krb5_int32 sec, usec; };
+struct time_now {
+    krb5_timestamp sec;
+    krb5_int32 usec;
+};
 
 #if defined(_WIN32)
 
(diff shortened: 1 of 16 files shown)


src/lib/krb5/os/c_ustime.c
```diff
@@ -73,7 +76,7 @@ get_time_now(struct time_now *n)
 static struct time_now last_time;
 
 krb5_error_code
-krb5_crypto_us_timeofday(krb5_int32 *seconds, krb5_int32 *microseconds)
+krb5_crypto_us_timeofday(krb5_timestamp *seconds, krb5_int32 *microseconds)
```
mrogers950: This one's also declared in k5-int.h with krb5_int32 *.

greghudson: Thanks, I'll change it there as well.

PR #656, opened 2017-05-19
Pass down the local address to the KDB AS request audit functions

cryptomilk:
As we already broke the API it might be useful to know over which IP address the client connection came in.

It also does some code cleanup to have consistent naming (local_addr, remote_addr) and use const when passing the buffer down.

### DIFF
--- a/src/lib/apputils/net-server.c
+++ b/src/lib/apputils/net-server.c
@@ -131,8 +131,8 @@ struct connection {
     struct sockaddr_storage addr_s;
     socklen_t addrlen;
     char addrbuf[56];
-    krb5_fulladdr faddr;
-    krb5_address kaddr;
+    krb5_address remote_addr_buf;
+    krb5_fulladdr remote_addr;
 
     /* Incoming data (TCP) */
     size_t bufsiz;
@@ -951,8 +951,8 @@ struct udp_dispatch_state {
     void *handle;
     const char *prog;
     int port_fd;
-    krb5_address addr;
-    krb5_fulladdr faddr;
+    krb5_address remote_addr_buf;
+    krb5_fulladdr remote_addr;
     socklen_t saddr_len;
     socklen_t daddr_len;
     struct sockaddr_storage saddr;
@@ -1084,10 +1084,12 @@ process_packet(verto_ctx *ctx, verto_ev *ev)
 
     state->request.length = cc;
     state->request.data = state->pktbuf;
-    state->faddr.address = &state->addr;
-    init_addr(&state->faddr, ss2sa(&state->saddr));
+
+    state->remote_addr.address = &state->remote_addr_buf;
+    init_addr(&state->remote_addr, ss2sa(&state->saddr));
+
     /* This address is in net order. */
-    dispatch(state->handle, ss2sa(&state->daddr), &state->faddr,
+    dispatch(state->handle, ss2sa(&state->daddr), &state->remote_addr,
              &state->request, 0, ctx, process_packet_response, state);
 }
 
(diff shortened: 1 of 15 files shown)


src/lib/apputils/net-server.c
```diff
@@ -1329,7 +1336,11 @@ process_tcp_connection_read(verto_ctx *ctx, verto_ev *ev)
     } else {
         /* msglen known. */
         socklen_t local_saddrlen = sizeof(struct sockaddr_storage);
-        struct sockaddr *local_saddrp = NULL;
+        krb5_address local_addr_buf = {0};
+        krb5_fulladdr local_addr = {
+            .address = &local_addr_buf,
```
greghudson: I will probably want to keep with the old pattern here of setting local_addr.address just before the call to init_addr(), rather than using a C99 initializer.

(I think maybe krb5_fulladdr should change so that the first field is krb5_address and not krb5_address *.  But that's out of scope for this PR.)

cryptomilk:
Updated changeset.

src/lib/apputils/net-server.c
```diff
@@ -1329,7 +1336,9 @@ process_tcp_connection_read(verto_ctx *ctx, verto_ev *ev)
     } else {
         /* msglen known. */
         socklen_t local_saddrlen = sizeof(struct sockaddr_storage);
-        struct sockaddr *local_saddrp = NULL;
+        krb5_address local_addr_buf;
+        krb5_fulladdr local_addr;
```
greghudson: I don't think we can use a local variable here.  dispatch() in the KDC will call process_as_req() which will save an alias to the pointer, and possibly return (due to async kdcpreauth modules) before finish_process_as_req() uses the alias.  I think the PR needs to add fields to struct connection like it does to struct udp_dispatch_state.  (struct tcp_dispatch_state would also work, but adding fields there would be inconsistent with how we deal with remote_addr for TCP.)

(I don't like how fragile this design is with regard to object lifetimes, but the same fragility as it relates to remote_addr predates this PR.)

cryptomilk:
I will look into it tomorrow.

src/lib/apputils/net-server.c
cryptomilk: I've added it to the connection struct now. Once this goes in we should probably first remove the pointer for the address in krb5_fulladdr. After that cleanup will be easier and also copying addresses around.

greghudson:
I pushed revised commits.  The notable changes are:

* Condense the changes to four commits and rewrite the commit messages.
* Eliminate the null local address case in schpw.c:dispatch() in the first commit.
* Fix the naming of remote and local address variables in schpw.c.  (The PR retained the name local_saddr after changing the type to krb5_fulladdr, and didn't change the name remote_faddr.)

I'd like @rharwood to have a look at the revised branch in case I introduced a bug or other mistake.

src/lib/apputils/net-server.c
```diff
@@ -1351,10 +1359,14 @@ process_tcp_connection_read(verto_ctx *ctx, verto_ev *ev)
...
-            local_saddrp = ss2sa(&state->local_saddr);
-
-        dispatch(state->conn->handle, local_saddrp, &conn->faddr,
+                        &local_saddrlen) < 0) {
+            krb5_klog_syslog(LOG_ERR, _("getsockname failed: %s"),
+                             error_message(errno));
+            goto kill_tcp_connection;
+        }
+        conn->local_addr.address = &conn->local_addr_buf;
+        init_addr(&conn->local_addr, ss2sa(&state->local_saddr));
```
greghudson: Hm.  I know I said to put local_addr in the connection object and not the state object, but I recently realized that init_addr() stores an alias from the socket address into the resulting krb5_fulladdr value, and it seems dangerous to have an alias from an object in conn to an object in state.  So we should either move local_saddr into conn (possibly initializing it in accept_tcp_connection()) or move local_addr into state.

frozencemetery:
Nothing seems amiss, other than the net-server.c discussion in progress.

PR #658, opened 2017-05-24
Free checksum data deterministically.

tkuthan:
In nomal course of execution, md5.contents allocated by
kg_checksum_channel_bindings() in make_ap_req_v1() is freed in
make_gss_checksum(). But when there is a failure in
krb5_mk_req_extended() or in make_gss_checksum() before free is called,
the memory leaks.

This patch frees the memory unconditionally in make_ap_req_v1().


tkuthan:
Leaked memory from md5.contents was reported from SMB testing.

There was no proof in the core file supplied, that the code flow indeed error-ed out before the call to xfree() in make_ap_req_v1(), but it seems quite probable. Either way, the memory should be freed on error too.

PR #659, opened 2017-05-24
kdc.conf(5): aes-sha2 enctypes are part of the aes family

frozencemetery:
The functional portion of this change was made in d1ec317288278d10ae34fde9b2414e4fca5c52dd.

(As an additional note, looking at the generated kdc.conf.man file, it looks like this may not have been regenerated recently - in addition to a lot of formatting churn, I see a note about the sha2 enctypes being new in 1.15, so I'm not sure it was updated prior to release.  I'd update it with this PR but for the churn.)


greghudson:
Our release procedures update the generated man pages on the release branch, but we don't have a regular procedure for updating them on master, so they tend to lag behind.  I can run a update after merging this PR.

frozencemetery:
Thanks.  In case I've botched the tags, I think this should be backported to the 1.15 release branch as well.

PR #660, opened 2017-05-26
Fix gmt_mktime for y2038

greghudson:
[One of our nightly test machines runs an old enough Solaris not to have timegm(), and its 64-bit variant picked up this omission in the y2038 work.]

gmt_mktime() is used as a fallback when the platform does not have
timegm().  Make it work for dates in the unsigned 32-bit range, not
the signed 32-bit range.


kaduk:
I guess the assert_time() calls were limiting this to the 32-bit range even when time_t is  64 bits?  So, it seems reasonable, then.

PR #662, opened 2017-06-11
Add option to specify salt to ktutil add_entry command

mkazia:
This change adds ability to specify an optional argument salt to the ktutil add_entry command. This helps in creating user keytabs where the salt is not derived from the user principal name. Some uses of this are in creating keytabs for IPA which uses random salting and for Active Directory where enterprise UPNs are used. If the salt argument is not specified then the original behavior of deriving salt from principal is used.

### DIFF
--- a/src/kadmin/ktutil/ktutil.c
+++ b/src/kadmin/ktutil/ktutil.c
@@ -141,6 +141,7 @@ void ktutil_add_entry(argc, argv)
     char *enctype = NULL;
     krb5_kvno kvno = 0;
     int use_pass = 0, use_key = 0, use_kvno = 0, i;
+    char *salt = NULL;
 
     for (i = 1; i < argc; i++) {
         if ((strlen(argv[i]) == 2) && !strncmp(argv[i], "-p", 2)) {
@@ -164,16 +165,22 @@ void ktutil_add_entry(argc, argv)
             use_key++;
             continue;
         }
+        if ((strlen(argv[i]) == 2) && !strncmp(argv[i], "-s", 2)) {
+            salt = argv[++i];
+            continue;
+        }
     }
 
-    if (argc != 8 || !(princ && use_kvno && enctype) ||
+    if (!((argc == 8 && princ && use_kvno && enctype) ||
+          (argc == 10 && princ && use_kvno && enctype && salt)) ||
         use_pass + use_key != 1) {
         fprintf(stderr, _("usage: %s (-key | -password) -p principal "
-                          "-k kvno -e enctype\n"), argv[0]);
+                          "-k kvno -e enctype [-s salt]\n"), argv[0]);
         return;
     }
 
-    retval = ktutil_add(kcontext, &ktlist, princ, kvno, enctype, use_pass);
+    retval = ktutil_add(kcontext, &ktlist, princ, kvno, enctype, use_pass,
+                        salt);
     if (retval)
         com_err(argv[0], retval, _("while adding new entry"));
 }
--- a/src/kadmin/ktutil/ktutil_funcs.c
+++ b/src/kadmin/ktutil/ktutil_funcs.c
@@ -87,13 +87,14 @@ krb5_error_code ktutil_delete(context, list, idx)
  * one first.
  */
 krb5_error_code ktutil_add(context, list, princ_str, kvno,
-                           enctype_str, use_pass)
+                           enctype_str, use_pass, salt_str)
     krb5_context context;
     krb5_kt_list *list;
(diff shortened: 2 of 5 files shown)


kaduk:
This implementation seems to only permit salts that can be represented as C strings (i.e., no embedded NUL), and would make it a bit exciting for a user to specify various control characters in salt strings.
When you say that IPA uses random salting, is it truly random, or just "random UTF-8" (a poorly specified concept) as is the consensus interpretation of the permitted values in RFC 4120 (i.e., of the KerberosString element in ETYPE-INFO2-ENTRY)?

greghudson:
FreeIPA generates random octets, then restricts each byte to the range 0x20-0x7E to ensure that it's printable ASCII.

I would ideally like an option to make an AS-REQ to get the ETYPE-INFO2 for the principal and use the parameters from the first one with an acceptable enctype, but our library interfaces don't currently make that easy.  This is probably an adequate stopgap feature, even if it won't cover every situation.  (Even if we handled any possible salt value as input, there's still s2kparams.)

mkazia:
The idea is to do an AS-REQ with kinit set to trace, gather the salt from the trace log and pass it to add_entry to create the keytab with right salt encryption types such as aes256-cts. If there is a way to make AS-REQ from the ktutil and if you can point me to it, I can incorporate that.

src/kadmin/ktutil/ktutil.c
```diff
@@ -149,6 +150,7 @@ void ktutil_add_entry(argc, argv)
         }
         if ((strlen(argv[i]) == 2) && !strncmp(argv[i], "-k", 2)) {
             kvno = (krb5_kvno) atoi(argv[++i]);
+            use_kvno++;
```
apatil: @mkazia Thanks for attending to this as well. It's useful to be able to use kvno 0.

greghudson: What's the use case for specifying kvno 0?  That's somewhat dangerous in our code base, as we use kvno 0 for "unspecified" in a lot of places, including the encoding of ASN.1 sequences that include kvno values.

Also, sneaking a second feature into a commit isn't great for our record-keeping, so I would probably want to separate this out.

mkazia: I started work on the CentOS 6.8 krb5 SRPM and then backported that code from there to this git tree. That change is include in a patch called krb5-kvno-230379.patch, for which the the top google reference is https://build.opensuse.org/package/view_file/home:garloff:OTC:curl/krb5-mini/krb5-kvno-230379.patch?expand=1. Since that patch and my change had conflict I did not make a good job of leaving that change alone. If that change is not desired I'll submit a new version without that change. Good catch @apatil.

apatil: Oh, I didn't realize this was unintentional. kvno 0 is useful for creating keytabs when you don't know the principal's current kvno and don't have a way to look it up. The version of krb5-workstation that yum installs on Rhel 7 has krb5-kvno-230379.patch, and the ability to specify kvno 0 is convenient. The kvno does not seem to be included in the AS-REP.

frozencemetery: Hi, that patch is originally from here: http://krbdev.mit.edu/rt/Ticket/Display.html?id=3349

greghudson: I opened PR #664 with a modified version of Luke's patch, so please remove any kvno 0 handling from this one.  (I will deal with any merge conflicts.)

mkazia: @greghudson The requested changes have been made. Let me know if there is anything else I can do.

src/kadmin/ktutil/ktutil_funcs.c
```diff
@@ -165,9 +166,13 @@ krb5_error_code ktutil_add(context, list, princ_str, kvno,
                                     &password.length);
         if (retval)
             goto cleanup;
-        retval = krb5_principal2salt(context, princ, &salt);
-        if (retval)
-            goto cleanup;
+        if (salt_str) {
+            salt = string2data(salt_str);
```
greghudson: This leaves salt in the precarious position of containing an alias to salt_str in one case and allocated data in the other.  Right now salt is leaked, which is a spearate bug.  I will adjust the commit to address this problem (and probably fix the memory leak), since our conventions for dealing with this kind of problem aren't easy to explain in text.

greghudson:
@frozencemetary: please re-review the revised changes to ktutil_funcs.c before I merge this.

PR #663, opened 2017-06-12
Fix errors from -Werror=maybe-unitialized with -Og

cipherboy:
When building for debug with:

    CFLAGS="-Og -ggdb" ./configure

gcc throws four errors due to -Og including the -Werror=maybe-uninitialized flag. This PR fixes those errors. Note that the if guard around the fclose() call is necessary due to fclose(NULL) being undefined behavior, potentially causing a [segmentation fault](https://opensource.apple.com/source/Libc/Libc-186/stdio.subproj/fclose.c). 

Thanks,
Alex

### DIFF
--- a/src/util/profile/prof_file.c
+++ b/src/util/profile/prof_file.c
@@ -309,7 +309,7 @@ errcode_t profile_update_file_data_locked(prf_data_t data, char **ret_modspec)
     unsigned long frac;
     time_t now;
 #endif
-    FILE *f;
+    FILE *f = NULL;
     int isdir = 0;
 
 #ifdef HAVE_STAT
@@ -368,7 +368,8 @@ errcode_t profile_update_file_data_locked(prf_data_t data, char **ret_modspec)
         retval = profile_process_directory(data->filespec, &data->root);
     } else {
         retval = profile_parse_file(f, &data->root, ret_modspec);
-        (void)fclose(f);
+        if (f != NULL)
+            (void)fclose(f);
     }
     if (retval) {
         return retval;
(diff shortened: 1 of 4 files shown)


src/util/profile/prof_file.c
```diff
@@ -368,7 +368,8 @@ errcode_t profile_update_file_data_locked(prf_data_t data, char **ret_modspec)
         retval = profile_process_directory(data->filespec, &data->root);
     } else {
         retval = profile_parse_file(f, &data->root, ret_modspec);
-        (void)fclose(f);
+        if (f != NULL)
```
greghudson: This condition is always true.  Is there actually a compiler warning if you omit it, or were you just adding it to be defensive?

frozencemetery: There's actually a warning on it being unitialized.  You're right that the NULL check isn't needed though, since it isn't actually uninitialized.

frozencemetery:
If we decide not to support building in this way that's fine by me; as an additional datapoint, this behavior is still present in gcc as of 7.1.1.

greghudson:
After discussion with Robbie, I'm going to close this without taking the changes, and just recommend against building with gcc -Og.

I have long wanted a gcc option that gives the same warnings/errors for uninitialized variables as gcc -O2, but doesn't apply any optimizations to the code, although of course compilation would be at least as slow as gcc -O2.  Unfortunately gcc -Og is not that; if gives different warnings/errors for uninitialized variables than -O2 does and does optimize the code, although supposedly not in a way that interferes with debugging.  The only benefit over just gcc -g is that the code runs faster.  That can definitely be important for some projects, but a krb5 build with gcc -g is plenty fast enough for development.

PR #668, opened 2017-06-19
Channel bindings

cipherboy:
This adds channel binding support per the [draft](https://tools.ietf.org/html/draft-ietf-kitten-channel-bound-flag-01).


src/lib/gssapi/generic/gssapi_ext.h
```diff
@@ -571,6 +571,31 @@ gss_import_cred(
...
+ */
+
+#define STUB_MAGIC_ID 0x00000ca7
+
+#define GSSINT_CHK_STUB(p) (!GSSINT_CHK_LOOP(p) && \
+    p->initial_ctx_id != NULL && \
+    ((stub_gss_ctx_id_rec *)p->initial_ctx_id)->magic_num == STUB_MAGIC_ID)
+
+
+typedef struct _stub_gss_ctx_id_rec {
```
nicowilliams: This and the above macros should be implementation details and should not be present in this header file.

src/lib/gssapi/krb5/delete_sec_context.c
```diff
@@ -50,6 +50,10 @@ krb5_gss_delete_sec_context(minor_status, context_handle, output_token)
     ctx = (krb5_gss_ctx_id_t) *context_handle;
     context = ctx->k5_context;
 
+    /* Ensure that the context we're trying to free is a krb5 context */
+    if (ctx->magic != KG_CONTEXT)
```
nicowilliams: I don't understand this.  Why shouldn't `gss_delete_sec_context()` delete empty contexts?

src/lib/gssapi/generic/gssapi_ext.h
nicowilliams: Also, I don't think a new structure was needed, just a flag in the existing context structures (union and for each provider).

src/lib/gssapi/mechglue/g_imp_sec_context.c
```diff
@@ -97,7 +97,7 @@ gss_ctx_id_t *		context_handle;
     /* Initial value needed below. */
     status = GSS_S_FAILURE;
 
-    ctx = (gss_union_ctx_id_t) malloc(sizeof(gss_union_ctx_id_desc));
+    ctx = (gss_union_ctx_id_t)calloc(sizeof(gss_union_ctx_id_desc), 1);
```
nicowilliams: IIRC the MIT style guide requires a space after a return value cast?  Or am I confusing it with the Illumos style guide?

Also, the arguments of `calloc()` can be in either order, but the first one is the one called `nmemb`, so the first one ought to be `1`, but this is a nit I don't usually pick and frankly idk why I'm picking it here!  :)

src/lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -706,8 +706,12 @@ kg_accept_krb5(minor_status, context_handle,
             goto fail;
         }
 
-        /* Use ap_options from the request to guess the mutual flag. */
-        gss_flags = GSS_C_REPLAY_FLAG | GSS_C_SEQUENCE_FLAG;
+        /*
+         * Use ap_options from the request to guess the mutual flag.
+         * In the case of a gss_set_context_flags() call, make sure to
```
nicowilliams: I don't think that's quite right.  We want to have a distinct field in the context for "requested flags" and "flags we actually got".

src/lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -783,9 +787,13 @@ kg_accept_krb5(minor_status, context_handle,
         xfree(reqcksum.contents);
         reqcksum.contents = 0;
 
-        /* Read the token flags.  Remember if GSS_C_DELEG_FLAG was set, but
-         * mask it out until we actually read a delegated credential. */
+        /*
+         * Read the token flags.  Remember if GSS_C_DELEG_FLAG was set, but
+         * mask it out until we actually read a delegated credential. If
+         * gss_set_context_flags() was called, include those flags.
```
nicowilliams: Ditto.

src/lib/gssapi/krb5/set_context_flags.c
```diff
@@ -0,0 +1,41 @@
...
+                                               GSS_C_SEQUENCE_FLAG |
+                                               GSS_C_DELEG_FLAG |
+                                               GSS_C_DCE_STYLE |
+                                               GSS_C_IDENTIFY_FLAG |
+                                               GSS_C_EXTENDED_ERROR_FLAG);
+    external_context->gss_flags |= GSS_C_TRANS_FLAG;
+    if (req_flags & GSS_C_DCE_STYLE)
+        external_context->gss_flags |= GSS_C_MUTUAL_FLAG;
+
+    external_context->req_flags = req_flags;
```
nicowilliams: I think this is backwards.  You want to set `external_context->req_flags` and _not_ `external_context->gss_flags` here.  Before this change `external_context->gss_flags` was just `ret_flags`, and in this change it should be renamed to `external_context->ret_flags`, but in no case should this function modify the `ret_flags`.  The `ret_flags` can only be computed at accept-time -- they must not be settable by any setter.

The rule in GSS is that a) all features are optional (except channel binding, which is what we're fixing here!), b) you can get more features than you've asked for (e.g., you may not have asked for replay-detection, but Kerberos providers always give you that even though they don't have to if you didn't ask for it).

BTW, (b) means that ret_flags can be different on the initiator and acceptor side.

src/lib/gssapi/mechglue/g_accept_sec_context.c
```diff
@@ -267,6 +267,16 @@ gss_cred_id_t *		d_cred;
 	    if (status != GSS_S_COMPLETE)
 		return status;
 	}
+
+	if (mech->gss_set_context_flags != NULL) {
```
nicowilliams: I don't understand this.  Why call this at all if the caller didn't?

src/lib/gssapi/krb5/set_context_flags.c
```diff
@@ -0,0 +1,41 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+
+#include "k5-int.h"
+#include "gssapiP_krb5.h"
+
+#include <stdlib.h>
+
+OM_uint32 KRB5_CALLCONV
+krb5_gss_set_context_flags(OM_uint32 *minor_status, gss_ctx_id_t context,
+                           uint64_t req_flags, uint64_t ret_flags)
```
nicowilliams: The I-D says this function gets just three arguments: the minor_status, the context, and `req_flags`.  There should be no way to set `ret_flags`!

src/lib/gssapi/mechglue/g_init_sec_context.c
```diff
@@ -121,6 +121,7 @@ OM_uint32 *		time_rec;
     gss_OID		selected_mech;
     gss_mechanism	mech;
     gss_cred_id_t	input_cred_handle;
+    OM_uint32		used_req_flags;
```
nicowilliams: I don't think you should need this variable at all.

src/lib/gssapi/mechglue/g_set_context_flags.c
```diff
@@ -0,0 +1,53 @@
...
+        return GSS_S_FAILURE;
+
+    ctx = (stub_gss_ctx_id_rec *)union_ctx->initial_ctx_id;
+    if (ctx == NULL)
+        return GSS_S_FAILURE;
+
+    if (ctx->magic_num != STUB_MAGIC_ID)
+        return GSS_S_FAILURE;
+
+    ctx->req_flags = req_flags;
```
nicowilliams: Ahh, right, you do need to track the `req_flags` in the union context because you can't tell which provider's setter to call until init/accept time.

src/lib/gssapi/mechglue/g_set_context_flags.c
```diff
@@ -0,0 +1,53 @@
...
+    ctx->ret_flags = ret_flags;
```
nicowilliams: See commentary elsewhere about there being no need to track `ret_flags` in the union context.

src/lib/gssapi/mechglue/g_set_context_flags.c
```diff
@@ -0,0 +1,53 @@
...
+ * Defined:
+ * https://tools.ietf.org/html/draft-ietf-kitten-channel-bound-flag-01
+ *
+ * Section 2.2
+ *
+ * See src/lib/gssapi/generic/gssapi_ext.h for type definitions.
+ */
+OM_uint32 KRB5_CALLCONV
+gss_set_context_flags(OM_uint32 *minor_status, gss_ctx_id_t context,
+                      uint64_t req_flags, uint64_t ret_flags)
```
nicowilliams: The I-D has only three arguments for this function.  There should be no way to _set_ `ret_flags` directly.

frozencemetery: The draft has a `ret_flags` parameter, and four arguments for this function.

greghudson:
Summary of voice discussion:

1. We don't need the stub structure.  We can just extend gss_union_ctx_id_desc to contain the new fields we need (for the moment req_flags and ret_flags_understood), and let its mech_type and internal_ctx_id fields be nothing for an empty context.  The union structure is internal to the mechglue, so we don't need to worry about what other mechanisms might do with it.

2. The krb5 mech should not assert any new mechanism attributes from the draft at this time, because RFC 4121 does not communicate any channel binding confirmation from the acceptor to the initiator.  In the future we hope to extend RFC 4121 so that the mutual-auth response (1) is encrypted in the initiator subsession key, not the ticket session key, and (2) communicates whether the acceptor checked the channel bindings.  But until we have done that, we can't even assert MAY_CONFIRM, because right now we don't confirm under any circumstances.  (A higher-layer protocol might do its own confirmation, making use of the new channel-bound return flag on the acceptor, but that's not a property of the GSS mechanism.)

3. In the krb5 mech, the only thing we should do with ret_flags_understood at this time is check whether the channel-bound flag is understood, and use that to implement the section 3 semantics in the draft.  In particular, we should not use ret_flags_understood to constrain or augment the flags returned to the application in ret_flags.

4. Although the channel-bound ret_flag semantics in the draft seem solid, we might see changes related to request flags and mechanism attributes.  This is mostly fodder for the kitten list but I'll mention them here as they could affect the PR:

* req_cb_confirmation_flag does not seem very useful at this time, because no mechanisms we know about should assert CBINDING_CONFIRM or in the future.  However, we probably want to reserve a req_flags value for use on the wire in RFC 4121 so that we can use it for the AP-REP extension described above.

* The krb5 mech with hypothetical extended AP-REP will confirm channel bindings if the initiator and acceptor are upgraded, but only if mutual auth is performed.  Does CBINDING_MAY_CONFIRM express this?  Do we need a third attribute, or a relaxation of the semantics of MAY_CONFIRM?  Do we think applications are actually going to do this kind of mechanism filtering?

src/lib/gssapi/krb5/set_context_flags.c
cipherboy: Could you specify in the draft what is supposed to happen with req_flags? Currently there is no language specifying what happens in the case of accept_sec_context(). There are technically four sets of flags: set_context_flags's req_flags, set_context_flags's ret_flags_understood, flags from an init_sec_context() sent on the wire, and the nebulous "actual negotiated flags/features used in this connect" (which likely lean towards init's wire flags versus the set_context_flags's flags). In the case of setting the CB flag for psuedo-mech negotiation, that flag's behavior is understood; however, in all other cases, that behavior is unspecified. (e.g., what happens if MUTUAL_FLAG is specified on accept() side, but not on init() side?).

src/lib/gssapi/krb5/set_context_flags.c
```diff
@@ -0,0 +1,41 @@
...
+{
+    krb5_gss_ctx_id_t external_context;
```
simo5: why "external" ?

src/lib/gssapi/mechglue/g_set_context_flags.c
simo5: @nicowilliams Isn't ret_flags here used to check that the actual flags returned by negotiation match the required returned flags ?
In that case you need a place to store what flags you require to be returned, no ?

greghudson: We do need to track this value in the union context (Nico was consulting an out-of-date version of his draft), but I want to make it clear that the fourth parameter here is "return flags understood".  Not required, or specified, but understood by the application.

To make that clear, I would like the parameter name and variable name to be consistently named "ret_flags_understood".  That's a big longer than we like identifiers to be in most cases, but anything else is likely to sow confusion.

simo5: what about understood_flags or known_flags ?

src/lib/gssapi/mechglue/g_imp_sec_context.c
frozencemetery: No space after cast in MIT.

cipherboy:
Okay, this should be updated with people's suggestions :)

src/lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -782,6 +783,7 @@ kg_accept_krb5(minor_status, context_handle,
                 goto fail;
             }
 
+            have_cb_success = 1;
```
greghudson: You can just set GSS_C_CHANNEL_BOUND_FLAG right here, and get rid of have_cb_success.  It doesn't need to be conditional on mutual auth or ret_flags_understood.  There should be no reason to unset to flag, since it should just never be set in gss_flags other than here.

src/lib/gssapi/krb5/init_sec_context.c
```diff
@@ -567,6 +567,9 @@ kg_new_connection(
     if (req_flags & GSS_C_DCE_STYLE)
         ctx->gss_flags |= GSS_C_MUTUAL_FLAG;
 
+    /* refuse to set channel bound flag by default */
+    ctx->gss_flags = ctx->gss_flags & (~GSS_C_CHANNEL_BOUND_FLAG);
```
greghudson: As in accept_sec_context, there should be no need to unset channel-bound flag; just don't set it.

src/lib/gssapi/krb5/init_sec_context.c
```diff
@@ -875,6 +878,18 @@ mutual_auth(
         *time_rec = ts_delta(ctx->krb_times.endtime, now);
     }
 
+    /* refuse to set channel bound flag by default */
+    ctx->gss_flags = ctx->gss_flags & (~GSS_C_CHANNEL_BOUND_FLAG);
+
+    /*
+     * only set CB success flag if it was specified as a ret flag, and
+     * if channel bindings were specified by the caller.
```
greghudson: We can't ever set the channel-bound flag in the initiator with the current krb5 mech, because there is no channel-bindings confirmation in the mutual auth response.  This commit shouldn't need to make any changes to init_sec_context.

src/lib/gssapi/spnego/gssapiP_spnego.h
```diff
@@ -49,6 +49,11 @@ extern "C" {
 #define	ERR_SPNEGO_NEGOTIATION_FAILED		0x20000004
 #define	ERR_SPNEGO_NO_TOKEN_FROM_ACCEPTOR	0x20000005
 
+
+#define SPNEGOINT_CHK_EMPTY(p) ((p != NULL) \
```
cipherboy: Please correct my understanding if it is wrong :)

Here is my set of assumptions:

1. The mechglue only knows which mech to call when the application calls init/accept_sec_context().
2. The mechglue stores the flags when the application calls create_sec_context() and set_context_flags() due to (1)
3. If the mech determined by the mechglue is SPNEGO, the mechglue doesn't know what mech is under SPNEGO 
4. SPNEGO, once called, has no way to contact the mechglue that called it (e.g., it has no handle to the union context)
5. In its current form, when the mechglue calls SPNEGO->create_sec_context and SPNEGO->set_context_flags(), this is before the call to SPNEGO->init/accept_sec_context()
6. SPNEGO doesn't know the mech under SPNEGO->init/accept_sec_context() is called, due to creds and other features that influence its mech selection process (which might later include mech attributes due to the CB flag). 

The problem I'm trying to solve with this spnego code is that of storing the flags until (6) happens under the order in (5). Unless (4) is invalid, and we can actually access fields on the parent union_ctx, SPNEGO thus needs somewhere to store flags. To me, this was mimicking the mechglue, creating an empty SPNEGO ctx, and setting the flags there.

If you have other ideas, I'd be interested in hearing them. :)

greghudson: Okay, I see the problem.  SPNEGO potentially creates several mechglue contexts as it tries different mechs, and passing through the create_sec_context/set_context_flags call before init/accept_sec_context would only work if we created at most one mechglue context.  You can ignore this comment.

src/lib/gssapi/mechglue/g_init_sec_context.c
```diff
@@ -205,6 +208,19 @@ OM_uint32 *		time_rec;
 	    if (status != GSS_S_COMPLETE)
 		goto end;
 	}
+
+	if (used_req_flags == 0) {
+	    used_req_flags = union_ctx_id->req_flags;
+	    if (mech->gss_set_context_flags != NULL) {
+		status = mech->gss_set_context_flags(
```
cipherboy: Not sure I follow here; in the last commit I added the create_sec_context() call. This is in a [branch](https://github.com/cipherboy/krb5/blob/aced581e1e16121f8eaa005632864922b858a332/src/lib/gssapi/mechglue/g_init_sec_context.c#L196-L224) where `union_ctx_id->internal_ctx_id == NULL`; thereafter create_sec_context() is called and if it falls, we exit. I think we should be safe from internal_ctx_id being null still when this is called. 

However, I do think it is wrong to wrap `set_context_flags()` in `used_req_flags == 0`, as that would not pass through ret_flags_understood, so I'll move the inner `if` statement outside.

src/lib/gssapi/mechglue/g_init_sec_context.c
cipherboy: I think this is necessary:

Consider the case when the mechglue has been updated to include `set_context_flags()`, but the particular mech used hasn't; in this case, an application could conceivably use `set_context_flags()` to set the req_flags it wants, and specifies 0 to `init_sec_context()`, expecting the flags from `set_context_flags()` to be used. However, if the mech doesn't support `set_context_flags()`, it has no way of learning of these flags, unless we detect that the argument `req_flags` to `init_sec_context()` is 0 and substitute in those from the `set_context_flags()` call (Line 213).

src/lib/gssapi/mechglue/g_init_sec_context.c
greghudson: You're right that there is code to create the empty mech context, and can ignore this comment.

src/lib/gssapi/mechglue/g_init_sec_context.c
greghudson: I think Nico might have been saying that you can just reassign the parameter req_flags rather than introducing a new variable, and I agree.

But I think I see a bug here.  If an application calls gss_create_sec_context(), then gss_set_context_flags(), then repeatedly calls gss_init_sec_context() with zero flags, the req_flags from gss_set_context_flags() should be used.  I believe this code would only use those flags in the first call to gss_init_sec_context(), which would be a problem for an old mech.

src/lib/gssapi/krb5/accept_sec_context.c
greghudson: On a closer look, I see two problems:

* The new code doesn't implement bullet point 4 of section 3 of the draft.  If the acceptor caller supports channel bindings (communicated via ret_flags_understood), the acceptor provided channel bindings, and the initiator didn't supply any (all bytes of the wire checksum are 0), then we shouldn't fail out; we should instead just not set the channel-bound flag in ret_flags.

* The existing code doesn't implement bullet point 2 of section 3 of the draft when ``authdat->checksum == NULL`` or when ``authdat->checksum->checksum_type != CKSUMTYPE_KG_CB``.

Also, there is significant technical debt present in the old code: the checksum processing code is open-coded into a function which is much too long even without it.  Because I don't like adding to existing technical debt, I'm going to insist on factoring out the checksum processing code before making it more complicated with the channel-bound flag.  That refactoring should take place in a second commit and shouldn't affect behavior.

src/lib/gssapi/mechglue/g_accept_sec_context.c
```diff
@@ -218,25 +220,40 @@ gss_cred_id_t *		d_cred;
...
 	if (status != GSS_S_COMPLETE) {
-	    free(union_ctx_id);
-	    return (status);
+	    return status;
 	}
 
-	/* set the new context handle to caller's data */
-	*context_handle = (gss_ctx_id_t)union_ctx_id;
+	if (mech->gss_create_sec_context != NULL) {
+	    status = mech->gss_create_sec_context(
```
cipherboy: Do you want that as part of this PR and a separate refactoring commit?

greghudson: I think it could even be part of the first commit, unless you find that the code changes are invasive.  It can certainly be part of this PR.

cipherboy: With respect to simplifying krb5, should I add `create_sec_context`/`set_context_flags`/(channel binding?) support to iakerb as well? At minimum, I'd need to add calls to `krb5_gss_create_sec_context` in the iakerb code in order to simplify any paths in the krb5 mech.

greghudson: I don't think you would need to add channel binding code for IAKERB other than ret_flags_understood bookkeeping; it devolves into the regular krb5 code for the actual AP-REQ/AP-REP exchange.  If it's not too hard to add create_sec_context/set_context_flags for IAKERB I would go ahead and do that.

cipherboy:
Okay. A short summary of changes: 

- Refactoring of `authdat->checksum` checking is done
    - `has_cb_success` left due to this [call](https://github.com/cipherboy/krb5/blob/channel_bindings/src/lib/gssapi/krb5/accept_sec_context.c#L587) which overwrites gss_flags. 
- `set_context_flags` only gets called in mechglue (accept/init) when either `req_flags` or `ret_flags_understood` is nonzero.
    - Same holds for SPNEGO
- `channel_bindings` test cases updated to more closely reflect the draft; behavior fixed to match.
- KRB5 init never returns CB success flag.
- Mixing of flags from set_context_flags no longer happens in init/accept code.
- Rebased from master

Thanks!

greghudson:
This isn't a complete re-review, but I want to comment early on the refactoring of the checksum processing.  The technical debt surrounding the existing code has several sources:

1. At a protocol level, the krb5 gss mech contains a gross hack: it hijacks the checksum field from the RFC 4120 Authenticator type to contain structured data instead of a checksum.  To make matters worse, the structured data is extensible (which tends to push its processing code towards having an open-ended contract), and some implementations violate the spec by including an actual checksum over an empty string, or omitting the checksum field altogether.  The existing code has some comments about the non-conforming implementations, but it doesn't explain the non-intuitive use of the checksum field in the first place.

2. If we see a non-conforming checksum field containing an actual checksum, we verify the checksum, to no useful end: we already know that the creator of the authenticator possesses the key, and as the checksum is over a zero-length string, we're not binding that key to any other value in the protocol exchange.  So we could drop that code and just ignore the checksum.

3. The code intermixes the byte-by-byte decoding of the structured checksum data with decision-making about the contents.  To draw an analogy, imagine how much harder it would be to maintain kg_accept_krb5() if, instead of calling decode_krb5_ap_req() and krb5_rd_req_decoded(), it mixed together code to decode the DER representation of the AP-REQ with the existing code.  That's where we are with the checksum.

I'm not sure whether we want to separate out the decoding of the checksum from its processing at this time.  Doing so would make the code clearer in some ways but also more heavyweight, as we would need to create an in-memory representation of the values that can reside in the checksum, and add to that representation as we implement extensions.

But we definitely need a comment before kg_process_checksum() explaining what it does and why we're reading complex structured data from a "checksum".  Usually a comment before a helper function can be brief, but in this case it's going to have to be specific as the contract is so complicated.

Also, only linker-visible functions in lib/gssapi/krb5 need the "kg_" prefix, although it has historically crept into a number of static helper function names.  We can drop it for this function.

greghudson:
More partial feedback, looking just at the "Chanel bindings changes" commit:

* I think we should separate out the mech attributes and SPNEGO changes into a separate commit.

* We should be able to simplify the section 3 bullet point 2 enforcement by retrospectively checking for missing wire channel bindings after we have processed the checksum, instead of checking for it in all three branches.

* We shouldn't need constant-time comparisons for channel bindings checks.  Every channel bindings hash we receive was generated by a legitimate holder of the ticket session key, so an attacker can't perform an oracle attack.  Using k5_bcmp() instead of memcmp() out of paranoia might be okay, but the open-coded loop isn't necessary.

cipherboy:
@greghudson -- Updated PR with feedback :)

- SPNEGO's MechAttrs + MechAttrs define in a separate commit
- Bullet 2 simplified
- k5_bcmp() instead of open loop.
- Detailed comment for process_checksum

cipherboy:
This PR has been closed in favor of [PR#685](https://github.com/krb5/krb5/pull/685).

PR #670, opened 2017-06-24
Public function to retrieve enctype, salt and s2kparams from KDC

mkazia:
AS_REQ message is sent to KDC and enctype, salt and s2kparams are retrieved from AS_REP/ERROR
message sent back from KDC.
This function will help ktutil use the correct salt and s2kparams associated with the principal when creating keytabs.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -7317,6 +7317,42 @@ krb5_get_init_creds_password(krb5_context context, krb5_creds *creds,
                              const char *in_tkt_service,
                              krb5_get_init_creds_opt *k5_gic_options);
 
+/**
+ * Retrieve enctype, salt and s2kparams from KDC
+ *
+ * @param [in]  context         Library context
+ * @param [in]  principal       Principal whose information is requested
+ * @param [in]  opt             Initial credential options
+ * @param [out] enctype_out     The enctype chosen by KDC
+ * @param [out] salt_out        Salt returned from KDC
+ * @param [out] s2kparams_out   String-to-key parameters returned from KDC
+ *
+ * Send an initial ticket request for @a principal and extract the encryption
+ * type, salt type, and string-to-key parameters from the KDC response.  If the
+ * KDC provides no etype-info, set @a enctype_out to @c ENCTYPE_NULL and set @a
+ * salt_out and @a s2kparams_out to empty.  If the KDC etype-info provides no
+ * salt, compute the default salt and place it in @a salt_out.  If the KDC
+ * etype-info provides no string-to-key parameters, set @a s2kparams_out to
+ * empty.
+ *
+ * @a opt may be used to specify options which affect the initial request, such
+ * as request encryption types or a FAST armor cache (see
+ * krb5_get_init_creds_opt_set_etype_list() and
+ * krb5_get_init_creds_opt_set_fast_ccache_name()).
+ *
+ * Use krb5_free_data_contents() to free @a salt_out and @a s2kparams_out when
+ * they are no longer needed.
+ *
+ * @version New in 1.17
+ *
+ * @retval 0 Success
+ * @return A Kerberos error code
+ */
+krb5_error_code KRB5_CALLCONV
+krb5_get_etype_info(krb5_context context, krb5_principal principal,
+                    krb5_get_init_creds_opt *opt, krb5_enctype *enctype_out,
+                    krb5_data *salt_out, krb5_data *s2kparams_out);
+
 struct _krb5_init_creds_context;
 typedef struct _krb5_init_creds_context *krb5_init_creds_context;
 
@@ -8485,7 +8521,6 @@ krb5_set_kdc_send_hook(krb5_context context, krb5_pre_send_fn send_hook,
 void KRB5_CALLCONV
 krb5_set_kdc_recv_hook(krb5_context context, krb5_post_recv_fn recv_hook,
                        void *data);
-
 
 #if defined(TARGET_OS_MAC) && TARGET_OS_MAC
 #    pragma pack(pop)
--- a/src/lib/krb5/krb/get_etype_info.c
+++ b/src/lib/krb5/krb/get_etype_info.c
@@ -1,0 +1,180 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* lib/krb5/krb/get_etype_salt_s2kp.c - Retrieve enctype, salt and s2kparams */
+/*
+ * Copyright (C) 2017 by Cloudera, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "k5-int.h"
+#include "fast.h"
+#include "init_creds_ctx.h"
+
+/* Extract etype info from the error message pkt into icc, if it is a
+ * PREAUTH_REQUIRED error.  Otherwise return the protocol error code. */
+static krb5_error_code
+get_from_error(krb5_context context, krb5_data *pkt,
+               krb5_init_creds_context icc)
+{
+    krb5_error *error = NULL;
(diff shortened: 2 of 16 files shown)


greghudson:
Thanks for the work on this.  A few notes:

* There's no need to allocate a gic options structure if we don't have any options to set.  We can just pass NULL to krb5_init_creds_init().

* Everything that gets cleaned up in a function's cleanup handler needs to be initialized.  Pointer variables (including context types which are typedefs to pointer types) can simply be assigned NULL in their declarations, while krb5_data variables can be assigned empty_data().  Occasionally it is necessary to stick a memset() statement at the beginning of the function body to zero-fill a structure, but for most types that's not needed.

* I think most of the clang warnings in the failed CI build relate to needing to initialize variables that get cleaned up, but there may be some other warnings that need to be fixed up as well.

* decode_krb5_padata_sequence() can fail, so its result needs to be checked.

* Functions with linker visibility inside libkrb5 should have the prefix k5_ to avoid conflicts.

* If you want to use this within ktutil but keep it a private interface (we could also consider making it public), it needs to be declared in k5-int.h, not int-proto.h.  It also needs to be listed as an exported symbol in libkrb5.exports and krb5_32.def.

I will probably also want to delay merging this until we also have code to use this within ktutil or elsewhere.

greghudson:
Something I miss on the first round of feedback: the interface should accept a krb5_principal and a zero-terminated array of enctypes, not strings to be parsed.

Also, I realized we don't build ktutil on Windows, so there's no need to export the symbol in krb5_32.def.  (If we did need to export it, it would need to be marked as ``; PRIVATE``, but we can just take it back out.)

greghudson:
I just talked to a developer who is interested in using this API to improve their domain-join facility.  Since that facility is programmatic and doesn't go through ktutil, I think we should make this a public API.

As a public API, it should have the prefix ``krb5_``, should be delared in krb5.hin with doxygen comments similar to those of other API functions, should use KRB5_CALLCONV, should be exported from krb5_32.def without any annotation, and should also be listed in doc/appdev/refs/api/index.rst.

mkazia:
Thank you for continuous feedback. I'm glad that there is use for this function outside of what we are trying to accomplish with ktutil. Few questions:

- Regarding the input parameter type change you requested, it currently functions in such a way that if a NULL is passed in the enctype_str, then the permitted enctype that is supported by KDC is chosen and the actual enctype used is returned in "enctype" out parameter. With the change you suggested with zero-terminated array of enctypes, how do you want to use this array? Is it used in-place of the internal enctypes array returned by krb5int_parse_enctype_list?

- For the function to be available in public api, I'm assuming that we have to declare it in include/krb5/krb5.h and no longer needed in include/k5-int.h

- In doc/appdev/refs/api/index.rst would this be classified as "Rarely used public interfaces"?

greghudson:
> With the change you suggested with zero-terminated array of enctypes, how do you want to use this array? Is it used in-place of the internal enctypes array returned by krb5int_parse_enctype_list?

Yes, but I also just noticed that you are using that enctype array by calling 
krb5_set_default_in_tkt_ktypes(), which is a side effect on the context.

Better would be to use krb5_get_init_creds_opt_set_etype_list()... which requires resurrecting that gic_opt option I earlier said you didn't need (because at the time you weren't setting any options).

Another acceptable option would be to remove that parameter entirely, and let the caller use krb5_set_default_in_tkt_ktypes() on the context if it doesn't want to use the default enctype list for the AS-REQ.  That would make the most sense if we don't expect very many callers to use the parameter.  I would think that is the case, but perhaps you have a different idea of how ktutil would use this API.

> For the function to be available in public api, I'm assuming that we have to declare it in include/krb5/krb5.h and no longer needed in include/k5-int.h

Yes.

> In doc/appdev/refs/api/index.rst would this be classified as "Rarely used public interfaces"?

Probably.  I don't agree with those categories and will probably do away with them at some point.

mkazia:
For ktutil, I was planning to use this function in two ways. 

1. When the user has specified the string representation of enctype on add_entry then find the corresponding krb5_enctype, salt and s2kparams.
2. When the user does not specify the enctype on add_entry then find the supported krb5_enctype from KDC  and return it along with salt and s2kparams.

The reason I did not include the krb5_context as the input parameter to the function was because ktutil was currently not using it. I didn't want to introduce initializing krb5_context, setting options, freeing etc. in ktutil code.

If you think we should leave the managing of krb5_context to the caller, then please let me know and I'll make the changes accordingly.

greghudson:
Oh, I missed that too.  As a public libkrb5 API, it definitely needs a context as the first parameter.

mkazia:
I think I have incorporated all the changes requested. Let me know if you need any additional changes.

src/lib/krb5/krb/get_etype_salt_s2kp.c
```diff
@@ -0,0 +1,118 @@
...
+        goto cleanup;
+    retval = krb5_init_creds_step(context, icc, &reply, &request, &realm,
+                                  &flags);
+    if (retval)
+        goto cleanup;
+    if (flags != KRB5_INIT_CREDS_STEP_FLAG_CONTINUE) {
+        retval = KRB5KRB_AP_ERR_MSG_TYPE;
+        goto cleanup;
+    }
+    retval = krb5_sendto_kdc(context, &request, &realm, &reply, &master, 1);
```
greghudson: Setting no_udp to 1 from the get-go certainly makes the code simpler, but will also make this function fail against UDP-only KDCs, which was the default configuration for the MIT krb5 KDC until recently.

src/lib/krb5/krb/get_etype_salt_s2kp.c
```diff
@@ -0,0 +1,118 @@
...
+    if (retval)
+        goto cleanup;
+
+    if (krb5_is_krb_error(&reply)) {
```
greghudson: I would like to see a helper function to extract padata from reply.  Currently the memory management appears to leak in a number of error cases, and I think using a helper function will help.

I would also be okay with keeping the krb5_is_krb_error() conditional in this function, and making two separate helpers for extracting padata from KRB-ERROR and AS-REP.  Whichever approach makes the code come out simpler is fine.

src/lib/krb5/krb/get_etype_salt_s2kp.c
```diff
@@ -0,0 +1,118 @@
...
+            goto cleanup;
+        retval = k5_get_etype_info(context, icc, asrep->padata);
+        if (retval) {
+            krb5_free_kdc_rep(context, asrep);
+            goto cleanup;
+        }
+        krb5_free_kdc_rep(context, asrep);
+    } else
+        return KRB5KRB_AP_ERR_MSG_TYPE;
+    *enctype = icc->etype;
```
greghudson: It's possible for the KDC reply not to include any etype-info or salt information.  In this case, k5_get_etype_info() will still return 0 (it will fall back to get_salt() which will return 0 after not finding PA-PW-SALT or PA-AFS3-SALT elements), but icc->etype won't have been set.  This can happen if the principal has no keys, for instance, and we need to handle it gracefully rather than return the wrong result.

The best way I can think of to detect this case is to set icc->etype to ENCTYPE_NULL before calling k5_get_etype_info(), and then check if it's still ENCTYPE_NULL afterwards.  (krb5_init_creds_init() sets icc->etype to the first request enctype, so we need the assignment.)

mkazia:
Requested changes are done.

greghudson:
I think that's pretty close to right.  If you add the ktutil support, I can review that and write automated test cases for the new feature.

greghudson:
I am pushing a revised commit with a test harness and tests.  Notable changes:

* The API is now named krb5_get_etype_info().
* The API accepts a krb5_get_init_creds_options object instead of an enctype list.  This will make it slightly less convenient for ktutil to use with a specified enctype, but means that arbitrary options can be specified including a FAST armor ccache.
* I added FAST processing of both error and AS-REP replies.
* If no etype-info is sent by the KDC (typically, when the principal requires preauth and has no keys), we return null values.
* If the default salt is indicated, the API computes the default salt.  (Only tested by hand since our KDC currently always includes the salt in etype-info values.)

I will add another commit with the actual ktutil feature (and tests using ktutil) before merging.

mkazia:
Thanks @greghudson for picking this up.

src/include/krb5/krb5.hin
```diff
@@ -7317,6 +7317,42 @@ krb5_get_init_creds_password(krb5_context context, krb5_creds *creds,
...
+ *
+ * @version New in 1.17
+ *
+ * @retval 0 Success
+ * @return A Kerberos error code
+ */
+krb5_error_code KRB5_CALLCONV
+krb5_get_etype_info(krb5_context context, krb5_principal principal,
+                    krb5_get_init_creds_opt *opt, krb5_enctype *enctype,
+                    krb5_data *salt, krb5_data *s2kparams);
```
frozencemetery: These names should match the comment and implementation.

src/lib/krb5/krb/get_etype_info.c
```diff
@@ -0,0 +1,185 @@
...
+                                     NULL);
+    if (ret)
+        return ret;
+    if (error->error != KDC_ERR_PREAUTH_REQUIRED) {
+        ret = ERROR_TABLE_BASE_krb5 + error->error;
+        goto cleanup;
+    }
+    ret = k5_get_etype_info(context, icc, padata);
+    if (ret)
+        goto cleanup;
```
frozencemetery: Could just fall-through here.

src/lib/krb5/krb/get_etype_info.c
```diff
@@ -0,0 +1,185 @@
...
+    ret = decode_krb5_as_rep(pkt, &asrep);
+    if (ret)
+        return ret;
+    ret = krb5int_fast_process_response(context, icc->fast_state,
+                                        asrep, &strengthen_key);
+    if (ret)
+        goto cleanup;
+    ret = k5_get_etype_info(context, icc, asrep->padata);
+    if (ret)
+        goto cleanup;
```
frozencemetery: Also here.

src/lib/krb5/krb/get_etype_info.c
```diff
@@ -0,0 +1,185 @@
...
+        ret = krb5_principal2salt(context, principal, &salt);
+    else if (icc->salt.length > 0)
+        ret = krb5int_copy_data_contents(context, &icc->salt, &salt);
+    if (ret)
+        goto cleanup;
+
+    if (icc->s2kparams.length > 0) {
+        ret = krb5int_copy_data_contents(context, &icc->s2kparams, &s2kparams);
+        if (ret)
+            goto cleanup;
```
frozencemetery: Here too, though I could see wanting to be explicit on this one.

greghudson: We shouldn't fall through here; if this copy fails, we don't want to assign partial results to the other two output variables.  (The other two places should fall through though; I am changing those.)

greghudson:
I have added a second commit with the ktutil addent feature.  I opted for a relatively conservative design, where ktutil only talks to the KDC if explicitly requested via the -f flag.

I wanted to add ktutil tests, but ran into an obstacle: if you just pipe input like "addent -password ...\npassword\nwkt filename" to ktutil, readline's internal buffering eats the password and it doesn't appear to the POSIX prompter.  I thought about adding ss_execute_cmd() support to ktutil like we did for kadmin, but individual ktutil commands are useless.  I guess ktutil -q "cmd1" -q "cmd2" etc. might work but I'm not really fond of it.  I tested manually, of course.

greghudson:
After some IRC conversation I realized that ktutil failing to properly read passwords from pipe input is a fixable bug.  I will fix it in a separate PR and add tests to the second commit.

greghudson:
Okay, tests for ktutil added.  I rebased this on PR #854 temporarily until that PR is merged.

PR #671, opened 2017-06-26
kadm5.acl code refactoring

greghudson:
These commits move and refactor the kadm5.acl code in preparation for adding the kadm5_auth interface.  There is a minor functional difference in the third commit, so I split it out and assigned it a ticket number.

The second commit doesn't leave much of the original code intact, so the changes to auth_acl.c will be difficult to review as a diff.  It's not a lot of code, so reviewing the refactored code as if it were new is probably the best approach.

### DIFF
--- a/src/kadmin/server/auth_acl.c
+++ b/src/kadmin/server/auth_acl.c
@@ -1,8 +1,8 @@
 /* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 /* kadmin/server/auth_acl.c */
 /*
- * Copyright 1995-2004, 2007, 2008 by the Massachusetts Institute of Technology.
- * All Rights Reserved.
+ * Copyright 1995-2004, 2007, 2008, 2017 by the Massachusetts Institute of
+ * Technology.  All Rights Reserved.
  *
  * Export of this software from the United States of America may
  *   require a specific license from the United States Government.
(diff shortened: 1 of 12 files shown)


src/kadmin/server/auth_acl.c
```diff
@@ -72,7 +66,6 @@ struct wildstate {
 
 struct acl_state {
```
frozencemetery: This struct only has a single element - is it worth keeping the type?

src/kadmin/server/auth_acl.c
```diff
@@ -0,0 +1,588 @@
...
+};
+
+struct acl_state {
+    struct acl_entry *list;
+};
+
+static struct acl_state aclstate;
+
+/*
+ * Get a line from the ACL file into buf, with maximum size bufsize.  Lines
```
frozencemetery: There's no parameter bufsize around; I think this is referring to chunksize, which is hardcoded to 128?

src/kadmin/server/auth_acl.c
```diff
@@ -0,0 +1,588 @@
...
+    { 'e', ACL_EXTRACT },
+    { '\0', 0 }
+};
+
+struct wildstate {
+    int nwild;
+    const krb5_data *backref[9];
+};
+
+struct acl_state {
```
frozencemetery: Is this worth keeping around as a type?  It's a struct with only one element, and not used elsewhere.

src/kadmin/server/auth_acl.c
```diff
@@ -0,0 +1,588 @@
...
+    struct acl_entry *list;
+};
+
+static struct acl_state aclstate;
+
+/*
+ * Get a line from the ACL file into buf, with maximum size bufsize.  Lines
+ * ending with \ are continued on the next line.  The caller should set *lineno
+ * to 1 and *incr to 0 before the first call.  On successful return, *lineno
+ * will be the line number of the line read.  Return true on success, false on
```
frozencemetery: Function does not return true/false; instead, it seems to return NULL on failure.

src/kadmin/server/auth_acl.c
greghudson: struct acl_state will become the module data type (and we will stop using any global state in this file) when auth_acl.c becomes a kadm5_auth handle.  We could use struct acl_entry ** as the module data type instead, but that is harder to think about.

src/kadmin/server/auth_acl.c
greghudson: A relic of an earlier rewrite of that function.  I will fix it.  chunksize is not visible to the caller.

src/kadmin/server/auth_acl.c
greghudson: Also a relic of an earlier rewrite.

PR #675, opened 2017-07-04
kadm5_auth pluggable interface

greghudson:
Three commits here: one to add the kadm5_auth interface and consumer code, a second to rework the kadmind calling code for the new design, and a third to add tests.

I hope to get feedback from Samba and FreeIPA developers on whether the interface meets their needs before merging this, or at least before the 1.16 release.


greghudson:
I added an "end" method to the kadm5_auth interface as discussed in the dev call yesterday.

Currently, I'm taking the shortcut of calling the end method in stub_cleanup(), which happens whether or not the stub function invoked an authorization method.  I think that's okay because it's easy for modules to ignore spurious end calls.

PR #676, opened 2017-07-06
Update autoconf archive macros

greghudson:
Update the macros imported from autoconf-archive to current versions.
Fixes an issue where ACX_PTHREAD fails on Solaris with the native
compiler because it detects an uninitialized read and treats it as an
error.


wfiveash:
Thanks for doing this!

PR #677, opened 2017-07-14
Preserve GSS context on init/accept failure

greghudson:
[@frozencemetery: please run interposer tests against this PR]

After gss_init_sec_context() or gss_accept_sec_context() has created a
context, don't delete the mechglue context on failures from subsequent
calls, even if the mechanism deletes the mech-specific context (which
is allowed by RFC 2744 but not preferred).  Check for union contexts
with no mechanism context in each GSS function which accepts a
gss_ctx_id_t.


frozencemetery:
Interposer tests and gssproxy tests passing.  Will try to review in a bit.

PR #678, opened 2017-07-17
Prevent KDC unset status assertion failures

greghudson:
Assign status values if S4U2Self padata fails to decode, if an
S4U2Proxy request uses invalid KDC options, or if an S4U2Proxy request
uses an evidence ticket which does not match the canonicalized request
server principal name.  Reported by Samuel Cabrero.
    
If a status value is not assigned during KDC processing, default to
"UNKNOWN_REASON" rather than failing an assertion.  This change will
prevent future denial of service bugs due to similar mistakes, and
will allow us to omit assigning status values for unlikely errors such
as small memory allocation failures.

[There is a second non-backport commit to remove status assignments for error conditions which don't look practically reachable to me.]

### DIFF
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -240,10 +240,8 @@ finish_process_as_req(struct as_req_state *state, krb5_error_code errcode)
     state->reply.ticket = &state->ticket_reply;
     state->reply_encpart.session = &state->session_key;
     if ((errcode = fetch_last_req_info(state->client,
-                                       &state->reply_encpart.last_req))) {
-        state->status = "FETCH_LAST_REQ";
+                                       &state->reply_encpart.last_req)))
         goto egress;
-    }
     state->reply_encpart.nonce = state->request->nonce;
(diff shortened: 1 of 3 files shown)


src/kdc/do_as_req.c
```diff
@@ -240,10 +240,8 @@ finish_process_as_req(struct as_req_state *state, krb5_error_code errcode)
     state->reply.ticket = &state->ticket_reply;
     state->reply_encpart.session = &state->session_key;
     if ((errcode = fetch_last_req_info(state->client,
-                                       &state->reply_encpart.last_req))) {
-        state->status = "FETCH_LAST_REQ";
+                                       &state->reply_encpart.last_req)))
```
greghudson: That would be out of scope for this PR.  Also, I hope that we'll flesh out fetch_last_req_info() some day, but as it already has a DB entry to work with, it should be an only-fail-on-ENOMEM type operation, so we shouldn't need a status code for it.

frozencemetery: Agreed with respect to status code removal for sure.

PR #680, opened 2017-07-18
Allow WARN_CFLAGS/WARN_CXXFLAGS override for Solaris compiles

wfiveash:
aclocal.m4 is modified to allow WARN_CFLAGS/WARN_CXXFLAGS to be
externally set to override the default values.


greghudson:
I can adjust the commit message to include that information while merging.  This commit just makes the Solaris compiler case respect the same user overrides as the gcc case.

frozencemetery:
Okay, thanks!

wfiveash:
Thanks Greg for that.

PR #682, opened 2017-07-28
Add KDC policy pluggable interface

frozencemetery:
Add the header include/krb5/kdcpolicy_plugin.h, defining a pluggable
interface for a single module to control AS+TGS request behavior.  This
replaces the policy.c functions with handles for pass/fail as well as
ticket lifetime and session key encryption type.

Add against_local_policy_as() and against_local_policy_tgs() as entry
functions.  Move against_local_policy_as() to finish_process_as_req() so
that both can have authentication indicator information available.

Add a test module in plugins/policy/test.  Create t_kdcpol.py with basic
policy tests.  Add plugin interface documentation in
doc/plugindev/policy.rst.

Also-authored-by: Matt Rogers <mrogers@redhat.com>

### DIFF
--- a/src/include/krb5/kdcpolicy_plugin.h
+++ b/src/include/krb5/kdcpolicy_plugin.h
@@ -1,0 +1,128 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* include/krb5/kdcpolicy_plugin.h - KDC policy plugin interface */
+/*
+ * Copyright (C) 2017 by Red Hat, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * Declarations for kdcpolicy plugin module implementors.
+ *
+ * The kdcpolicy pluggable interface currently has only one supported major
+ * version, which is 1.  Major version 1 has a current minor version number of
+ * 1.
+ *
+ * kdcpolicy plugin modules should define a function named
+ * kdcpolicy_<modulename>_initvt, matching the signature:
+ *
+ *   krb5_error_code
+ *   kdcpolicy_modname_initvt(krb5_context context, int maj_ver, int min_ver,
+ *                            krb5_plugin_vtable vtable);
+ *
+ * The initvt function should:
+ *
+ * - Check that the supplied maj_ver number is supported by the module, or
+ *   return KRB5_PLUGIN_VER_NOTSUPP if it is not.
+ *
+ * - Cast the vtable pointer as appropriate for maj_ver:
+ *   maj_ver == 1: Cast to krb5_kdcpolicy_vtable
+ *
+ * - Initialize the methods of the vtable, stopping as appropriate for the
+ *   supplied min_ver.  Optional methods may be left uninitialized.
+ *
+ * Memory for the vtable is allocated by the caller, not by the module.
+ */
+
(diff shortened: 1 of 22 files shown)


greghudson:
What are some use cases for having a policy module decide the session key enctype?  I'm not completely comfortable with the idea.

It's unusual for us to only support one module in a restriction-applying interface.  I can see how having modules decide on the session key enctype makes it more difficult to define natural accumulator behavior for multiple policy modules.

I'd like the two main functions to be optional rather than mandatory, so that a module can implement only one of them.

frozencemetery:
Marked functions as optional.

We thought it would be potentially useful to be able to, e.g., choose a stronger skey enc based on the auth indicator (e.g., AES256 for OTP, or something).  There isn't a use case we have that requires this though, so I can remove it if you want.

As for multiple modules - it's not just the skey_enc that has restriction problems.  What do we do if one module says "yes" and another says "no"?  (We could add a "maybe"/"pass"-style return, but that doesn't solve the problem.)  Or if multiple modules specify different lifetimes?

greghudson:
> We thought it would be potentially useful to be able to, e.g., choose a stronger skey enc based on the auth indicator (e.g., AES256 for OTP, or something).

The KDC picks the strongest enctype (as determined by the client preference list) that we think the server supports.  A policy module could only pick a stronger enctype than the KDC if the client preference list was misordered, or if the policy module knew more about server session key support than is indicated by the string attribute or key set.

We could supply the session key enctype to the policy module and give the policy module a chance to say no, in case the policy is something like "no use of DES, 3DES, or RC4 session keys for admin service principals even if we accidentally added one of those keys to its key set".  Since we'd only expect that to crop up with both a misconfigured client and a misconfigured server principal, we shouldn't need to go out of our way to recover by selecting a different session key enctype.

> As for multiple modules - it's not just the skey_enc that has restriction problems. What do we do if one module says "yes" and another says "no"? (We could add a "maybe"/"pass"-style return, but that doesn't solve the problem.)

This is an easy variant of the kind of problem we address with multiple modules for pwqual, kadm5_auth, localauth, etc..  We don't need a tristate return since we're not looking for an affirmative okay from the policy modules--the KDC has already validated the request, and we're just giving policy modules a chance to say no.  If any module says no, we can stop consulting other modules and reject the request.

> Or if multiple modules specify different lifetimes?

Take the minimum of the lifetimes specified by modules.

greghudson:
The ticket end time should be returned from modules as a krb5_deltat, not a string.  krb5_string_to_deltat() should only be found in UI-adjacent code.

If a module can affect the ticket end time, it should also be able to affect the ticket renew-until time.

frozencemetery:
- Removed enctype control.  I don't think we're worried about misconfiguration here.
- Added support for multiple modules.
- Switched to using a `krb5_deltat` for endtimes.
- Added handle for ticket renew lifetime.

frozencemetery:
Per out-of-band review:
* policy.c now handles ticket time objects instead of having the processing logic in do_{as,tgs}_rew.c.  Moved `against_local_policy_tgs()` to after flag checking as part of this change.
* Removed typedef of `krb5_db_entry` in kdcpolicy_plugin.h to be consistent with #686.

src/kdc/do_as_req.c
```diff
@@ -208,6 +209,15 @@ finish_process_as_req(struct as_req_state *state, krb5_error_code errcode)
 
     state->ticket_reply.enc_part2 = &state->enc_tkt_reply;
 
+    errcode = against_local_policy_as(kdc_context, state->request,
+                                      state->client, state->server,
+                                      state->auth_indicators, state->kdc_time,
+                                      &status, &state->enc_tkt_reply.times);
+    if (errcode) {
+        state->status = status;
```
greghudson: Why the local variable here?  Other calls use &state->status.

frozencemetery:
(Removed extraneous local variable.)

src/include/krb5/kdcpolicy_plugin.h
```diff
@@ -0,0 +1,109 @@
...
+ * - tkt_end_out, if set, specifies the maximum lifetime
+ * - tkt_renew_out, if set, specifies the maximum renewable time
+ */
+typedef krb5_error_code
+(*krb5_kdcpolicy_local_as_fn)(krb5_context context,
+                              krb5_kdcpolicy_moddata moddata,
+                              krb5_kdc_req *request,
+                              struct _krb5_db_entry_new *client,
+                              struct _krb5_db_entry_new *server,
+                              const char **auth_indicators,
```
greghudson: This should be ``const char *const *auth_indicators``, I think.

src/kdc/policy.c
```diff
@@ -1,67 +1,257 @@
...
-        *status = "NO ADDRESS";
-        return KRB5KDC_ERR_POLICY;
+    krb5_error_code ret = 0;
+    char **ais = NULL;
+    size_t i, ai_count;
+
+    *ais_out = NULL;
+
+    if (auth_indicators == NULL || auth_indicators[0] == NULL)
+        goto done;
```
greghudson: I don't see a reason to make modules handle null auth_indicators lists, when we could just construct a zero-element null-terminated list.

src/plugins/policy/test/main.c
```diff
@@ -0,0 +1,112 @@
...
+static krb5_error_code
+test_local_as(krb5_context context, krb5_kdcpolicy_moddata moddata,
+              krb5_kdc_req *request, krb5_db_entry *client,
+              krb5_db_entry *server, const char **auth_indicators,
+              const char **status, krb5_deltat *tkt_end_out,
+              krb5_deltat *tkt_renew_out)
+{
+    if (request->client != NULL &&
+        strncmp(request->client->data->data, "fail",
+                request->client->data->length) == 0) {
```
greghudson: This comparison will succeed if the first client data element is the empty string, "f", "fa", or "fai".  Although that's harmless in test code, we have had CVEs resulting from this incorrect construction in the past.  We have a helper data_eq_string() in k5-int.h to make these kind of comparisons easier.

frozencemetery:
* Adjusted language on krb5_db_entry.  Not totally happy with it, but couldn't come up with a better way to say it.
* Renamed end->lifetime and friends.
* Adjusted type of auth indicators strings.
* Mark plugin name as const
* Improve behavior of empty auth indicator lists
* Extracted helper function for updating times
* Using data_eq_string
* Switched t_kdcpol to use test preauth module for auth indicators

greghudson:
This PR needs to be rebased after the addition of the kadm5_auth interface; once that is done I will work on merging it.

frozencemetery:
(Rebased)

greghudson:
I made some revisions:

* Fixed some cases where the interface name "policy" was used instead of "kdcpolicy", especially in the documentation.

* Documented how to inspect principal entries in a module, and the associated compatibility risk.

* Added kdcpolicy.rst to doc/plugindev/index.rst.

* Used a standard pluggable interface leading comment.  (I will fix certauth_plugin.h to also use a standard comment; this apparently slipped past me.)

* Allowed modules to abort KDC startup, but also to simply decline to initialize using KRB5_PLUGIN_NO_HANDLE.

* Made the form of the handle list match other pluggable interfaces.  The prior code could leave empty or partially initialized handles in the list if modules fail.

* Avoided the use of "local" in identifiers, as kdcpolicy modules do not necessarily enforce local policy.

* Miscellaneous finicky edits.

PR #683, opened 2017-07-31
Display an error message if pkinit_kdc_ocsp is specified

frozencemetery:
pkinit_kdc_ocsp is non-functional in the OpenSSL backend, so remove most
traces of it, including its man page entry.  Detect its presence and error out
instead of silently ignoring the realm entirely.

I'm happy to leave the man page entries around if that would make it more clear.  If we want to add support back into the KDC in the future, it may also make sense to leave the struct entries and such around.

### DIFF
--- a/doc/admin/conf_files/kdc_conf.rst
+++ b/doc/admin/conf_files/kdc_conf.rst
@@ -770,9 +770,6 @@ For information about the syntax of some of these options, see
     pkinit is used to authenticate.  This option may be specified
     multiple times.  (New in release 1.14.)
 
-**pkinit_kdc_ocsp**
-    Specifies the location of the KDC's OCSP.
-
 **pkinit_pool**
     Specifies the location of intermediate certificates which may be
     used by the KDC to complete the trust chain between a client's
--- a/src/man/kdc.conf.man
+++ b/src/man/kdc.conf.man
@@ -891,9 +891,6 @@ Specifies an authentication indicator to include in the ticket if
 pkinit is used to authenticate.  This option may be specified
 multiple times.  (New in release 1.14.)
 .TP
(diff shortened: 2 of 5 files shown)


greghudson:
Removing the documentation and struct entries is correct; half-implemented functionality doesn't belong in the tree.

I think ocsp_check is leaked on error in pkinit_init_kdc_profile().

If I understand correctly, if a KDC sets pkinit_kdc_ocsp, currently it will start up successfully and offer PKINIT preauth.  After this commit it will put an "OCSP is not supported" message in kdc.log, will still start up successfully, but will not serve PKINIT.  I am not sure I like that behavior change; perhaps we can discuss it on a call.

frozencemetery:
Fixed leak, thanks.

Right now what happens is `krb5kdc[3149]: preauth pkinit failed to initialize: No realms configured correctly for pkinit support` (comes from [here](https://github.com/krb5/krb5/blob/master/src/plugins/preauth/pkinit/pkinit_srv.c#L1675), due to an earlier `ENOTSUP` from `pkinit_server_plugin_init_realm()` raised by the ocsp check in `pkinit_identity_prompt()` that I remove).

So I don't think there's a behavior change with respect to pkinit.  Personally I'd be inclined to fail the KDC start, but that doesn't seem to be krb5's style.

greghudson:
Okay, if we're not changing the functional behavior then I have no objection to minimizing the amount of code involved.

Aborting KDC startup on improper kdcpreauth module configuration is a larger topic; while I would be okay with making that happen in some cases, it doesn't feel right to start with this particular edge case and not review configuration errors more generally.

frozencemetery:
Agreed.  I'm very in favor of configuration verification generally, but my memory is that we need some changes to that code before such is realistic.

PR #684, opened 2017-08-08
Log failure details on mutex error assertions

frozencemetery:
Unfortunately compilers will often eat these return codes and so even if I get a coredump, I don't get the return value.  Presenting upstream something I use for debugging downstream.

### DIFF
--- a/src/include/k5-thread.h
+++ b/src/include/k5-thread.h
@@ -134,6 +134,10 @@
      More to be added, perhaps.  */
 
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


greghudson:
I was a bit concerned by the impact of this change on code size, although it's less than 1%.  Can we put the added code inside ``#ifndef NDEBUG`` so that it goes away when the code is built without assertions?

frozencemetery:
Sure, that's fine with me since we build with assertions.  Updated accordingly.

PR #685, opened 2017-08-10
Channel Binding Signalling support

frozencemetery:
(The continuation of #668 for logistical reasons.)  This PR adds support for channel binding as per [the IETF draft](https://tools.ietf.org/html/draft-ietf-kitten-channel-bound-flag-01).  Currently we are waiting on a new version of the draft before considering merging.


greghudson:
Did you make any changes, or are these the same commits?  (I have a bunch of revisions prepared, but am waiting until I merge #677 to push them to the PR branch.)

frozencemetery:

> Did you make any changes, or are these the same commits?  (I have a bunch of revisions prepared, but am waiting until I merge #677 to push them to the PR branch.)

These are the exact same commits (should match, including hashes).

greghudson:
I pushed some revisions.  In the first two commits:

* I rebased after merging PR #677, which coincidentally addresses the concern that we need to check for empty contexts in most of the mechglue GSSAPI entry points.  (After PR #677, a union context with no mech context could be either empty or partially destroyed, but returning GSS_S_NO_CONTEXT is the right answer in either case for most API functions.)

* In the mechs, I simplified the relationship between the main init/accept functions and helper functions to make it clear that the helper functions aren't responsible for creating or deleting contexts.  I'm still considering whether krb5 mech contexts should create the krb5_context field at gss_create_sec_context() time and use a new field to track if the context is empty.

* In the krb5 mech, I moved the new functions to gssapi_krb5.c where we keep other short functions.

* In IAKERB, I switched to using a flag to keep track of whether the underlying krb5 context had been started, rather than look at the krb5 context.  I also fixed some error type confusion in  iakerb_gss_create_sec_context() where a krb5_error_code was returned as a GSS status code, and I fixed a memory leak in an error case in that function.

* In SPNEGO, I don't think clean_spnego_ctx() is needed, so I took it out.  acc_ctx_hints() doesn't do much of interest to a SPNEGO context (though it needs a simple guard to avoid leaking memory if a broken initiator triggers it multiple times), and on hard failure we can still delete the context and null out the handle.

  I also noted a large block of repeated code in the two call_init() functions.  I factored that out into a new prepare_mech_context() helper, and also skipped pre-creating a context if we don't have flags to set on it (so the helper is only introduced in the second commit).

  Most importantly, sc->ctx_handle can remain unset after a call to the SPNEGO gss_accept_sec_context(), either because the optimistic token isn't for the negotiated mech or because the caller requested hints.  So we can't use that field to test if a SPNEGO context is empty.  We can use internal_mech == GSS_C_NO_OID to test if a SPNEGO context is empty or has only been used for hints, so I switched to doing that instead.  (There is a side effect that we will now service multiple successive hint requests, which would previously result in an error.  I think that's harmless.)

* I added the new APIs to lib/gssapi32.def for Windows.

In the third commit, I did some further simplification to process_checksum(), including using the newer and safer k5-input.h in favor of the TREAD_ macros.  I also edited the explanatory comment and moved it to the function definition instead of the call site.

frozencemetery:
Changes are friendly.  Question, though - does gssapi32.def need `GSS_C_MA_CBINDING_CONFIRM` / `GSS_C_MA_CBINDING_MAY_CONFIRM`?  What about `GSS_C_CHANNEL_BOUND_FLAG`?

greghudson:
Yes on the mech attributes, no on the flag (which is just a define).  I pushed an update.

frozencemetery:
Thanks!

greghudson:
I rebased this branch.  If I am following everything correctly, we need to change the "Implement channel mech binding attributes" commit to (1) only include GSS_C_MA_CBINDING_CONFIRM, and (2) in SPNEGO, prefer (but do not filter) GSS_C_MA_CBINDING_CONFIRM-enabled mechs in the initiator's preference list if GSS_C_CB_CONFIRM_FLAG is specified.

We still can't merge this until an oid assignment is made for GSS_C_MA_CBINDING_CONFIRM.

frozencemetery:
@cipherboy has provided updates for -03.

greghudson:
I'm working on some revisions.  There's nothing in the draft about SPNEGO setting GSS_C_MUTUAL_FLAG if the caller asks for GSS_C_CB_CONFIRM_FLAG, so I would like to take that out unless I missed something.  (I get the reasoning, but I don't think that job belongs to the SPNEGO mech.)

PR #687, opened 2017-08-17
Add ccselect module based on server hostname

frozencemetery:
Extend fallback logic to prefer the ccache for which the predicted
realm is the closest subdomain.  This causes, for instance, a service
service1/foo.bar.example.com to prefer ccaches for, in order,
BAR.EXAMPLE.COM, EXAMPLE.COM, and COM.  Previously this behavior was
gated solely by the `realm_try_domains` setting (for DNS-discovered
realms) and the [domain_realm] mappings in krb5.conf.

### DIFF
--- a/src/lib/krb5/ccache/ccselect_hostname.c
+++ b/src/lib/krb5/ccache/ccselect_hostname.c
@@ -1,0 +1,146 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* lib/krb5/ccache/ccselect_hostname.c - hostname ccselect module */
+/*
+ * Copyright (C) 2017 by Red Hat, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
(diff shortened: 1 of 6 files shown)


greghudson:
strcspn() can only be used on a C string, and the data member of a krb5_data object is not a C string.  memchr() should be used instead.

The title of the commit ("Try subdomains of predicted realm") is backwards about the subdomain relationship (a parent domain is not a subdomain), although the commit message and the code comments seem to avoid that error.  The title also needs to be shorter, so that might be something I should take care of when merging anyway.

The larger (and unfortunately squishy) issue is one of semantics.  When we last discussed this by voice, my intent was that if the server realm didn't match any of the ccache client principals, we would go back to the uppercased server hostname (the second component of server if it is of type NT-SRV-HST) and look for matches of parent domains in client principal realms.  I'll call that behavior parents-of-server-hostname.  This commit instead implements parents-of-server-realm.

The behavior is identical if the server realm  is the last fallback (B.C.D where the hostname was a.b.c.d).  The behavior differs if the server principal isn't host-based, or if configuration (``[domain_realm]``, a TXT record, a hostrealm module, etc.) has assigned a realm to the server hostname which isn't the immediate parent domain of the hostname.  I have more thoughts about this than really fit into a github comment, so I will want another round of voice discussion to hash out which behavior is best.

frozencemetery:
Well, the branch is now kind of unfortunately named, but... updated.

- Switched to doing hostname rather than realm matching.
- Moved to a separate plugin so as not to disrupt existing realm matching fallback.
- Stopped inappropriately using string functions.
- Added another commit to remove the `choose()` function.  If this is unnecessary cleanup feel free to not take it (or tell me to not do it).
- deps file not updated for ccselect_hostname.c; my autotools foo is not strong enough to perform this operation.

src/lib/krb5/ccache/ccselect_hostname.c
```diff
@@ -0,0 +1,149 @@
...
+    char *host, *host_cursor, *host_cursor_new;
+    krb5_cccol_cursor col_cursor;
+    krb5_ccache ccache, tmp_ccache, best_ccache = NULL;
+    krb5_principal princ, tmp_princ, best_princ = NULL;
+    unsigned int cursor_len;
+
+    *ccache_out = NULL;
+    *princ_out = NULL;
+
+    if (krb5_is_referral_realm(&server->realm))
```
greghudson: I don't think we need to check this as the hostname module doesn't operate on ``server->realm``, but we do need to check that ``server->type`` is ``KRB5_NT_SRV_HST`` and that ``server->length`` is at least 2.

frozencemetery:
- Fixed check in `hostname_choose()`
- Added behavior documentation to krb5_conf.rst.
- Changed my mind on the krb5_data use... if nothing else it allows me to have fewer things with "cursor" in the name.

PR #690, opened 2017-08-19
Fix kdcpolicy tests for POSIX getopt()

greghudson:
"klist ccachname -e" does not work with a POSIX getopt()
implementation.  Fix t_kdcpolicy.py to use "klist -e ccachename".


wfiveash:
t_kdcpolicy.py does not run correctly on Solaris as a result of the klist argument order.

greghudson:
I folded in PR #692.  After pushing that to master, the Travis build found a another bug in t_kdcpolicy.py, where it fails due to the clock second value rolling over between kinit and kvno, producing this klist output:

    08/21/17 20:50:29  08/21/17 21:50:29  krbtgt/KRBTEST.COM@KRBTEST.COM
    	renew until 08/21/17 22:50:29, Etype (skey, tkt): aes256-cts-hmac-sha1-96, aes256-cts-hmac-sha1-96 
    08/21/17 20:50:30  08/21/17 21:50:29  host/testing-gce-be37eba8-d562-4003-928c-abd66ab02339@KRBTEST.COM
    	renew until 08/21/17 22:50:29, Etype (skey, tkt): aes256-cts-hmac-sha1-96, aes256-cts-hmac-sha1-96 

I'd like to fix that bug in the same commit, but I need to think about how.  I think the tests are not actually exercising the kdcpolicy TGS time constraints as the service ticket end time is constrained by the initial ticket end time and not by the one-hour limit imposed by the module.  To exercise the TGS time constraints, they need to be lower than the AS constraints, perhaps by a factor of two.

greghudson:
Okay, I made the test TGS policy divide lifetimes by 2, so that service tickets in the test script have a predictable and distinct lifetime.

PR #691, opened 2017-08-19
Fix krb5_conf.rst syntax

greghudson:
Commit 92a1a7efe2fc43337416098f2227038a72f1e35a used the wrong header
underline in krb5_conf.rst for the kadm5_auth interface subsection.
Fix it.


greghudson:
I'm going to squash this with PR #688 as it fixes a bug in the same commit.

PR #692, opened 2017-08-21
Use correct trace SKIP macro in kdcpolicy initialization

frozencemetery:


greghudson:
I will squash this with PR #690 as it fixes a bug in the same commit.

PR #693, opened 2017-08-23
Fix AIX build issues

greghudson:
In k5-platform.h, only test for SHARED and define the finalizer as
static if we don't expect linker options to be used for finalizers.
SHARED is not a robust test (it isn't defined when building objects
for shared libraries on AIX, OSF/1, or sometimes IRIX because they
don't use separate PIC objects), and as linker finalizer options are
only applied when shared libraries are created, we don't have to worry
about finalizers happening for static libraries.

In expand_path.c, remove the unnecessary structure tag from "struct
token" as it conflicts with <net/if_arp.h> on AIX.

In localaddr.c, initialize output parameters at the beginning of
get_ifreq_array().  Otherwise, gcc cannot be sure that they are always
set when get_ifreq_array() returns 0, because we use errno as a return
value in one case.  Also remove two unused variables.

Based on patches from Tony Reix.

ticket: 8608 (new)

### DIFF
--- a/src/appl/simple/client/sim_client.c
+++ b/src/appl/simple/client/sim_client.c
@@ -62,7 +62,7 @@ int
 main(int argc, char *argv[])
 {
     int sock, i;
-    unsigned int len;
+    socklen_t len;
     int flags = 0;                      /* flags for sendto() */
     struct servent *serv;
     struct hostent *host;
--- a/src/appl/simple/server/sim_server.c
+++ b/src/appl/simple/server/sim_server.c
@@ -33,6 +33,7 @@
  */
 
 #include "krb5.h"
+#include "port-sockets.h"
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
@@ -64,7 +65,7 @@ int
 main(int argc, char *argv[])
 {
     int sock, i;
-    unsigned int len;
+    socklen_t len;
     int flags = 0;                      /* for recvfrom() */
     int on = 1;
     struct servent *serv;
--- a/src/include/k5-platform.h
+++ b/src/include/k5-platform.h
@@ -361,16 +361,7 @@ typedef struct { int error; unsigned char did_run; } k5_init_t;
 
 
 
-#if !defined(SHARED) && !defined(_WIN32)
-
-/*
- * In this case, we just don't care about finalization.  The code will still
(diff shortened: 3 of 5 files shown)


greghudson:
Tony has pointed out two more build issues which would be nice to fix in the same commit, if possible:

* k5tls fails to link unless -lpthreads is added to the link line.  This is a tough one, and we might need to leave it unfixed.  On platforms with shared library dependencies, we don't want to add -lpthreads since we don't use pthreads directly.  I asked Tony to run ``pkg-config --libs libssl`` and he says it outputs ``-L/opt/freeware/lib -lssl``, which wouldn't work.

* appl/simple uses ``int *`` instead of ``socklen_t *`` for the length argument to getsockname().  Our usual answer to this is ``GETSOCKNAME_ARG3_TYPE``, but appl/simple/server tries not to use our portability headers in order to be a better example.  (appl/simple/client does not, for whatever reason.)  So some thought is required here.

frozencemetery:
Greg Hudson writes:

> * appl/simple uses ``int *`` instead of ``socklen_t *`` for the length
> argument to getsockname().  Our usual answer to this is
> ``GETSOCKNAME_ARG3_TYPE``, but appl/simple/server tries not to use our
> portability headers in order to be a better example.
> (appl/simple/client does not, for whatever reason.)  So some thought
> is required here.

It already has a `#define GETSOCKNAME_ARG3_TYPE`, and includes
"fake-addrinfo.h".  I feel like once it's got that, we might as well
pull in those, we probably should just use the portability layer, since
otherwise we're just duplicating the work.

greghudson:
> It already has a `#define GETSOCKNAME_ARG3_TYPE`, and includes
"fake-addrinfo.h".

I don't see that in appl/simple/server/sim_server.c.

frozencemetery:
Sorry, I should have been a lot more clear with my pronouns.  Rephrasing: I think for consistency we should use the portability we already have elsewhere, rather than the halfway state some of the sample code seems to be in.

greghudson:
I dug around and discovered that we're just using socklen_t more often in recent code, so I used that for sim_client.c and sim_server.c.

I think this is ready to merge now.  I'm just going to ignore the libssl pthread dependency issue since I don't see a good resolution for it.

PR #694, opened 2017-08-24
Make certauth eku module restrictive-only

greghudson:
[I noticed this when looking into feedback on the certauth interface.  The impact isn't as bad as I initially thought, because a PKINIT client cert with the EKU will typically also contain a PKINIT SAN, and the san module will reject the cert if PKINIT SANs are present and don't match.  But it would still have been a CVE bug, so it's fortunate that I ran into this before the release.]

The PKINIT certauth eku module should never authoritatively authorize
a certificate, because an extended key usage does not establish a
relationship between the certificate and any specific user; it only
establishes that the certificate was created for PKINIT client
authentication.  Therefore, pkinit_eku_authorize() should return
KRB5_PLUGIN_NO_HANDLE on success, not 0.


greghudson:
I combined in a fix for the san module, which was explicitly rejecting certs too aggressively, and added a test case using a no-extensions client cert.  (A generic client cert doesn't actually trigger the san module bug because it has no SANs at all, but it is one of the primary use cases for the dbmatch module, so it seemed like a problem that we didn't have a test case for it.)

pedrohc:
CVE-2017-7562 was issued for this bug.

PR #695, opened 2017-08-24
Renewable ticket processing changes

greghudson:
The second commit here is a trivial y2038 fix that I noticed while preparing the first commit.  The impact of the bug would have been miniscule.

The first commit addresses some feedback we got years ago about KDC behavior around issuing renewable tickets, and a recent observation from Weijun that we have a minor bug where we issue inconsistent renewed tickets.  I don't think the bug is important enough for a backport so I combined the fix with the behavior change.

### DIFF
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1790,6 +1790,7 @@ kdc_get_ticket_renewtime(kdc_realm_t *realm, krb5_kdc_req *request,
 {
     krb5_timestamp rtime, max_rlife;
 
+    clear(tkt->flags, TKT_FLG_RENEWABLE);
     tkt->times.renew_till = 0;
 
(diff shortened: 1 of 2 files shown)


greghudson:
I'm going to hold back the second commit and combine it with some other y2038 fixes, since Weijun reported a couple of bugs on krbdev recently.

greghudson:
I might need to add some slop allowance to the checks in t_renew.py.  I just saw:

    *** Failure: maxrenewlife client 1: expected life 14400, got 14401

Since KDC-REQ-BODY contains an absolute end time and not a lifetime, the requested lifetime may not exactly match the resulting ticket lifetime, if the lifetime is not constrained by any KDC policy.  I'm not quite sure why the resulting ticket lifetime is longer (I would expect it to be shorter, if the clock second advanced between the client request and the KDC servicing it), but in hindsight it's not surprising that it's one second different.

PR #698, opened 2017-08-29
Bump bundled libverto for 0.2.7 release

frozencemetery:
[Notably, this contains the fix for #294.]

Local changes:
  - Update upstream URL for fedorahosted deprecation.
  - Add verto_cleanup() to exports.

Upstream changes:
  - Add verto_cleanup() call.
  - Fix memleak when memory allocator is realloc().
  - Fix leaks of filenames during load.
  - Fix many unused variable warnings
  - Factor out mutability check.
  - Properly handle _GNU_SOURCE checking.

### DIFF
--- a/src/util/verto/README
+++ b/src/util/verto/README
@@ -36,5 +36,5 @@ BUILTIN_MODULE define.
 
 The libverto and libev upstream project pages are at:
 
-  https://fedorahosted.org/libverto/
+  https://github.com/latchset/libverto/
   http://software.schmorp.de/pkg/libev.html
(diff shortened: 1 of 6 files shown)


greghudson:
verto.c:605:5: error: ‘for’ loop initial declarations are only allowed in C99 mode

We could potentially change our gcc compiler flags to work around this, but it would be easier if verto didn't use that construct.

frozencemetery:
Hmm, I had assumed c99 would be fine here due to the presence of `va_copy()`, but it looks like that was never getting built by krb5.  (And gcc doesn't warn about it, but that's not a relevant issue here.)

I'll make a new release and get back to you.

PR #699, opened 2017-08-31
Limit ticket lifetime to 2^31-1 seconds

greghudson:
Although timestamps above 2^31-1 are now valid, intervals must not
exceed 2^31-1 seconds.  The initially computed interval in
kdc_get_ticket_endtime() could be negative, causing the function to
yield an incorrect result.  (With the new larger value of
kdc_infinity, this could specifically happen if an KDC-REQ contains a
zero till field.)  Cap the interval at the maximum valid value.
Reported by Weijun Wang.

Also use a y2038-safe timestamp comparison in set_request_times() when
comparing the requested renewable end time to the requested ticket end
time.

### DIFF
--- a/src/clients/ksu/ccache.c
+++ b/src/clients/ksu/ccache.c
@@ -282,7 +282,7 @@ krb5_error_code krb5_check_exp(context, tkt_time)
 
     }
 
-    if (ts_delta(currenttime, tkt_time.endtime) > context->clockskew) {
+    if (ts_after(currenttime, ts_incr(tkt_time.endtime, context->clockskew))) {
         retval = KRB5KRB_AP_ERR_TKT_EXPIRED ;
         return retval;
     }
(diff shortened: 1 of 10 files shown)


greghudson:
The KDC change isn't quite right (it doesn't correctly handle "kinit -l -1s", as the test suite rightly notices); hold off before reviewing this PR.

greghudson:
Okay, this should be ready to review.

frozencemetery:
It looks like a similar change might need to be made to `krb5_check_exp()` (ksu/ccache.c).  There are also a couple of other checks which compare a delta and a clockskew with `<>` elsewhere.

greghudson:
Amended to include more stuff.  Some clockskew checks are now 1 second more tolerant because we were inconsistent before.

STALE() in kdc/replay.c really only needs a unidirectional check (we're not looking to expire lookaside entries from the future), but I wasn't sure about changing the logic that much in this commit.

PR #700, opened 2017-08-31
Allow small errors in t_renew.py tests

greghudson:


frozencemetery:
Do we need this in t_kdcpolicy.py as well?

greghudson:
No, because the lifetime restrictions in t_kdcpolicy.py are all determined by KDC-internal limits, never by the request.  So they should always be exact.

PR #707, opened 2017-10-24
Fix PKINIT cert matching data construction

greghudson:
Rewrite X509_NAME_oneline_ex() and its call sites to use dynamic
allocation and to perform proper error checking.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -5002,33 +5002,29 @@ out:
     return retval;
 }
 
-/*
- * Return a string format of an X509_NAME in buf where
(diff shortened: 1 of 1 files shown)


frozencemetery:
Red Hat has assigned this CVE-2017-15088

greghudson:
On self-review I noticed some minor exception-handling mistakes in the new function definition (memory leaks on error, not checking the return value of BIO_new()).  Please re-review.  I think the candidate fix should be okay for practical purposes.

frozencemetery:
Right, this is better than the original, so +1.

(I believe `BIO_new()` failures aren't realistic for Linux, and the BIO leak on error isn't worth the delay a respin would require, so I agree it's fine for what I'm doing with it, and I'll plan to fix that later.)

PR #708, opened 2017-11-13
Expose context errors in pkinit_server_plugin_init

frozencemetery:
3ff426b9048a8024e5c175256c63cd0ad0572320 attempted to display an error
when OCSP support was requested, but this error message was suppressed
in pkinit_server_plugin_init().  Other error messages from
pkinit_init_kdc_profile(), missing pkinit_identity or pkinit_anchors,
are also now exposted.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_srv.c
+++ b/src/plugins/preauth/pkinit/pkinit_srv.c
@@ -1680,16 +1680,23 @@ pkinit_server_plugin_init(krb5_context context,
 
     for (i = 0, j = 0; i < numrealms; i++) {
         TRACE_PKINIT_SERVER_INIT_REALM(context, realmnames[i]);
-        retval = pkinit_server_plugin_init_realm(context, realmnames[i], &plgctx);
-        if (retval == 0 && plgctx != NULL)
+        krb5_clear_error_message(context);
+        retval = pkinit_server_plugin_init_realm(context, realmnames[i],
+                                                 &plgctx);
+        if (retval)
+            TRACE_PKINIT_SERVER_INIT_FAIL(context, realmnames[i], retval);
+        else
             realm_contexts[j++] = plgctx;
     }
 
     if (j == 0) {
-        retval = EINVAL;
-        krb5_set_error_message(context, retval,
-                               _("No realms configured correctly for pkinit "
-                                 "support"));
+        if (numrealms == 1) {
+            k5_prependmsg(context, retval, "PKINIT initialization failed");
+        } else {
+            retval = EINVAL;
+            k5_setmsg(context, retval,
+                      _("No realms configured correctly for pkinit support"));
+        }
         goto errout;
     }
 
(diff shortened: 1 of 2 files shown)


greghudson:
I have two concerns:

1. "Is there an extended error message" isn't a great proxy for "is there a good error message".  pkinit_server_plugin_init_realm() could return a code whose default message is very descriptive, without setting an extended error.  If there are reachable return points for pkinit_server_plugin_init_realm() which result in bad messages, we should just fix them.  On top of that, context->err.msg being non-null doesn't imply that there is an extended message for the code currently in hand, just for some code some time in the past.

2. The current test doesn't make sense for the multi-realm case.  I don't want us to do any complicated aggregation in that case (since it's overwhelmingly more common to have just one realm), but passing through the last message seems likely to be anti-helpful.  For example, if the first realm was supposed to have PKINIT configured and the second wasn't, we would wind up passing through the less interesting message.

What I would suggest is:

* if numrealms is 1, call k5_prependmsg() (an alias for krb5_prepend_error_message()) with "PKINIT initialization failed".

* otherwise, do what we currently do (don't pass through any per-realm messages).

We could also (or instead) add a trace log for each realm initialization failure.

frozencemetery:
Okay, updated.  I added a trace call as well.  Messages now look like:

```
Nov 13 16:11:47 hostname krb5kdc[7624](Error):** preauth pkinit failed to initialize: PKINIT initialization failed: OCSP is not supported: (realm: EXAMPLE.COM)
```

which is a bit verbose but that's fine.

greghudson:
Use krb5_clear_error_message() when operating on a context, not k5_clear_error().

I don't think that trace macro will work; it appears to use context->err as a string, when in fact it is a structure.  You need to pass in the failure code to the trace macro; there is no way to reliably get that information from the context.  The message also seems a bit off; I would suggest "PKINIT server initialization failed for realm {str}: {str}".

I think the logic at the end of the loop would read better as a if...else than with "continue".

For the krb5_set_error_message() call which is now indented one level further, use k5_setmsg() (an  internal alias added a few years back) to avoid the need for awkward wrapping.

frozencemetery:
- Switched to `krb5_clear_error_message()`.
- Changed logic to use an if/else.
- Switched to `k5_setmsg()`.
- Adjusted trace macro to work now.  However, I'm not sure what you mean about the failure code in the trace macro, since I'm not manipulating that at all?  I've added it and its display as a third part of the macro, in case that's what's intended.

greghudson:
The trace macro definition cannot assume that c->err.msg has anything to do with the error code returned by pkinit_server_plugin_init_realm().  Use {kerr} and retval for the message.

Does pkinit_server_plugin_init_realm() actually ever return 0 with its output set to null?

frozencemetery:
Looks like not.  Removed that check and should be properly using `{kerr}` now:
```[26791] 1511286973.811292: PKINIT server initialization failed for realm EXAMPLE.COM: 95/OCSP is not supported: (realm: EXAMPLE.COM)```

PR #709, opened 2017-11-18
Suggestion for code style fix.

fmmedeiros:
Adding brackets to show explicitly that the else of line 185 belongs to the if statement of line 183, and the first if starting at line 181.


kaduk:
Note that this part of the tree is bundled from the external libverto project and will periodically get re-imported from upstream libverto as we have a need for any updates/fixes from there.
We do make some additional patches on top of upstream, and I agree that your proposal is better style, but it is unclear whether the benefit outweighs the risk of a harder import in the future.

fmmedeiros:
Ok, Thanks!

frozencemetery:
@fmmedeiros Should you (or anyone else) wish to contribute to libverto, I maintain the repository on github at [latchset/libverto](https://github.com/latchset/libverto/).

greghudson:
I believe this particular change was to libev, not libverto.

PR #713, opened 2017-12-06
Add ASN.1 encoders and decoders for SPAKE types

greghudson:
[The single-factor SPAKE code is nearing readiness.  The associated ASN.1 code hasn't needed any changes for quite some time, so can be reviewed now.]

Add a new internal header k5-spake.h.  Add ASN.1 encoder and decoder
functions and an internal free function for SPAKE types.  Add ASN.1
tests and asn1c test vectors the new types.

The additions to to make-vectors.c use C99 designated initializers in
order to initialize unions.  This is okay since make-vectors.c is only
compiled as part of "make test-vectors" and not as part of the regular
build.


src/include/k5-spake.h
```diff
@@ -0,0 +1,107 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* include/k5-spake.h - SPAKE preauth mech declarations */
+/*
+ * Copyright (C) 2015 by the Massachusetts Institute of Technology.
```
frozencemetery: Is this the right year?

greghudson: Looks like yes, based on the commit author date.  We've been at this for a while.

PR #715, opened 2018-01-03
Hardware preauth salt fixes

greghudson:
If a principal has the requires_hwauth bit set, include PA-ETYPE-INFO or PA-ETYPE-INFO2 padata in the PREAUTH_REQUIRED error, as preauth mechs involving hardware tokens may also use the principal's Kerberos password.

When looking up the client long-term key in the securid_sam2 preauth module, look for any salt type, not just the default salt type.

### DIFF
--- a/src/kdc/kdc_preauth.c
+++ b/src/kdc/kdc_preauth.c
@@ -144,7 +144,7 @@ static preauth_system static_preauth_systems[] = {
     {
         "etype-info",
         KRB5_PADATA_ETYPE_INFO,
-        0,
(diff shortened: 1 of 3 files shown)


greghudson:
My eventual plan for that file is to refactor it so that it no longer has a static_preauth_systems table, and we just add informational preauth entries to the list in the code.  I did a refactoring along those lines to lib/krb5/krb/preauth2.c back in commit 5c23bce0e8d3328bb36bc85ee10cfac486b8ae9b.

PR #717, opened 2018-01-11
Fix grouping of GGF extensions in gssapi_ext.h

frozencemetery:
Move gss_import/export_cred to be with the rest of the GGF extensions.
Also adds clarifying comments that gss_set_cred_option and
gssspi_mech_invoke are not part of the GGF extensions.

### DIFF
--- a/src/lib/gssapi/generic/gssapi_ext.h
+++ b/src/lib/gssapi/generic/gssapi_ext.h
@@ -169,12 +169,31 @@ OM_uint32 KRB5_CALLCONV gss_set_sec_context_option
 	 const gss_OID /*desired_object*/,
 	 const gss_buffer_t /*value*/);
 
+/*
+ * Export import cred extensions from GGF, but using Heimdal's signatures
+ */
+OM_uint32 KRB5_CALLCONV gss_export_cred
+	(OM_uint32 * /* minor_status */,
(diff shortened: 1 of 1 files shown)


greghudson:
Based on https://k5wiki.kerberos.org/wiki/Projects/Export_import_cred it looks like our gss_export_cred/gss_import_cred function signatures were based on Heimdal's, not the GGF draft.

frozencemetery:
Adjusted header comment accordingly.

greghudson:
I was going to argue that gss_export/import_cred weren't related to the GGF draft, but when I went back and found the commit log for the Heimdal implementation (commit fa502c6648c5d3836cfc1f4eacbd33c785e380fb, July 2009) I found:

    As documented by IBM (they have the wrong prototype though)
    and GGF (GSS-API Extensions) back in 2001

draft-engert-ggf-gss-extensions was in 2003.  There may have been an earlier version without the extra export/import_cred parameters, though I can't easily find it if so.  The IBM documentation for these functions currently uses the same prototype as Heimdal/MIT unless I fail at "spot the difference", but there's evidence that IBM changed the documentation in 2014, so that might not have been the case in 2009.

Anyway, to keep a long story short, I will accept the PR as-is.

PR #721, opened 2018-01-20
Initialize uninitialized variables (make gcc 7 happy)

nehaljwani:
Without this change, a build with gcc 7 fails with:
```
appdefault.c: In function 'krb5_appdefault_string':
appdefault.c:163:20: error: 'string' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         *ret_value = string;
         ~~~~~~~~~~~^~~~~~~~
cc1: some warnings being treated as errors
make[3]: *** [appdefault.so] Error 1
```

### DIFF
--- a/src/lib/krb5/krb/appdefault.c
+++ b/src/lib/krb5/krb/appdefault.c
@@ -44,6 +44,8 @@ appdefault_get(krb5_context context, const char *appname, const krb5_data *realm
     krb5_error_code retval;
     const char * realmstr =  realm?realm->data:NULL;
 
+    *ret_value = NULL;
+
     if (!context || (context->magic != KV5M_CONTEXT))
         return KV5M_CONTEXT;
 


nehaljwani:
Initialize {curr,end}time variables
    
Without this patch, GCC 7 build fails with:
```    
    sendto_kdc.c: In function 'service_fds':
    sendto_kdc.c:347:56: error: 'curtime' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         interval = (curtime < endtime) ? endtime - curtime : 0;
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
    sendto_kdc.c:342:13: note: 'curtime' was declared here
         time_ms curtime, interval;
                 ^~~~~~~
    sendto_kdc.c:1326:13: error: 'endtime' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         endtime += interval;
                 ^~
```

greghudson:
We got rid of -Wmaybe-uninitialized (or at least intended to) in commit 473536392a00e76642e8770b8163a501d10c6961.  Is this still happening on master, or is this a build of a prior release?

Our usual practice here isn't to initialize the locals, but to make sure that the called function sets its output parameter, e.g. by adding ``*ret_value = NULL;`` at the beginning of appdefault_get().  Does that approach also make the errors go away?

nehaljwani:
@greghudson Yes, before submitting this PR, I saw that that flag has been gotten rid of, but IMO that's just trying to avoid the problem. I didn't see any false positives while fixing the code, so I thought I should send a PR. Yes, this was caught during the build of a release prior to that commit.

In the first commit, the compiler isn't complaining about `ret_value` being uninitialized. It is complaining about `string` being uninitialized. I just did what was already done in another function `krb5_appdefault_boolean()` at https://github.com/krb5/krb5/blob/krb5-1.16-final/src/lib/krb5/krb/appdefault.c#L142

greghudson:
In krb5_appdefault_string(), if appdefault_get() returns 0, it should have set string to something.  So either appdefault_get() isn't correct, or the gcc warning is a false positive.  Setting *ret_value at the top of appdefault_get() conforms to our documented practice ( http://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling ) and ought to convince the compiler that string can't be left uninitialized after the call to appdefault_get() in krb5_appdefault_string().  krb5_appdefault_boolean() does initialize string, but shouldn't need to.

Similar reasoning applies to the callers of get_curtime_ms() in sendto_kdc.c.  In this case, get_curtime_ms() already sets *time_out unconditionally as far back as 1.13, so I would guess this is from a build of 1.12 or earlier.

greghudson:
Ah, looking more closely, I see that 1.14.4 has the ``*time_out = 0;`` line at the top of get_curtime_ms(), but 1.14.2 does not.  I was only looking at the branch heads of past releases.

nehaljwani:
Yes! Setting the return value in appdefault_get() makes gcc 7 happy!

Thank you for the great explanation @greghudson .

I have updated the PR.

PR #722, opened 2018-01-22
fix a few German translations (prepositions)

pseyfert:


tiran:
The plural form ``msgstr "Warnung: Ihr Passwort wird in %d Stunden%s am %s ablaufen."`` will not work correctly. It's Stunde (singlular) and Stunden (plural).

pseyfert:
correct, I still work on that in a separate branch following https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html (except there I actually touch the .c, need to regenerate the .pot, actually build … so i was hesitant to put out the PR as quickly - and actually got distracted with other things)

pseyfert:
see https://github.com/pseyfert/krb5/commit/64077750cccf3983ef1369b4443c3f6b112b8cd2 for where I'm going with the plural treatment

greghudson:
Thanks, I'll merge this now, and you can submit the preposition changes as a separate PR when you're ready.  Thanks to @tiran for reviewing this PR.

PR #723, opened 2018-01-26
PKINIT certid fixes

greghudson:
This is http://krbdev.mit.edu/rt/Ticket/Display.html?id=8636 converted to a PR, without any changes to the code diffs yet.

I want to poke around for less verbose approaches to converting from hex to binary, or possibly add something to libkrb5support for hex conversion.  Hex conversion is simple enough that just repeating it everywhere isn't totally unreasonable, but once you add in unit tests that stops being the case.

(It would be possible to use the OpenSSL bignum functions correctly, but by the time we added the necessary length adjustments it would be about as much code as converting from hex ourselves.  And semantically, converting a hex string to a bignum to a binary string isn't very elegant.)

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -4636,6 +4636,43 @@ reassemble_pkcs11_name(pkinit_identity_opts *idopts)
     return ret;
 }
 
+static int
+hex_string_to_bin(const char *str, int *bin_len_out, CK_BYTE **bin_out)
+{
+    size_t str_len, i;
+    CK_BYTE *bin;
+    char *endptr, tmp[3] = { '\0', '\0', '\0' };
+    long val;
+
+    *bin_len_out = 0;
+    *bin_out = NULL;
+
+    str_len = strlen(str);
+    if (str_len % 2 != 0)
+        return EINVAL;
+    bin = malloc(str_len / 2);
+    if (bin == NULL)
+        return ENOMEM;
+
+    errno = 0;
+    for (i = 0; i < str_len / 2; i++) {
+        tmp[0] = str[i * 2];
+        tmp[1] = str[i * 2 + 1];
+
+        val = strtol(tmp, &endptr, 16);
+        if (val < 0 || val > 255 || errno != 0 || endptr != &tmp[2]) {
+            free(bin);
+            return EINVAL;
+        }
+
+        bin[i] = (CK_BYTE)val;
+    }
+
+    *bin_len_out = str_len / 2;
+    *bin_out = bin;
+    return 0;
+}
+
 static krb5_error_code
 pkinit_get_certs_pkcs11(krb5_context context,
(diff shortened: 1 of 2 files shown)


src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -4623,6 +4623,53 @@ reassemble_pkcs11_name(pkinit_identity_opts *idopts)
     return ret;
 }
 
+int hex_string_to_bin(const char *str, int *bin_len_out, CK_BYTE **bin_out)
```
frozencemetery: Should be static to pass clang.

greghudson:
I didn't find a way to decode hex with OpenSSL without also converting to a bignum.  Not really surprising as OpenSSL isn't a data format library.

I found more instances of hex encoding and decoding in the krb5 code base than I expected.  Many were in test programs, but there's also hex stuff in kdb5_util, ktutil, libkdb_ldap, and kproplog.  (Also a few cases where we convert hex to integers.)

My plan is to:

* Introduce hex functions into libkrb5support in a separate PR.

* Remove the tests from Sumit's patch (which lets us make hex_string_to_bin static) and accept it as a for-backport commit.

* In a follow-on commit (which I will add to this PR), replace hex_string_to_bin with a call to the libkrb5support function.

greghudson:
Updated as planned.  I also made a few changes to hex_string_to_bin() for brevity, even though it's not sticking around on the master branch.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -4636,6 +4636,42 @@ reassemble_pkcs11_name(pkinit_identity_opts *idopts)
...
 
+static int
+hex_string_to_bin(const char *str, int *bin_len_out, CK_BYTE **bin_out)
+{
+    size_t str_len, i;
+    CK_BYTE *bin;
+    char *endptr, tmp[3] = { '\0', '\0', '\0' };
+    long val;
+
+    *bin_out = NULL;
```
frozencemetery: Shouldn't we either also initialize `bin_len_out` here or initialize neither?

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -4636,6 +4636,42 @@ reassemble_pkcs11_name(pkinit_identity_opts *idopts)
...
+
+    str_len = strlen(str);
+    if (str_len % 2 != 0)
+        return EINVAL;
+    bin = malloc(str_len / 2);
+    if (bin == NULL)
+        return ENOMEM;
+
+    errno = 0;
+    for (i = 0; i < str_len / 2; i += 2) {
```
frozencemetery: I think the loop update should be `i++` here?

greghudson: Whoops, that would have been bad.  To mitigate the risk of having introduced other bugs, I temporarily brought back Sumit's tests after fixing this (and the bin_len_out initialization) and made sure they pass.

PR #724, opened 2018-01-29
Process included directories in alphabetical order

frozencemetery:
readdir() and FindFirstFile()/FindNextFile() do not define any
ordering on the entries they return.  Use sorted scandir() instead on
Unix-likes.

Please note that I have only implemented this for not-WIN32.  Windows doesn't seem to have an equivalent of `scandir()`.  I assume the way to implement this would be to gather (and filter) the filenames into a list and then sort the list before processing, but I don't have a Windows machine to test this.  Apologies for the inconvenience.

### DIFF
--- a/src/util/profile/prof_parse.c
+++ b/src/util/profile/prof_parse.c
@@ -241,12 +241,18 @@ static int valid_name(const char *filename)
     }
     return 1;
 }
+#ifndef _WIN32
+static int valid_name_scandir(const struct dirent *d)
+{
+    return valid_name(d->d_name);
+}
+#endif
 
 /*
  * Include files within dirname.  Only files with names ending in ".conf", or
  * consisting entirely of alphanumeric characters, dashes, and underscores are
  * included.  This restriction avoids including editor backup files, .rpmsave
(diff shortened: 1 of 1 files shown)


greghudson:
scandir() is also a relatively recent addition to POSIX (I think it was added in 2008), and before that it has had some historical changes to the types of the select and comparison function parameters (void * vs. const void * vs. const struct dirent **).  Every build platform I have immediate access to (including our Solaris buildbot worker) appears to have a POSIX 2008-conformant implementation of it, but there is a i mild risk that we might run into at least a type warning on some Unix-like platform out there.

For Windows, Heimdal has a readdir implementation in libroken which we could adapt, and we could borrow a scandir() implementation from one of the BSDs, so that the profile code can just use the POSIX interfaces and not have ifdefs.  That will take a substantial amount of plumbing work, so I'm not sure when I'll get to it.

frozencemetery:
Noted.  I'll go ahead and carry this patch rather than waiting for merge.

frozencemetery:
Closing in favor of #784.

PR #726, opened 2018-02-10
Make public headers work with gcc -Wundef

greghudson:
The C standard allows undefined symbols in #if statements, giving them
the value 0 (C99 section 6.10.1).  However, some software builds with
gcc -Wundef, which issues a warning on undefined symbols in
preprocessor expressions.  Make the public headers safe for that
environment by using #ifdef instead of #if for TARGET_OS_MAC.
Reported by Ben Kaduk.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -106,7 +106,7 @@
 
 KRB5INT_BEGIN_DECLS
 
-#if TARGET_OS_MAC
+#if defined(TARGET_OS_MAC) && TARGET_OS_MAC
 #    pragma pack(push,2)
 #endif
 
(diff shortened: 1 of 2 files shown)


tlyu:
The other `TARGET_OS_*` macros in TargetConditionals.h in my Xcode installation can be defined to 0, so `#if defined(TARGET_OS_MAC) && TARGET_OS_MAC` (or equivalent using an intermediate macro that we define) is probably safer, even though nothing currently in that file does `#define TARGET_OS_MAC 0`.

On the other hand maybe we can take this change as-is and fix it if someone complains in the future.

greghudson:
I was assuming an autoconf-like discipline for TARGET_OS_MAC being defined, but I guess not.  I'll use the safer form.

frozencemetery:
Do we support building anywhere `TARGET_OS_MAC` is 0?

greghudson:
I don't think we're aware of any environments where TARGET_OS_MAC is defined to zero, but if Apple is using an "always define to 1 or 0" approach to these symbols (as opposed to autoconf's "undefine or define to 1"), they could conceivably introduce such an environment in the future.

PR #727, opened 2018-02-14
KDC padata and preauth refactoring

greghudson:
This work is preliminary to PKINIT freshness.  It should not introduce any behavior changes.  The main goals are to eliminate the static preauth systems table so that we can explicitly decide when to add informational padata, and to incrementally allocate the hint and AS-REP padata lists instead of preallocating them.  (Preallocation is of course slightly more efficient, but it's risky unless we have a robust way of predetermining the maximum list size, and without the static preauth systems table we would be changing that each time we added a new piece of informational padata.)

### DIFF
--- a/src/kdc/kdc_preauth.c
+++ b/src/kdc/kdc_preauth.c
@@ -101,107 +101,13 @@ typedef struct preauth_system_st {
     krb5_kdcpreauth_loop_fn loop;
 } preauth_system;
 
+static preauth_system *preauth_systems;
+static size_t n_preauth_systems;
+
 static krb5_error_code
 make_etype_info(krb5_context context, krb5_preauthtype pa_type,
                 krb5_principal client, krb5_key_data *client_key,
                 krb5_enctype enctype, krb5_pa_data **pa_out);
-
-static void
-get_etype_info(krb5_context context, krb5_kdc_req *request,
-               krb5_kdcpreauth_callbacks cb, krb5_kdcpreauth_rock rock,
-               krb5_kdcpreauth_moddata moddata, krb5_preauthtype pa_type,
-               krb5_kdcpreauth_edata_respond_fn respond, void *arg);
-
-static krb5_error_code
-return_etype_info(krb5_context, krb5_pa_data *padata,
-                  krb5_data *req_pkt, krb5_kdc_req *request,
-                  krb5_kdc_rep *reply, krb5_keyblock *encrypting_key,
-                  krb5_pa_data **send_pa, krb5_kdcpreauth_callbacks cb,
-                  krb5_kdcpreauth_rock rock, krb5_kdcpreauth_moddata moddata,
-                  krb5_kdcpreauth_modreq modreq);
-
-static krb5_error_code
-return_pw_salt(krb5_context, krb5_pa_data *padata,
-               krb5_data *req_pkt, krb5_kdc_req *request, krb5_kdc_rep *reply,
-               krb5_keyblock *encrypting_key, krb5_pa_data **send_pa,
-               krb5_kdcpreauth_callbacks cb, krb5_kdcpreauth_rock rock,
-               krb5_kdcpreauth_moddata moddata, krb5_kdcpreauth_modreq modreq);
-
-
-
-static preauth_system static_preauth_systems[] = {
-    {
-        "FAST",
-        KRB5_PADATA_FX_FAST,
-        PA_HARDWARE,
-        NULL,
-        NULL,
-        NULL,
-        NULL,
-        NULL,
-        0
-    },
-    {
-        "etype-info",
-        KRB5_PADATA_ETYPE_INFO,
-        PA_HARDWARE,
-        NULL,
-        NULL,
-        NULL,
-        get_etype_info,
-        0,
-        return_etype_info
-    },
-    {
-        "etype-info2",
-        KRB5_PADATA_ETYPE_INFO2,
-        PA_HARDWARE,
-        NULL,
-        NULL,
-        NULL,
-        get_etype_info,
-        0,
-        return_etype_info
-    },
-    {
-        "pw-salt",
-        KRB5_PADATA_PW_SALT,
-        PA_PSEUDO,              /* Don't include this in the error list */
-        NULL,
-        NULL,
-        NULL,
-        0,
-        0,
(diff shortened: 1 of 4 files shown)


src/kdc/kdc_preauth.c
```diff
@@ -799,7 +790,7 @@ hint_list_finish(struct hint_state *state, krb5_error_code code)
     kdc_realm_t *kdc_active_realm = state->realm;
 
     if (!code) {
-        if (state->pa_data[0] == 0) {
+        if (state->pa_data == NULL) {
```
frozencemetery: This is checking a different thing.  That's intended?

src/kdc/kdc_preauth.c
```diff
@@ -820,20 +811,26 @@ hint_list_next(struct hint_state *arg);
...
-            pa->magic = KV5M_PA_DATA;
-            pa->pa_type = state->pa_type;
+            ret = alloc_pa_data(state->pa_type, 0, &pa);
+            if (ret)
+                goto error;
         }
-        *state->pa_cur++ = pa;
+        ret = add_pa_data_element(&state->pa_data, pa);
+        if (ret)
+            goto error;
```
frozencemetery: I think this leaks `pa` (if `code ==0` but `pa == NULL`).

src/kdc/kdc_preauth.c
```diff
@@ -888,17 +885,24 @@ get_preauth_hint_list(krb5_kdc_req *request, krb5_kdcpreauth_rock rock,
...
+    state->pa_data = NULL;
+    state->ap = preauth_systems;
 
-    state->pa_data = calloc(n_preauth_systems + 1, sizeof(krb5_pa_data *));
-    if (!state->pa_data) {
-        free(state);
-        (*respond)(arg);
-        return;
-    }
+    if (alloc_pa_data(KRB5_PADATA_FX_FAST, 0, &pa) != 0)
```
frozencemetery: Nit: these should be using an intermediate variable (e.g., `ret`).

src/kdc/kdc_preauth.c
```diff
@@ -888,17 +885,24 @@ get_preauth_hint_list(krb5_kdc_req *request, krb5_kdcpreauth_rock rock,
...
+        goto error;
+    if (add_pa_data_element(&state->pa_data, pa) != 0)
+        goto error;
+    if (add_etype_info(kdc_context, rock, &state->pa_data) != 0)
+        goto error;
```
frozencemetery: If `add_etype_info` fails, don't we end up freeing an already-added pa_data element?  These should probably happen in the other order.

src/kdc/kdc_preauth.c
```diff
@@ -1300,26 +1299,18 @@ return_padata(krb5_context context, krb5_kdcpreauth_rock rock,
...
 
-    if ((send_pa_list = malloc((size+1) * sizeof(krb5_pa_data *))) == NULL)
-        return ENOMEM;
-    if ((pa_order = malloc((size+1) * sizeof(int))) == NULL) {
-        free(send_pa_list);
-        return ENOMEM;
-    }
+    pa_order = k5calloc(size + 1, sizeof(int), &retval);
+    if (pa_order == NULL)
+        goto cleanup;
```
frozencemetery: `original_key` hasn't been initialized here.

src/kdc/kdc_preauth.c
```diff
@@ -1300,26 +1299,18 @@ return_padata(krb5_context context, krb5_kdcpreauth_rock rock,
...
 
     retval = krb5_copy_keyblock_contents(context, encrypting_key,
                                          &original_key);
-    if (retval) {
-        free(send_pa_list);
-        free(pa_order);
-        return retval;
-    }
+    if (retval)
+        goto cleanup;
```
frozencemetery: Here either.

src/kdc/kdc_preauth.c
```diff
@@ -1349,29 +1340,37 @@ return_padata(krb5_context context, krb5_kdcpreauth_rock rock,
                 }
             }
         }
+        send_pa = NULL;
```
frozencemetery: This can happen in the declaration block; it's not used before here.

src/kdc/kdc_preauth.c
greghudson: Yes; with new helper functions an empty pa_data list is represented as a null pointer, while the old code would have preallocated an array and just not filled any entries yet.

src/kdc/kdc_preauth.c
greghudson: Since add_pa_data_element() claims pa, I had it do so on success and failure for simplicity.  (Having it copy so that it never claims pa would be more reasonable for a library API, but within the KDC I figure the looser, more efficient memory management is acceptable as long as it's consistent between success and failure.)

src/kdc/kdc_preauth.c
greghudson: If we were going to do anything with the result besides test it for success, I would agree, but get_preauth_hint_list() does nothing on failure besides stop (this dates back to it being a function returning void in 1.0 and is certainly questionable).

src/kdc/kdc_preauth.c
greghudson: Only insofar as we free the whole state->pa_data list.  I'm not sure what order change is being proposed.

src/kdc/kdc_preauth.c
greghudson: Good catch.  I added a memset at the start, and checked that the other cleaned up variables are initialized.

src/kdc/kdc_preauth.c
greghudson: But we're inside a loop, and we need to make sure it's null before the call on each iteration.

frozencemetery: Whoops, yes, you're right.

src/kdc/kdc_preauth.c
greghudson: I decided that perhaps designing a list accumulator to claim caller memory should carry the cost of a comment at each call site.  Ordinarily I don't like documenting function behavior at every call site, but unusual memory management may be an exception.  So I've added comments like ``/* add_pa_data_element() claims pa on success or failure. */`` to each call.

PR #728, opened 2018-02-20
KerbSubmitTicket() should check malloc() return for NULL

Sashan:
Patch has been sent earlier from my alexandr.nedvedicky company email.
more details on [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012902.html)

### DIFF
--- a/src/lib/krb5/ccache/cc_mslsa.c
+++ b/src/lib/krb5/ccache/cc_mslsa.c
@@ -385,6 +385,8 @@ CacheInfoEx2ToMITCred(KERB_TICKET_CACHE_INFO_EX2 *info,
      * not a NULL list of addresses.
      */
     creds->addresses = (krb5_address **)malloc(sizeof(krb5_address *));
+    if (creds->addresses == NULL)
+        return FALSE;
     memset(creds->addresses, 0, sizeof(krb5_address *));
 
     return TRUE;
@@ -739,13 +741,14 @@ KerbSubmitTicket( HANDLE LogonHandle, ULONG  PackageId,
 {
     NTSTATUS Status = 0;
     NTSTATUS SubStatus = 0;
-    KERB_SUBMIT_TKT_REQUEST * pSubmitRequest;
+    KERB_SUBMIT_TKT_REQUEST * pSubmitRequest = NULL;
     DWORD dwRequestLen;
-    krb5_auth_context auth_context;
+    krb5_auth_context auth_context = NULL;
     krb5_keyblock * keyblock = 0;
     krb5_replay_data replaydata;
     krb5_data * krb_cred = 0;
     krb5_error_code rc;
+    BOOL rv = FALSE;
 
     if (krb5_auth_con_init(context, &auth_context)) {
         return FALSE;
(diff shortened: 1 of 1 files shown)


frozencemetery:
Thanks for your patches.  Normally we're not so nit-picky, but you mentioned you were trying to learn the github workflow, so I'm being extra-fussy.

- It would be better if they were grouped into logical units, rather than all being their own PRs.  I think either one or two PRs make sense in total.
- Clean git history.  There should be no merge commits in the submission.  (`git rebase` and `git push --force` will help with this.)
- Please conform to the coding style.  In particular, most of the time this will mean using spaces (four) rather than tabs for indentation.
- We prefer `goto cleanup` and the like for early exit from functions rather than copy-pasting the same cleanup code many times.
- Commit messages should mention why the change was made in addition to what the change was.

Again, thanks for your patches!

Sashan:
> Clean git history. There should be no merge commits in the submission. (git rebase and git push --force will help with this.)

OK. I've just read the `git help rebase` will make sure to use it next time

>Please conform to the coding style. In particular, most of the time this will mean using spaces (four) rather than tabs for indentation.

Is there some c-style check I can use? looks like my vim settings expand 8spaces to tab. will be more careful next time.

thanks for all your hints hope I'll have a chance to follow them in future pull requests.

greghudson:
http://k5wiki.kerberos.org/wiki/Coding_style/Style_checker documents our style checker script.

Whitespace issues aside, this patch looks correct.  However, I see two other unchecked malloc() calls in cc_mslsa.c at lines 387 and 801, and I would prefer to fix that issue across the whole file in one commit.

Sashan:
I've fixed white space issue. also the updated version introduces a `cleanup:` branch to `KerbSubmitTicket()` ticket. Unfortunately I have no windows machine around to try out new code.

src/lib/krb5/ccache/cc_mslsa.c
```diff
@@ -787,18 +794,14 @@ KerbSubmitTicket( HANDLE LogonHandle, ULONG  PackageId,
...
-    if (rc) {
-        krb5_auth_con_free(context, auth_context);
-        if (keyblock)
-            krb5_free_keyblock(context, keyblock);
-        if (krb_cred)
-            krb5_free_data(context, krb_cred);
-        return FALSE;
-    }
+    if (rc)
+        goto cleanup;
```
frozencemetery: `pSubmitRequest` needs to be initialized `NULL` otherwise this (and the next `goto cleanup`) cause an unitialized `free()`.

src/lib/krb5/ccache/cc_mslsa.c
```diff
@@ -832,15 +833,18 @@ KerbSubmitTicket( HANDLE LogonHandle, ULONG  PackageId,
...
                                            &SubStatus
     );
+
+    rv = (!FAILED(Status) && !FAILED(SubStatus));
+
+cleanup:
     free(pSubmitRequest);
     if (keyblock)
         krb5_free_keyblock(context, keyblock);
+    if (krb_cred)
```
greghudson: Let's de-conditionalize the krb5_free_keyblock() and krb5_free_data() calls while we're introducing a cleanup label.  (Both functions can handle null pointers.)

Sashan: revision `ef0f174` fixes nits pointed out by Robbie and Greg

PR #729, opened 2018-02-20
KDC should release resources when it is about to bail out on error

Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012896.html) few days ago from my Alexandr.Nedvedicky company mail

### DIFF
--- a/src/kdc/main.c
+++ b/src/kdc/main.c
@@ -983,6 +983,8 @@ int main(int argc, char **argv)
     retval = load_kdcpolicy_plugins(kcontext);
     if (retval) {
         kdc_err(kcontext, retval, _("while loading KDC policy plugin"));
+	unload_authdata_plugins(kcontext);
+	load_preauth_plugins(&shandle, kcontext, ctx);
         finish_realms();
         return 1;
     }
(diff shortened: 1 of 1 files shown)


greghudson:
I'm inclined not to take this patch.  We generally don't try to free everything when exiting from a process with an error.  (There's an argument for using ``exit(1);`` instead of ``return 1;`` if we're going to drop some memory onto the floor, but I don't know if that trivial change is worth the commit.)  We also try to avoid repeating cleanup code in multiple error cases.  I think the existing code calls finish_realms() in each error case because we want to make sure to call the database module finalizer before exiting in case there is some database cleanup required beyond freeing memory or closing sockets.

At a detail level, load_preauth_plugins() seems like the wrong function to clean up preauth module resources, and the changes don't clean up everything (no call to krb5_klog_close() or krb5_free_context(), for instance).

Sashan:
> I'm inclined not to take this patch. We generally don't try to free everything when exiting from a process with an error.

I'm fine with this decision. So I'll drop this patch to floor in my tree too. Thanks for looking at it.

PR #730, opened 2018-02-20
pkinit should free realm identity certificates

Sashan:
KDC does not seem to attempt to release identity credentials on exit at all.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -949,8 +949,31 @@ pkinit_init_certs(pkinit_identity_crypto_context ctx)
 }
 
 static void
+pkinit_free_cred(pkinit_cred_info creds)
+{
+    if (creds != NULL) {
+	if (creds->cert != NULL) {
+	    X509_free(creds->cert);
+	    creds->cert = NULL;	
+	}
+
+	if (creds->key != NULL) {
+	    EVP_PKEY_free(creds->key);
+	    creds->key = NULL;
+	}
+
+	free(creds->name);
+	creds->name = NULL;
+
+	free(creds);
+    }
+}
+
+static void
 pkinit_fini_certs(pkinit_identity_crypto_context ctx)
 {
+    unsigned int i;
+
     if (ctx == NULL)
         return;
 
@@ -968,6 +991,11 @@ pkinit_fini_certs(pkinit_identity_crypto_context ctx)
 
     if (ctx->revoked != NULL)
         sk_X509_CRL_pop_free(ctx->revoked, X509_CRL_free);
+
+    for (i = 0; i < MAX_CREDS_ALLOWED; i++) {
+	pkinit_free_cred(ctx->creds[i]);
+	ctx->creds[i] = NULL;
+    }
 }
 
 static krb5_error_code


Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012895.html) from my Alexandr.Nedvedicky company email

greghudson:
I have concerns about this patch on three levels, each of which might render the following ones moot:

1. I wouldn't expect the KDC to populate id_crypto_context->creds; I think that field should only be used for clients.  If you're detecting a KDC memory leak here, I would like to figure out what KDC code path is allocating the memory in ->creds.  I'm not sure if your tooling can reveal that information, but can you point to where in the test suite the memory leak was detected by libumem?

2. There is an existing free_cred_info() function, and crypto_free_cert_info() to free all the creds in a pkinit_identity_crypto_context object, so the new functions appear to duplicate code.

3. There is some unnecessary nulling in the new code.  If we're in the process of freeing an object, there's no need to null out the fields as we free them.  (crypto_free_cert_info() nulls creds after freeing them because it does not destroy the enclosing pkinit_identity_crypto_context object.)

greghudson:
Okay, I guess the KDC certificate gets loaded into id_crypto_context->creds.

So I'm guessing we just need to make pkinit_server_plugin_fini_realm() call crypto_free_cert_info() on the identity crypto context.  (There's a small matter of the req_cryptoctx parameter to crypto_free_cert_info()... which is unused, along with the plg_cryptoctx parameter.  We could take those parameters out.)

greghudson:
After the analysis of PR #735, I think the fix for this problem might be less obvious than freeing up the creds array on KDC exit.

Unlike pkinit_identity_initialize(), pkinit_identity_prompt() still frees the creds array internally after loading certs for the second time.  So it is surprising to see memory from pkinit_identity_prompt -> crypto_load_certs surviving to the end of the KDC process.  One possible explanation is that crypto_cert_select_default() can appropriate some pointers from the first element of the creds array, storing it into ``id_cryptoctx->my_certs`` and ``id_cryptoctx->my_key``.  But those pointers are supposed to be freed in pkinit_fini_certs() as called by pkinit_fini_identity_crypto(), so I'm not sure.  I will do some more investigation.

greghudson:
Hm, I ran a PKINIT-configured KDC under valgrind and was able to find the leak for PR #735 (the creds[0] structure allocated during pkinit_identity_initialize() and later overwritten when certs are loaded for the second time) and the leak for PR #736 (client_pubkey), but I can't see any memory leaking which was allocated during pkinit_identity_prompt().

Sashan:
> can't see any memory leaking which was allocated during pkinit_identity_prompt(). [ETA: same holds if I run the whole of t_pkinit.py through valgrind.]

That's correct there is no memory leak coming from `pkinit_identity_prompt()`. The `pkinit_identity_prompt()` just overwrites pointer to `creds[cindex]` without releasing them first.

I'm sorry my comment and pasted callstack was confusing. The callstack just show the culprit, which overwrites the pointer to buffer effectively causing the leak. Sorry I was more clear about it earlier.

greghudson:
Okay, thanks.  Then I think all we need for this issue is for pkinit_identity_initialize() to take responsibility for freeing the creds array, as described in my comments on PR #735.

(That doesn't leave behind a totally satisfactory design, but I think throwing in extra free-the-creds calls isn't really an improvement.   A better design direction would be to separate out the creds collection into a separate type from pkinit_identity_crypto_context, and manage its lifetime with local variables in pkinit_identity_initialize() and pkinit_identity_prompt().  Not really an urgently needed change in my opinion.)

Sashan:
> Okay, thanks. Then I think all we need for this issue is for pkinit_identity_initialize() to take responsibility for freeing the creds array, as described in my comments on PR #735.

I'm just testing this change, I'll update the pull requests as soon as I'll have results.

greghudson:
Okay, looking at the diff:

* The static prototype for free_cred_info() seems to be a remnant of a previous revision and can go.

* The change to pkinit_fini_identity_crypto() should be unnecessary.

* The call to crypto_free_cert_info() in pkinit_client_prep_questions() should be removed, as the cred array lifetime is now managed within pkinit_identity functions.

Sashan:
thanks for pointing out all those nits. the patch should be clean now.

PR #731, opened 2018-02-20
always initialize keyblock to zero

Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012901.html) from my Alexandr.Nedvedicky company email.

### DIFF
--- a/src/kadmin/dbutil/dump.c
+++ b/src/kadmin/dbutil/dump.c
@@ -104,6 +104,7 @@ master_key_convert(krb5_context context, krb5_db_entry *db_entry)
     krb5_kvno kvno;
     int i, j;
 
+    memset(&v5plainkey, 0, sizeof(krb5_keyblock));
     is_mkey = krb5_principal_compare(context, master_princ, db_entry->princ);
 
     if (is_mkey) {
--- a/src/kadmin/dbutil/kdb5_create.c
+++ b/src/kadmin/dbutil/kdb5_create.c
@@ -373,6 +373,7 @@ tgt_keysalt_iterate(ksent, ptr)
     krb5_int32          ind;
     krb5_data   pwd;
 
+    memset(&key, 0, sizeof(krb5_keyblock));
     iargs = (struct iterate_args *) ptr;
     kret = 0;
 
--- a/src/kadmin/ktutil/ktutil_funcs.c
+++ b/src/kadmin/ktutil/ktutil_funcs.c
@@ -111,6 +111,7 @@ krb5_error_code ktutil_add(context, list, princ_str, kvno,
     int i, tmp;
     unsigned int pwsize = BUFSIZ;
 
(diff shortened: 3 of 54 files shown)


greghudson:
In general, the circumstances where we need to memset() a structure variable to zero are the same as the circumstances where we need to initialize a pointer to NULL--most commonly, when the variable is cleaned up in the cleanup label.  We don't have a requirement to always initialize variables of any particular type independent of how they are used.

I looked over the first ten or so of the changes in this patch, and they looked unnecessary, or in a few cases redundant with existing memsets (e.g. kdc/cammac.c).

frozencemetery:
Echoing that, does applying this patch fix some problem for you?  If so, what?

Sashan:
> Echoing that, does applying this patch fix some problem for you? If so, what?

the patch as such does not fix any bug. I've implemented the change as I was suspecting similar type of leak as we've seen in #735. Unfortunately in this case the assert was firing because of uninitialized memory. I used simple grep to find all occurrences of keyblock and initialized them to 0.

perhaps I should invest more time and address only those locations, which trigger the assert, which I put to `krb5_dbe_def_decrypt_key_data()`:
```
diff --git a/src/lib/kdb/decrypt_key.c b/src/lib/kdb/decrypt_key.c
index 541064d88..e064072ac 100644
--- a/src/lib/kdb/decrypt_key.c
+++ b/src/lib/kdb/decrypt_key.c
@@ -108,6 +108,7 @@ krb5_dbe_def_decrypt_key_data( krb5_context     context,
         dbkey->magic = KV5M_KEYBLOCK;
         dbkey->enctype = key_data->key_data_type[0];
         dbkey->length = tmplen;
+       assert(dbkey->contents == NULL);
         dbkey->contents = (krb5_octet *) plain.data;
     }
```

greghudson:
krb5_dbe_def_decrypt_key_data() is the default implementation of krb5_dbe_decrypt_key_data().  dbkey and keysalt are output structure parameters, with keysalt being optional.  Per current practices, the implementation (1) should not assume or require anything about the contents of the structure pointed to by the caller's pointer before the call, and (2) make the contents of the structure well-defined on success or failure.  Currently, (1) is satifisfied but (2) is not, so it would be a positive change to make krb5_dbe_decrypt_key_data() (the entry point) zero out the output structures at the beginning.  We should not impose any new requirement that the caller initialize the output keyblock before calling krb5_dbe_decrypt_key_data().

Sadly, the inverse function, krb5_dbe_encrypt_key_data(), does not appear to respect (1) for its key_data output parameter; the default implementation frees and nulls the key_data_contents pointers in the output structure at the start.  There is some risk in changing that, although I wouldn't really expect any callers to depend on the freeing behavior.

Sashan:
I see. So we can close the request as such. I can keep the patch aside in my repo.

PR #733, opened 2018-02-20
process_tgs_req() must release encrypting_key

Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012903.html) few days ago from Alexandr.Nedvedicky company email

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -144,6 +144,7 @@ process_tgs_req(struct server_handle *handle, krb5_data *pkt,
     memset(&reply_encpart, 0, sizeof(reply_encpart));
     memset(&ticket_reply, 0, sizeof(ticket_reply));
     memset(&enc_tkt_reply, 0, sizeof(enc_tkt_reply));
+    memset(&encrypting_key, 0, sizeof(encrypting_key));
     session_key.contents = NULL;
 
     retval = decode_krb5_tgs_req(pkt, &request);
@@ -719,8 +720,6 @@ process_tgs_req(struct server_handle *handle, krb5_data *pkt,
 
     errcode = krb5_encrypt_tkt_part(kdc_context, &encrypting_key,
                                     &ticket_reply);
-    if (!isflagset(request->kdc_options, KDC_OPT_ENC_TKT_IN_SKEY))
-        krb5_free_keyblock_contents(kdc_context, &encrypting_key);
     if (errcode)
         goto cleanup;
     ticket_reply.enc_part.kvno = ticket_kvno;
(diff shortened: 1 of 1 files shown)


Sashan:
Travis (clang) found a bug in my patch. the fixed patch looks as follows
```
diff --git a/src/kdc/do_tgs_req.c b/src/kdc/do_tgs_req.c
index cc5a69236..a5a842a67 100644
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -114,6 +114,7 @@ process_tgs_req(struct server_handle *handle, krb5_data *pkt,
     int newtransited = 0;
     krb5_error_code retval = 0;
     krb5_keyblock encrypting_key;
+    krb5_keyblock *cleanup_encrypting_key = NULL;
     krb5_timestamp kdc_time, authtime = 0;
     krb5_keyblock session_key;
     krb5_keyblock *reply_key = NULL;
@@ -877,6 +878,10 @@ cleanup:
         krb5_free_pa_data(kdc_context, reply_encpart.enc_padata);
     if (enc_tkt_reply.authorization_data != NULL)
         krb5_free_authdata(kdc_context, enc_tkt_reply.authorization_data);
+    if (cleanup_encrypting_key != NULL) {
+       free(cleanup_encrypting_key->contents);
+       cleanup_encrypting_key->contents = NULL;
+    }
     krb5_free_pa_data(kdc_context, e_data);
     k5_free_data_ptr_list(auth_indicators);
 
```

greghudson:
encrypting_key is currently freed at line 723, conditional on ``!isflagset(request->kdc_options, KDC_OPT_ENC_TKT_IN_SKEY)``.  It could leak if a ``goto cleanup;`` statement executes between when it is allocated and when it is freed, which I assume is what you detected.  This could happen on various transited failures, on a handle_authdata() failure, or a U2U second ticket mismatch.  I'm guessing the test suite triggered a handle_authdata() failure and libumem detected the leak.  It looks like this bug was introduced in release 1.7.

The patch currently in this PR doesn't seem like it would have any effect.  It introduces cleanup_encrypting_key and cleans it up, but otherwise doesn't use it.  (Also, there is no need to null out pointers freed in a cleanup label.)

Like the PKINIT leak, I think this bug calls for a conservative backport commit and a separate change to improve code maintainability.  The conservative fix is to move the conditionalized free into the cleanup label and to initialize encrypting_key with memset() so that it's always safe to clean up.  That should fix the bug, but leaves encrypting_key in the poor state of sometimes containing an alias pointer and sometimes containing a storage pointer.  To fix that, we will want one ``krb5_keyblock`` variable which is always storage, and one ``krb5_keyblock *`` variable which is always an alias.

greghudson:
I pushed two revised commits to do what I described in my last comment.  Robbie, please have a look if you have time.

PR #734, opened 2018-02-20
finish realm should free keys with 0 length too

Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012900.html) few days ago from my Alexandr.Nedvedicky company email

### DIFF
--- a/src/kdc/main.c
+++ b/src/kdc/main.c
@@ -161,11 +161,7 @@ finish_realm(kdc_realm_t *rdp)
     if (rdp->realm_context) {
         if (rdp->realm_mprinc)
             krb5_free_principal(rdp->realm_context, rdp->realm_mprinc);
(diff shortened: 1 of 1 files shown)


greghudson:
It looks like this trivial bug happens when the test suite uses the test KDB module, which returns a zeroed master key when asked for one.  Looking at the code which frees the realm master key, I'd like to see that whole conditional block to free realm_mkey replaced with:

    zapfree(rdp->realm_mkey.contents, rdp->realm_mkey.length);

Sashan:
OK. the most recent version uses `zaprfree()`

greghudson:
zapfree() checks the pointer for null, so we can just call it without the conditional.

Sashan:
> `zapfree()` checks the pointer for null, so we can just call it without the conditional.

thanks, I've just killed the `if ()`

PR #735, opened 2018-02-20
pkinit_load_fs_cert_and_key() must not blindly overwrite pointer

Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012897.html) few days ago from my Alexandr.Nedvedicky company email.

Not sure if change is correct fix. It seems to me it rather fixes symptoms.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -4446,6 +4446,11 @@ pkinit_load_fs_cert_and_key(krb5_context context,
         goto cleanup;
     }
 
+    if (id_cryptoctx->creds[cindex] != NULL) {
+	pkinit_free_cred(id_cryptoctx->creds[cindex]);
+	id_cryptoctx->creds[cindex] = NULL;
+    }
+
     id_cryptoctx->creds[cindex] = malloc(sizeof(struct _pkinit_cred_info));
     if (id_cryptoctx->creds[cindex] == NULL) {
         retval = ENOMEM;
@@ -4495,8 +4500,9 @@ pkinit_get_certs_fs(krb5_context context,
     }
 
     retval = pkinit_load_fs_cert_and_key(context, id_cryptoctx,
-                                         idopts->cert_filename,
-                                         idopts->key_filename, 0);
+					 idopts->cert_filename,
+					 idopts->key_filename, 0);
+
 cleanup:
     return retval;
 }


Sashan:
Travis fails, because the change in this pull request depends on pull [request](https://github.com/krb5/krb5/pull/730), which adds pkinit_free_cred() function. Perhaps,
I should merge those two pull requests together.

greghudson:
This trivial bug was introduced in release 1.12.  Before then, there was only pkinit_identity_initialize(), which would internally call:

1. crypto_load_certs() to populate the creds array, potentially prompting (on the client side only) for passwords which protect the private keys
2. pkinit_cert_matching() or crypto_cert_select_default() to choose a cert from the creds array
3. crypto_free_cert_info() to free the creds array

We had to complicate this design to support the clpreauth responder (introduced in release 1.11, but not supported by PKINIT until 1.12).  To support the responder, a clpreauth module needs to assemble a list of questions in the prep_questions() method; then in the process() method, it must look for answers from the responder, and only issue prompter callbacks if the answers weren't given.

To make this work, we added a ``defer_id_prompts`` flag to crypto_load_certs().  If set, crypto_load_certs() will register questions when it needs a password; if not set it will try to use passwords from the responder callback.  pkinit_identity_initialize() now stops after loading certs with prompts deferred, and the new function pkinit_identity_prompt() loads the certs again without deferred prompts.

For reasons unclear to me, pkinit_identity_initialize() no longer frees the creds array after loading certs.  One might hypothesize that the state is needed to correlate responder answers with certs, but that clearly not the case, because the client-side caller of pkinit_identity_initialize() immediately frees the creds array after pkinit_identity_initialize() returns.  So there's no reason that couldn't have been handled internally.  pkinit_identity_prompt() continues to be responsible for freeing the creds array after loading certs.

On the KDC side, there can be no prompting for passwords, nor is there choosing between certificates (the creds array only holds one cert, which is then chosen by crypto_cert_select_default()).  The new KDC code just calls pkinit_identity_initialize() and pkinit_identity_prompt() in sequence, and loading the certs twice is just a small amount of extra overhead that we tolerate for the sake of the client.  But unlike on the client, we don't free the creds array after pkinit_identity_initialize(), so we wind up overwriting the first element creds array the second time we load certificates, leaking memory.

I think the fix is to make pkinit_identity_initialize() once again responsible for freeing the creds array it loads, by calling crypto_free_cert_info() after crypto_load_certs().  The call to crypto_free_cert_info() in pkinit_client_prep_questions() should be removed.  I believe that change should render this PR and PR #730 unnecessary.

Sashan:
I think we can close this pull request, as the fix here is kind of dead end. the final version of fix got merged to #730

PR #736, opened 2018-02-20
client_pubkey always leaks in server_process_dh()

Sashan:
Same patch has been sent to [krbdev](http://mailman.mit.edu/pipermail/krbdev/2018-February/012898.html) few days ago from my Alexandr.Nedvedicky company address.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -2979,6 +2979,7 @@ server_process_dh(krb5_context context,
 
     retval = 0;
 
+    BN_free(client_pubkey);
     if (dh_server != NULL)
         DH_free(dh_server);
     return retval;


greghudson:
I think I introduced this bug in commit e5c77a11341a79e6af1e5aef7c587a5b75a9e378 (OpenSSL 1.1 support, first appearing in release 1.15).  Prior to that commit, the client public key gets stored in a DH handle referenced from cryptoctx, and is later freed when the DH handle is freed in pkinit_fini_req_crypto().

(I can see several opportunities for simplification here.  The server code shouldn't need to use two DH handles, and server_check_dh() and server_process_dh() could be combined, with the result saved in pkinit_req_crypto_context instead of the DH handle.  But that's well beyond the scope of a memory leak fix.)

greghudson:
Looking further, it seems that I actually accounted for the changed handling of client_pubkey in the cleanup code, but didn't notice that the cleanup label isn't reached if the function returns successfully.

Per current practice, the cleanup label should always run, and output parameters should be handled as shown in http://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling .  That will require a little bit more surgery to this function than I'm comfortable with for a backport commit.

For the backport commit, we should free client_pubkey where we free dh_server in the successful path, right at the end.  In general we try to clean up stuff near the end of functions, rather than freeing it as soon as it is no longer needed.

(I can make all of the necessary changes here; just making notes.)

Sashan:
I've updated the branch and changed commit comment.
The BN_free() moved to DH_free(). I've also killed the if () test, since same
check is implemented in DH_free(). Hope it is OK.

greghudson:
The conditional for DH_free() is indeed unnecessary, not just because DH_free() can handle null inputs, but because dh_server will never be null on the success path--we needed it to do the function's work, after all.  But backport commits are generally as conservative as possible, so I would be inclined to leave it alone.

PR #737, opened 2018-02-22
client_pubkey always leaks in server_process_dh()

Sashan:
This is a fix for memory leak of client_pubkey() in server_process_dh(), the fix
follows the coding-style [guidelines](http://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling)

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -2911,10 +2911,10 @@ server_process_dh(krb5_context context,
                   pkinit_identity_crypto_context id_cryptoctx,
                   unsigned char *data,
                   unsigned int data_len,
-                  unsigned char **dh_pubkey,
-                  unsigned int *dh_pubkey_len,
-                  unsigned char **server_key,
-                  unsigned int *server_key_len)
+                  unsigned char **dh_pubkey_out,
+                  unsigned int *dh_pubkey_len_out,
+                  unsigned char **server_key_out,
+                  unsigned int *server_key_len_out)
 {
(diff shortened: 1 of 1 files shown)


greghudson:
This looks like the right general approach.  A few notes:

* Every pointer which is cleaned up in a cleanup label needs to be initialized to NULL when it is declared.  In this case, dh_pubkey and server_key.

* There's a prototype in pkinit_crypto.h which should probably be updated to make the parameter names match.  (There are also these unsightly ``/* IN */`` and ``/* OUT */`` comments in that prototype which are rendered unnecessary by the ``_out`` parameter names, but every other prototype in that file has the same thing, so I guess they should stay for now.)

* It would probably make sense to hit client_process_dh() and client_create_dh() at the same time.  (server_check_dh() seems closer to okay as-is since it doesn't have any output parameters.)  Again, I'm happy to do that work before merging.

From a github perspective, I would have combined this with PR #736 as a second commit, but that does require some familiarity with how to manipulate commit history in git.  For now it's fine to keep it separate.

It also looks like the dejagnu tests are failing, so there's probably a bug introduced somewhere in here.  I can chase that down.

greghudson:
Ah, the bug is that ``DH_free(dh_server);`` is now called twice in the cleanup label, which is a double free.

Sashan:
I've just submit new PR #738, which addresses double free, but still ignores your comments.
I'll get back to my patch and will try to address them. I'll take it as an exercise.

greghudson:
Closing this in favor of #738.  When you have revisions to a pull request branch, you can just force-push it to the same branch in your github fork to update the PR, instead of opening a new PR.

PR #741, opened 2018-03-05
SPAKE preauth support

greghudson:
In the past I would have divided a change of this size into multiple commits for easier review, generally separating out the tests and the documentation or something.  I'm not sure that's really useful here.  The bulk of the review work will be on the code anyway, so I'm leaving this all in one commit.  The ASN.1 support has already been merged.  There is one initial commit to extend k5_sha256() to support multiple data inputs.

The largest chunks of diff here are edwards25519.c and edwards25519_tables.h; all of the other files are just a few hundred lines each.  edwards25519_tables.h does not need review.  For edwards25519.c, review should only be needed on the big initial comment and on the last few functions.  I don't think there's any need (or resources) to double-check the mathematical correctness--this all gets tested against the test vectors which came from a separate Python implementation, and most of the code from BoringSSL was formally verified using Coq (which is why it looks strange with a ton of open braces).


doc/admin/conf_files/krb5_conf.rst
```diff
@@ -365,6 +365,21 @@ The libdefaults section may contain any of the following relations:
...
+    are:
+
+    ============ ================================
+    edwards25519 Edwards25519 curve (:rfc:`7748`)
+    P-256        NIST P-256 curve (:rfc:`5480`)
+    P-384        NIST P-384 curve (:rfc:`5480`)
+    P-521        NIST P-521 curve (:rfc:`5480`)
+    ============ ================================
+
+    The default value for the client is ``edwards2559``.  The default
```
abbra: the default value is different from the table above, perhaps this is a typo?

greghudson: Yes, thanks, I will fix that.

frozencemetery:
When do we plan to turn on SPAKE by default?  When second factors hit, or ...?

greghudson:
I was thinking after one release, with the hope of hearing about some user experiences in the intervening year.  Second factor support is not a requirement.

KDC performance is a concern for enabling SPAKE by default, but less of one than when I started the work.  Last week I updated the edwards25519 code to the latest from BoringSSL, and the new 64-bit code is quite fast.

greghudson:
Pushed changes to:

* Use expected_trace in t_spake.py, instead of the file-local expect() and trace() helpers.  I probably wrote those tests before expected_trace existed.
* Disable the asserts in edwards25519.c by default, with a note that they should be enabled for a test run whenever that code is updated.
* Fix a grammar error in doc/formats/cookie.rst ("zero or" -> "zero or more").
* Remove TRACE_SPAKE_BAD_KDC_CHALLENGE_GROUP and instead error out of SPAKE kdcpreauth initialization in that case.  Previously we were issuing the trace message and not doing any other error handling, which doesn't seem right.
* Fix a memory leak in derive_wbytes(), and while I was there, make the code shorter by preallocating the PRF input string instead of using k5buf.

npmccallum:
Overall, this looks great. Here's my one gripe:

I'm not enthralled with the way that `spake_preauth_indicator` is implemented.

1. I suspect the ability to specify more than one indicator would be helpful.

2. I would hope that 2FA plugins would be able to add indicators. It isn't obvious to me how this could be done with the current code structure. I may be missing something.

greghudson:
Thanks for the review.

Right now PKINIT and OTP allow multiple indicators from the profile (plus additional indicators from certauth modules for PKINIT), while encrypted challenge allows just the one.  I didn't think it all that likely that admins would want to configure indicators for 1FA SPAKE as it doesn't prove any additional credential knowledge, so I tried to keep the code simple and followed the encrypted challenge approach.  But it likely wouldn't be more than a couple of lines of additional code to handle multiple indicators, so I can do that.

I also expect that 2FA modules would need to be able to add indicators (more likely than for 1FA SPAKE, in fact).  One option there is to pass cb/rock into 2FA modules so that they can call cb->add_auth_indicator() themselves; another is to get the indicators back from them via the 2FA pluggable interface.  The first option might make the most sense as 2FA modules might benefit from other callbacks like get_string() and client_entry().

It does occur to me, though, that the 1FA SPAKE indicators from the profile will need to be added in verify_encdata() as well as verify_response().  We wouldn't want to duplicate that much code between the two functions, so we would either need a helper function, or to move the code somewhere else.

One obvious candidate for another place is spake_return(), as the return_padata method is only (and always) invoked when we're reply with an AS-REP.  But in an apparent oversight, adding auth indicators from the return_padata method doesn't entirely work, as kdc/do_as_req.c calls check_indicators() and check_kdcpolicy_as() before return_padata().  It might be reasonable to insist that adding auth indicators is part of verification, but I didn't document the requirement, so I probably just didn't think about it.

Adding indicators at the end of spake_verify() on success won't work because of the asynchronous function delegation to verify_response() and friends.  Switching to normal synchronous delegation (collecting results and calling (*respond)() at the end of spake_verify() in all cases) would preclude having asynchronous 2FA modules.  So that won't work.

I will move the 1FA indicator code into a helper function for now, unless there's sentiment that I should make adding indicators from the return_padata method work.

npmccallum:
This seems reasonable.

One thing I want to see for 2FA is that indicators are added at the same time as the authentication itself. Currently, in OTP, we have two steps:
1. Perform authentication.
2. Add indicators.

As we (FreeIPA) implement it, this results in two queries to LDAP. Thus, if LDAP changes between these two steps we get wrong indicators. Thus, I'd like to be able to have a single step to avoid this problem in the future.

greghudson:
I made those changes in a separate commit so Nathaniel can have a look (since the existing commit is quite large); I will fold them into the prior commit before merging.

PR #744, opened 2018-03-13
krb5: let k5_pac_validate_client handle S4U2Self with enterprise prin…

frenche:
…cipals

Tested-by: Isaac Boukris <iboukris@gmail.com>

References:
http://mailman.mit.edu/pipermail/krbdev/2017-August/012806.html
https://lists.samba.org/archive/samba-technical/2018-March/126192.html

I didn't manage to add a test , as I can't find a way to get the returned pac-principal with an '@' sign.
(otherwise note that S4U2Self test in t_authdata.py uses 'kvno', which I think always parses client name as enterprise principal).

Also, please let me know if a bug is needed in order to allow for back-ports.
Thank you.

### DIFF
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -413,6 +413,7 @@ k5_pac_validate_client(krb5_context context,
     krb5_ui_2 pac_princname_length;
     int64_t pac_nt_authtime;
     krb5_principal pac_principal;
+    int flags;
 
     ret = k5_pac_locate_buffer(context, pac, KRB5_PAC_CLIENT_INFO,
                                &client_info);
@@ -440,8 +441,12 @@ k5_pac_validate_client(krb5_context context,
     if (ret != 0)
         return ret;
 
-    ret = krb5_parse_name_flags(context, pac_princname,
-                                KRB5_PRINCIPAL_PARSE_NO_REALM, &pac_principal);
+    /* Parse the UTF-8 name as an enterprise principal if we are matching
+     * against one; otherwise parse it as a regular principal with no realm. */
+    flags = KRB5_PRINCIPAL_PARSE_NO_REALM;
+    if (principal->type == KRB5_NT_ENTERPRISE_PRINCIPAL)
+        flags |= KRB5_PRINCIPAL_PARSE_ENTERPRISE;
+    ret = krb5_parse_name_flags(context, pac_princname, flags, &pac_principal);
     if (ret != 0) {
         free(pac_princname);
         return ret;


greghudson:
I don't like conditionalizing on whether the PAC string contains a '@'.  Conditionalizing on ``principal->type == KRB5_NT_ENTERPRISE_PRINCIPAL`` as you did in your patch would be fine, although I would go with the approach in this PR of building up a flags variable rather than embedding a conditional in the krb5_parse_name_flags() argument.

With the conditional changed, I guess I like this approach better than the Heimdal approach of unparsing and doing a text comparison, but only slightly.

It is not necessary to file a bug report to get a PR backported.  But I would like to know more about when enterprise principal names appear in PACs.  [MS-PAC] just says that PAC_CLIENT_INFO "contains the client's account name".

greghudson:
Oh, and can you elaborate on "I can't find a way to get the returned pac-principal with an '@' sign"?  I'm not sure quite what operation you need or how that fits into a test case.

frenche:
Hi, I've updated with v2 patch.

I think the enterprise-name would appears in PAC when it is sent in PA-FOR-USER as such (passed to krb5_get_credentials_for_user()).
In MS-KILE it says PAC_CLIENT_INFO is the cname:
https://msdn.microsoft.com/en-us/library/ee896749.aspx
In MS-SFU it says, the cname is taken from PA-FOR-USER, see:
https://msdn.microsoft.com/en-us/library/ff634450.aspx

Tshark output of the packet exchange:
https://pastebin.com/ECwB64Eh
Samba net command debug output including krb5_trace:
https://pastebin.com/uk0bbQcb

Regarding the test, my understanding was that I need to trigger an mspac returned in S4U2Self flow, where PAC_CLIENT_INFO is a string with an '@' sign to cause the same error (has realm present), however I can't figure how feasible it is.

greghudson:
I don't think any of the in-tree DB modules can synthesize a PAC at all, even the test module, so yeah, it would take a fair amount of work to create a test case.

greghudson:
I don't understand why the KDC wouldn't canonicalize the client principal when processing an S4U2Self request (it would for an AS request, right?), but that packet capture confirms that it doesn't.

frenche:
Yeah, it would canonicalize for AS request, if the canonicalize flag is on, but never for S4U2Self.

Maybe it must return the exact username to assure the client it supports the extension (otherwise, how would the client know it wasn't the impersonator that got canonicalized).
Or maybe canonicalization doesn't work in TGS, just guessing.

btw, the client could do the canonicalization by him self, from PAC data.

PR #742, opened 2018-03-12
Implement PKINIT freshness tokens (RFC 8070)

greghudson:
This work naturally divided into client and KDC support.  I put tests and documentation in with the KDC commit, as the KDC side is the bulk of what's being tested and documented.

PR #746, opened 2018-03-15
Include preauth name in trace output where available

frozencemetery:
During tracing, query the preauth context to convert the
krb5_preauthtype numbers into human-readable names for debugging.
Also add a new trace type for single padata numbers.

This is most definitely a hack.  However, since the whole preauth system is dynamic, I'm not sure how we can do better approach-wise.

A more invasive change to enable us to always have the name around a trace time is probably doable, if this/that is the correct approach.  (I'd want it in a separate commit though.)

### DIFF
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -75,6 +75,7 @@
  *   {cksum}       const krb5_checksum *, display cksumtype and hex checksum
  *   {princ}       krb5_principal, unparse and display
  *   {ptype}       krb5_int32, krb5_principal type, display name
+ *   {patype}      krb5_preauthtype, a single padata type number
  *   {patypes}     krb5_pa_data **, display list of padata type numbers
  *   {etype}       krb5_enctype, display shortest name of enctype
  *   {etypes}      krb5_enctype *, display list of enctypes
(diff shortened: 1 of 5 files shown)


greghudson:
This doesn't seem like a hack to me, but what did you have in mind to always have the name around?  We could just have a table that we need to update with new pa-types; that would let us handle things like 133 (PA-FX-COOKIE) and also display protocol constant names rather than module names.

Minor concerns: we generally use k5_ rather than krb5int_ for new internal functions; some of the changes to the tests might create long lines (I can have a look and decide what to do); krb5int_pa_module_name() should have a comment noting that the lifetime of the returned name is restricted to the lifetime of the context.  (Because of RTLD_NODELETE that's not necessarily true, but not all platforms have that feature.)

frozencemetery:
> This doesn't seem like a hack to me, but what did you have in mind to always have the name around?

I considered making the surrounding functions perform the lookup before tracing (since often they then proceed to perform lookups after), but it seemed too invasive.

A pa-types table would probably be superior, but I don't know how feasible that is to construct.

> some of the changes to the tests might create long lines

I don't think it's worth changing the output format for this (though I'm of course open to changing it if there are other reasons).  I've included one possibility for less long lines without breaking the strings in this update.

greghudson:
We have a function in trace.c named principal_type_string; we could just add a padata_type_string() below it.  Here's an example based on the IANA table and the constants we have in krb5.h, with the expectation of changing the last two entries once pkinit-freshness and SPAKE are merged:

https://gist.github.com/greghudson/df0db00830d1b750bd82ac22d70d75a9

frozencemetery:
Updated.  There's one line that's now too long that can't really be fixed.  I've removed the dynamic logic entirely; the static is sufficient for my purposes, and any custom plugin is going to stick out by being a number (as the test one does).

greghudson:
My planned merge order is PR #747 (this is occasionally breaking CI tests), PR #742, this PR (with the PA_AS_FRESHNESS entry using the constant, and the SPAKE entry omitted), and then PR #741 (adding a SPAKE entry to the table).

I'm not a big fan of hanging indents (as a purely personal aesthetic opinion).  For long lines in the tests, my usual approach when writing these tests in the first place was to trim the amount of text we're looking for.  Instead of "Produced preauth for next request: ..." we could look for "for next request: ...", and instead of "Preauth tryagain input types: ..." we could look for "tryagain input types: ...".  That trimming alone is enough to address all of the long lines for this patch, after undoing the changes to the indentation style.

I also think I should have used the more concise variable name "msgs" instead of "expected_trace" when setting a list of trace messages before a call to realm.run() or similar.  This rename would yield more horizontal space without having to drop to the next line after the open paren.  Doing that addresses some of the long lines, but we'd still have to trim at least some of the "Produced preauth for next request" lines, at which point it makes sense to trim all of them for consistency.  I could be happy with any combination of the two approaches.

I noticed 109 appearing in one of the trace messages with no name.  109 is the number for TD-DH-PARAMETERS.  TYPED-DATA was introduced in RFC 4120 as a suggested encoding for KRB-ERROR e-data, was mostly only used in PKINIT (and maybe some Microsoft extensions we don't implement; I'm not sure), and was then basically folded into PA-DATA by RFC 6113 section 7, so that IANA has a single registry for both type numbers.  Ideally we would move the PKINIT typed-data declarations from pkinit.h to krb5.hin with KRB5_ prefixes and add them to the translation table, but I think it's probably best to leave that can of worms closed for the moment.

frozencemetery:
Updated accordingly.  I've also rebased onto your pkinit-freshness branch to have the symbol for the moment.

PR #749, opened 2018-03-19
Report extended errors in kinit -k -t KDB:

greghudson:
In kinit, if we recreate the context using kinit_kdb_init(), also
reset the global errctx so that we use the new context to retrieve
extended error messages.

[kinit should just use a global context variable like klist does.  But there are 50+ uses of "k5->ctx" which would need to be changed, which is far more than appropriate for a bugfix commit.  There is some other residual complexity left behind from supporting v4 and v5 kinit which should probably be simplified out at the same time as moving to a global context.]


frozencemetery:
We could also think about removing v4 support...

greghudson:
We removed v4 support back in 1.7, aside from some principal conversion APIs which I believe AFS needs.  But kinit has some residual structure which made sense in 1.6, but doesn't make sense now.  (However, if we simply combine k5_begin() and k5_kinit() into one function, it's a really long function; I am looking into how it can be refactored.)

PR #751, opened 2018-03-21
ASN.1 encoding simplifications

greghudson:
asn1buf.c is legacy code, somewhat trimmed down (it used to be used for decoding) but still messy in design and partly redundant with the newer k5buf.c.  I have long been interested by the Heimdal approach to DER encoding, which encodes backwards into a precisely allocated buffer.  The danger of Heimdal's approach is that it uses separate counting and encoding functions for each primitive, and any disagreement between the two would result in a memory error.

These commits implement what I think is a safer and less verbose variant of the Heimdal approach.  The same functions are used for counting and encoding, with the two new asn1buf primitives conditionally adding bytes and always counting them.

I haven't done any performance measurements, but I expect the impact to be negligible.  We're making more function calls and memory reads from the representation, but we're doing fewer allocations and memory writes.  ASN.1 encoding is almost always adjacent to network I/O which likely dwarfs it in cost anyways.


src/lib/krb5/asn.1/asn1_encode.c
```diff
@@ -26,105 +26,100 @@
...
 /**** Functions for encoding primitive types ****/
 
-krb5_error_code
-k5_asn1_encode_bool(asn1buf *buf, intmax_t val, size_t *len_out)
+/* Insert one byte into buf going backwards. */
+static inline void
+insert_byte(asn1buf *buf, uint8_t o)
+{
+    if (buf->ptr != NULL)
+        *--buf->ptr = o;
```
frozencemetery: Could we do this as a two-liner instead?

src/lib/krb5/asn.1/asn1_encode.c
```diff
@@ -26,105 +26,100 @@
...
     int digit;
 
     valcopy = val;
     do {
         digit = valcopy & 0xFF;
-        ret = asn1buf_insert_octet(buf, digit);
-        if (ret)
-            return ret;
-        len++;
+        insert_byte(buf, digit);
```
frozencemetery: `digit` is an `int` at the time of this call, but the function takes a `uint8_t`.  I think usually we eschew this kind of cast, but it's being used fairly often in this PR.

src/lib/krb5/asn.1/asn1_encode.c
greghudson: Sure:

```C
    if (buf->ptr != NULL) {
        buf->ptr--;
        *buf->ptr = o;
    }
    buf->count++;
```

src/lib/krb5/asn.1/asn1_encode.c
greghudson: I don't mind making `digit` uint8_t in this function and k5_asn1_encode_uint(), but in general I'm okay with implicit downcasting as long as we have reason to believe the value will be in the range of the target integer type, as is the case for all of the insert_byte() calls.

frozencemetery: Thanks for clarifying.

PR #756, opened 2018-03-30
Document comments in krb5.conf

greghudson:


greghudson:
I had thought about that and decided that this much was enough to say on the topic.

kaduk:
That was basically my thinking, too.

PR #760, opened 2018-04-10
Merge duplicate subsections in profile library

frozencemetery:
Modify profile_add_node() to return the existing node, rather than
making a new one, when adding subsection configuration.

This fixes an issue where the first instance of a subsection will hide
the second instance entirely.  In particular, it was previously
impossible to split realm-specific configuration across multiple
config files.

ticket: 7863


greghudson:
I added a test case.

frozencemetery:
Great, thanks!

PR #761, opened 2018-04-14
Remove ulog_replay database lock

greghudson:
Additional context for the PR: in release 1.13 we introduced hierarchical iprop, which means a KDC might simultaneously be running kpropd to receive updates and occasional dump files from an upstream KDC, and kadmind -proponly to serve updates and occasional dump files to downstream KDCs.  At that time I added a number of locking changes including commits 3c3638e1cfafa6ec5595e6c2c06ca209cf9d0d09, 444ef5fe9ec8d64a5db27b3a8aaf6813dd7ef0e0, 71d028f1054deb186807e7c8048218b82b478422, and a2ac57b0ec230efed06fabc2d55db1fcbc6f7ea3.  The general theme was to lock around every ulog operation to ensure consistency, and to avoid simultaneous ulog and DB locks when possible.

A few operations inherently needed to hold simultaneous ulog and DB locks: ulog_replay() and, at the time) kdb5_util dump -iprop.  Because of that we needed a consistent locking order to prevent deadlock.  kdb5_util dump -iprop was taking out an explicit read lock on the DB before reading the ulog header so that we knew the dump file was a snapshot matching the serial number.  Because of that, the consistent order needed to be DB-then-ulog.  ulog_replay() naturally would have been ulog-then-DB, because it holds a lock on the ulog for the whole operation and each DB operation implicitly acquires a short-lived lock.  To rectify that, I added code to take out an explicit write lock on the DB before locking the ulog.

We have since removed the DB lock for iprop dumps, and have just allowed those dump operations to reflect changes postdating the serial number recorded in the dump file.  This is commits 8f752f88bf1b1284d8e2b8119031625f2b703ab7 and 71dff2ba654260a870dfc2adc35b2a90a458ce35, and relates to these tickets:

  http://krbdev.mit.edu/rt/Ticket/Display.html?id=7904
  http://krbdev.mit.edu/rt/Ticket/Display.html?id=7977

Because of that, ulog_replay() is the only operation which simultaneously holds DB and ulog locks, and ulog-then-DB is an okay order for lock acquisition.  So we don't need the explicit write lock.  As I would like to introduce a new KDB module with no support for explicit locking, it makes sense to get rid of it.

(On reflection, it should be possible to avoid holding simultaneous ulog and DB locks in ulog_replay(); it could replay each DB operation with no lock on the ulog, then lock, store the update, and unlock.  But we don't need to make that change right now.)  

After commit 71dff2ba654260a870dfc2adc35b2a90a458ce35, it is no longer
necessary for ulog_replay() to lock the database before the ulog to
avoid deadlock.  Remove the lock so that the iprop upstream KDB is not
required to support explicit locking.


greghudson:
I did a full review of the operations which call locking ulog functions.  In so doing, I realized that any operation which takes out an explicit DB lock (with krb5_db_lock()) and then performs DB writes will be doing DB-then-ulog locking.

Besides this one in ulog_replay(), there are currently only two other operations which call krb5_db_lock(): ``kdb5_util load -update`` and ``kadmin.local lock``.  Neither of those operations have any business happening on a secondary KDC; the only database writers on such a KDC should be ulog_replay() (as invoked by kpropd) and ``kdb5_util load -i`` (also as invoked by kpropd, and not at the same time as ulog_replay()).  Still, I am not comfortable with having two different locking orders in the code.  So I think I will change ulog_replay() to avoid holding simultaneous DB and ulog locks.

greghudson:
Updated to do that.  The code to sanity-check the incoming update serial number needs to be inside the lock (because it accesses a ulog field and because it calls reset_ulog(), which is an unlocked function), so I moved it to after the database update.  For similar reasons, the on-failure-reset-the-ulog code at the end needs to use ulog_init_header() instead of reset_ulog().

PR #763, opened 2018-04-18
Fix segfault in finish_dispatch()

frozencemetery:
dispatch() doesn't necessarily initialize state->active_realm which
led to an explicit NULL dereference in finish_dispatch().

Additionally, fix make_too_big_error() so that it won't subsequently
dereference state->active_realm.

tags: pullup
target_version: 1.16-next
target_version: 1.15-next

### DIFF
--- a/src/kdc/Makefile.in
+++ b/src/kdc/Makefile.in
@@ -85,6 +85,7 @@ check-cmocka: t_replay
 check-pytests:
 	$(RUNPYTEST) $(srcdir)/t_workers.py $(PYTESTFLAGS)
 	$(RUNPYTEST) $(srcdir)/t_emptytgt.py $(PYTESTFLAGS)
+	$(RUNPYTEST) $(srcdir)/t_bigreply.py $(PYTESTFLAGS)
 
 install:
 	$(INSTALL_PROGRAM) krb5kdc ${DESTDIR}$(SERVER_BINDIR)/krb5kdc
--- a/src/kdc/dispatch.c
+++ b/src/kdc/dispatch.c
@@ -124,7 +124,7 @@ dispatch(void *cb, const krb5_fulladdr *local_addr,
          verto_ctx *vctx, loop_respond_fn respond, void *arg)
 {
     krb5_error_code retval;
-    krb5_kdc_req *as_req;
+    krb5_kdc_req *req = NULL;
     krb5_data *response = NULL;
     struct dispatch_state *state;
     struct server_handle *handle = cb;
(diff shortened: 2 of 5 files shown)


frozencemetery:
I've minimized the commit and will file a ticket in a moment.

greghudson:
I added a test case by setting [kdcdefaults] kdc_max_dgram_reply_size = 10.

The test currently fails because the KDC fails to encode the error message when errpkt.server is NULL, as the KRB-ERROR sname and realm fields aren't optional.  (This is kind of a botch in RFC 4120; what if the service name is so long that the RESPONSE_TOO_BIG error doesn't fit into a UDP reply?)  So while kvno works, the KDC never responds to the UDP request and the client waits a second before trying TCP.

greghudson:
Since 1.0, we have had this weird asymmetry in the KDC code where process_tgs_req() is responsible for decoding the request packet but process_as_req() is not.  (Both functions need the DER representation for various reasons; process_as_req() receives both the decoded and DER form of the request as parameters.)

Before commit 0a2f14f752c32a24200363cc6b6ae64a92f81379 this asymmetry didn't have much effect; the global kdc_active_realm was set either way.  After that commit, we have the problem where the dispatch.c state->kdc_active_realm is only set for the AS code path.

I think to properly fix this bug, we need to fix the asymmetry and make dispatch() decode both AS and TGS requests and call setup_serve_realm(), changing the signature of process_tgs_req() so that it receives the decoded request and active realm instead of the server handle.  Both kinds of requests decode into krb5_kdc_req structures, so it should be possible to avoid duplicating any code for the two kinds of requests.

Arguably we should also be putting request->server into too-big errors instead of the active realm's TGS principal (which was not necessarily named in the KDC request).  To do that we would also want the decoded request in dispatch.c for both AS and TGS requests.  But that behavior change is out of scope for a backport commit, and doesn't seem very important anyway.

greghudson:
I added the test case back in (and fixed it up; the expected trace messages didn't quite work for the TGS case).  I also made a few amendments:

* Eliminated code duplication in dispatch() by using a single variable to hold the decoded KDC request, and by freeing the request in the done label (setting it to NULL in the case where process_tgs_req() frees it).

* process_tgs_req() no longer needs the server handle, if we just change ``handle->kdc_err_context`` to ``kdc_context`` in a few places.

* Edited the commit message to make it clearer when the null dereference occurred.

To eliminate the ownership transfer to the processing function (in the future), we need to add the request to the dispatch state structure and alter the contract of kdc_find_fast() so it doesn't free and replace the request.  For the latter change, I think in both processing functions we would want the decoded request to arrive as ``outer_request``, possibly be unwrapped by kdc_find_fast() into ``inner_request``, and ``request`` or ``state->request`` would become an alias to one of those.

I am also kind of interested in making dispatch() responsible for checking the msg_type field, so the processing functions don't have to.  But that's an unnecessary change for this bugfix, so I didn't throw it in.

frozencemetery:
Thanks!

PR #764, opened 2018-04-18
- better user experience with 'kadmin -proponly'

Sashan:
I forgot to add 'iprop_enable = true' to kdc.conf for my test realm.
Then I tried to run kadmin with -proponly flag. It gave me message:
	root@kdc1:~#  /usr/lib/krb5/kadmind -nofork -proponly
	kadmind: kadmind: Invalid argument while initializing network, aborting

the message is not helpful much. My first idea was to let kadmin
fail with more user friendly message such as:
	'-proponly requires to put 'iprop_enable = true' for realm in kdc.conf

thinking more about I believe the '-noprop' flag should just override
iprop_enable = false.

### DIFF
--- a/src/kadmin/server/ovsec_kadmd.c
+++ b/src/kadmin/server/ovsec_kadmd.c
@@ -468,6 +468,10 @@ main(int argc, char *argv[])
         fail_to_start(0, _("Missing required realm configuration"));
     if (!(params.mask & KADM5_CONFIG_ACL_FILE))
         fail_to_start(0, _("Missing required ACL file configuration"));
+    if (proponly && !params.iprop_enabled) {
+        fail_to_start(0, _("-proponly can only be used when "
+                           "iprop_enable is true"));
+    }
 
     ret = setup_loop(&params, proponly, &vctx);
     if (ret)


greghudson:
iprop_enable also affects the operation of kdb5_util dump, so I don't think it would be safe to make kadmind -proponly imply iprop_enable.  Improving the error message would be worthwhile, though.

src/kadmin/server/ovsec_kadmd.c
```diff
@@ -166,7 +166,7 @@ setup_loop(int proponly, verto_ctx **ctx_out)
             return ret;
     }
 #ifndef DISABLE_IPROP
-    if (handle->params.iprop_enabled) {
+    if ((handle->params.iprop_enabled) || (proponly)) {
```
frozencemetery: No parens around `proponly`.

frozencemetery:
Left some feedback.  I think it's clear that there should be a clearer error message here at the very least, but am not sure whether we want the override behavior.

Sashan:
OK, understood. So I'm adding an 'else' branch, which prints error message with enough hints.

src/kadmin/server/ovsec_kadmd.c
```diff
@@ -173,6 +173,10 @@ setup_loop(int proponly, verto_ctx **ctx_out)
                                    krb5_iprop_prog_1);
         if (ret)
             return ret;
+    } else if (proponly) {
+        fprintf(stderr, _("%s: add 'iprop_enable = true' to kdc realm "
+                          "configuration, when using '-proponly' switch\n"),
```
frozencemetery: I don't think this is quite what you mean.  The thing the error should be capturing is the conflict between `iprop_enable = false` (which happens to be the default) and the user explicitly requesting `-proponly`.

Sashan:
I've moved the test to `main()`. This is the error message displayed:
```
kadmind: kadmind: '-proponly' can't be used when 'iprop_enable' is set to false, aborting
```

PR #772, opened 2018-04-28
LMDB KDB module

greghudson:
The testing commit has a lot of churn, particularly in t_iprop.py where a lot of existing code needed to be indented one level.  Reviewing it with git diff -w might be easiest.  I chose not to make the kadmin/gssrpc tests honor the K5TEST_LMDB environment variable because the necessary shell code wouldn't have been as simple as I would have liked, and because I think they don't add much to DB code coverage.

I wasn't sure whether to add a new admin doc section for KDB types and adding a subsection to database.rst.  I went with a new section for now, partly because I don't like the current contents of database.rst (it copies a bunch of reference material with ``.. include`` directives) and therefore couldn't figure out where the new material would best fit.

Some notes on places where it might be possible to factor out similar code but I decided against it:

* The lockout code is duplicated among the three KDB modules.  I filed issue 8667 about this; we could make small libkdb5 helper functions but we can't factor out the majority of the code.

* destroy_file() is duplicated with DB2 (although the version here is simplified a bit).  We could put it in kdb5.c but it would feel out of scope there.  Putting it in libkrb5support might be interesting (we could use it for kdestroy, for instance), but then we have to worry about whether it builds on Windows.

* klmdb_iterate() and klmdb_iter_policy() have some repeated structure, but they have different callbacks and principal iteration has to fetch the lockout information and to support reverse iteration, so I didn't think a common function would work very naturally.

* klmdb_open() and klmdb_create() wind up in similar states as far as open environments and DBs, but get there in sufficiently different ways that it didn't look feasible to make a common function.


src/plugins/kdb/lmdb/kdb_lmdb.c
```diff
@@ -0,0 +1,1142 @@
...
+{
+    return 0;
+}
+
+static krb5_error_code
+klmdb_fini(krb5_context context)
+{
+    klmdb_context *dbc;
+
+    if (context->dal_handle->db_context == NULL)
```
frozencemetery: Can just be `if (dbc == NULL)` if the assignment is moved above this check.

src/plugins/kdb/lmdb/kdb_lmdb.c
```diff
@@ -0,0 +1,1142 @@
...
+        return 0;
+    dbc = context->dal_handle->db_context;
+    mdb_txn_abort(dbc->read_txn);
+    mdb_txn_abort(dbc->load_txn);
+    mdb_env_close(dbc->env);
+    mdb_env_close(dbc->lockout_env);
+    free(dbc->path);
+    free(dbc->lockout_path);
+    free(dbc);
+    context->dal_handle->db_context = NULL;
```
frozencemetery: `dbc = NULL`

src/plugins/kdb/lmdb/kdb_lmdb.c
```diff
@@ -0,0 +1,1142 @@
...
+            goto cleanup;
+    }
+
+cleanup:
+    mdb_cursor_close(cursor);
+    mdb_txn_abort(txn);
+    return ret;
+lmdb_error:
+    ret = klerr(context, err, _("LMDB principal iteration failure"));
+    goto cleanup;
```
frozencemetery: I'm not really a fan of this use of `goto` for control flow... how would you feel about putting `lmdb_error:` above `cleanup:` for fall-through there and using an explicit `goto cleanup;`after the `for` loop?

src/plugins/kdb/lmdb/lockout.c
```diff
@@ -0,0 +1,180 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* plugins/kdb/lmdb/lockout.c */
+/*
+ * Copyright (C) 2009 by the Massachusetts Institute of Technology.
```
frozencemetery: Should the year update here?

src/plugins/kdb/lmdb/kdb_lmdb.c
greghudson: That wouldn't have the same effect (or any effect, since it's right before a return statement).

greghudson:
Thanks for the review.  I updated for everything except the comment I responded to.

src/plugins/kdb/lmdb/kdb_lmdb.c
frozencemetery: Right, got too excited.

PR #773, opened 2018-05-03
Remove unrecognized "-nodes" argument from cert scripts

frozencemetery:
Apparently openssl never recognized this option, but previously didn't
complain about it.  Not using DES is the default, so it's safe to
remove.  This fixes make-certs.sh to run on openssl-1.1.0h.

### DIFF
--- a/src/tests/dejagnu/pkinit-certs/make-certs.sh
+++ b/src/tests/dejagnu/pkinit-certs/make-certs.sh
@@ -114,7 +114,7 @@ extendedKeyUsage = $CLIENT_EKU_LIST
 EOF
 
 # Generate a private key.
-openssl genrsa $KEYSIZE -nodes > privkey.pem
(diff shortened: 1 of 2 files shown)


vdukhovni:
Note: It was never an "option" it was placed after a positional argument!  The old code ignored extra positional arguments, the new code does not (which is a good thing).

greghudson:
From looking at various versions of OpenSSL:

* ``-nodes`` is an option for ``openssl req`` and ``openssl pkcs12`` but not ``openssl genrsa``; genrsa defaults to no encryption for the private key.

* openssl has never honored options after positional arguments like gnu libc's getopt() does, so Viktor is correct that the ``-nodes`` was being ignored as an excess positional argument, which is now treated as an error.

So, the patch change is good; I will edit the commit message a bit.

PR #775, opened 2018-05-07
Windows build fixes and cleanups

greghudson:
These commits fix the build (including Leash but not the wix installer) for Visual Studio 2017 Community, cleans up some detritus, and adds .gitignore entries to aid in Windows development.

Some additional context on the Leash out2con.cpp change: this code was added by Kevin Wasserman in 2011 as a debugging aid; I believe the idea is that ``leash -console`` will create a console window and any stdio output (which there isn't any of by default, but a developer might add printf statements for debugging) will be duplicated to that window as well as the original stdout.  In PR #312, @tsondergaard concluded that we should do this with _dup() and _dup2() on the underlying file descriptors rather than violate stdio abstractions.  I experimentally modified the code to do that; unfortunately, I was unable to get either the old code (with the _flag assignment removed) or the new code to output anything to the console window.  It is inconvenient for me to test the old code under VS2010 (so far I have been using VS2017 Community with no MSDN license) to verify that it ever worked.  It is possible that the right answer is to remove out2con.cpp and the -console flag entirely.


greghudson:
I added a commit to minimally update the installer for VS2017 and current Wix toolset.  There are a bunch of deprecation warnings not fixed by this commit; fixing those leads down a rabbit hole of apparently unnecessary registry entries, which will take more investigation to resolve.

PR #776, opened 2018-05-07
Log a message when non-root ksu authorization fails

frozencemetery:
If non-root user attempts to ksu but is denied by policy, log to
syslog at LOG_WARNING in keeping with other failure messages.

ticket: 8270

### DIFF
--- a/src/clients/ksu/main.c
+++ b/src/clients/ksu/main.c
@@ -414,6 +414,16 @@ main (argc, argv)
     if (hp){
         if (gb_err) fprintf(stderr, "%s", gb_err);
         fprintf(stderr, _("account %s: authorization failed\n"), target_user);
+
+        if (cmd != NULL) {
+            syslog(LOG_WARNING,
+                   "Account %s: authorization for %s for execution of %s failed",
+                   target_user, source_user, cmd);
+        } else {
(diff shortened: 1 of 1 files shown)


greghudson:
Ticket 8270 talks about successful operations, while this commit is about failed operations.  Is another change needed for successful ksu command operations using .k5users?

greghudson:
I looked over the bugzilla report ( https://bugzilla.redhat.com/show_bug.cgi?id=1270927 ) and I'm a little lost.  It sounds like there might have a disagreement about whether there is a syslog on success (when running a specific command).  There is some discussion of root vs. non-root cases, but I can't keep track of when that's talking about the running uid versus the target uid.

I think there is also a Fedora patch to use PAM in use, which could be muddying the waters slightly.

frozencemetery:
That wasn't my reading, but I can double-check.  (I read it as never seeing the command get logged at all, in the success case because the wrong component was being looked at, and in the failure case because no message ends up being logged.)

The PAM changes are an updated version of [Ticket 5939](http://krbdev.mit.edu/rt/Ticket/Display.html?id=5939).  (I think we've discussed this previously and there wasn't interest in having this upstream, but if I'm wrong on that I'm happy to submit a PR for it.)  I will poke a bit and see what difference not having it makes.

PR #777, opened 2018-05-11
Windows sources cleanup

greghudson:
The first commit here stops building an old ticket manager; the second removes a bunch of unused files.


greghudson:
Yes; there was a merge conflict with the SCLIB removal and I apparently failed at git.  cns/Makefile.in and wintel/Makefile.in are now removed in the second commit.

kaduk:
Cool, thanks for the update (which looks good).

PR #781, opened 2018-05-23
Use PKG_PROG_PKG_CONFIG instead of a hardcoded pkg-config command

tgurr:
In cross environments the pkg-config command may be prefixed with the
host triplet, e.g. x86_64-pc-linux-gnu-pkg-config.

### DIFF
--- a/src/configure.in
+++ b/src/configure.in
@@ -22,6 +22,8 @@ AC_SUBST(KRB5_VERSION)
 
 AC_REQUIRE_CPP
 
+PKG_PROG_PKG_CONFIG
+
 AC_CHECK_HEADER([stdint.h], [],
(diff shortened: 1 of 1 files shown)


greghudson:
If PKG_PROG_PKG_CONFIG fails to find pkg-config (so PKG_CONFIG is set to ""), we could set it to "pkg-config", expecting that it won't be found in the path, but the failing commands will be clear (and will behave exactly as they do currently).

Alternatively, we could use PKG_CHECK_EXISTS for resolv_wrapper and PKG_CHECK_MODULES for libedit and libverto.  They check for pkg-config not existing, and even let the user of configure set module_CFLAGS and module_LIBS to sidestep the pkg-config requirement.  I will look into that.

greghudson:
Addressed by PR #787.

PR #782, opened 2018-06-01
Prune Leash sources

greghudson:
When Leash was imported in 2011-2012, some of its functionality was
removed either at compile-time or runtime--most notably krb4 support,
AFS integration, and editing of "properties" (registry values).
Remove most of the unnecessary material.

[This does not include removing the -noribbon flag and its associated code; there should be no user-visible changes from this commit.]


kaduk:
I thought that the htmlhelp tree was "real", in some sense (we wanted to have the viewable files in the tree as opposed to just in the opaque HTML Help Editor format?), and there was some hassle-y workflow that we used to do the conversion.

greghudson:
Viewable files in the source tree, or in the filesystem after the msi installer is used?  We only copy ``windows\leash\htmlhelp\*.chm`` in nmake install, and files.wxi doesn't mention any .htm files.

I left behind all of the HTML files actually referenced in MITKerberosHelp.hhp, and the images they reference.

kaduk:
I believe viewable in the source tree (and potentially so as to be viewable on our website as well).
I might have more state in my email archives, but can't really look right now.

greghudson:
I do see extracted help files in http://web.mit.edu/kerberos/kfw-4.1/kfw-4.1/kfw-4.1-help.  The HTML files and images there appear to match the ones I left behind, except with lowercase filenames.  http://k5wiki.kerberos.org/wiki/Kerberos_for_Windows_Release_Engineering says we extract the HTML files from the CHM file using arCHMage and then run a perl one-liner to fix up the image links.

kaduk:
Ah, that matches up sufficiently well with my recollection to leave me confident that it is what I'm thinking of.  Thank you for doing the work to find it, and sorry for making you do the extra work!
(I accordingly remove any objection I had raised to this pull request.)

PR #790, opened 2018-06-13
Make docs build python3-compatible

frozencemetery:
python3-sphinx removes the execfile() function, which we use for
loading version data and paths information in docs.  Add our own
replacement wrapper instead.

See also: https://portingguide.readthedocs.io/en/latest/builtins.html#removed-execfile


greghudson:
Does it still work if we just exec the file contents without calling compile()?  exec can take a string rather than a code object.

frozencemetery:
It sure does, updated!

PR #792, opened 2018-06-14
Add documentation on dictionary attacks

greghudson:


doc/admin/dictionary.rst
```diff
@@ -0,0 +1,81 @@
...
+encryption type and salt string of the attacker's choosing.  Any of the
+following methods can prevent dictionary attacks by active attackers:
+
+* Enabling SPAKE preauthentication and setting the
+  **disable_encrypted_timestamp** variable to ``true`` in the
+  :ref:`realms` subsection of the client configuration.
+
+* Using an HTTPS proxy, configured in the client's krb5.conf realm
+  configuration.  Using :ref:`KDC discovery <kdc_discovery>` may be
+  defeated with DNS spoofing.
```
kaduk: Should probably expound a little bit more on the second sentence, about who is defeating what.

greghudson: Done, along with a few other adjustments.

greghudson:
I think it's worth knowing about; it's a pretty large constant factor.

greghudson:
I can deemphasize OTP and PKINIT pretty easily.  Starting with solutions would require a complete rewrite.  The best solution for a deployment depends on that deployment's flexibility and on the likely threats.  It's pretty important that admins set ``+requires_preauth -allow_srv`` on user principals if they can't do anything better, but it's almost important that they understand what the remaining risks are once they do that.

greghudson:
To deemphasize OTP/PKINIT, I struck the initial clause and added a short paragraph about them to the end.

PR #797, opened 2018-06-16
Eliminate use of the 'register' keyword

tsondergaard:
'register' is a reserved and unused keyword in C++17 so having it
present in the public headers presents a a compatibility issue. Also
in C the 'register' keyword is mostly obsolete, so remove all uses of
it.

Specifically this fixes the following warning from MSVC 2017 (15.7.3) when using the /std:c++17 switch:
    
```krb5\include\krb5/krb5.h(3544): warning C5033: 'register' is no longer a supported storage class```

### DIFF
--- a/src/lib/rpc/xdr.c
+++ b/src/lib/rpc/xdr.c
@@ -212,7 +212,7 @@ xdr_u_long(XDR *xdrs, u_long *ulp)
  * XDR short integers
  */
 bool_t
-xdr_short(register XDR *xdrs, short *sp)
+xdr_short(XDR *xdrs, short *sp)
 {
 	long l;
 
@@ -243,7 +243,7 @@ xdr_short(register XDR *xdrs, short *sp)
  * XDR unsigned short integers
  */
 bool_t
-xdr_u_short(register XDR *xdrs, u_short *usp)
+xdr_u_short(XDR *xdrs, u_short *usp)
 {
 	u_long l;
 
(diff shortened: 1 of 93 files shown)

### CODE REVIEW

File: src/lib/rpc/xdr.c [hunk 1]
```diff
@@ -607,7 +607,7 @@ xdr_union(
 bool_t
 xdr_string(XDR *xdrs, char **cpp, u_int maxsize)
 {
-	register char *sp = *cpp;  /* sp is the actual string pointer */
+	char *sp = *cpp;  	/* sp is the actual string pointer */
```
frozencemetery: Looks like unintentional whitespace change?

src/util/ss/execute_cmd.c
```diff
@@ -53,14 +53,14 @@
...
     int argc;
     char *argv[];
     int sci_idx;
 {
-    register ss_request_entry *request;
-    register ss_data *info;
-    register char const * const * name;
+    ss_request_entry *request;
+    ss_data *info;
+    char const * const * name;
```
frozencemetery: `char const *const *name;`?

frozencemetery:
libss comes from e2fsprogs, right?  Is there a plan to upstream these changes?

src/lib/rpc/xdr.c
greghudson: Inline comments are supposed to start at column 32, I think in the Solaris style as well as ours.  However, having two spaces before the tab is anomalous even in a file which uses tabs.  I changed it to be just a tab.

src/util/ss/execute_cmd.c
greghudson: Yes, fixed.

greghudson:
e2fsprogs contains a version of libss, which is typically the one used on Linux distributions as the system libss.  But we don't treat it as an upstream and I think there's a fair amount of divergence.

tsondergaard:
Quick response. Thanks!

PR #800, opened 2018-06-19
A round of (mostly leak) fixes

frozencemetery:
Since there have been a pile of leak fixes recently, I figured I'd get these out rather than having them duplicated.  Now that the `register` work has merged, there shouldn't be conflicts with master.

### DIFF
--- a/src/appl/gss-sample/gss-server.c
+++ b/src/appl/gss-sample/gss-server.c
@@ -781,6 +781,7 @@ main(int argc, char **argv)
                 /* Accept a TCP connection */
                 if ((work->s = accept(stmp, NULL, 0)) < 0) {
                     perror("accepting connection");
+                    free(work);
                     continue;
                 }
 
--- a/src/appl/user_user/client.c
+++ b/src/appl/user_user/client.c
@@ -53,7 +53,7 @@ int main (int argc, char *argv[])
 
     if (argc < 2 || argc > 4) {
         fputs ("usage: uu-client <hostname> [message [port]]\n", stderr);
-        return 1;
+        exit(1);
     }
 
     retval = krb5_init_context(&context);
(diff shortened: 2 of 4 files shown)


greghudson:
Focusing on the appl/user_user/client commit for just one moment:

In a lot of the "goto done" cases, retval won't have been set to a non-zero value (because the condition is something like "gethostbyname() returned NULL") and therefore main() will incorrectly return zero.

I believe as a matter of programming practice that it should be okay for a program to call exit(N) for some non-zero N without freeing allocated memory or closing file descriptors, but that calling "return N;" from main() without freeing memory accessible from variables in main() is less okay.  So it might be reasonable to just change all of the "return N;" to "exit(N);".  If the tooling that motivated this change was Coverity, I think that change should shut it up, although I'm not 100% sure.

frozencemetery:
That seems okay.  Updated

greghudson:
Please pull the first commit (otp_state.c:callback) out into its own PR.  We will need to create a ticket describing the bug as it appears to the user and probably add a test case.

Everything else looks fine for this PR.  For "Fix garbage return in extract_cammacs()", I verified through code inspection that this code error can't currently manifest as a user-visible bug (extract_cammacs() won't be called with a null or zero-length list).  All of the other commits appear to be leak fixes in places where leaks don't matter (ksu or test programs).

greghudson:
I oversimplified in my feedback on user_user/client.c.  The final "return 0;" should stay, and any memory the code doesn't currently free on the success path should be freed.  Only the failure paths should change to exit().

frozencemetery:
I *think* (and the analyzer seems to agree) that no memory leaks on successful exit.  Closed the socket, though.

frozencemetery:
(Deferring ksu commits per IRC conversation.)

greghudson:
I looked around in ksu a bit and concluded that it was likely written under the assumption that leaking memory in a short-running program is okay.  Reversing that assumption might be worthwhile, but I would like it to be done all at once rather than piecemeal to avoid a pile of mostly unimportant commits in the same part of the tree.  (Coverity Scan finds at least 17 resource leaks in ksu, for reference.)  So I wound up asking Robbie to defer the ksu commits in this PR for now.

PR #801, opened 2018-06-21
Correctly handle success in otp_state.c's callback()

frozencemetery:
In otp_state.c:callback(), avoid invoking the failure callback on the
success path.  Since request_send() consumes the request, don't try to
free it.

(I've filed a ticket for this.)


greghudson:
I added a test case (relying on a fixed pyrad, which unfortunately hasn't hit Debian, so I did ``pip install --user --upgrade pyrad`` to stop skipping the latter half of t_otp.py).  I also edited the commit message since it isn't the success path which is buggy.

frozencemetery:
Thanks for adding a test, that looks good!

PR #804, opened 2018-07-01
Fix bugs with concurrent use of MEMORY ccaches

greghudson:
[This bug is serious enough to deserve a backport, but the fix is more complicated than I'm normally comfortable backporting.  I will probably decide to backport it but I'm still considering.]

A memory ccache iterator stores an alias into the cache object's
linked list of credentials.  If the cache is reinitialized while the
iterator is active, the alias becomes invalid.  Also, multiple handles
referencing the same memory ccache all use aliases to the same data
object; if one of the handles is destroyed, the other contains a
dangling pointer.

Fix the first issue by adding a generation counter to the cache and to
cursors, incremented each time the cache is initialized or destroyed.
Check the generation on each cursor step and end the iteration if the
list was invalidated.  Fix the second issue by adding a reference
count to the cache object, counting one reference for the table slot
and one for each open handle.  Empty the cache object on each destroy
operation, but only release the object when the last handle to it is
destroyed or closed.

Add regression tests for the two issues to t_cc.c.

The first issue was reported by Sorin Manolache.

ticket: 8707

### DIFF
--- a/src/lib/krb5/ccache/cc_memory.c
+++ b/src/lib/krb5/ccache/cc_memory.c
@@ -102,18 +102,20 @@ extern krb5_error_code krb5_change_cache (void);
 typedef struct _krb5_mcc_link {
     struct _krb5_mcc_link *next;
     krb5_creds *creds;
-} krb5_mcc_link, *krb5_mcc_cursor;
+} krb5_mcc_link;
 
 /* Per-cache data header.  */
 typedef struct _krb5_mcc_data {
     char *name;
     k5_cc_mutex lock;
     krb5_principal prin;
-    krb5_mcc_cursor link;
+    krb5_mcc_link *link;
     krb5_timestamp changetime;
     /* Time offsets for clock-skewed clients.  */
     krb5_int32 time_offset;
     krb5_int32 usec_offset;
+    int refcount;               /* One for the table slot, one per handle */
+    int generation;             /* Incremented at each initialize */
 } krb5_mcc_data;
 
 /* List of memory caches.  */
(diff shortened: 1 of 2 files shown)


greghudson:
Updated to keep the cache locked until krb5_mcc_next_cred() is done; otherwise the pointer in the cursor could be freed by another thread (via krb5_cc_initialize() or krb5_cc_destroy() on another handle referencing the same cache) after the generation is checked.  This race wasn't initially obvious to me because the code after the generation check doesn't appear the access the cache, only the cursor, so I opted for an explicit comment.

frozencemetery:
I think the ticket number is wrong - [8707 is the ulog resource leak](http://krbdev.mit.edu/rt/Ticket/Display.html?id=8707).  I think it should be [8202](http://krbdev.mit.edu/rt/Ticket/Display.html?id=8202).

src/lib/krb5/ccache/cc_memory.c
```diff
@@ -185,39 +205,34 @@ krb5_mcc_initialize(krb5_context context, krb5_ccache id, krb5_principal princ)
...
- *
- * Errors:
- * system errors (locks related)
  */
 krb5_error_code KRB5_CALLCONV
 krb5_mcc_destroy(krb5_context context, krb5_ccache id)
 {
     krb5_mcc_list_node **curr, *node;
-    krb5_mcc_data *d;
+    krb5_mcc_data *d = id->data;
```
frozencemetery: This is duplicated five lines below (right before the `for` loop).

greghudson:
Fixed both issues.

PR #805, opened 2018-07-01
Define missing posix socket error: EAFNOSUPPORT

nehaljwani:
Windows build (vs2008) (some old SDK?) fails with:
```
if not . == . echo .  1>>new-exports 
	..\..\config\rm.bat libkrb5support.exports
	ren new-exports libkrb5support.exports
	cl    -I.\..\..\include -I.\..\..\include\krb5 -I..\..\windows\include -DWSHELPER=1 -DKRB5_DNS_LOOKUP=1 -DWIN32_LEAN_AND_MEAN -DKRB5_DEPRECATED=1 -DKRB5_PRIVATE -D_CRT_SECURE_NO_DEPRECATE -DUSE_LEASH=1  /Os /MD -nologo /EHsc /W3 -Fdobj\AMD64\rel\\ -FD  -Foobj\AMD64\rel\\ -c ...
threads.c
.\..\..\include\port-sockets.h(98) : error C2065: 'EAFNOSUPPORT' : undeclared identifier
.\..\..\include\port-sockets.h(98) : error C2051: case expression not constant
.\..\..\include\port-sockets.h(134) : error C2065: 'EAFNOSUPPORT' : undeclared identifier
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\BIN\amd64\cl.EXE"' : return code '0x2'
Stop.
NMAKE : fatal error U1077: 'for' : return code '0x1'
Stop.
NMAKE : fatal error U1077: 'for' : return code '0x1'
Stop.
```

### DIFF
--- a/src/include/port-sockets.h
+++ b/src/include/port-sockets.h
@@ -72,6 +72,9 @@ typedef WSABUF sg_buf;
 #ifndef ETIMEDOUT
 #define ETIMEDOUT WSAETIMEDOUT
 #endif
+#ifndef EAFNOSUPPORT
+#define EAFNOSUPPORT WSAEAFNOSUPPORT
+#endif
 
 /* Translate posix_error to its Winsock counterpart and set the last Winsock
  * error to the result. */


nehaljwani:
Well, as long as there is no python27 module which requires the dll provided by kerberos ...

greghudson:
Sorry, can you explain how that would constrain the Visual Studio version used to build the KfW sources?

nehaljwani:
Python 2.7 compilation on Windows requires compiling with Visual Studio 2008 (VS2008). This is highly unfortunate, but is the reality. And that is why all modules that python2.7 packages would load, will be required to be built with VS2008 so that runtimes are not mixed. This puts a hard requirement on all the libraries that the python C extensions link against or are imported by python2.7 to be built using VS2008.

greghudson:
What goes wrong if runtimes are mixed?  It doesn't seem tenable to require that every DLL used by a process be built with the same version of Visual Studio.

(If it's a malloc/free issue, we provide APIs to free all memory we allocate so that malloc and free take place in the same build unit.)

greghudson:
I did some research on mixing C runtimes on Windows.  The most comprehensive treatment seems to be this article:

http://siomsystems.com/mixing-visual-studio-versions/

There are basically three possible reactions a developer could have to this state of affairs:

1. We have to build everything with the same Visual Studio version, because surely allocate-free will happen across API borders or file handles will be passed across API borders.

2. We have to make sure that allocate/free happen in the same build unit and avoid passing libc objects like file handles across API borders, because surely people will need to use DLLs built with different versions of Visual Studio, or some code units will be built in debugging mode while others are not.

3. We have to give up on Windows, because both of the above "surely" clauses are correct.

It sounds like the Python community may have adopted the first attitude, while we have historically adopted the second.  (The linked article adopts the first attitude but offers up the second as a workaround.)  There is no use of FILE * in krb5.h or gssapi.h, and every interface we build on Windows is careful to provide functions like krb5_free_unparsed_name() to free memory allocated by other functions in that library.

The first and second approaches can be intermixed as long as the first isn't taken to the absolute.  Python 2.7 modules can all be built with VS2008 and can allocate/free memory across boundaries; at the same time, some of the modules can use the KfW DLLs built with a later version of VS as long as they properly use the API.

It also appears that Microsoft has mitigated the malloc/free aspect of this problem in VS2010+:

http://siomsystems.com/mixing-visual-studio-versions/#comment-11167

That doesn't help the Python 2.7 module ecosystem, but it does substantially reduce the pain on an ongoing basis.

PR #808, opened 2018-07-04
Fix some omissions in make clean

greghudson:
I also looked into modifying travis.yml to detect future omissions in make clean.  I got as far as:

    git ls-files -o -x Makefile -x "*.pc" -x config.log -x config.status -x autoconf.h -x krb5-config

which we would then have to check for non-empty output.  The various -x options could be replaced with -X filename if we wanted to add a file containing the patterns.  Integrating that into the .travis.yml script (which is a long sequence of "a && b && ..." because travis doesn't abort on error) seems a bit messy.


frozencemetery:
For other projects, I've ended up making a shell script which Travis calls out to instead (`sh -ex` etc.) to handle the abort.

PR #810, opened 2018-07-10
Convert Python tests to Python 3

greghudson:
Look for python3 in configure.in and verify that we got it.  Convert
test code to conform to Python 3.


frozencemetery:
Awesome, thank you!

PR #812, opened 2018-07-13
Fix freeing string with secret without zapping memory

Dylan-MSFT:
If you call krb5_c_string_to_key with an encryption type of ENCTYPE_ARCFOUR_HMAC, you will filter down to this function with a password in the "string" input. This password is copied into utf8 and freed without being zeroed. You then end up with a plaintext password floating through unallocated memory.


greghudson:
Merged as commit 5d19f90f3216309c311507acbca2c8d0a286d632.

PR #813, opened 2018-07-16
Eliminate preprocessor-disabled dead code

frozencemetery:
It's been policy for a while now not to create "dead hunks" like
these.  A great deal of this code simply doesn't work because it
hasn't been kept up-to-date, and may never have worked.  Eliminate
these dead hunks along with the complexity to support them.

The instances of this construct (and similar) which are preserved are:
- Many Makefile.in use `!if 0` on DOS to avoid *nix-specific code
- krb5.hin indicates an unused assignment
- getdate.y hides certain timezones, which are visible in source for completeness
- aesopt.h is left alone to match the configuration interface in the source
- def-check.pl and depfix.pl use `\nif 0` as a hack to detect shell vs. the perl interpreter
- profile_tcl.c is a generated file
- ev.c is copied from libev


src/util/support/fnmatch.c
```diff
@@ -33,11 +33,7 @@
  */
 
 #if defined(LIBC_SCCS) && !defined(lint)
-#if 0
```
greghudson: Also leave alone.

src/util/support/getopt.c
```diff
@@ -35,10 +35,6 @@
  */
 
 #if !defined(HAVE_GETOPT) || !defined(HAVE_UNISTD_H)
-#if 0
```
greghudson: And here.

src/util/support/getopt_long.c
```diff
@@ -129,27 +129,6 @@ getopt_internal(nargc, nargv, ostr)
 	return (optopt);			/* dump back option letter */
 }
 
-#if 0
```
greghudson: This is an upstream file which we modify as little as possible.

frozencemetery:
Sounds good, updated.

PR #814, opened 2018-07-16
CI updates for python3

frozencemetery:
- python-cjson seems dead upstream and in particular has no python3 version: https://github.com/AGProjects/python-cjson/issues/6
- Update paste dependency to python3
- Fix pyrad-depdendent tests

### DIFF
--- a/src/tests/jsonwalker.py
+++ b/src/tests/jsonwalker.py
@@ -1,10 +1,5 @@
 import sys
-try:
-    import cjson
-except ImportError:
-    print("Warning: skipping audit log verification because the cjson module" \
-          " is unavailable")
-    sys.exit(0)
+import json
 from collections import defaultdict
 from optparse import OptionParser
 
@@ -72,7 +67,7 @@ class Parser(object):
         """
         Generator that works through dictionary.
         """
-        for a,v in adict.iteritems():
+        for a,v in adict.items():
             if isinstance(v,dict):
                 for (attrpath,u) in self._walk(v):
                     yield (a+'.'+attrpath,u)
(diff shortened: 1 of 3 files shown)


src/tests/jsonwalker.py
```diff
@@ -1,10 +1,6 @@
 import sys
-try:
-    import cjson
-except ImportError:
-    print("Warning: skipping audit log verification because the cjson module" \
-          " is unavailable")
-    sys.exit(0)
+import json
+
```
greghudson: We don't normally put a blank line here.

src/tests/jsonwalker.py
```diff
@@ -93,17 +89,15 @@ def _walk(self, adict):
         with open(options.filename, 'r') as f:
             content = list()
             for l in f:
-                content.append(cjson.decode(l.rstrip()))
+                content.append(json.loads(l.rstrip()))
         f.close()
     else:
-        print('Input file in jason format is required')
+        print('Input file in json format is required')
```
greghudson: Capitalize JSON while we're fixing this.

src/tests/jsonwalker.py
```diff
@@ -93,17 +89,15 @@ def _walk(self, adict):
...
         exit()
 
     defaults = None
     if options.defaults is not None:
-        with open(options.defaults, 'r') as f:
-            defaults = cjson.decode(f.read())
-        f.close()
+        defaults = json.load(open(options.defaults, 'r'))
```
greghudson: Why remove the ``with ...``?  Leaking the file is harmless but there's no reason to add a harmless leak in a fix like this.  (However, the ``f.close()`` was wrong; I'm not sure f is even in scope at that point.)

frozencemetery: Whoops, meant to make one change or the other and did both... will fix, thanks!

frozencemetery:
Updated.  The easiest solution I could think of was modifying krad's Makefile to filter output, but that seemed not the right approach.  Open to suggestion.

PR #818, opened 2018-07-25
Allow u2u requests when KRB5_KDB_DISALLOW_SVR is set

chrishecker:
Here is the change to allow u2u tickets with -allow_svr.  As briefly discussed on the krbdev list, this trues up the code and the error message (the KDC errors for u2u requests with KDC_ERR_MUST_USE_USER2USER, which is confusing).  I also found an old bug for this issue while searching, so I've added that link below.

I've been testing this change in production for years on 1.9, and just recently ported it to the latest code.

One question: should I make the same check in check_tgs_svc_deny_all()?  It looks like it, but I never had to in my testing...I'm not exactly sure what that code is for, it looks like some kind of policy check function table?  It'd be a trivial fix to make obviously.

The other place KRB5_KDB_DISALLOW_SVR is checked is in kdc_get_server_key(), but there's no krb5_kdc_req passed to that function, so I'm not sure how you'd check for KDC_OPT_ENC_TKT_IN_SKEY if it was necessary there.

Thanks, let me know what you think,
Chris

-------

When checking KRB5_KDB_DISALLOW_SVR, also check if it is a u2u request so the KDC_ERR_MUST_USE_USER2USER error message is correct.  In other words, we can still issue u2u tickets for princs that aren't allowed to be normal services.

This fixes Ticket #2641 "KRB5_KDB_DISALLOW_SVR flag unnecessarily prevents User2User"
http://krbdev.mit.edu/rt/Ticket/Display.html?id=2641

### DIFF
--- a/doc/admin/admin_commands/kadmin_local.rst
+++ b/doc/admin/admin_commands/kadmin_local.rst
@@ -297,8 +297,9 @@ Options:
 
 {-\|+}\ **allow_dup_skey**
     **-allow_dup_skey** disables user-to-user authentication for this
-    principal by prohibiting this principal from obtaining a session
-    key for another user.  **+allow_dup_skey** clears this flag.
+    principal by prohibiting others from obtaining a service ticket
+    encrypted in this principal's TGT session key.
+    **+allow_dup_skey** clears this flag.
 
 {-\|+}\ **requires_preauth**
     **+requires_preauth** requires this principal to preauthenticate
@@ -325,7 +326,9 @@ Options:
 
 {-\|+}\ **allow_svr**
     **-allow_svr** prohibits the issuance of service tickets for this
-    principal.  **+allow_svr** clears this flag.
+    principal.  In release 1.17 and later, user-to-user service
+    tickets are still allowed unless the **-allow_dup_skey** flag is
+    also set.  **+allow_svr** clears this flag.
 
 {-\|+}\ **allow_tgs_req**
     **-allow_tgs_req** specifies that a Ticket-Granting Service (TGS)
--- a/doc/admin/conf_files/kdc_conf.rst
+++ b/doc/admin/conf_files/kdc_conf.rst
@@ -134,9 +134,8 @@ The following tags may be specified in a [realms] subsection:
         the principal within this realm.
 
     **dup-skey**
-        Enabling this flag allows the principal to obtain a session
-        key for another user, permitting user-to-user authentication
-        for this principal.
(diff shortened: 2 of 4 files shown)


chrishecker:
> One question: should I make the same check in check_tgs_svc_deny_all()?
Okay, looks like I need to make this change now too, my new build is failing here.  I'll update this.

Chris

chrishecker:
Okay, I updated tgs_policy.c and that fixed it in my tests...I guess that function wasn't there in 1.9 or something.

Chris

greghudson:
As far as I know, AS requests cannot be U2U requests, so I am confused as to how the original patch solved a problem in 1.9.

chrishecker:
I think I see what happened.  In 1.9, both validate_tgs_request and validate_as_request were in kdc_util.c, and both had basically identical KRB5_KDB_DISALLOW_SVR checks, and indeed as you imply my patch was to validate_tgs_request.  When I applied my 1.9 patch to the master kdc_util.c it applied just fine with an offset, but it looks like patch found the validate_as_request check since validate_tgs_request had been moved to tgs_policy.c.  It's my fault for not reviewing it more carefully, it's just the code looked identical at the point of the diff and it built fine, so I just did a quick search for other KRB5_KDB_DISALLOW_SVR checks in the src/kdc directory and mentioned the other one above, and then found I needed that one during my tests.  I should have actually stopped to understand what was going on.  I'm going to remove the AS one, and keep the other, and repush this change.  Sorry about that.  Gimme a minute and I'll fix it.

chrishecker:
Okay, I pushed this change to remove the validate_as_request change, so it's just the single if-statement in tgs_policy.c.  I'm building and testing locally, but it's a subset so it should be okay.  Again, sorry about being so sloppy, great way to start off this group of pull requests, huh?  : )

greghudson:
Better to be consistent with the proximal code, which means not using isflagset().

chrishecker:
No problem, I'll fix it tomorrow and repush.

chrishecker:
Once more, with feeling!

Thanks for the test case, I'll look at how you did that and add them for future fixes I submit!

Thanks,
Chris

greghudson:
Since the commit message referenced ticket 2641 (which is a nice catch; I did a search and didn't come up with anything), I went back and reread that ticket, finding some suggested documentation updates from Patrick Moore.  I added some of those to the commit, with edits.

chrishecker:
Looks good, thanks!

PR #820, opened 2018-07-27
add -v switch to show KVNO for displayed tickets in the ccache

chrishecker:
also fixes a minor extra-space formatting bug with klist -ed

### DIFF
--- a/src/clients/klist/klist.c
+++ b/src/clients/klist/klist.c
@@ -52,7 +52,7 @@
 int show_flags = 0, show_time = 0, status_only = 0, show_keys = 0;
 int show_etype = 0, show_addresses = 0, no_resolve = 0, print_version = 0;
 int show_adtype = 0, show_all = 0, list_all = 0, use_client_keytab = 0;
-int show_config = 0;
+int show_config = 0, show_kvno = 0;
 char *defname;
 char *progname;
 krb5_timestamp now;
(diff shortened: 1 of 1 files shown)


chrishecker:
This one is super minor but sometimes useful.

chrishecker:
Hmm, I just spotted this comment:

>      /* V = version so v can be used for verbose later if desired. */

I could move it to another letter I suppose?

Chris

greghudson:
I guess the first question is: what use case does this satisfy that kvno does not?  kvno will prefer a ticket in the cache to going out and fetching one.

When adding new UI or API features it's usually a good idea to check if Heimdal has anything similar, so that we aren't needlessly incompatible.  In this case it does not, but here is the Heimdal klist usage message:

```
Usage: klist [-ftsT5Alvanh] [--cache=string] [-c string] [--test] [--tokens]
   [--v5] [--all-content] [--list-all] [--json] [--verbose] [--version]
   [--hidden] [--help] 
-c string, --cache=string credential cache to list
-f                        list flags
-t, --test                test for having tickets
-T, --tokens              display AFS tokens
-5, --v5                  display v5 credential tokens
-A, --all-content         List all caches with their content
-l, --list-all            List all caches
--json                    JSON formated output
-v, --verbose             Verbose output
--version                 version
--hidden                  Verbose output
```

So using ``-v`` is contraindicated since Heimdal already uses it for verbose.  We have long option support now (see how it is used in kinit.c), and this seems specialized enough that it probably doesn't need a short option.

Among long option names, ``--kvno`` and ``--show-kvno`` are the two obvious choices.  Since Heimdal's ``-f`` has no corresponding long option, there is really no precedent to fall back on, so we could use either one.

chrishecker:
> I guess the first question is: what use case does this satisfy that kvno does not?
kvno actually gets a ticket if it's not there, though, right?  I needed this for debugging what had ended up in the ccache after running my app and doing some kadmin stuff.

I'm fine with --kvno or --show-kvno, once it's long it can be longer.  If you don't have a preference I'll go with --kvno since it's the klist program and all it does is list, so the "show" is a bit redundant.

Chris

greghudson:
I don't have a preference; ``--kvno`` should be fine.

greghudson:
As for the use case: yes, kvno gets a ticket if it's not already present in the cache.  But you could use klist to see what tickets are in the cache and then kvno to see what version each ticket is.  That's not really satisfactory for any kind of automated script, but for manual purposes it seems adequate.

PR #821, opened 2018-08-01
Extend waiting time from 10s to 30s

alejandro-perez:
There are some GSS mechanisms (such as GSS EAP - RFC 7055) where User interaction is required and 10s is too little to allow users to eg. select the identity they want to use or type their password.

### DIFF
--- a/src/appl/gss-sample/gss-misc.c
+++ b/src/appl/gss-sample/gss-misc.c
@@ -115,7 +115,7 @@ read_all(int fildes, void *data, unsigned int nbyte)
 
     FD_ZERO(&rfds);
     FD_SET(fildes, &rfds);
-    tv.tv_sec = 10;
+    tv.tv_sec = 300;
     tv.tv_usec = 0;
 
     for (ptr = buf; nbyte; ptr += ret, nbyte -= ret) {


greghudson:
I have no objection to raising the gss-sample timeout, or even making it indefinite, but I am curious about the use case.  Is gss-server/gss-client being used with GSS EAP just for testing purposes, or does it fill another role?

alejandro-perez:
Within the Moonshot community gss-server/gss-client is used for testing purposes, as it is allegedly the simplest application using GSS-API that is shipped in most (all?) Linux distributions.

Now I'm testing 2FA and in the time I type the password and the OTP code it has expired most of the times.

I know raising it now won't help with currently shipped Kerberos, but at least will be of help for future ones.

I think that probably having a new command line parameter would be even better, because I guess some times having an early timeout would be better than a long one. Let me try to extend this PR to provide support for that if that sounds ok for you.

greghudson:
Thanks for the explanation.

When would having a low timeout be desirable?

alejandro-perez:
If one is performing automated tests where no interactivity is required, you'd probably want that if gss-server is unresponsive (crash?) gss-client does not wait forever. But right, since this is using TCP it will eventually realise it nonetheless. Hence, I guess that indefinite time would be great.

greghudson:
Well, it might be good for gss-server to clean up dead connections, since the TCP layer won't always recognize that.  But a timeout of 30 minutes or so should be adequate for that purpose.

alejandro-perez:
Yes, even 5 minutes will do. I don't think that any testing user would expect that if a password has not been supplied within 5 minutes the application would be waiting for them. I'm not sure how much SSH waits for the user to provide a password, but surely not 30 minutes :). 

I just think something reasonable that gives you time to maybe create a new identity if none was created yet (something you can do with Moonshot), and write your password, and the application is still there. 5 minutes should be ok.

PR #823, opened 2018-08-03
Stop matching u2u credentials on regular searches

greghudson:
Here is my fix for ticket 8718.  Review it commit-by-commit as the third commit is a style fix (the second commit is minimal but leaves the function in an awkward state).

Chris Hecker recently floated the idea of a kvno option to get user-to-user creds; if we had that, we could write a test case for this fix.  So there is a chance I will defer merging this until he or I implements that kvno option.

### DIFF
--- a/src/lib/krb5/ccache/cc_retr.c
+++ b/src/lib/krb5/ccache/cc_retr.c
@@ -164,6 +164,11 @@ pref (krb5_enctype my_ktype, int nktypes, krb5_enctype *ktypes)
 krb5_boolean
 krb5int_cc_creds_match_request(krb5_context context, krb5_flags whichfields, krb5_creds *mcreds, krb5_creds *creds)
 {
+    /* Only match a user-to-user credential if explicitly asked for, since the
+     * ticket won't work as a regular service ticket. */
+    if (! set(KRB5_TC_MATCH_IS_SKEY) && creds->is_skey)
+        return FALSE;
+
     if (((set(KRB5_TC_MATCH_SRV_NAMEONLY) &&
           srvname_match(context, mcreds, creds)) ||
          standard_fields_match(context, mcreds, creds))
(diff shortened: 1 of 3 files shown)


src/lib/krb5/ccache/cc_retr.c
```diff
@@ -162,41 +159,52 @@ pref (krb5_enctype my_ktype, int nktypes, krb5_enctype *ktypes)
...
 {
-    /* Only match a user-to-user credential if explicitly asked for. */
-    if (! set(KRB5_TC_MATCH_IS_SKEY) && creds->is_skey)
+    krb5_boolean is_skey;
+
+    if ((whichfields & KRB5_TC_MATCH_SRV_NAMEONLY)) {
+        if (!srvname_match(context, mcreds, creds))
+            return FALSE;
+    } else {
+        if (!standard_fields_match(context, mcreds, creds))
```
frozencemetery: `else if`?

src/lib/krb5/ccache/cc_retr.c
```diff
@@ -162,41 +159,52 @@ pref (krb5_enctype my_ktype, int nktypes, krb5_enctype *ktypes)
...
+            return FALSE;
+    }
+
+    /* Only match a user-to-user credential if explicitly asked for, since the
+     * ticket won't work as a regular service ticket. */
```
frozencemetery: This comment is added in the first commit, but gets expanded in the third one.  Probably should just be added wholesale in the first?

src/lib/krb5/ccache/cc_retr.c
greghudson: The idea is to keep the two if statements parallel, but I agree that it causes a double-take.

Thinking about it, maybe there should be a single helper function (match_princs) that accepts whichfields as a parameter. I'll try that out.

greghudson:
Updated; have a look at princs_match() and its call site.  I made the comment in the second commit match the one in the third.

greghudson:
Added a test case to t_u2u.py in the second commit using PR #830.  (Also added the missing success() at the end of t_u2u.py; I botched that earlier.)

PR #824, opened 2018-08-06
asan fixes

greghudson:
A new dev machine means a new version of clang, which means new things detected by clang.  One of these is a harmless-in-practice memory bug in krb5_read_password() which has been present since 2004: if prompt2 is given, k5prompt.reply is pointed at a krb5_data object inside an inner scope, and then `k5prompt.reply->length`` is referenced just after that object goes out of scope.  Since there are no new stack allocations in the interim, it works.  I didn't make the fix minimal since there's no need to backport it.

### DIFF
--- a/src/lib/krb5/os/read_pwd.c
+++ b/src/lib/krb5/os/read_pwd.c
@@ -42,38 +42,37 @@ krb5_read_password(krb5_context context,
                    const char *prompt, const char *prompt2,
                    char *return_pwd, unsigned int *size_return)
 {
-    krb5_data reply_data;
-    krb5_prompt k5prompt;
+    krb5_data reply_data, verify_data = empty_data();
+    krb5_prompt k5prompt, vprompt;
     krb5_error_code retval;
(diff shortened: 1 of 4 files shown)


src/lib/krb5/os/read_pwd.c
```diff
@@ -42,38 +42,37 @@ krb5_read_password(krb5_context context,
...
-        retval = krb5_prompter_posix(NULL,
-                                     NULL,NULL, NULL, 1, &k5prompt);
-        if (retval == 0) {
-            /* compare */
-            if (strncmp(return_pwd, (char *)verify_data.data, *size_return))
+    retval = krb5_prompter_posix(NULL, NULL, NULL, NULL, 1, &k5prompt);
+
+    if (retval == 0 && prompt2 != NULL) {
+        retval = alloc_data(&verify_data, *size_return);
+        if (retval) {
```
frozencemetery: If you're already doing a modernize commit, can you flatten the control flow using something like `goto done`?

PR #825, opened 2018-08-08
Add kdestroy -p option

greghudson:
[Requested by simo.]

Add an option to destroy a cache within a collection by principal
name.  This option can be used together with -c to specify the
collection.  Also document that kdestroy -A and -c can be used
together (ticket 8602).

### DIFF
--- a/src/clients/kdestroy/kdestroy.c
+++ b/src/clients/kdestroy/kdestroy.c
@@ -49,10 +49,12 @@ char *progname;
 static void
 usage()
 {
-    fprintf(stderr, _("Usage: %s [-A] [-q] [-c cache_name]\n"), progname);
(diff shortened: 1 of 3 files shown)


simo5:
Looks great!
Thanks!

greghudson:
I meant to call out that as implemented in the PR, kdestroy -p suppresses the "Other credential caches present" warning, on the theory that if you're searching for a specific client principal you probably know that already.

src/clients/kdestroy/kdestroy.c
```diff
@@ -122,6 +134,11 @@ main(int argc, char *argv[])
         }
     }
 
+    if (all && princ_name != NULL) {
+        fprintf(stderr, _("-a option is exclusive with -p option\n"));
```
frozencemetery: Capitalized - `-A`

greghudson: Fixed, thanks.

PR #826, opened 2018-08-08
Update many documentation links to https

greghudson:
[web.mit.edu is supposed to change to not require client certs for https in two days.  I will not merge this until that cutover has finished.]

Use secure URLs in the documentation where possible.  The Sphinx web
site does not appear to have a valid server cert, but does have a more
official URL.


greghudson:
The web.mit.edu cutover to anonymous https has occurred, so it should now be possible to verify these links.

PR #827, opened 2018-08-10
Hash table for memory ccache resolution

greghudson:
We got a request to use a hash table for resolving MEMORY ccaches so that they perform reasonably when a process contains many of them.  I could only find one other use of an in-memory hash table in the code.  Two uses is enough to justify a libkrb5support facility, so I added one.

I discovered that murmurhash3 isn't all that resistant to hash flooding attacks and that general-purposes languages with dictionaries now use siphash.  There some liberally-licensed implementations of siphash, but it's also easy to implement from the paper, and that way I could tailor the implementation to fit our platform primitives and the hash table code.

The hash table code is dead simple, partly because with only two callers it doesn't have to handle very diverse requirements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -559,6 +559,7 @@ local.properties
 
 /src/util/support/libkrb5support.exports
 /src/util/support/t_base64
+/src/util/support/t_hashtab
 /src/util/support/t_hex
 /src/util/support/t_json
 /src/util/support/t_k5buf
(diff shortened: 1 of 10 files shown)


greghudson:
Okay, I made k5_hashtab_remove() return true or false.  And, good catch on the incorrect ``#include`` in the second commit; I fixed that.

PR #836, opened 2018-08-30
Update cmocka to version 1.1.2 in Travis.

mmattioli:


greghudson:
Relevant history:

* Updating the cmocka version came up in PR #647.  At that time Robbie and I decided to leave the version at 1.0.1.

* Several months later, our Travis builds started failing because 1.0.1 would no longer build.  I updated the version to 1.1.1 without a PR (and without using the @mmattioli commit; I guess I had forgotten about PR #647 at that time).  I unfortunately didn't include a justification in the commit message.  I dug up this IRC conversation from the krbdev channel logs (2017-07-17):

```
[15:45] <ghudson> Our Travis builds started erroring because of a new distribution or compiler throwing warnings in cmocka 1.0.1.
[15:45] <ghudson> If that's the only problem, I will bump the cmocka version in .travis.yml to 1.1.1.
[15:45] <ghudson> (Without review.)
[15:45] <ghudson> If there are other problems I will file a PR.
[15:47] <ghudson> (I think the difference is Trusty vs. Precise, as Travis is finally catching up to 2014.)
[16:14] <rharwood> seems fine
[16:14] <rharwood> any chance we could move to the packaged version?
[16:15] <rharwood> mmm, I guess that's 1.0.1, so probably not
[16:20] <ghudson> Well, we wouldn't be compiling it, so we wouldn't run into compiler warnings treated as errors.
[16:21] <ghudson> So if there's a packaged version in Trusty, that would be fine.
[16:21] <rharwood> ugh right we're on trusty
[16:21] <rharwood> not usefully
```

The arguments in PR #647 still apply, I think; we don't need to bump the cmocka version in Travis builds solely because a new version is available.

mmattioli:
My reasoning was the two bug fixes in the [changelog](https://git.cryptomilk.org/projects/cmocka.git/diff/ChangeLog?h=cmocka-1.1.2&id=8ed891eb285dd1fc62020fc2e0b1b684591b3a60):

- Fixed fixture error reporting
- Fixed compiler flags detection

If neither of these are issues which affect how cmocka is currently used then I suppose it can be left alone.

frozencemetery:
I don't have an objection to those as reasons.  Just wanted it in the commit message :)

mmattioli:
@frozencemetery fixed 😄

SaharahSarah:
It would probably be a good idea to have the commit message mention specifically the problem krb5 was having with Travis, mainly the fact that the Travis build was failing due to compiler errors.  Also, please following the [krb5 version control practices for the commit message](http://k5wiki.kerberos.org/wiki/Coding_style/Version_control_practices). For instance, by removing the period in the subject etc.

mmattioli:
@SaharahSarah ...third time's the charm?

mmattioli:
Any other changes needed?

greghudson:
Sorry to let this back-burner.

The issue we were having with Travis builds with 1.0.1 was a combination of us specifying -Werror in the build matrix for clang, the warning flags chosen by cmocka's build system (correctly, as far as I know), and some minor defect in the cmocka code.  Whatever issues with detecting compiler flags are corrected in 1.1.2, they don't affect the Travis build.

mmattioli:
I suppose I'll close this PR since these changes are irrelevant.

PR #839, opened 2018-09-03
Get password expiration through api

mrandybu:
This function allow get the expiration time of the kerberos password using api. I use it in krb5-ticket-watcher to notify the user in advance that the approaching expiration of the password. It will help avoid problems with the sudden disconnection of domain's services.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -7266,6 +7266,19 @@ krb5_get_init_creds_opt_set_responder(krb5_context context,
                                       krb5_responder_fn responder, void *data);
 
 /**
+* Get the lifetime remaining of kerberos password.
+* @param [in]  context         Library context
+* @param [in]  client          Client principal
+* @param [in]  password        Password
+* @param [out] rem_pw_exp_time Lifetime remaining of password
+*
+* This function allow get the lifetime remaining of kerberos password.
+*/
+krb5_error_code KRB5_CALLCONV
+krb5_get_pw_exp(krb5_context context, krb5_principal client,
+                const char *password, krb5_timestamp *rem_pw_exp_time);
+
+/**
  * Get initial credentials using a password.
  *
  * @param [in]  context         Library context
--- a/src/lib/krb5/krb/gic_pwd.c
+++ b/src/lib/krb5/krb/gic_pwd.c
@@ -582,3 +582,50 @@ krb5_get_in_tkt_with_password(krb5_context context, krb5_flags options,
             return (retval);
     return retval;
 }
+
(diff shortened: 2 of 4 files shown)


greghudson:
Is krb5_get_init_creds_opt_set_expire_callback() not adequate for this use case?

mrandybu:
krb5_get_init_creds_opt_set_expire_callback is not very suitable for my task. I would like to receive  the remaining lifetime of the password, not the expiration date. And use it when checking the lifetime of tickets. I think that this function should work separately, without using krb5_get_init_creds_password, like as krb5_get_init_creds_opt_set_expire_callback.

greghudson:
How does the remaining lifetime of the password differ from the expiration date?   (Sorry if I'm missing something; just trying to do due diligence before adding a new public API.)

mrandybu:
I did not correctly put it. I mean, not the date when the password lifetime ends, but the time of life, that is, how many milliseconds are left before the end date. it is convenient to use to warn the user, for example for a few days, and invite him to change his password. I hope this is a more understandable description :)

greghudson:
So given the expiration date, the lifetime can be obtained just by subtracting the current time, right?

My question is whether this new API adds a new capability for the application, or only a convenience.  If an application can obtain the password expiration lifetime using krb5_get_init_creds_opt_set_expire_callback() and krb5_get_init_creds_password(), I would rather not add a new public API that does essentially that.

mrandybu:
Well, how can I at any time get a timestamp for the expiration of the password,
using only krb5_get_init_creds_opt_set_expire_callback and krb5_get_init_creds_password?

greghudson:
You would create a krb5_get_init_creds_opt object with krb5_get_init_creds_opt_alloc(), set an expire callback, call krb5_get_init_creds_password(), and record the values passed to the callback.  Then throw away the creds without storing them if you don't want them.  This is essentially what the proposed new API does, albeit more verbose.  It is also more flexible: it can work when getting initial creds requires more information for pre-authentication than just the password (by supplying a prompter), it can retrieve the account expiration time as well as the password expiration time, and it can get the password expiration time at the same time as storing credentials if that is desired.

In the context of krb5-ticket-watcher, you could make v5::initCredential() record the password expiration time, so that the expiration time is updated each time a new ticket is requested.  I don't know if that is enough for your requirements.  (As long as you are creating a krb5_get_init_creds_opt object, you could also use krb5_get_init_creds_opt_set_out_ccache() and then remove the code to explicitly store the credential after krb5_get_init_creds_password() succeeds.)

mrandybu:
Yes, I managed the standard methods, of course it turned out more cumbersome, but still works, thank you (:

PR #843, opened 2018-09-16
Some S4u2self related fixups

frenche:


greghudson:
We don't use the realm from the result of convert_to_enterprise() (see line 581, `sprinc = *eprinc;` followed by `sprinc.realm = tgtptr->server->data[1];`), so we shouldn't need to copy in the original realm there.

frenche:
Ok, changed to just ignore the realm.

btw, in case the server name is an SPN (or already enterprise) we probably don't need to convert to enterprise (I have not tested yet, just thinking).

frenche:
Commenting on my comment; I've tested and while SPN seem to work ok with forest-trust, the enterprise way has apparently the advantage of working with external-trust too (I wonder where are these details documented by MS).

greghudson:
If it's okay, I will rewrite the test case to be in tests/gssapi/t_s4u.py, so that it's not in a script about authdata tests.  I will try to make it do an actual cross-realm request (the request will fail since we don't have the necessary KDC support, but we can check trace logs to verify that the client is doing the right thing) and to start with a non-enterprise principal name.  That way the test doesn't depend on convert_to_enterprise() being called when it isn't necessary.

frenche:
Sure, please, I think it could be a good basis for further development in this area (sorry for the wrong location of the test, I was looking for an invocation of 'kvno -U' to reuse).

greghudson:
I pushed revised commits.  Turns out there was already a cross-realm S4U2Self test in t_s4u.py so I only needed to add the default_realm removal to it.

frenche:
Oh, looks good.. I guess there is no point having an additional test with default realm.
In the commit message, maybe "spurious error if *a default* one.." reads better.

btw, I'm curious what KDC support (roughly) is missing to make it work (say without PAC).

greghudson:
I edited the commit message to be clearer about what hypothetically isn't configured.

I'm not sure what all is missing in the KDC support.  On the remote realm (r2 in this test) we need search_sprinc() to issue a referral.  Maybe it can do that whenever it sees an enterprise principal name in the server principal in a TGS request?  I'm not confident of that; we might need to also recognize the presence of S4U2Self pa-data in the request.  For the final query back on the local realm, I don't know offhand exactly what the KDC sees and whether it needs a PAC to securely fulfill the S4U2Self request.

frenche:
Thanks for your thoughts.
Regarding search_sprinc(), note that I didn't have to change anything in samba's get_principal() to get a the referral, either when the service principal was an spn or an enterprise-name, so I guess that could be set up.

PR #849, opened 2018-09-25
Fix getdate.y compatability with POSIX yacc

frozencemetery:
Commit 28fd0a934cdc7b3b42ce213c6d334d4edf1ab591 muted a warning from bison
about shift/reduce conflicts in the grammer.  However, the extension for
suppressing the warning is bison-only.  Revert that portion of the change and
live with the warning rather than adding additional conditionalization.
Reported by Michael Osipov.

### DIFF
--- a/src/kadmin/cli/getdate.y
+++ b/src/kadmin/cli/getdate.y
@@ -176,8 +176,9 @@ static time_t	yyRelSeconds;
 
 %}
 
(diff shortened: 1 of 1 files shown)


frozencemetery:
(Updated to Michael's suggestion of leaving the `%expect` commented.)

PR #851, opened 2018-09-27
Use "replica" for downstream kprop/iprop KDCs

greghudson:
All of the functional changes (just two, and they are pretty innocuous) are in the second commit.  The first commit is documentation and internal name changes.


frozencemetery:
There's a reference in LICENSE to kpropd_rpc.c which has the wrong path.  Other than that (and translations), it looks good to me.

greghudson:
In lib/gssapi/LICENSE?  That's a Sun license grant, referring to the path in the Solaris source tree.

frozencemetery:
Oh, I see.  That's fine then.

PR #852, opened 2018-09-29
Let sign and verify PAC including the realm component

frenche:
This is needed for KDC handling of cross-realm S4U2Self tickets,
see MS-SFU 3.2.5.x (only KDB plugins should need it).

Note, we currently do not allow to re-sign a PAC (that was
signed without realm) with realm, but only to create a new one.

See discussion at:
http://mailman.mit.edu/pipermail/krbdev/2018-September/012992.html


frenche:
I've added more torture like tests for S4U2Self PAC combinations, enterprise and cross-realm, using PACs generated by Windows 2008 KDC for the same user principal. The PACs were collected by printing them and their keys from the service and kdc code, see (not for the faint-hearted).
https://github.com/frenche/krb5/commit/dc99b3a05b985f21a33cd12a2ef4a7849c0932f3

frenche:
I've pushed a slight update to the second commit, to print the index of the PAC being checked in case of failure.

greghudson:
Revised.  We have some _ext() functions in the API and no _ex() functions, so I renamed the new ones.  We generally keep output parameters at the end, so I moved the new flag argument one position back in krb5_pac_sign_ext().  I also added documentation and Windows export directives.

An alternative would be to name the new functions krb5_pac_sign_realm() and krb5_verify_pac_realm() and remove the flag parameters, with the realm behavior being always-on for the new functions.  I have no strong opinion on whether that's better.

frenche:
> An alternative would be to name the new functions krb5_pac_sign_realm() and krb5_verify_pac_realm() and remove the flag parameters, with the realm behavior being always-on for the new functions. I have no strong opinion on whether that's better.

That was my initial take, I changed it to make it easier for the caller to just pass the boolean instead of conditioning on it what function to use.

Other then that it look ok, though I'd like to add a comment on why KRB5_PRINCIPAL_UNPARSE_DISPLAY is used, and that perhaps we should use it generally if another use case arises (I don't have any atm).
I'm in between flights right now so I'll add it later.

frenche:
I've added a comment explaining the why unparse_display is needed for enterprise name with realm.

PR #853, opened 2018-10-02
Allow referrals for S4U2Self requests

frenche:
as needed for cross realm protocol-transition to work.
See MS-SFU 3.2.5.1.1 for details.

This has been previously explicitly disabled in order
to preserve old behaviour which seem like a bug.

See discussion at:
http://mailman.mit.edu/pipermail/krbdev/2018-September/012992.html

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -269,6 +269,8 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
     errcode = kdc_process_s4u2self_req(kdc_active_realm,
                                        request,
                                        header_enc_tkt->client,
+                                       header_ticket->server,
+                                       is_referral,
                                        server,
                                        subkey,
                                        header_enc_tkt->session,
@@ -288,16 +290,8 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
 
     if (errcode)
         goto cleanup;
-    if (s4u_x509_user != NULL) {
+    if (s4u_x509_user != NULL)
         setflag(c_flags, KRB5_KDB_FLAG_PROTOCOL_TRANSITION);
-        if (is_referral) {
-            /* The requesting server appears to no longer exist, and we found
-             * a referral instead.  Treat this as a server lookup failure. */
-            errcode = KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN;
-            status = "LOOKING_UP_SERVER";
-            goto cleanup;
-        }
-    }
 
     /* Deal with user-to-user and constrained delegation */
     errcode = decrypt_2ndtkt(kdc_active_realm, request, c_flags,
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1441,6 +1441,8 @@ krb5_error_code
 kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
                          krb5_kdc_req *request,
                          krb5_const_principal client_princ,
+                         krb5_const_principal header_srv_princ,
+                         krb5_boolean issuing_referral,
                          const krb5_db_entry *server,
                          krb5_keyblock *tgs_subkey,
                          krb5_keyblock *tgs_session,
@@ -1450,6 +1452,7 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
                          const char **status)
 {
     krb5_error_code             code;
+    krb5_boolean                is_local_tgt;
     krb5_pa_data                *pa_data;
     int                         flags;
     krb5_db_entry               *princ;
@@ -1541,6 +1544,14 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
     if (request->kdc_options & AS_INVALID_OPTIONS) {
         *status = "INVALID AS OPTIONS";
         return KRB5KDC_ERR_BADOPTION;
+    }
+
+    is_local_tgt = !is_cross_tgs_principal(header_srv_princ);
+    if (is_local_tgt && issuing_referral) {
+        /* The requesting server appears to no longer exist, and we found
+         * a referral instead.  Treat this as a server lookup failure. */
+        *status = "LOOKING_UP_SERVER";
+        return KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN;
     }
 
     /*
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1546,6 +1546,19 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
         return KRB5KDC_ERR_BADOPTION;
     }
 
+    /*
+     * Valid S4U2Self requests can occur in the following combinations:
+     *
+     * (1) local TGT, local user, local server
+     * (2) cross TGT, local user, issuing referral
+     * (3) cross TGT, non-local user, issuing referral
+     * (4) cross TGT, non-local user, local server
+     *
+     * The first case is for a single-realm S4U2Self scenario; the second,
+     * third, and fourth cases are for the initial, intermediate (if any), and
+     * final cross-realm requests in a multi-realm scenario.
+     */
+
     is_local_tgt = !is_cross_tgs_principal(header_srv_princ);
     if (is_local_tgt && issuing_referral) {
         /* The requesting server appears to no longer exist, and we found
@@ -1561,6 +1574,13 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
                            (*s4u_x509_user)->user_id.user)) {
         krb5_db_entry no_server;
         krb5_pa_data **e_data = NULL;
+
+        if (!is_local_tgt && !issuing_referral) {
+            /* A local server should not need a cross-realm TGT to impersonate
+             * a local principal. */
+            *status = "NOT_CROSS_REALM_REQUEST";
+            return KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN; /* match Windows error */
+        }
 
         code = krb5_db_get_principal(kdc_context,
                                      (*s4u_x509_user)->user_id.user,
@@ -1584,6 +1604,14 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
         }
 
         *princ_ptr = princ;
+    } else if (is_local_tgt) {
+        /*
+         * The server is asking to impersonate a principal from another realm,
+         * using a local TGT.  It should instead ask that principal's realm and
+         * follow referrals back to us.
+         */
+        *status = "S4U2SELF_CLIENT_NOT_OURS";
+        return KRB5KDC_ERR_POLICY; /* match Windows error */
     }
 
     return 0;
--- a/src/kdc/kdc_util.h
+++ b/src/kdc/kdc_util.h
@@ -269,6 +269,8 @@ krb5_error_code
 kdc_process_s4u2self_req (kdc_realm_t *kdc_active_realm,
                           krb5_kdc_req *request,
                           krb5_const_principal client_princ,
+                          krb5_const_principal header_srv_princ,
+                          krb5_boolean issuing_referral,
                           const krb5_db_entry *server,
                           krb5_keyblock *tgs_subkey,
                           krb5_keyblock *tgs_session,


src/kdc/do_tgs_req.c
```diff
@@ -290,12 +290,21 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
...
-            goto cleanup;
+
+        if (!is_cross_tgs_principal(header_ticket->server)) {
+            /* If the tgt is local, then both the impersonated
+             * client and the server must be local */
+            if (is_referral || client == NULL) {
+                errcode = KDC_ERR_POLICY;
+                goto cleanup;
+            }
+        }
```
frozencemetery: We cuddle `else`

src/kdc/do_tgs_req.c
```diff
@@ -290,12 +290,21 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
...
+        else {
+            /* Otherwise, atleast one of them must be foreign */
+            if (!is_referral && client != NULL) {
```
frozencemetery: Combine this into an `else if`

src/kdc/do_tgs_req.c
```diff
@@ -290,12 +290,21 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
...
-            /* The requesting server appears to no longer exist, and we found
-             * a referral instead.  Treat this as a server lookup failure. */
-            errcode = KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN;
-            status = "LOOKING_UP_SERVER";
-            goto cleanup;
+
+        if (!is_cross_tgs_principal(header_ticket->server)) {
+            /* If the tgt is local, then both the impersonated
+             * client and the server must be local */
+            if (is_referral || client == NULL) {
```
frozencemetery: Combine this with the parent `if` conditional

src/kdc/do_tgs_req.c
frenche: Ok, I'll change.

src/kdc/do_tgs_req.c
frenche: Duh, will fix.

src/kdc/do_tgs_req.c
frenche: That would change the meaning of the 'else' to include the case of local tgt && (!is_referral  && client  != NULL), which is a valid case that I think will break.

Apart of that, I think I better understand the previous comment, and will rather preserve the error for the is_referral case.

frenche:
I've updated the commit.
Note about the third case (the else-if), I think it is like the protocol-transition equivalent of the lineage check we do in kdc_process_tgs_req(), which is rightly skipped there for S4U2Self, see:
https://buildfarm.opencsw.org/source/xref/krb5/src/kdc/kdc_util.c#307

greghudson:
`errcode` is expected to hold a com_err constant, not a k5-int.h protocol error code.  Set it to KRB5KDC_ERR_POLICY, not KDC_ERR_POLICY.  (Some code added in 2013 compares `errcode` to KDC_ERR_* constants; this code does not work, but no one has noticed since it only affects the audit interface.)

(I still need to think carefully about the logic here.  This change also expands the already enormous process_tgs_req() by about 20 lines and therefore invites refactoring; I will think about whether there is a reasonable opportunity to do that.)

frenche:
I fixed the error code. Note that the error looked ok in wireshark, same as Windows reply. Looking at the code now, I'd guess I just copied the error from a couple of lines above it, where it is used for errcode..

greghudson:
> Note that the error looked ok in wireshark, same as Windows reply.

That's surprising.  In the cleanup handler, we do:

        errcode -= ERROR_TABLE_BASE_krb5;
        if (errcode < 0 || errcode > KRB_ERR_MAX)
            errcode = KRB_ERR_GENERIC;

so an original value of KDC_ERR_POLICY (12) would seem to result in KRB_ERR_GENERIC (69) on the wire.

frenche:
Ah, that's right. I think I might been comparing the same capture file, no wonder it was a perfect match.

I now see that we also send e-text "UNKNOWN_REASON" when status isn't set, so maybe worth setting it.

frenche:
I split the commit into three for now, as each is a change with its own logic.
Also updated it slightly - more below.

I've been a bit hasty with two other changes, trying to limit to what's necessary for the protocol to work, but I didn't check what happens in these cases before my changes, and how windows KDC behaves in the third case.

Note: my lab consists of a Windows domain ACME.COM and a Samba one (with MIT krb5 backend) MIT.NET,  the two connected with forest trust.

About the second commit; if we do not error out as before the commit, then it fails with KRB5KDC_ERR_GENERIC (e-text: HANDLE_AUTHDATA), due to pac signature mismatch (issued by my samba plugin) same as I have on heimdal (for now), while Windows return KRB5KDC_ERR_POLICY.
```log
kdc log:
Oct 04 17:30:22 f krb5kdc[30082](info): TGS_REQ : handle_authdata (-1765328250)
Oct 04 17:30:22 f krb5kdc[30082](info): TGS_REQ (3 etypes {18 17 23}) 192.168.47.101: HANDLE_AUTHDATA: authtime 1538663422,  FDC$@MIT.NET for FDC$@MIT.NET, Principal FDC$ is missing required realm
```

As regarding the third commit; before the change we'd fail with KRB5KDC_ERR_POLICY (e-text: BAD_TRANSIT), while Windows surprisingly fails with: KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN.
```log
kdc log:
Oct 04 17:18:49 f krb5kdc[30082](Error): unexpected error checking transit from 'FDC$@MIT.NET' to 'FDC$@MIT.NET' via 'ACME.COM': Cannot find ticket for requested realm
Oct 04 17:18:49 f krb5kdc[30082](info): TGS_REQ (6 etypes {18 17 20 19 16 23}) 192.168.47.101: BAD_TRANSIT: authtime 1538660028,  FDC$@MIT.NET for FDC$@MIT.NET, KDC policy rejects request
```


The way I was able to test the third case, was with heimdal's  kgetcred tool, by first doing kinit to one realm, say ACME.COM, and then by running:
```code
$ kgetcred --out-cache=out krbtgt/ACME.COM@MIT.NET
```
Which makes the tool first to ask 'krbtgt/MIT.NET@ACME.COM' from ACME's KDC, and then 'krbtgt/ACME.COM@MIT.NET' from MIT's KDC.
This works fine (both ways, Windows and Samba), but then I use the cross-realm ticket in 'out' cache to ask a s4u2self ticket from my own realm, with:
```code
$ KRB5CCNAME=out kgetcred --impersonate=administrator@ACME.COM FDC\$@ACME.COM
```

greghudson:
I would like to explore moving all three checks into kdc_process_s4u2self_req(), perhaps passing more information to that function, instead of adding more verbiage to process_tgs_req().  In hindsight the check added in commit 8a9909ff9ef6b51c5ed09ead6713888fbb34072f should possibly have been added to kdc_process_s4u2self_req() instead.

frenche:
I think it's a good idea, as we may eventually want to use the is_referral flag in kdc_process_s4u2self_req(), in order to possibly skip the "is-self" check (to allow more flexible matching based on DB lookup).

I moved the three checks to kdc_process_s4u2self_req(), hope it looks better.

greghudson:
I sat down and wrote down the four cases of valid S4U2Self requests from the KDC's perspective:

* single realm: local TGT, no referral, local user
* cross-realm initial: cross TGT, referral, local user
* cross-realm intermediate: cross TGT, referral, non-local user
* cross-realm final: cross TGT, no referral, non-local user

The first and third proposed changes seem to impose correct constraints: if the header ticket contains a local TGT, we need to have found the requested server principal (not a referral) and the S4U2Self pa-data must name a local user.

The second proposed change seems like it would break the final request in a cross-realm S4U2Self sequence.  For this request, the server queries its own KDC, but its header ticket is cross-realm (the referral ticket issued by the client realm or an intermediate realm).  There may be a more specific constraint to apply.

Here are fuller notes on the four cases:
```
single realm (at KDC for REALM)
header ticket is local TGT: server@REALM -> krbtgt/REALM@REALM
server entry (not referral): server@REALM
S4U2Self padata: user@REALM (local)

cross-realm initial (at KDC for UREALM):
header ticket is cross TGT: server@SREALM -> krbtgt/UREALM@SREALM
sname = NT_ENTERPRISE server@SREALM
server entry (referral): krbtgt/SREALM@UREALM (or krbtgt/IREALM@UREALM)
S4U2Self padata: user@UREALM (local)

cross-realm intermediate (at KDC for IREALM):
header ticket is cross TGT: server@SREALM -> krbtgt/IREALM@UREALM
sname = NT_ENTERPRISE server@SREALM
server entry (referral): krbtgt/SREALM@IREALM
S4U2Self padata: user@UREALM (not local)

cross-realm final (at KDC for SREALM):
header ticket is cross TGT: server@SREALM -> krbtgt/SREALM@UREALM
                                         (or krbtgt/SREALM@IREALM)
sname = server
server entry (not referral): server
S4U2Self padata: user@UREALM (not local)
```

frenche:
> The second proposed change seems like it would break the final request in a cross-realm S4U2Self sequence. For this request, the server queries its own KDC, but its header ticket is cross-realm (the referral ticket issued by the client realm or an intermediate realm).

The location of this change is where we know that the impersonated user *is* local, and therefore we can conclude that a cross realm TGT should not have been used (the condition is less obvious than before, but the comment emphasizes it). And it works in my manual tests, i.e. a server from my Samba MIT realm can impersonate a user from ACME.COM realm.

frenche:
Hmm, I tried to make sure not to leak (and not to have to free unnecessarily), but I think I missed it in the second change since I have to db_free the user 'princ' entry.
I'll move the change a little bit upwards to avoid having to free.

greghudson:
Revised.  (I picked up the change to avoid the memory leak before making changes.)  I added a comment describing the four valid combinations to the commit adding the new constraints.

frenche:
Looks good, much clearer, thanks!
In the first commit message, 'explicitly prefers' should read 'explicitly prevents' I suppose.

greghudson:
Fixed, thanks.

greghudson:
[Side note: I'm now kind of curious why the client doesn't just set srealm to krbtgt/SERVER_REALM in the first place.  Can you verify that Windows clients do that for cross-realm S4U2Self?  It don't think it's because the KDC needs to know the eventual server name, since that is in the header ticket client.  It could be a negotiation hint, in that we otherwise can't really tell whether the client realm's KDC supports S4U2Self and included the necessary PAC in the referral TGT.]

I think that logic eventually belongs in the core KDC code, so that the KDC can locate an alternate TGS (using capaths etc.) if the one indicated by the server principal isn't in the KDB.  I don't really want to make this PR conditional on doing that work, though.

For this test case the test KDB module needs to be augmented to support server referrals (kdb_test.c:334), also needs a way to set cross-realm keys that can be shared with another KDB.  Implementing stackable KDB modules would be much harder, and isn't necessarily a direction I'm comfortable with.

frenche:
> Side note: I'm now kind of curious why the client doesn't just set srealm to krbtgt/SERVER_REALM in the first place. Can you verify that Windows clients do that for cross-realm S4U2Self?

Yea, that's what Windows client appears to be doing. See capture generated by Windows client at (the first request is local, the second one is a cross realm):
https://github.com/frenche/krb5/blob/hack_extract_pacs/win_s4u2self.pcap

It was collected by logging in to a domain member Win7 machine with a user that has been assigned an SPN (thus has 'delegation' tab in ADUC), and by using powershell as follows:
```log
C:\Users\apache>powershell
Windows PowerShell
Copyright (C) 2009 Microsoft Corporation. All rights reserved.

PS C:\Users\apache> $Null = [Reflection.Assembly]::LoadWithPartialName('System.IdentityModel')
PS C:\Users\apache> $Ident = New-Object System.Security.Principal.WindowsIdentity @('ayala@acme.com')
PS C:\Users\apache> $Ident = New-Object System.Security.Principal.WindowsIdentity @('nava@cdom.acme.com')
```

I found the powershell command at (after failing to achieve it using native LsaLogonUser()):
http://www.harmj0y.net/blog/activedirectory/s4u2pwnage/

> It don't think it's because the KDC needs to know the eventual server name, since that is in the header ticket client. It could be a negotiation hint, in that we otherwise can't really tell whether the client realm's KDC supports S4U2Self and included the necessary PAC in the referral TGT.

One possible explanation, is in order to allow for the remote KDC to do the 'is-self' check, but I think it is really unlikely.

> I think that logic eventually belongs in the core KDC code, so that the KDC can locate an alternate TGS (using capaths etc.) if the one indicated by the server principal isn't in the KDB. I don't really want to make this PR conditional on doing that work, though.

Ok (note that when I tried that initially, I broke Samba plugin quite badly, so we'd need to be careful).

> For this test case the test KDB module needs to be augmented to support server referrals (kdb_test.c:334), also needs a way to set cross-realm keys that can be shared with another KDB.

I'm looking into this, thanks once again for the leads!

frenche:
One more comment about the side-note; MS-SFU 3.1.5.1.1.1 implies the same, that the service should ask a ticket for self from the remote KDC, not a cross-realm TGT:
Using the TGT to the TGS in the user's realm, Service 1 requests a service ticket to *itself*.

greghudson:
I will go ahead and merge this without automated tests, although I'd be happy to see automated tests for this in a future PR.

frenche:
Thanks! Still working on the tests but I'm being slow right now, will update asap.

>> It don't think it's because the KDC needs to know the eventual server name, since that is in the header ticket client.
>
> One possible explanation, is in order to allow for the remote KDC to do the 'is-self' check, but I think it is very unlikely.

I was curious to test asking a krbtgt explicitly from the remote KDC, so I have change the client code to do so (like this: https://github.com/frenche/krb5/commit/746be73a97d0e3afd63aa81a90d19bd98aff5979), and got KRB_AP_ERR_BADMATCH error from the remote KDC.

According to the below MS resource, this error may mean "There is an account mismatch during protocol transition", see:
https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4769

So I think I might been wrong to assume that the "is-self" check can be skipped on the remote KDC.
I'll try perhaps to consult with MS's dochelp and CC krbdev (if no objection).

>> It could be a negotiation hint, in that we otherwise can't really tell whether the client realm's KDC supports S4U2Self and included the necessary PAC in the referral TGT.

I read this line again today, and I'm not sure I actually understand how it would be a hint for that. I should note that getting a referral for an enterprise sname works regardless of protocol transition. I have re-verified it using kgetcred as such:
$ kgetcred --name-type=ENTERPRISE_PRINCIPAL apache\\@acme.com@ACME.COM

otoh, the realm component in the PAC client info field of s4u2self referrals may be a hint for that.

greghudson:
dochelp usually requires a fair amount of back and forth, especially for something like this, so I would suggest asking privately (CCing me is fine) and then summarizing to krbdev.

To explain what I meant about negotiation hints: since S4U2Self uses pa-data, there is always the possibility that an old KDC just ignores the pa-data and processes the request like a standard KDC would.  The client should be able to detect this in order to issue a meaningful diagnostic.  For the local realm case, it is easy to recognize when this happens--you get a ticket to yourself instead of to the requested users.  For the foreign realm case, it is also easy to recognize this case under the MS protocol as written.  But if the client asked for krbtgt/SERVER-REALM, it would be hard to recognize this case since you can't look inside the referral TGT to see what's in the PAC.

frenche:
> dochelp usually requires a fair amount of back and forth, especially for something like this, so I would suggest asking privately (CCing me is fine) and then summarizing to krbdev.

Ok thanks, will do.

> For the foreign realm case, it is also easy to recognize this case under the MS protocol as written. 

I think that would be true only if we assume that if the KDC knows how to handle enterprise names and to issue referrals, then it also knows how to process pa-data for_user. Otherwise, if we do not assume that then the response from such KDC (that knows enterprise names and referrals but not pa-data for_user) would be exactly the same, that is a referral TGT to the realm of the server and the client has still no way to verify.

PR #855, opened 2018-10-04
Fix 64-bit Windows socket write error handling

greghudson:
Add casts to ensure that the result type of SOCKET_WRITEV() on Windows
can represent -1.  Otherwise it will be treated as 2^32-1 when cast to
ssize_t on 64-bit Windows.

### DIFF
--- a/src/include/port-sockets.h
+++ b/src/include/port-sockets.h
@@ -40,8 +40,9 @@ typedef WSABUF sg_buf;
  */
 /* WSASend returns 0 or SOCKET_ERROR.  */
 #define SOCKET_WRITEV_TEMP DWORD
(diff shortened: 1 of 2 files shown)


greghudson:
Oops, I meant to write ssize_t and mis-entered it when I retyped the change.  I will fix it to ssize_t, and make the suggested change to net_write.c.

greghudson:
The current KfW release line (4.1) is based on the 1.13 branch.  While I don't necessarily plan to issue another KfW 4.1 release, the branch should be updated with this bugfix for people who need to build something similar to KfW 4.1 from source.

(I am somewhat hoping for positive testing results from the person who recently reported information on this bug, but will likely merge this soon regardless.)

PR #856, opened 2018-10-08
Prefer TCP to UDP for password changes

frozencemetery:
When password changes are performed over UDP, spotty networks may
cause the client to retransmit.  This leads to replay errors if the
kpasswd server receives both requests, which hide the actual request
status and make it appear that the password has not been changed, when
it may in fact have been.  Use TCP instead with UDP fallback to avoid
this issue.

~~Modernize change_set_password and make a helper function,
change_set_password_proto.~~

ticket: 7905

### DIFF
--- a/src/lib/krb5/os/changepw.c
+++ b/src/lib/krb5/os/changepw.c
@@ -59,13 +59,12 @@ struct sendto_callback_context {
 
 static krb5_error_code
 locate_kpasswd(krb5_context context, const krb5_data *realm,
-               struct serverlist *serverlist, krb5_boolean no_udp)
+               struct serverlist *serverlist)
 {
     krb5_error_code code;
 
     code = k5_locate_server(context, realm, serverlist, locate_service_kpasswd,
-                            no_udp);
-
+                            FALSE);
     if (code == KRB5_REALM_CANT_RESOLVE || code == KRB5_REALM_UNKNOWN) {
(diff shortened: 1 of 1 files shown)


frozencemetery:
(7905 proposes four solutions to the UDP replay issue.  Solutions which require modifying the kadmind code won't work, because they don't fix the problem for non-MIT KDCs.  This leaves the alternate approach of "try an AS-REQ with the new password when we get a replay error", which I dislike because an AS-REQ may not be possible without some additional policy, like a second factor; it also would leave any logging from a replay around on the KDC.)

greghudson:
I think we can use a single call to k5_sendto() with the strategy UDP_LAST (always passing no_udp=FALSE to k5_locate_server()).

frozencemetery:
Updated.  Since this removes an outer loop, it's best viewed to ignore whitespace in the diff.

greghudson:
Let's go ahead and get rid of the final parameter to locate_kpasswd() as well, since it only has the one caller.

Use the constants TRUE and FALSE for parameters of type krb5_boolean, rather than 1 or 0 (this will come up when changing locate_kpasswd() to always pass false to k5_locate_server()).

frozencemetery:
Makes sense, updated.

greghudson:
I'm not seeing a change to locate_kpasswd() in the updated commit.

frozencemetery:
Once more, this time actually committing the change.

PR #857, opened 2018-10-09
Introduce GSS_KRB5_NT_ENTERPRISE_NAME name OID

frenche:
We've discussed this a while ago, see:
https://www.mail-archive.com/kerberos@mit.edu/msg21978.html

I got stuck back then with the canonicalization part, and it got stalled.
I'm opening this PR to consult and follow up on this - I'll revise it soon and update.


frenche:
I've updated the PR and changed the name to GSS_KRB5_NT_ENTERPRISE_NAME as it only applies to KRB5.

I also dropped the canonicalization part for now, as I think it is a problem of its own.
For the impersonate use case it won't matter since Windows never canonicalize the impersonated name.
While regarding acquire_with_password; first, as Greg noted, is an extension, not a standardized part of the API. And second, it would have the same issue if canonicalize flag is enabled via krb5.conf in least two scenarios regardless of if enterprise name being used.

For instance, if the username is entered with the wrong case then Windows would canonicalize it.
The other case is, when the UPN suffix matches the realm name (default setup), then kinit works with UPN even when the RHS of the UPN is different than the canonical name (and although enterprise name is not used) . In this case too, if the canonicalize flag is on then Windows would return the canonical name.

In both of these two cases, when the name is canonicalized, then acquire_with_password fails with:
major: Invalid credential was supplied
minor: Principal in credential cache does not match desired name

Let me know what do you think of that, and also please advise how to go about choosing an OID - thanks!

greghudson:
For the OID, I see you used 1.2.840.113554.1.2.2.7 as the placeholder.  I was going to suggest the same thing with .6 at the end (after looking at the MIT-internal registry), but you presumably picked ...7 for a reason; is ...6 already in use?

greghudson:
I don't see any problems with the code, other than that the existing import function has some `(char *)` casts of expressions which already have type `char *`, including one in the krb5_parse_name() call which gets changed to krb5_parse_name_flags().  We can get rid of that cast while changing it.

For tests, it looks like tests/gssapi/t_imp_name gets built by the build system but isn't run as part of "make check".  It would be good to make it verify its results and run it in the check-unix target.

We should add documentation to doc/appdev/gssapi.rst in the "Name types" section.

frenche:
> For the OID, I see you used 1.2.840.113554.1.2.2.7 as the placeholder. I was going to suggest the same thing with .6 at the end (after looking at the MIT-internal registry), but you presumably picked ...7 for a reason; is ...6 already in use?

No reason that I can remember, will change to 6.

> I don't see any problems with the code, other than that the existing import function has some (char *) casts of expressions which already have type char *, including one in the krb5_parse_name() call which gets changed to krb5_parse_name_flags(). We can get rid of that cast while changing it.

I'll look into that.

> For tests, it looks like tests/gssapi/t_imp_name gets built by the build system but isn't run as part of "make check". It would be good to make it verify its results and run it in the check-unix target.

import_name() already verifies the result of gss_import_name(), and I've already added it to check-unix (in a fixup, missed it initially), or I'm missing something else?

> We should add documentation to doc/appdev/gssapi.rst in the "Name types" section.

Will do - thanks for looking at this.

greghudson:
The OID entry should have a comment giving the expanded OID, similar to the one for GSS_KRB5_NT_PRINCIPAL_NAME.

import_name() verifies the return code, but that's not a very comprehensive test.  Instead of printing out the display name and name type (which will be ignored in the make check output), it should verify those results against expected values.

frenche:
I've addressed all the comments, hope it shows up soon in github. Note that I had to check explicitly for enterprise type in kdb_test because we don't set KRB5_KDB_FLAG_ALIAS_OK for the impersonated user.

greghudson:
I think github might have a server sync issue at the moment.  I see your comment and the revised commit sometimes but not every time I reload the page.

I've pushed a revised commit to your branch.  Be careful using strncmp() alone to compare a counted string to a C string; using it alone will accept prefixes of the C string.  I added a helper function to check the length and use memcmp().

greghudson:
I am wondering if kdc_process_s4u2self_req() should pass KRB5_KDB_FLAG_ALIAS_OK.  Alternatively, the get_principal documentation in kdb.h could be changed to say that it's okay to return in-realm aliases for enterprise names even if ALIAS_OK isn't set.

frenche:
The changes look good, thanks.

> I am wondering if kdc_process_s4u2self_req() should pass KRB5_KDB_FLAG_ALIAS_OK. Alternatively, the get_principal documentation in kdb.h could be changed to say that it's okay to return in-realm aliases for enterprise names even if ALIAS_OK isn't set.

I think the alternative is ok.

BTW, note that regardless of what we set 'princ' to, in get_principal() the kdc returns the name from pa-data for_user in cname (I think it may impact in AS request flow though). This is the same as I've observed in Windows behavior for S4U2Self.

I also noticed some sort of a bug, that using 'e:enterprise@abc@NOREALM' causes the new test to fail as it tries to contact a KDC in NOREALM. According to MS-SFU 3.1.5.1.1.1 it should instead start with the server realm, quote: "Service 1 sends a KRB_AS_REQ without any pre-authentication to Service 1's Key Distribution Center (KDC)".
This has the advantage to work even when the specified (or more often the default) realm is wrong, while the down side is that it will add a round trip in case where the API consumer sets the correct realm of the user.

I'm in favor of changing the code to start looking in the KDC of the server, and I'm trying to see how it could be done. For testing it, we could just change the impersonated enterprise user to end with '@NOREALM'.

greghudson:
I need to look more carefully at the permutations for returning an alias entry for an enterprise name without ALIAS_OK specified.  You've said that on a Windows KDC, an AS-REQ for an enterprise name without canonicalization will issue a ticket for the enterprise name.  I guess that can be handled by the KDC code rather than the KDB, though.

Our client code was originally written for the local realm case only, and was modified to handle cross-realm but for an already canonicalized principal rather than an enterprise name.  So we simply don't implement the realm determination check described in [MS-S4U] 3.1.5.1.1.1 yet.  Our code happens to work if the realm attached to the enterprise name (which is usually just the default realm as you noted) is the realm of the client, but [MS-S4U] assumes that an enterprise name has no realm attached to it.

Assuming we do implement this step, I would suggest that when the code looks for a KDC_ERR_PREAUTH_REQUIRED (indicating that the principal is within the realm), it also accepts an AS-REP.  In the Windows universe there are no client principals without preauth-required, but in our universe that's more normal.

frenche:
> I need to look more carefully at the permutations for returning an alias entry for an enterprise name without ALIAS_OK specified. You've said that on a Windows KDC, an AS-REQ for an enterprise name without canonicalization will issue a ticket for the enterprise name. I guess that can be handled by the KDC code rather than the KDB, though.

Yea, right now the kdb cannot tell if the canonicalize flag was set because the kdc always pass KRB5_KDB_FLAG_CANONICALIZE for enterprise names, so a change in kdc would be needed to allow making a distinction in kdb or it could be handled entirely by the kdc as suggested.

> Our client S4U2Self code was originally written for the local realm case only. It was later augmented to handle cross-realm, but for an already canonicalized principal rather than an enterprise name. So we simply don't implement the realm determination process described in [MS-S4U] 3.1.5.1.1.1 yet.

Are you sure of that? s4u_identify_user() seem to follow client referrals for enterprise names, what isn't implemented exactly? Also traffic capture of running the test seem alright, see:
https://github.com/frenche/krb5/blob/1b3df4276ae100eae62bd5f412caf5694b0f064f/src/tests/gssapi/mit_xrealm_ent_test.pcap

> Our code happens to work if the realm attached to the enterprise name (which is usually just the default realm as you noted) is the realm of the client, but [MS-S4U] assumes that an enterprise name has no realm attached to it.

In the test the default realm is of the server SREALM, but it works with UREALM as well, it just skips one AS request in that case.

> Assuming we do implement this step, I would suggest that when the code looks for a KDC_ERR_PREAUTH_REQUIRED (indicating that the principal is within the realm), it also accepts an AS-REP. In the Windows universe there are no client principals without preauth-required, but in our universe that's more normal.

It already does that in s4u_identify_user() (i'd need to add that to my heimdal PR as i didn't understand it initially).
btw, windows has an account-option "do not require kerberos preauthentication" though maybe it's not the same (I'll try).

greghudson:
You're right, I had forgotten about the whole s4u_identify_user() code path.  I guess it just needs to make a shallow copy of in_creds->client (in the positive case of the conditional at line 105) and then overwrite the realm with the server realm.

frenche:
I've added a precedent commit to start the lookup with the server realm, and adjusted the test.

> [Side note: kind of mystified about the anonymous case at line 80. Does it ever make sense to do S4U2Proxy on the anonymous principal, and is that the right way to do it?]

I didn't figure it either, my vague assumption was it relates to FAST but couldn't find any doc or info about it.

> btw, windows has an account-option "do not require kerberos preauthentication" though maybe it's not the same (I'll try).

I've tested and it does exactly what you'd expect, and windows client deals fine with AS-REP as well.

greghudson:
Revised, with trace messages added to the new test.  I expect to merge this soon, and submit a separate PR to update kdb.h to document the alias behavior we're exhibiting in the test KDB module.

src/lib/gssapi32.def
```diff
@@ -185,3 +185,5 @@ EXPORTS
 ; Added in 1.16
 	GSS_KRB5_GET_CRED_IMPERSONATOR			@148	DATA
 	GSS_C_SEC_CONTEXT_SASL_SSF			@149	DATA
+; Added in 1.17
+	GSS_KRB5_NT_PRINCIPAL_NAME			@150 	DATA
```
frenche: Shouldn't this be enterprise?

frenche:
Otherwise, looks good to me.
Thanks!

PR #859, opened 2018-10-12
Prevent SIGPIPE from socket writes on UNIX-likes

frozencemetery:
Pass MSG_NOSIGNAL so that writes to a disconnected socket merely
return EPIPE, rather than raising a SIGPIPE that the caller is
unlikely to be ready for.

In order to pass flags, switch from writev to sendmsg.


greghudson:
Seems about right, but I think MSG_NOSIGNAL is a relatively recent POSIX addition, so we need to conditionalize its use.

Per https://github.com/lpeterse/haskell-socket/issues/8#issuecomment-115650974 it seems like the BSD world has a socket option SO_NOSIGPIPE which we might want to set when available, as it seems to exist in some operating systems that don't have MSG_NOSIGNAL.  (However, not on Linux, and it doesn't appear to be in POSIX, so we can't do just SO_NOSIGPIPE.)

frozencemetery:
Updated.

greghudson:
We `#ifdef` socket constants a lot, and I'm inclined to do that here.  I don't see a guarantee in POSIX that constants are `#define`d rather than enum'd, but I think it's pretty universal.  (On Linux the MSG_ constants are defined in an enum but then also `#define`d to themselves so that `#ifdef` works.)

frozencemetery:
Some day I will remember to pass `-a` to `--amend`... updated.

greghudson:
Revised.  connect() operates on an existing socket; it does not create a new one.  I added a second commit to actually use SOCKET_CONNECT (which was not used anywhere); otherwise the SO_NOSIGPIPE change is inoperative.

kaduk:
Is this in a state where testing on a BSD system would be useful?

greghudson:
Yes, although "it doesn't break anything in make check" is all I would really expect from testing.  Simulating a SIGPIPE situation and disabling the MSG_NOSIGNAL part of the patch seems like a lot.

frozencemetery:
`make check` passes on my frebsd 11.2 VM.

PR #860, opened 2018-10-13
 Add a test for cross realm S4U2Self using referrals

frenche:
Also fixup some of the comment in the recently added PAC API tests.

### DIFF
--- a/src/lib/krb5/krb/t_pac.c
+++ b/src/lib/krb5/krb/t_pac.c
@@ -511,12 +511,12 @@ check_pac(krb5_context context, int index, const unsigned char *pdata,
     ret = krb5_pac_verify_ext(context, pac, auth_time, p, server_key, kdc_key,
                               with_realm);
     if (ret)
-        err(context, ret, "[pac: %d] krb5_pac_verify_ex", index);
+        err(context, ret, "[pac: %d] krb5_pac_verify_ext", index);
 
     ret = krb5_pac_sign_ext(context, pac, auth_time, p, server_key,
                             kdc_sign_key, with_realm, &data);
     if (ret)
-        err(context, ret, "[pac: %d] krb5_pac_sign_ex", index);
+        err(context, ret, "[pac: %d] krb5_pac_sign_ext", index);
 
     krb5_pac_free(context, pac);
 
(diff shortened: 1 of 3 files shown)


frenche:
BTW, following on our previous discussion from PR #853, I did more tests and I think I figured out why Windows remote KDC returns KRB_AP_ERR_BADMATCH when asked explicitly for "krbtgt/SREALM" with pa-data for_user.

The first test I did was to instead ask the remote KDC for some "HTTP/some.other.host" from the server realm, and that worked just fine - which proves IMO that the remote KDC does *not* perform the "is-self" check (as it returned a referral TGT with the correct PAC, which the KDC in the server realm accepted and issued the final service ticket).
Then I tried to ask instead the remote KDC for a "krbtgt/SOME_TRANSIT_REALM", and there again I got the expected referral TGT to the server realm, which worked to obtained the final service ticket from the KDC in that realm.

Then I realized that the problem with asking "krbtgt/SREALM", was because the remote KDC checks if the requested server is from its own realm, and so refuses to give the ticket since the header client is not local and there for this is clearly an invalid S4U2Self request.
To confirm that, I tested asking for "HTTP/host.in.remote.kdc.realm" (or the enterprise name of that service) and got the same KRB_AP_ERR_BADMATCH error.

It might be a bug in Windows logic to consider the cross-realm "krbtgt/SREALM" as local principal, but I think this somewhat clarifies Windows KDC request processing, and helps to confirm that a full "is-self" check is not required.

frenche:
> So it considers krbtgt/SREALM to be a local principal but not krbtgt/OTHER-REALM ?
SREALM is directly trusted, so it finds 'krbtgt/SREALM' when it looks it up in DB, while OTHER_REALM is transitively trusted.

> I could see it specifically not liking a remote client asking for a cross TGT for its own realm, as that would certainly be a nonsensical request outside of the S4U2Self.

Given the referenced MS doc on KRB_AP_ERR_BADMATCH, I think the error in this context is specific to protocol transition (also, the same error is given for none krbtgt request, just any service from the remote KDC realm).

greghudson:
Okay, that makes more sense.

greghudson:
Revised.  I factored out the construction of the cross-realm TGT name in kdb_test.c to retain more readability in the get_principal function.  The new test made the old test using DB2/LMDB redundant, so I combined the two tests by adding the removal of the default realm.  I added expected trace messages to verify that the client is making all of the expected queries.

frenche:
Looks good to me.
Thanks for the refactoring.

PR #861, opened 2018-10-15
Environment variable documentation update

frozencemetery:
First commit brings back the contents of `kerberos(1)` as it stood at 0f81e372a2830c9170f6e08dfa956841d0ebdfb1, converted to ReST and installed as `kerberos(7)`.

Second commit modernizes that text a bit, and third commit changes all other command man pages to point to it.

Fourth commit fixes an incorrect build instruction in src/man's README.


frozencemetery:
An unresolved question here is what to do with env_variables.rst - it's probably best to not keep this content in two places.

kaduk:
> An unresolved question here is what to do with env_variables.rst - it's probably best to not keep this content in two places.

Having not looked at the contents thereof, I'll note that ReST does have an "include" mechanism, which we already use for some of the command-line utilities' manpages, IIRC.

greghudson:
The first commit should include src/man/kerberos.man so it doesn't break the build.  (The second commit will have to update it, of course.)

frozencemetery:
Updated.

greghudson:
I pushed revised commits.  src/Makefile.in needed a change to create the new man7/cat7 directories or make install breaks.  The translated man pages contained broken references to rlogin and rsh; I changed those to just be words.  I made a few minor wording changes.

kerberos.rst after the second commit still says "Kerberos uses this ticket for network utilities such as rlogin and rcp."  We could change that to just say ssh, or mention something else.

PR #862, opened 2018-10-15
kdb5_util doc fixups and a small behavior change

greghudson:
This is follow-up from http://mailman.mit.edu/pipermail/kerberos/2018-September/022066.html .

### DIFF
--- a/doc/admin/admin_commands/kdb5_util.rst
+++ b/doc/admin/admin_commands/kdb5_util.rst
@@ -12,10 +12,12 @@ SYNOPSIS
 [**-r** *realm*]
 [**-d** *dbname*]
 [**-k** *mkeytype*]
-[**-M** *mkeyname*]
 [**-kv** *mkeyVNO*]
-[**-sf** *stashfilename*]
+[**-M** *mkeyname*]
 [**-m**]
+[**-sf** *stashfilename*]
+[**-P** *password*]
+[**-x** *db_args*]
 *command* [*command_options*]
 
 .. _kdb5_util_synopsis_end:
@@ -79,6 +81,10 @@ COMMAND-LINE OPTIONS
     expose the password to other users on the system via the process
     list.
 
+**-x** *db_args*
+    specifies database-specific options.  See :ref:`kadmin(1)` for
+    supported options.
+
 .. _kdb5_util_options_end:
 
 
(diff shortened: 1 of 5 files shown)


doc/admin/admin_commands/kdb5_util.rst
```diff
@@ -90,13 +96,18 @@ create
 
 .. _kdb5_util_create:
 
-    **create** [**-s**]
+    **create** [**-s**] [**-h**]
```
frozencemetery: Should we mention `-W` here?

Also, I don't think it's actually possible to engage the `-h` behavior right now; it's not in the `getopt()` call: https://github.com/krb5/krb5/blob/master/src/kadmin/dbutil/kdb5_create.c#L167

doc/admin/admin_commands/kdb5_util.rst
```diff
@@ -197,8 +209,8 @@ load
 
 .. _kdb5_util_load:
 
-    **load** [**-b7**\|\ **-ov**\|\ **-r13**] [**-hash**]
-    [**-verbose**] [**-update**] *filename* [*dbname*]
+    **load** [**-b7**\|\ **-ov**\|\ **-r13**\|\ **-r18**] [**-hash**]
```
frozencemetery: Comment here should be updated or removed: https://github.com/krb5/krb5/blob/master/src/kadmin/dbutil/dump.c#L1497 (also same deal with `-i` and `-c`).

doc/admin/admin_commands/kdb5_util.rst
greghudson: Interesting.  Commit 21ddb0df0b5394279f82480257c0cf32f45810a1 attempted to add kdb5_util create -h, but in the intervening 21 years no one has ever noticed that it never made it into the getopt string.  I will change the PR to remove the switch case for the option instead of documenting it.

(This also suggests that we could get rid of kdb5_util load -hash without bothering anyone, but I won't go that far.)

greghudson:
The -i and -c options to load and dump are used internally by the iprop system.  One could make an argument for documenting them anyway but I don't think it's important to do so.

doc/admin/admin_commands/kdb5_util.rst
greghudson: I decided to update this comment as there is a similar up-to-date one for dump.

PR #863, opened 2018-10-16
Retry KCM writes once

frozencemetery:
This is a follow-on to #859 and is based on that (only the top commit is new).  Requested by @jhrozek from SSSD.

### DIFF
--- a/src/lib/krb5/ccache/cc_kcm.c
+++ b/src/lib/krb5/ccache/cc_kcm.c
@@ -299,13 +299,36 @@ kcmio_unix_socket_write(krb5_context context, struct kcmio *io, void *request,
 {
     char lenbytes[4];
     sg_buf sg[2];
+    int ret;
+    krb5_boolean reconnected = FALSE;
 
     SG_SET(&sg[0], lenbytes, sizeof(lenbytes));
     SG_SET(&sg[1], request, len);
     store_32_be(len, lenbytes);
-    if (krb5int_net_writev(context, io->fd, sg, 2) < 0)
-        return errno;
-    return 0;
+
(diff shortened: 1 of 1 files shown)


greghudson:
Can you summarize the scenario that leads to needing a retry?  I get from the comment that it involves a user taking a while to type a password, but I don't understand how that leads to a KCM server closing the connection.

frozencemetery:
Downstream this is [rhbz#1537866 comment 30](https://bugzilla.redhat.com/show_bug.cgi?id=1537866#c30), sssd-kcm had the client idle timeout too low and hung up our connections to it.  @jhrozek may be able to give a better answer.

greghudson:
Okay, the impedance mismatch seems to be that sssd has a kcm client idle timeout of 60 seconds, while the client code does not expect the server to close the connection (neither does Heimdal's client code, when it uses Unix socket RPC).  Any time a ccache handle is held open for longer than 60 seconds with no activity (kinit with a long password wait is just one example), it becomes nonfunctional when the server disconnects.

I don't know why the kcm server has an idle timeout (something to do with socket activation?).  But it does seem important to be able to restart the kcm daemon for an upgrade without breaking Kerberos operations in progress, so perhaps we should add the reconnect feature anyway.  However, a single reconnect isn't guaranteed to work, as there will likely be a window during the process restart while connections can't be accepted on the Unix domain socket.  (Perhaps that's not an issue with socket activations?)  I'm not certain what the best approach is since a delay-loop seems clumsy; I will think about it.

greghudson:
I pushed a revised commit; please check if the comment and commit message are still agreeable.

PR #865, opened 2018-10-24
Document aliases for enterprise get_principal

greghudson:
[As noted in PR #857, the alternative here is to change kdc_process_s4u2self_req() to pass ALIAS_OKAY to get_principal when looking up the requested client principal.  We always use the client's requested name for the ticket, so there probably wouldn't be any harm in that.]

Enterprise principals are always aliases.  In most contexts when we
see them we pass KRB5_KDB_FLAG_ALIAS_OK to the KDB module's
get_principal method, but for S4U2Proxy clients we currently do not.
Document that a KDB module may return an alias for enterprise
principals regardless of flags.

### DIFF
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -1018,9 +1018,10 @@ typedef struct _kdb_vftabl {
      *     requested; also set by the admin interface.  Determines whether the
      *     module should return in-realm aliases.
      *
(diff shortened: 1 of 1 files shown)


frenche:
In the comment, S4U2Proxy should read S4U2Self.
Otherwise, looks good to me.

PR #872, opened 2018-11-29
Add ksu option for non-forwardable tickets

greghudson:
Add ksu -F and -P options to explicitly not request forwardable and
proxiable tickets.

### DIFF
--- a/doc/user/user_commands/ksu.rst
+++ b/doc/user/user_commands/ksu.rst
@@ -12,7 +12,8 @@ SYNOPSIS
 [ **-c** *source_cache_name* ]
 [ **-k** ]
 [ **-r** time ]
-[ **-pf** ]
+[ **-p** | **-P**]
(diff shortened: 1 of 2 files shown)


doc/user/user_commands/ksu.rst
```diff
@@ -12,7 +12,8 @@ SYNOPSIS
 [ **-c** *source_cache_name* ]
 [ **-k** ]
 [ **-r** time ]
-[ **-pf** ]
```
frozencemetery: `-pf` is still listed on line 250.  (I'd put this comment there, but github doesn't seem to let me do that anymore.)

greghudson: Fixed.  I also changed the usage message in the C code not to say ``-pfPF``, since there's room to say ``[-p | -P] [-f | -F]``.

PR #875, opened 2018-12-04
Change dns_canonicalize_hostname profile option

simo5:
Turns it into a three-state that allows the value "fallback" on top
of the true/false booleans supported previously.

When the option is set to fallback we dealy canonicalization and attempt
it only in krb5_get_credentials() if the KDC responds that the requested
server principal name is unknown.

### DIFF
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -201,6 +201,10 @@ The libdefaults section may contain any of the following relations:
     means that short hostnames will not be canonicalized to
     fully-qualified hostnames.  The default value is true.
 
+    If this option is set to ``fallback`` (new in release 1.18), DNS
+    canonicalization will only be performed the server hostname is not
+    found with the original name when requesting credentials.
+
 **dns_lookup_kdc**
     Indicate whether DNS SRV records should be used to locate the KDCs
     and other servers for a realm, if they are not listed in the
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -1158,6 +1158,12 @@ k5_plugin_register_dyn(krb5_context context, int interface_id,
 void
 k5_plugin_free_context(krb5_context context);
 
+enum dns_canonhost {
+    CANONHOST_FALSE = 0,
+    CANONHOST_TRUE = 1,
+    CANONHOST_FALLBACK = 2
+};
+
 struct _kdb5_dal_handle;        /* private, in kdb5.h */
 typedef struct _kdb5_dal_handle kdb5_dal_handle;
 struct _kdb_log_context;
@@ -1221,7 +1227,7 @@ struct _krb5_context {
 
     krb5_boolean allow_weak_crypto;
     krb5_boolean ignore_acceptor_hostname;
-    krb5_boolean dns_canonicalize_hostname;
+    enum dns_canonhost dns_canonicalize_hostname;
 
     krb5_trace_callback trace_callback;
     void *trace_callback_data;
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -191,6 +191,9 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
 #define TRACE_FAST_REQUIRED(c)                                  \
     TRACE(c, "Using FAST due to KRB5_FAST_REQUIRED flag")
 
+#define TRACE_GET_CREDS_FALLBACK(c, hostname)                           \
+    TRACE(c, "Falling back to canonicalized server hostname {str}", hostname)
+
 #define TRACE_GIC_PWD_CHANGED(c)                                \
     TRACE(c, "Getting initial TGT with changed password")
 #define TRACE_GIC_PWD_CHANGEPW(c, tries)                                \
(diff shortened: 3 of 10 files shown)


simo5:
Note: this compiles but is still untested.
Throwing it here in case the direction is off.

greghudson:
I added a test for the krb5_get_credentials() fallback, as discussed.  I noticed two things:

1. klist reports a redundant cache entry after canonicalization is performed:
```
12/10/18 12:42:44  12/11/18 12:42:44  krbtgt/R1@R1
12/10/18 12:42:44  12/11/18 12:42:44  host/www.kerberos.org@R1
12/10/18 12:42:44  12/11/18 12:42:44  host/ptr-mismatch.kerberos.org@R1
12/10/18 12:42:44  12/11/18 12:42:44  host/www.kerberos.org@R1
```
2. If the hostname as given isn't present before canonicalization, the first time you call krb5_get_credentials(), the output cred contains the canonicalized name, but the second time the output cred contains the original name (because it was retrieved from the cache).  Not sure if we can do anything about that, or if it will cause any problems.

simo5:
1. Should we check that the credential is not already in the cache when we do canonicalization ?
2. Should we return the cred with the original name instead of the canonicalized name when we fallback ?

greghudson:
1. No, that wouldn't solve the problem and shouldn't be necessary (if the cred was already present, we would have used it rather than make a TGS request).  The problem is that krb5_cc_store_cred() (the wrapper, in ccfns.c) stores credentials for both the cred server and the ticket server, if they differ.  So when we ask it to store the entry for the originally requested server, we wind up storing it again for the ticket server as an undesired side-effect.

The simplest fix is to pass KRB5_GC_NO_STORE to the state machine on the second try, so we can do our own store of both the requested server and ticket server.  That also means we won't store three different entries when referrals are in the mix.  Long-term, I am not sure that it was good for the double-store-for-referrals logic to be as deep in the stack as krb5_cc_store(), but it's something we can work with.

2. Yes; it looks like the old code takes care to do that in the referrals case (where the requested server has no realm, but the resulting server obviously has one).  I will shortly push a change to modify gcred to report the ticket server instead of the output cred server, so that the new tests won't break with this change.

simo5:
Ok, so the plan is for me to change the code to pass KRB5_GC_NO_STORE, and then manually store both the new and the old cred myself in the wrap-up code at the end of the function ?

greghudson:
> Ok, so the plan is for me to change the code to pass KRB5_GC_NO_STORE, and then manually store both the new and the old cred myself in the wrap-up code at the end of the function ?

You only have to store under the original name.  krb5_cc_store_cred() will look at the ticket server and store it under that name as well, if it is different.

simo5:
I pushed a WIP commit with the change as I understood it.
Is this what you meant ?

greghudson:
I was thinking we would only pass GC_NO_STORE to the state machine when looking up the canonicalized name, and only do our own store if we did canonicalization.  However, I'm not sure why your way doesn't work just as well (it causes a failure in existing S4U tests), so I will look into that.

greghudson:
Okay.  I think the problem with your way is that after the first try with the state machine, we don't know whether we got the cred from the cache or from the KDC.  Blindly storing in that situation will keep writing already-cached credentials to the cache each time krb5_get_credentials() is called.

If we only store when we fell back to the canonicalized name, we don't have that problem, because we know that the original name wasn't found in the cache.

simo5:
Uhm I have a patch that conditionally does the store/no store thing, but I see the same s4u failure.
I am going to push my current WIP

greghudson:
It passes Travis tests and passes the s4u tests for me.

simo5:
Odd, oh well, I'll squash and repush, and re-run test locally.

simo5:
Note: test passed locally for me too, must have done something wrong the previous time.

greghudson:
I pushed some changes.  Aside from some minor naming and wording quibbles, I refactored the lookup of the tristate variable in init_ctx.c, refactored the krb5_get_credentials() changes to avoid the backwards goto, and introduced k5_expand_hostname() so we don't need to copy and tamper with the context.

simo5:
Changes look good to me.

greghudson:
The disadvantages are making extra TGS requests (two of them, because of referrals) in common cases, and maybe causing some rare edge-case problems.

There is no security advantage unless the TGS requests are protected (with a KDC proxy; FAST unfortunately doesn't help).

There is a usability advantage, in that in cases like the one Simo raised, clients (well, newer MIT krb5 clients) will find the alias name without needing specific configuration.  That might be a good enough reason to change the default.

That question does remind me that we should have a trace log message for when we retry with canonicalization, and the tests should probably have some expected_trace clauses.  I will make those changes.

greghudson:
I have updated the PR with a trace log message.

Another concern with setting the default to ``fallback`` is that it will change the behavior for non-TGS uses of krb5_sname_to_principal()  (enumerated in https://k5wiki.kerberos.org/wiki?title=Projects/Server_Hostname_Canonicalization).  That would be especially concerning without shortname qualification, but even after that is added (which I hope to do soon), the change has significant potential to introduce failures on servers.

simo5:
I am wondering if we should consider the option of introducing a separate krb5_sname_to_principal()-like function to use in GSSAPI, but OTOH if we do that then we have diverging behavior and we do not help user transitioning all of their krb5 use away from DNS based canonicalization ... so perhaps we should just not turn fallback on by default in the MIT distribution and experiment with turning it on downstream (Fedora/RHEL) and see if it is as painful as you think ?

PR #876, opened 2018-12-18
imporomper initialization breaks compilation on Solaris

Sashan:
gcc we use on Solaris complains about empty initialization. This small patch
makes it happy.

I gave a try to krb5-1.17-beta2

### DIFF
--- a/src/configure.in
+++ b/src/configure.in
@@ -842,6 +842,10 @@ AC_CHECK_TYPES([struct rt_msghdr], , , [
 #include <net/route.h>
 ])
 
+# Tests for 64-bit edwards25519 code.
+AC_CHECK_SIZEOF([size_t])
+AC_CHECK_TYPES([__int128_t, __uint128_t])
+
 # stuff for util/profile
 
 # AC_KRB5_TCL already done
--- a/src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
+++ b/src/plugins/kdb/ldap/libkdb_ldap/ldap_principal2.c
@@ -628,7 +628,7 @@ update_ldap_mod_auth_ind(krb5_context context, krb5_db_entry *entry,
     int i = 0;
     krb5_error_code ret;
     char *auth_ind = NULL;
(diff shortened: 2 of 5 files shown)


greghudson:
I squashed in #877, #880, and #881 and made some minor changes.  Please check if the revised commit works for you.

I'm not philosophically opposed to an autoconf test for __int128_t support.  I was trying to adhere closely to the BoringSSL code, but they have different portability requirements.

The other half of that test is whether the platform is 64-bit.  (Otherwise I believe it's more efficient to use the 32-bit code even if the compiler supports a 128-bit type.)  I guess that's easily determined with `AC_CHECK_SIZEOF([size_t])` and `#if SIZEOF_SIZE_T >= 8`.

Sashan:
> I squashed in #877, #880, and #881 and made some minor changes. Please check if the revised commit works for you.
>
this still works. I'm with squashing all those pull requests to single changeset. And also thank you for making my changes to look better, especially `update_thash()`.

 
> I'm not philosophically opposed to an autoconf test for __int128_t support. I was trying to adhere closely to the BoringSSL code, but they have different portability requirements.
> 
> The other half of that test is whether the platform is 64-bit. (Otherwise I believe it's more efficient to use the 32-bit code even if the compiler supports a 128-bit type.) I guess that's easily determined with `AC_CHECK_SIZEOF([size_t])` and `#if SIZEOF_SIZE_T >= 8`.

OK I'll try to cook something up and we will see if it will be a good fit or not.

FYI: the test failure I've mentioned in #880, covered my back. I've messed up
`sha256()` glue/wrapper of Kerberos to our internal `libucrypto`. I was passing *in.length
instead of in[i].length to `_digest_update()` function. I was fortunate enough the bug was noticed by unit tests. thank you for keeping kerberos in good shape.

greghudson:
Do the autoconf tests in this updated commit work for you?

Sashan:
> Do the autoconf tests in this updated commit work for you?

yes, I need to run `autoconf` in order to get fresh `configure` script

PR #877, opened 2018-12-18
typecast required to stop compiler to complain on Solaris

Sashan:
Patch makes compiler happy on Solaris 11.4. I gave a try to krb5-1.17-beta2.
The function argument comes as (const krb5_data *), we need to relax it to
krb5_data * when dereferencing the pointer.

### DIFF
--- a/src/plugins/preauth/spake/util.c
+++ b/src/plugins/preauth/spake/util.c
@@ -91,8 +91,8 @@ update_thash(krb5_context context, groupstate *gstate, int32_t group,
 
     /* Set up the data array and hash it with the group's hash function. */
     dlist[0] = *thash;
-    dlist[1] = (data1 != NULL) ? *data1 : empty_data();
-    dlist[2] = (data2 != NULL) ? *data2 : empty_data();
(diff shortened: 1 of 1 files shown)


greghudson:
Hm.  C99 section 6.5.15 (Conditional Operator) says:

```
One of the following shall hold for the second and third operands:
— both operands have arithmetic type;
— both operands have the same structure or union type;
— both operands have void type;
— both operands are pointers to qualified or unqualified versions of compatible types;
— one operand is a pointer and the other is a null pointer constant; or
— one operand is a pointer to an object or incomplete type and the other is a pointer to a
qualified or unqualified version of void.
```

So, `*((data1 != NULL) ? data1 : &empty)` (where empty is of type krb5_data) would be okay, because a qualified and an unqualified pointer to the same type is okay.  (The result has the union of the qualifiers.)  A `const int` and `int` in the two operands would be okay, because they are arithmetic types.  So I guessqualified and unqualified versions of the same structure type is an omission in the standard--it should work because qualifiers only have an impact on lvalues, but it isn't specified to work.

greghudson:
Closing this PR as I've squashed it into #876.

PR #880, opened 2018-12-19
developer studio/solaris studio does not support int128

Sashan:
patch below makes code to compile with Solaris/developer studio.
I did try to use developerstudio12.6, but it still lacks to support int128
as supported by clang & gcc

### DIFF
--- a/src/plugins/preauth/spake/edwards25519.c
+++ b/src/plugins/preauth/spake/edwards25519.c
@@ -102,6 +102,10 @@
 #endif
 
 /*
+ * Solaris/Developer Studio lacks support for __int128_t
+ */
+#ifndef __SUNPRO_C
+/*
  * These preprocessor conditionals are derived the BoringSSL
  * include/openssl/base.h (OPENSSL_64_BIT) and crypto/internal.h
  * (BORINGSSL_HAS_UINT128).
@@ -113,6 +117,7 @@ typedef __int128_t int128_t;
 typedef __uint128_t uint128_t;
 #endif
 #endif
+#endif	/* __SUNPRO_C */
 
 #ifndef EDWARDS25519_ASSERTS
 #define assert_fe(f)


Sashan:
Although the code compiles, the change uncovers some bug on amd64 platform. The `make check` trips over assert here in `spake/t_vectors.c` unit test:
```
422     /* Verify transcript hash. */
423     hash = empty_data();
424     check(update_thash(ctx, gstate, t->group, &hash, support, challenge));
425     check(update_thash(ctx, gstate, t->group, &hash, S, &empty));
426     assert(data_eq(*thash, hash));
427     krb5_free_data_contents(ctx, &hash);
```

I'm still investigating, what's going on here.

greghudson:
Unfortunately, debugging test vector failures is usually a painstaking process of comparing intermediate values between a working and non-working platform, to isolate the operation that isn't producing the expected result.

To see if this was simply an endian issue (is amd64 big-endian?), I tried builing on a Sparc Solaris build machine (using gcc) and manually disabling the 64-bit flag, but the tests still pass for me.

Sashan:
> Unfortunately, debugging test vector failures is usually a painstaking process of comparing intermediate values between a working and non-working platform, to isolate the operation that isn't producing the expected result.
> 
> To see if this was simply an endian issue (is amd64 big-endian?), I tried builing on a Sparc Solaris build machine (using gcc) and manually disabling the 64-bit flag, but the tests still pass for me.

to keep you updated about my progress: the compiler bug is unlikely. I'm still investigating the issue. I did give it a try with gcc 7.3.0 and got the same failure. I'm keeping fairly large diff to customize kerberos for Solaris. I suspect kind of mismerge. So the change in this pull request is OK.

Speaking of this pull request: Would it make sense to write autoconf test and introduce something like `HAVE_INT128`? I can give it a try.

Sashan:
I narrowed the search area for culprit further down. The brief summary is: code in `edwards25519.c` and compiler(s) are both correct.

The test failure is caused by `libucrypto`. Mit Kerberos shipped with Oracle Solaris is linked with libucrypto (we build with autoconf option `--with-crypto-impl=ucrypto`). The library makes `t_vectors` unit test to fail. I still need to figure out what exactly is going wrong. However this is entirely my own business. Thanks for all your help.

greghudson:
Closing this as I've squashed it into #876.

PR #881, opened 2018-12-19
zap.so is required when unit tests are not compiled with clang or gcc

Sashan:
k5-platform.h does not pick up memset() for zap() when unit tests
are being built by Solaris/developer studio. In this case we
must link with zap.so explicitly.

not sure if proposed change in this pull request is what we
also want to have for gcc & clang


greghudson:
All of the modified lines use .o files, not .so files.  Does adding zap.o work?

greghudson:
Closing this as I've squashed it into #876.

PR #882, opened 2018-12-19
Remove incorrect KDC assertion

greghudson:
The assertion in return_enc_padata() is reachable because
kdc_make_s4u2self_rep() may have previously added encrypted padata.
It is no longer necessary because the code uses add_pa_data_element()
instead of allocating a new list.

CVE-2018-20217:

In MIT krb5 1.8 or later, an authenticated user who can obtain a TGT
using an older encryption type (DES, DES3, or RC4) can cause an
assertion failure in the KDC by sending an S4U2Self request.

[ghudson@mit.edu: rewrote commit message with CVE description]

### DIFF
--- a/src/kdc/kdc_preauth.c
+++ b/src/kdc/kdc_preauth.c
@@ -1683,7 +1683,6 @@ return_enc_padata(krb5_context context, krb5_data *req_pkt,
     krb5_error_code code = 0;
     /* This should be initialized and only used for Win2K compat and other
      * specific standardized uses such as FAST negotiation. */
-    assert(reply_encpart->enc_padata == NULL);
     if (is_referral) {
         code = return_referral_enc_padata(context, reply_encpart, server);
         if (code)
--- a/src/tests/gssapi/t_s4u.py
+++ b/src/tests/gssapi/t_s4u.py
@@ -147,6 +147,14 @@ if 'auth1: user@' not in out or 'auth2: user@' not in out:
 
 realm.stop()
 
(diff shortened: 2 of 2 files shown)


greghudson:
The full history of this bug (I originally put this in the commit message, but decided it was too much):

Encrypted padata was introduced to the KDC in commit 0ba5ccd7bb3ea15e44a87f84ca6feed8890f657d with only one user, return_svr_referral_data().  This function performed its own allocation of the list, and accordingly verified with an assert() that no code had previously populated the encrypted padata field.

Commit 0e39f8a3ad915eeb0131fb4a87b0fef304101cfd added another use of encrypted padata, kdc_make_s4u2self_rep(), without adjusting the original use to accomodate existing encrypted padata.  At this point an assertion failure could only arise when returning a referral for an S4U2Self request, because return_svr_referral_data() was only called on the TGS path when issuing a referral.

Commits acd672680307a3f1bad828e6573fa34bd7779203, 7a04b369aa71b0796f057cf5da57070a9c37c9c8, and 11079f43e990d8150870a2c894e17e286f46c96c add a third use of encrypted padata, FAST negotiation.  An assertion remains in the new function
return_enc_padata() that no prior code has set encrypted padata.  Since return_enc_padata() is now called unconditionally on the TGS path, the KDC will always abort if kdc_make_s4u2self_rep() adds
encrypted padata.  This happens if the TGT session key uses an older encrypted type (DES, DES3, or RC4).

PR #885, opened 2018-12-28
SPNEGO: Pass mechanism error token status to application

lhoward:
Fixes an issue where if a SPNEGO initiator receives a `NegTokenResp` where `negState` is `reject`, it will immediately return to the caller, rather than processing any (mechanism-specific) context error tokens from the acceptor.

NB: still checking I have read RFC4178 correctly and it is OK to return a `responseToken` with a reject response.

### DIFF
--- a/src/lib/gssapi/spnego/gssapiP_spnego.h
+++ b/src/lib/gssapi/spnego/gssapiP_spnego.h
@@ -21,7 +21,7 @@ extern "C" {
 #define	ACCEPT_INCOMPLETE 1
 #define	REJECT 2
 #define REQUEST_MIC 3
-#define	ACCEPT_DEFECTIVE_TOKEN 0xffffffffUL
+#define	UNSPECIFIED 0xffffffffUL
 
 /*
  * constants for der encoding/decoding routines.
--- a/src/lib/gssapi/spnego/spnego_mech.c
+++ b/src/lib/gssapi/spnego/spnego_mech.c
@@ -119,21 +119,19 @@ init_ctx_new(OM_uint32 *, spnego_gss_cred_id_t, send_token_flag *,
 	     spnego_gss_ctx_id_t *);
 static OM_uint32
 init_ctx_nego(OM_uint32 *, spnego_gss_ctx_id_t, OM_uint32, gss_OID,
-	      gss_buffer_t *, gss_buffer_t *,
-	      OM_uint32 *, send_token_flag *);
+	      gss_buffer_t *, gss_buffer_t *, send_token_flag *);
 static OM_uint32
 init_ctx_cont(OM_uint32 *, spnego_gss_ctx_id_t, gss_buffer_t,
 	      gss_buffer_t *, gss_buffer_t *,
 	      OM_uint32 *, send_token_flag *);
 static OM_uint32
 init_ctx_reselect(OM_uint32 *, spnego_gss_ctx_id_t, OM_uint32,
-		  gss_OID, gss_buffer_t *, gss_buffer_t *,
-		  OM_uint32 *, send_token_flag *);
+		  gss_OID, gss_buffer_t *, gss_buffer_t *, send_token_flag *);
 static OM_uint32
 init_ctx_call_init(OM_uint32 *, spnego_gss_ctx_id_t, spnego_gss_cred_id_t,
 		   gss_name_t, OM_uint32, OM_uint32, gss_buffer_t,
 		   gss_OID *, gss_buffer_t, OM_uint32 *, OM_uint32 *,
-		   OM_uint32 *, send_token_flag *);
+		   send_token_flag *);
 
 static OM_uint32
 acc_ctx_new(OM_uint32 *, gss_buffer_t, spnego_gss_cred_id_t, gss_buffer_t *,
@@ -697,7 +695,7 @@ init_ctx_new(OM_uint32 *minor_status,
 	*sc_out = sc;
 	sc = NULL;
 	*tokflag = INIT_TOKEN_SEND;
-	ret = GSS_S_CONTINUE_NEEDED;
+	ret = GSS_S_COMPLETE;
 
 cleanup:
 	release_spnego_ctx(&sc);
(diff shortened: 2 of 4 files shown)


greghudson:
I can manifest this bug by modifying tests/gssapi/t_err.c to use SPNEGO instead of krb5.  If we can fix the bug, we should add a SPNEGO switch to t_err.c and run it both ways in t_gssapi.py.

I can't imagine that SPNEGO was intentionally designed to be unable to send error tokens, but I do need to check that the initiator can determine what mech to process the error with.  I guess if there's an error token in the first response, it can only be for the optimistic mech, and if there's an error token in a later message then negotiation will have determined the mech.

lhoward:
> I can't imagine that SPNEGO was intentionally designed to be unable to send error tokens, but I do need to check that the initiator can determine what mech to process the error with. I guess if there's an error token in the first response, it can only be for the optimistic mech, and if there's an error token in a later message then negotiation will have determined the mech.

That makes intuitive sense but – further thoughts are welcome.

greghudson:
I think we're pretty confident of the protocol constraints at this point, so I'm looking at the patch.  To summarize what it does:

* In spnego_gss_init_sec_context(), if init_ctx_cont() returns an error with minor status ERR_SPNEGO_NEGOTIATION_FAILED and a response token, we continue on to init_ctx_call_init() instead of returning immediately.

* In init_ctx_call_init(), if `*negState` is REJECT and we have an existing context handle, we check for the unexpected case that the mechanism doesn't see the response token as an error, and return GSS_S_FAILURE with a minor status of GSS_S_BAD_MECH.

Two questions:

1. Why check if we have an existing context handle in init_ctx_call_init()?  It seems like the logic applies whenever *negState is REJECT.

2. Shouldn't that minor status be ERR_SPNEGO_NEGOTIATION_FAILED?  GSS_S_BAD_MECH is a major code.

lhoward:
Very sloppy on my part, returning GSS_S_BAD_MECH as a major code, I have no idea what I was thinking there.

lhoward:
(I'm sorry to reply when I have the context swapped out and am not in front of my work machine, so I hope this is actually helpful.)

Re: point 1, I think I was just being defensive, you can't get an error token from the acceptor unless you've already sent something to it (which would mean you have a valid mechanism context handle).

lhoward:
I will investigate further when I am in front of my work machine.

greghudson:
> Re: point 1, I think I was just being defensive, you can't get an error token from the acceptor unless you've already sent something to it (which would mean you have a valid mechanism context handle).

It looks like spnego_gss_init_sec_context() initializes negState to REJECT and doesn't change it in the call to init_ctx_new(), so I think you weren't just being defensive--the first call to init_ctx_call_init() will have negState=REJECT and sc->ctx_handle=GSS_C_NO_CONTEXT.

I will think about clearer ways to express this.

greghudson:
Other observations, some only loosely related to the scope of this PR:

* RFC 4178 says GSS_S_BAD_MECH will be returned on a failure in the negotiation phase; that's presumably what Luke was going for in the ee7775369585c9edbba2dae4ef83714340f5a6b0 force-push.  (I think the spec is misguided on this point; GSS_S_BAD_MECH should tell the caller something about the local GSS implementation or the credentials, not something about the server.  But I'm not proposing to ignore the spec.)  In our code, I think that means we should return GSS_S_BAD_MECH from init_ctx_cont() on a REJECT token, but only on the first response token (!sc->nego_done).  On subsequent responses, REJECT doesn't suggest a mech negotiation failure.

* If an acceptor sends REJECT with an error token, and the initiator mech doesn't return an error, this PR proposes to return the same error as we would have for a REJECT with no error token.  I'm inclined to return GSS_S_DEFECTIVE_TOKEN instead, on the assumption that the error token wasn't really an error token.  There is the possibility that the mech isn't very good about recognizing or processing error tokens, but I think that's unlikely and would be a bug in the mech.

* If the acceptor sends ACCEPT_COMPLETE and our mech produces another token, we ignore the acceptor's state and produce a token to send to the acceptor (who is not expecting it).  I think returning a GSS_S_DEFECTIVE_TOKEN error would be better in this situation.  Similarly if the acceptor sends ACCEPT_INCOMPLETE and our mech doesn't produce another token.  (Our acceptor code isn't any better; it completely ignores the initiator's negState and just forces the negotiation state to ACCEPT_INCOMPLETE.)

* RFC 4178 says supportedMech "SHALL only be present in the first reply from the target", with no exception for REJECT tokens.  I'm not sure if the "only" there is supposed to make that a purely negative statement; our comment near the beginning of init_ctx_nego() suggests that an implementor interpreted it as a positive statement for initial tokens (at least non-REJECT tokens).  Obviously if the acceptor doesn't support any of the initiator mechanisms, it can't fill in a valid value there.  Right now our acceptor code doesn't appear to send a supportedMech value with an error token; perhaps it should.

* RFC 4178 says negState is optional after the first reply; if it's not set, the code is supposed to infer the state from the mechanism state.  I think our initiator code conforms to this, but in kind of a weird way; get_negTokenResp() sets negState to ACCEPT_DEFECTIVE_TOKEN if there is no negState field, and because init_ctx_cont() doesn't pay as much attention to negState as it could, that is treated similarly to ACCEPT_COMPLETE or ACCEPT_INCOMPLETE.

lhoward:
Great analysis as usual @greghudson. @nicowilliams and I took a quick look at fixing this in Heimdal and we should make sure we match your comments above. One thing I did to was to make the inferring negState (when not present in the token) more explicit, but that mightn't be as easy to do in the MIT code.

greghudson:
Based on the needs above, I would like to modify the code to remember the acceptor's negState for longer in the initiator path, and to simplify/delay the determination of the negState value for the token we generate.  Of the four possible negState values we could generate:

* We never generate REQUEST_MIC; that value is only valid for the acceptor on its first reply.
* We do appear to generate rejection tokens on the initiator path in several cases.  (It's not clear that we really need to do this, but I don't propose to change that behavior.)  A negState value of REJECT always accompanies a send_token status of ERROR_TOKEN_SEND, so that is easy to recognize at the end of processing.
* A comment in init_ctx_call_init() suggests that we should always generate ACCEPT_INCOMPLETE for interoperability with Microsoft, even if we don't expect another token back from the acceptor.  I haven't found any details on this interop issue beyond the comment.
* However, it looks like currently we can generate ACCEPT_COMPLETE when sending a MIC token, if we already received a MIC from the acceptor.  (The code path is handle_mic() detecting that `sc->mic_sent` and `sc->mi_rcvd` are true, and `*mic_out` contains a token.)

If that last case is intended, we will need a bit of logic to generate ACCEPT_COMPLETE when sending a final MIC.  Aside from that we want REJECT in the error case and ACCEPT_INCOMPLETE in the successful case.

Aside from simplifying how we generate the outgoing negState, we can (with the remembered acceptor negState) add sanity checks for the following unexpected cases:

* The acceptor negState is REJECT and the client mech doesn't generate an error (required for this PR)
* The acceptor negState is ACCEPT_COMPLETE and the client mech generated a token
* The acceptor negState is ACCEPT_INCOMPLETE and we have nothing further to say (no mech token and no MIC).

I plan to change the initial setting of negState from REJECT to ACCEPT_INCOMPLETE so that the first call to init_ctx_call_init() can more simply determine when it's processing an error token from the acceptor.

I also plan to rename ACCEPT_DEFECTIVE_TOKEN to UNSPECIFIED (or NEGSTATE_UNSPECIFIED if there is a conflict), since an omitted negState isn't necessarily a defect.  No behavior change there, though.

greghudson:
Okay, here is my candidate.

It's possible that we could refactor the acceptor code in a similar way to how I refactored the initiator code, so that handle_mic() has the same responsibilities for both.  But that would require another round of careful analysis.

lhoward:
Thanks Greg, I will check with NegoEx shortly.

PR #886, opened 2018-12-30
Address some optimized-out memset() calls

greghudson:
Ilja Van Sprundel reported a list of memset() calls which gcc
optimizes out.  In krb_auth_su.c, use zap() to clear the password, and
remove two memset() calls when there is no password to clear.  In
iakerb.c, remove an unnecessary memset() before setting the only two
fields of the IAKERB header structure.  In svr_principal.c, use
krb5_free_key_keyblock_contents() instead of hand-freeing key data.
In asn1_k_encode.c, remove an unnecessary memset() of the kdc_req_hack
shell before returning.


greghudson:
There shouldn't be anything sensitive (keys or passwords) at the top level of a kdc_req structure.

PR #887, opened 2018-12-31
Fix S4U2Self using X509 certificate

iboukris:
This is still WIP as I don't understand much of the logic surrounding the must_preauth=0 change and I'd also want to try add a basic auto test that shows we get the expected error from KDC.

I managed to get LsaLogonUser() to work with a certificate on Windows to see how it does it, and have attached the code, packet captures and dissected packet logs (using a patched wireshark) as a temporary commit (though perhaps we can keep the kvno changes as a new flag, alternative to -U).


greghudson:
I think the must_preauth situation is this: under normal circumstances, if we received a PREAUTH_REQUIRED error from the KDC, we have to generate a preauth token from some mechanism or there's no point in sending another request.  In this case, we got a PREAUTH_REQUIRED error from the KDC, but we don't plan to continue; we are just trying to determine the client identity.

There may be a more elegant way to handle this (I will think about it), but that's why you're having to set must_preauth to 0 to make it work.

iboukris:
Okay I see, but how does the enterprise lookup case works without it?
Actually I don't even remember what exactly failed, i need to test again.

For the testing, there are four interesting combinations we can invoke the API with (with their corresponding expectations):
- cert only -> AS with cert, TGS with cert.
- cert and princ -> TGS with both.
- cert and empty princ (realm only) -> TGS with cert.
- cert and enterprise princ -> AS with princ, TGS with both.

iboukris:
Update: the empty princ case isn't interesting as it sends a TGS with both cert and empty name and fails.
I added a draft test for the three use cases, all work against Windows as expected:
- cert only -> AS with cert, TGS with cert.
- cert and princ -> TGS with both.
- cert and enterprise princ -> AS with princ, TGS with both.

In testrealm environment I get KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN for all three, I guess it would suffice to check for that (we don't have to read a file, but send dummy data instead).

Regarding must_preauth=0, if I omit it, then we do not add KRB5_PADATA_S4U_X509_USER padata to the AS request, so Windows replies with KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN and we do not continue with TGS but fail.

iboukris:
> Regarding must_preauth=0, if I omit it, then we do not add KRB5_PADATA_S4U_X509_USER padata to the AS request, so Windows replies with KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN and we do not continue with TGS but fail.

I suspect the AS request without X509 padata is some kind of fallback, that could explain how it works in the enterprise lookup case as it doesn't need the padata for the fallback to work (since there is a cname).
But I don't follow the code clearly yet, it's tricky.

greghudson:
> I suspect the AS request without X509 padata is some kind of fallback

Right, here's what happens:

* s4u_identify_user() sets an optimistic preauth type (krb5_get_init_creds_opt_set_preauth_list()).

* init_creds_step_request() sees the optimistic padata list in ctx and calls k5_preauth() with must_preauth=TRUE.

* k5_preauth() sees the KRB5_PADATA_S4U_X509_USER type and calls add_s4u_x509_user_padata().  However, it then calls process_pa_data() with must_preauth=TRUE and gets an error because no authenticating padata is generated, so returns an error with no padata list.

* init_creds_step_request() falls back to sending an unauthenticated request.

So, it's an impedance mismatch between the get_in_tkt.c model of optimistic preauth and the way we want to use it in s4u_creds.c.  Perhaps we can flag the need for S4U_X509_USER padata without using the optimistic preauth list, and deal with it as part of the `if (ctx->info_pa_permitted)` conditional instead.

iboukris:
Here's the trace, with must_preauth=0:
```log
$ KRB5_TRACE=/dev/stderr LD_LIBRARY_PATH=/home/admin/git/krb5/src/lib tests/s4u2self_cert ayala@ACME.COM ~/ayala_no_san.cer
[22860] 1546443277.791589: Getting initial credentials for @ACME.COM
[22860] 1546443277.791591: Attempting optimistic preauth
[22860] 1546443277.791592: Processing preauth types: PA-FOR-X509-USER (130)
[22860] 1546443277.791593: Produced preauth for next request: PA-FOR-X509-USER (130)
[22860] 1546443277.791594: Sending request (1898 bytes) to ACME.COM
[22860] 1546443277.791595: Resolving hostname 192.168.47.120
[22860] 1546443277.791596: Initiating TCP connection to stream 192.168.47.120:88
[22860] 1546443277.791597: Sending TCP request to stream 192.168.47.120:88
[22860] 1546443277.791598: Received answer (222 bytes) from stream 192.168.47.120:88
[22860] 1546443277.791599: Terminating TCP connection to stream 192.168.47.120:88
[22860] 1546443277.791600: Sending DNS URI query for _kerberos.ACME.COM.
[22860] 1546443277.791601: No URI records found
[22860] 1546443277.791602: Sending DNS SRV query for _kerberos-master._udp.ACME.COM.
[22860] 1546443277.791603: Sending DNS SRV query for _kerberos-master._tcp.ACME.COM.
[22860] 1546443277.791604: No SRV records found
[22860] 1546443277.791605: Response was not from master KDC
[22860] 1546443277.791606: Received error from KDC: -1765328359/Additional pre-authentication required
[22860] 1546443277.791609: Preauthenticating using KDC method data
[22860] 1546443277.791610: Processing preauth types: PA-PK-AS-REQ (16), PA-PK-AS-REP_OLD (15), PA-ETYPE-INFO2 (19), PA-ENC-TIMESTAMP (2)
[22860] 1546443277.791611: Selected etype info: etype aes256-cts, salt "ACME.COMayala", params ""
[22860] 1546443277.791612: PKINIT client has no configured identity; giving up
[22860] 1546443277.791613: PKINIT client has no configured identity; giving up
[22860] 1546443277.791614: Preauth module pkinit (16) (real) returned: 22/Invalid argument
[22860] 1546443277.791615: PKINIT client ignoring draft 9 offer from RFC 4556 KDC
[22860] 1546443277.791616: Preauth module pkinit (15) (real) returned: -1765328360/Preauthentication failed
[22860] 1546443277.791617: Preauth module encrypted_timestamp (2) (real) returned: -1765328174/No user identity options specified
[22860] 1546443277.791618: Getting credentials apache@ACME.COM -> krbtgt/ACME.COM@ACME.COM using ccache FILE:/tmp/krb5cc_1000
[22860] 1546443277.791619: Retrieving apache@ACME.COM -> krbtgt/ACME.COM@ACME.COM from FILE:/tmp/krb5cc_1000 with result: 0/Success
[22860] 1546443277.791620: Get cred via TGT krbtgt/ACME.COM@ACME.COM after requesting apache@ACME.COM (canonicalize on)
[22860] 1546443277.791621: Generated subkey for TGS request: aes256-cts/4E63
[22860] 1546443277.791622: etypes requested in TGS request: aes256-cts, aes128-cts, aes256-sha2, aes128-sha2, des3-cbc-sha1, rc4-hmac, camellia128-cts, camellia256-cts, des-cbc-crc, des, des-cbc-md4
[22860] 1546443277.791624: Encoding request body and padata into FAST request
[22860] 1546443277.791625: Sending request (5143 bytes) to ACME.COM
[22860] 1546443277.791626: Resolving hostname 192.168.47.120
[22860] 1546443277.791627: Initiating TCP connection to stream 192.168.47.120:88
[22860] 1546443277.791628: Sending TCP request to stream 192.168.47.120:88
[22860] 1546443277.791629: Received answer (1370 bytes) from stream 192.168.47.120:88
[22860] 1546443277.791630: Terminating TCP connection to stream 192.168.47.120:88
[22860] 1546443277.791631: Sending DNS URI query for _kerberos.ACME.COM.
[22860] 1546443277.791632: No URI records found
[22860] 1546443277.791633: Sending DNS SRV query for _kerberos-master._udp.ACME.COM.
[22860] 1546443277.791634: Sending DNS SRV query for _kerberos-master._tcp.ACME.COM.
[22860] 1546443277.791635: No SRV records found
[22860] 1546443277.791636: Response was not from master KDC
[22860] 1546443277.791637: Decoding FAST response
[22860] 1546443277.791638: TGS reply is for ayala@ACME.COM -> apache@ACME.COM with session key aes256-cts/16D5
[22860] 1546443277.791639: Got cred; 0/Success
Code: 0
```
And without the change:
```log
$ KRB5_TRACE=/dev/stderr LD_LIBRARY_PATH=/home/admin/git/krb5/src/lib tests/s4u2self_cert ayala@ACME.COM ~/ayala_no_san.cer
[24377] 1546443647.948395: Getting initial credentials for @ACME.COM
[24377] 1546443647.948397: Attempting optimistic preauth
[24377] 1546443647.948398: Processing preauth types: PA-FOR-X509-USER (130)
[24377] 1546443647.948399: Sending unauthenticated request
[24377] 1546443647.948400: Sending request (181 bytes) to ACME.COM
[24377] 1546443647.948401: Resolving hostname 192.168.47.120
[24377] 1546443647.948402: Sending initial UDP request to dgram 192.168.47.120:88
[24377] 1546443647.948403: Received answer (88 bytes) from dgram 192.168.47.120:88
[24377] 1546443647.948404: Sending DNS URI query for _kerberos.ACME.COM.
[24377] 1546443647.948405: No URI records found
[24377] 1546443647.948406: Sending DNS SRV query for _kerberos-master._udp.ACME.COM.
[24377] 1546443647.948407: Sending DNS SRV query for _kerberos-master._tcp.ACME.COM.
[24377] 1546443647.948408: No SRV records found
[24377] 1546443647.948409: Response was not from master KDC
[24377] 1546443647.948410: Received error from KDC: -1765328378/Client not found in Kerberos database
Code: 96c73a06
```

greghudson:
I'm looking into how to uncouple add_s4u_x509_user_padata() from the preauth list, and I am confused about how the client principal is currently communicated back to s4u_creds.c.  add_s4u_x509_user_padata() makes a copy of the request client and puts it in userid->user (where the caller can see it), but when that runs, won't the request client just be the dummy principal?  The KDC doesn't put S4U_X509_USER padata in the reply to the AS request, so it shouldn't run again.

iboukris:
I think it only cares about the realm, which supposedly gets updated upon receiving wrong_realm, so the dummy princ works fine for this purpose.

greghudson:
Okay, and after the referral, we process the optimistic preauth data again (because restart_init_creds_loop() is called and sets that up again) and update the caller-visible principal with the new realm.  Now it makes sense.

There is a cache check at s4u_creds.c:699 which uses the full client principal as reported by s4u_identify_user(), but that only runs when we have an input client principal.

iboukris:
hmmm, I have another bug, when both certificate and enterprise name are given we should do AS with the principal only and TGS with both, but we currently do AS with the certificate (thought I tested that..).

iboukris:
@greghudson, I fixed the cert+enterprise bug, and rebased.
I moved the tester code back to kvno as I think it could be useful for testing against Windows KDC. How do you suggest to do the read_file(), is there a better way?
Regarding must_preauth=0, I'll wait for your advise or code changes.

iboukris:
I'll drop "kvno: remove useless limitation .." commit for now, as now the service name is completely ignored which makes the code incoherent.

greghudson:
On the KDC side, I assume we would need to add code to the AS processing to find the subject cert in the padata, and probably a KDB method to look up a principal by cert.  Does that sound right, and were you planning to implement that at some point?

If we had that, we could add a feature to the test module and have a proper test for the client code.

iboukris:
Yes, these were exactly my thoughts by now (it would also help to test cross-realm based on certificate only, which I have not tested against Windows).

We'd need to invoke that new KDB method also in TGS request, when we process X509 padata if it comes with a certificate (I'll probe Windows KDC with mismatched principal and cert, to see which one is being preferred).

iboukris:
Yes, the KDC from service realm replies with preauth_required instead of wrong_realm, so it tries to get a ticket from its own realm and fails, added traces at:
https://github.com/iboukris/S4U/blob/master/win10_s4u2self_cross_realm_cert_only_no_san.pcapng

I may be missing some configuration that would have made the KDC of the server realm aware of the mapping configured in the client realm or something else, but I didn't find anything.
It shouldn't matter much, as we leave the mapping policies up to the KDB plugin. It should be trivial to implement some useful optimistic routing, based on certificate data.

greghudson:
>  Instead, I'll make it an alternative to get_principal() with no need to return entry/key, just a principal, then the server returns preauth_required if the principal is local, or wrong_realm if not.

We can't yield a proper preauth_required method-data list without a principal entry.  But I guess we don't need one if we see a subject certificate.  We'll want a good comment, since in this case the KDC knows it's abusing the preauth_required error code.

iboukris:
Sigh, I thought that for AS lookup with certificate Windows always return preauth_required without etype_info (even if the account does not require preauth). But now I see that it only does that to Windows client while to MIT client it responds with etype_info or with a ticket.
Not sure how it tells, I'll try to look more carefully but I think either way will be fine. An alternative get_pricipal() with optional returned entry, would be simple and flexible.

iboukris:
> But now I see that it only does that to Windows client while to MIT client it responds with etype_info or with a ticket. Not sure how it tells ...

It tells by the fact that MIT client sends an empty cname in AS lookup, if I change it to the subject certificate DN like Windows clients do, then I get the same reply. If you think it's worth the efforts to behave the same, then I'll add some code to decode the ASN1 and read the subject certificate DN (and perhaps check for SAN as well), but I think empty cname is fine as long as it works.

I pushed a revised version, there are still some todos and the doc is missing, but it's mostly ready I think.

greghudson:
Huh, that doesn't sound so much like non-Windows client detection as being conservative about omitting etype-info.  Kind of a weird thing to check, though; the presence of the pa-data type should be indication enough.

I think it's not worth the effort to decode the X.509 cert to set the cname, at least for now.

iboukris:
Yeah, I didn't mean to say that it does that by design, was just trying to pinpoint what makes the difference (it could even be just a bug).

iboukris:
The funny thing is that any string, with or without DN format, yields the same result as empty string. Only the proper subject certificate like windows client send, cause it to send an 'empty' preauth, even if client is not in the realm, which breaks cross realm.
Unlike what I said before, that no-SAN (only) breaks cross realm, what appears to be causing the break is when the subject cert is right, if I use MIT client to send no-SAN with empty string, I get proper preauth_require and cross realm works too (sorry, initially I only tested no-san exactly like windows client does). 
So yes, sending the subject cert DN would actually be a bad idea, as it will break cross realm, like Windows client is currently broken.

A slight change I'll do however, is to pass the cname in AS to the KDB handler as in_princ so the plugin can take a look if it desires.

iboukris:
Regarding the KDB method, perhaps it may be cleaner if I revert to the original suggestion so it only resolves cert to principal, but once the client principal gets resolved this way, the server also sets a new flag, like KRB5_KDB_FLAG_X509_PROTOCOL_TRANSITION. So if the plugin can only determine the realm of the client in AS stage, it can return \@UREALM and later check for the flag in get_principal() and return the same realm. Though I'm unsure if it will indeed be simpler.

src/plugins/kdb/test/kdb_test.c
```diff
@@ -535,7 +579,7 @@ test_check_allowed_to_delegate(krb5_context context,
 
 kdb_vftabl PLUGIN_SYMBOL_NAME(krb5_test, kdb_function_table) = {
     KRB5_KDB_DAL_MAJOR_VERSION,             /* major version number */
-    0,                                      /* minor version number 0 */
+    2,                                      /* minor version number 0 */
```
frozencemetery: Should be 1, yeah?

src/include/kdb.h
```diff
@@ -1398,7 +1404,19 @@ typedef struct _kdb_vftabl {
      */
     void (*free_principal_e_data)(krb5_context kcontext, krb5_octet *e_data);
 
-    /* End of minor version 1 for major version 6. */
+    /* End of minor version 1. */
+
+    /*
+     * Optional: TODO
```
frozencemetery: What's this for?

src/kdc/kdc_util.c
```diff
@@ -1623,14 +1636,22 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
...
         *princ_ptr = princ;
-    } else if (is_local_tgt) {
-        /*
-         * The server is asking to impersonate a principal from another realm,
-         * using a local TGT.  It should instead ask that principal's realm and
-         * follow referrals back to us.
-         */
-        *status = "S4U2SELF_CLIENT_NOT_OURS";
-        return KRB5KDC_ERR_POLICY; /* match Windows error */
+    } else {
```
frozencemetery: `return 0` at the end of the `if`, and then you can de-indent everything in the `else` block.

src/lib/kdb/kdb5.c
```diff
@@ -323,6 +323,8 @@ copy_vtable(const kdb_vftabl *in, kdb_vftabl *out)
     out->refresh_config = in->refresh_config;
     out->check_allowed_to_delegate = in->check_allowed_to_delegate;
     out->free_principal_e_data = in->free_principal_e_data;
+    /* TODO: should we first check minor version? */
```
frozencemetery: Yes, if I understand the question?

greghudson:
I've started reviewing these commits in detail, and will probably file a few tickets detailing what's currently wrong before revising the PR.  A few new thoughts:

* krb5_get_credentials_for_user() isn't a public API (it should have had a k5_ name, but it was added before we were being careful about that).  For applications, the only entry point is via gss_acquire_cred_impersonate_name(), which always passes a null subject cert.  So from an external perspective, the client side of this PR provides no new functionality except for kvno -F.  It doesn't need to be part of this PR, but did you have GSS support in mind for the future?

* kvno -F simply reads the file and puts it on the wire, so the file must contain the DER encoding of the cert.  I guess that's not hard to create with openssl (`openssl x509 -in cert.pem -outform DER -out cert.der`), but it wouldn't be too hard to read a PEM file as we have base64 functions in libkrb5support.  (PKCS12 support would be much harder, without using OpenSSL.)  I see PEM files in /etc/ssl/certs on my machine and it seems to be the default format for OpenSSL, so I think that might be a friendlier option than using the DER format.

* I'm not sure we've used the DAL min_ver field before.  Bumping the major version would be simplest, though I don't know how much of a hassle it's been downstream (in Samba/IPA/etc.) when we bump the DAL major version.

src/plugins/kdb/test/kdb_test.c
iboukris: Right. I was mislead by the comment in kdb.h ("End of minor version 1."), but I think it got decremented when the major version was incremented (see both commits with 'git log -S free_principal_e_data').

src/include/kdb.h
iboukris: You mean the missing doc? I'll do asap, still thinking the API a bit.

src/kdc/kdc_util.c
iboukris: Yeah, that would be cleaner. Will do.

src/lib/kdb/kdb5.c
iboukris: I'll add it the same check as free_principal_e_data had initially.

iboukris:
- GSS support is a good idea, but yeah let's leave it for a follow-up.

- Ok about base64, should it be internally in krb5_get_credentials_for_user() or in kvno.c ?

- See my comment above about min_ver, it seems we used it to add free_principal_e_data API.

greghudson:
> Ok about base64, should it be internally in krb5_get_credentials_for_user() or in kvno.c ?

For this PR I'd say put it in kvno.c.  If we decide to have a public API which takes a PEM filename, we can later move it into a library.

> See my comment above about min_ver, it seems we used it to add free_principal_e_data API.

I'd forgotten about that.  If we've used min_ver before, we can definitely use it again.

frozencemetery:

> * I'm not sure we've used the DAL min_ver field before.  Bumping the
> major version would be simplest, though I don't know how much of a
> hassle it's been downstream (in Samba/IPA/etc.) when we bump the DAL
> major version.

We have - we were 6.1 prior to being 7.0.

Dependency-wise, I just have to remember to upgrade them together, which
requires some coordination, but is doable.  I export a symbol
`kdbversion` into the package namespace that IPA &co. depend on.

The full change to IPA to go from 6.1 to 7.0 was this:
https://github.com/freeipa/freeipa/commit/9f8700fceead6e7b4947dc86f161e78dabb5d186

iboukris:
I've addressed the comments and todos. It looks ready to me now, please take a look.

I'm having second thoughts about PEM format though. I now realize that it's not just base64 decode, but we also need to parse the begin/end and maybe some more. I'd prefer to use a library for that if we really want it. Meanwhile, I changed the man page to require DER format.

greghudson:
I filed http://krbdev.mit.edu/rt/Ticket/Display.html?id=8777 with the consolidated list of deficiencies in the client code for this path.

It took me a while to figure out what the "￼Fix S4U2Self with X509 certificate TGS-REQ" commit was doing (and the corresponding description in the ticket is hard to follow).  I will consider how to untangle this part of the code to make it less fragile; if I don't come up with anything, I will add some comments.

iboukris:
Yes, that's a comprehensive and coherent list. Thanks for that.

greghudson:
> I'm having second thoughts about PEM format though. I now realize that it's not just base64 decode, but we also need to parse the begin/end and maybe some more.

I expected the begin and end line.  I assume "some more" refers to encapsulated headers, which would be expected in a privacy-enhanced mail message, but I don't think they're expected in certificates:

    $ cd /etc/ssl/certs
    $ grep ':' *.pem
    $

I revised the relevant commit to read PEM files and spot-tested it: https://github.com/greghudson/krb5/commit/e3593834991d1c2cd73d9d11e771cc626070049f

Of course it will break your tests, but can you decide if that seems like a reasonable amount of code?

iboukris:
I was not clear on the details initially, but your revision looks fine to me. I updated the commit and amended the tests accordingly. Also tested it with a real PEM certificate against Windows KDC.

greghudson:
I pushed the kinit man page fix to master since it was trivial and tangential to this PR.

I squashed the four client commits since they were of similar scope, and the resulting commit didn't seem too big to understand.

I still need to go over the last three commits (the KDC changes and the tests).

iboukris:
Not about "KDC: Fix cross-realm S4U2Self request with cert only", technically it shouldn't say "cert only" but "when using pa S4U_X509_USER only (without pa FOR_USER)", as implementation could chose to use only that, but in practice I've only seen it with cert-only.

greghudson:
I finished going over all of the commits.  The tests need a little bit of work.  I noticed a potentially thorny issue with caching when we identify the user by cert: we cache the ticket under the discovered principal name, but we don't know that name when we check the cache.  That's a recipe for unconstrained cache growth:

    $ kvno -F .../fake_cert user@SREALM
    user@SREALM: kvno = 1
    $ kvno -F .../fake_cert user@SREALM
    user@SREALM: kvno = 1
    $ klist
    [...]
    02/06/19 10:57:05  02/07/19 10:56:56  user@SREALM
	for client other@SREALM
    02/06/19 10:57:09  02/07/19 10:56:56  user@SREALM
	for client other@SREALM

Ideally we want to cache the entry under the cert so that we can find it again next time.  Barring that, we don't want to cache the result, since we should only cache things the caller asked for.  (Maybe this is why Microsoft constructs a principal using the cert subject.)

src/include/kdb.h
```diff
@@ -1389,16 +1395,35 @@ typedef struct _kdb_vftabl {
...
+     * not just a realm), the module should verify that it is the same as the
+     * lookup result for client_cert.  The module should not return a referral.
+     */
+    krb5_error_code (*get_s4u_x509_principal)(krb5_context kcontext,
+                                              const krb5_data *client_cert,
+                                              krb5_const_principal princ,
+                                              unsigned int flags,
+                                              krb5_db_entry **entry_out);
+
+    /* End of minor version 1 for major version 9. */
```
iboukris: Oups, this should be version 7. I'll fix.

iboukris:
> I finished going over all of the commits. The tests need a little bit of work. I noticed a potentially thorny issue with caching when we identify the user by cert: we cache the ticket under the discovered principal name, but we don't know that name when we check the cache. That's a recipe for unconstrained cache growth:
> 
> ```
> $ kvno -F .../fake_cert user@SREALM
> user@SREALM: kvno = 1
> $ kvno -F .../fake_cert user@SREALM
> user@SREALM: kvno = 1
> $ klist
> [...]
> 02/06/19 10:57:05  02/07/19 10:56:56  user@SREALM
> for client other@SREALM
> 02/06/19 10:57:09  02/07/19 10:56:56  user@SREALM
> for client other@SREALM
> ```
> Ideally we want to cache the entry under the cert so that we can find it again next time. Barring that, we don't want to cache the result, since we should only cache things the caller asked for. (Maybe this is why Microsoft constructs a principal using the cert subject.)

Hmmm, I had not thought of that. I'll look into it, and will try to clarify the test a bit. Thanks!

iboukris:
> we cache the ticket under the discovered principal name, but we don't know that name when we check the cache. That's a recipe for unconstrained cache growth

I think in theory, the same could have happened if the impersonate name (not realm) would have been canonicalized by the KDC (which never happens in practice).

> Ideally we want to cache the entry under the cert so that we can find it again next time. Barring that, we don't want to cache the result, since we should only cache things the caller asked for. (Maybe this is why Microsoft constructs a principal using the cert subject.)

Not sure I understand your thoughts above. I think we'd need to use the returned principal when we 'accept' the ticket as the service, or if we want to use it for S4U2Proxy since it matches the client inside the ticket. 
In my opinion we should just remove the old entry from cache if we already got the same for_user principal, and instruct the caller to save 'out_creds->client' and use it with the certificate next time so we can shortcut if we find it in cache.

iboukris:
BTW, the certificate-subject could be just a CN so it may not be unique, we could perhaps hash the certificate data, but I think simply overriding the old ticket would be fine.

greghudson:
Our existing ccache types do not allow us to remove entries.  Heimdal has code to "remove" FILE ccache entries by making them invisible, but they still take up space and have to be iterated over.  Therefore, our higher-level caching strategies can't rely on being able to remove or replace old entries.

iboukris:
Than we can avoid saving the new ticket and return success. They'd both have the same expiry based on the original TGT anyway, no? Not sure that's always the case though (btw, kvno saving the ticket in the same cache isn't very useful as it was pointed out recently on ML).

iboukris:
I added a patch in that spirit, let me know what do you think of this approach or what are our options here. I'd leave the test for duplicate tickets to later when we add gssapi, so that we could use check_ticket_count() we have added back then.

greghudson:
> (btw, kvno saving the ticket in the same cache isn't very useful as it was pointed out recently on ML).

Reference?

I can see three options:

1. (What you implemented) Check the cache after we get back the answer, and do not store a new entry if there is an old one matching the discovered client.  Further calls using only the cert will make KDC requests, but will not grow the cache.  Further calls using the cert and discovered client name will use the cached entry.

2. Don't cache the result.  Further calls using only the cert will make KDC requests as in (1).  Further calls using the cert and discovered client name will make one additional KDC request and cache it normally, and then use the cached entry.

3. Cache the result using a hash of the cert name.  Inelegant.  We don't check name type when comparing principals in the cache, so there's a vanishingly rare possibility of colliding with a real principal name.  Further calls using only the cert will used the cached entry.

I lean towards (2).  It's easy, I suspect that most usage patterns don't involve repeatedly asking for the same cert in the same cache, and it fits with our recent caching strategy of "cache only what the caller asked us for".  (3) also fits with that strategy, but requires a bunch of detail work.

iboukris:
http://mailman.mit.edu/pipermail/kerberos/2018-December/022132.html
An out-cache option could been useful, and I was trying to think if could help here as well.

iboukris:
I'd vote (1) so it stays similar to current API, and kvno -F results can intuitively be examined with klist.

Note that we do save the canonicalized client name in cache when doing kinit -C, and I think this case is somewhat analogous (getting new creds, and asking for client name canonicalization which is implicit here).

iboukris:
I did some testing on Windows with the lsa tester, while monitoring the traffic and the cache with klist.

- If I use username, then subsequent request is answered from cache (I need to purge, to trigger a request) and the cache does not grow.
- If I use certificate only (no-san) then it triggers a request on wire every time and the cache grows every time!
- If however, I clean the cache and use certificate followed by username (same as returned by kdc) then I get answer from cache, no second request is triggered and the cache does not grow.

greghudson:
Thanks for testing.  That sounds consistent with option 1, although consistency with LSA isn't really a requirement for caching behavior.  (Also, it doesn't sound like Windows using the cert subject as the client principal has anything to do with caching.)

Tangentially, why would a caller pass the cert if it has the principal name?  I get that LSA allows both to be passed and part of this PR is aligning our behavior with LSA's behavior in that case, but is there a reason not to just pass the principal if you have both?

iboukris:
> Tangentially, why would a caller pass the cert if it has the principal name? I get that LSA allows both to be passed and part of this PR is aligning our behavior with LSA's behavior in that case, but is there a reason not to just pass the principal if you have both?

I thought about it but I'm not sure. Maybe it helps for two factor auth, say using cert+radius that must be the same principal. Or maybe when reusing the principal from cache to ensure it matches the cert.

iboukris:
> I think in theory, the same could have happened if the impersonate name (not realm) would have been canonicalized by the KDC (which never happens in practice).

I was able to test this by forcing the KDC to modify the returned principal and the cache grows indeed, so I've expanded the check to include this case.

iboukris:
I've revised the tests and addressed the todos in the commit message.

iboukris:
> I'd vote (1) so it stays similar to current API, and kvno -F results can intuitively be examined with klist.

If we go with (2), then we could still call get_creds twice in kvno to force caching (unless that's what we want to avoid). Need to think of gssapi usage too. At any case I think we should keep the post get_ticket check, as the returned cname is unprotected and if modified in transit could cause the cache to grow.

iboukris:
> At any case I think we should keep the post get_ticket check, as the returned cname is unprotected and if modified in transit could cause the cache to grow.

Actually, the check isn't going to protect against this, since the interceptor can change the cname to a different one each time and the cache will still grow even with the check in place.

The only ways I can think of to protect against it are either:
- Never save s4u2self in cache (2's wise), instead instruct the caller to only save the ticket in cache after 'accepting' it as server, which makes sure the returned cname matches the one inside the ticket.
- Always assume that the KDC does not canonicalize the impersonate principal and check that the returned cname matches the requested impersonate principal (won't work for certificate).

greghudson:
In this case we discover the client name from S4U-X509-USER pa-data, which contains a checksum over the S4UUserID sequence in the TGT session key.  Is that not adequate protection?

greghudson:
I made a few minor updates:

* Edited some commit messages and comments I had written to use the hyphenated form of pa-data type names (I had been a little confused as to Microsoft's naming of those data types and had been using underscore forms as a result).

* Put the WIP caching commit before the tests commit since the tests don't pass without it.

* In the test script, saved a copy of the cache file and restored it instead of running kinit before each test.

* Expanded out the expected trace messages in the tests.  It's repetitive, but I think it's easier to understand what the test wants to see if the messages aren't pieced together.

iboukris:
> In this case we discover the client name from S4U-X509-USER pa-data, which contains a checksum over the S4UUserID sequence in the TGT session key. Is that not adequate protection?

That would have been true if we had compared the returned cname to the one in the returned S4U-X509-USER pa-data, which we currently don't. But maybe it's a good idea to do so (although not strictly part of the protocol afaict), at least for the certificate case. I'll look into it.

iboukris:
Forgot to mention on the PR, I added the above check to the wip commit and moved it where it should be squashed. Also moved the kvno commit a little down the line.

greghudson:
Sorry for the delay.

> That would have been true if we had compared the returned cname to the one in the returned S4U-X509-USER pa-data, which we currently don't. But maybe it's a good idea to do so (although not strictly part of the protocol afaict), at least for the certificate case.

Can we do that in all cases?  We might have to change how we remember the S4U-X509-USER principal from the first KDC.

I'd like to see this check be a separate clause from the current conditional at the end of the referral loop; folding "did we get the server we asked for as opposed to a referral" together with "does the unprotected client principal match the protected one" makes it hard to understand.  We can return KRB5_KDCREP_MODIFIED if we detect a mismatch.

iboukris:
> Can we do that in all cases? We might have to change how we remember the S4U-X509-USER principal from the first KDC.

Yeah, that make sense, it is simpler and more rigorous. Note that we now essentially prohibit client name canonicalization by old KDC who do not support S4U-X509-USER pa-data, but afaict this never happens anyway (apart of the certificate case, which requires support of this padata).

I've updated the WIP commit accordingly and unfolded the check, though I left it where it was to avoid conditioning on the error code again.

greghudson:
Okay, rebased to master, then squashed and adjusted slightly.  I think this is ready but will do a final pass over it before merging.

iboukris:
The adjustments look good, thanks. I've also ran a couple of tests against Windows and it looks ok.

iboukris:
I had another pass and pushed two tiny cleanups, see forced-pushed diff.

greghudson:
asan found a one-byte memory leak.  k5_make_tgs_req() copies the input padata such that a 0/NULL input turns into a 0/1-byte-allocation copy, so we need to free it in build_pa_s4u_x509_user().  I've pushed a fix.  (Also clarified the comment and shortened the post-search assertion.)

iboukris:
> k5_make_tgs_req() copies the input padata such that a 0/NULL input turns into a 0/1-byte-allocation copy

Okay I see, I didn't realize that. I could have caught it if I had asserted that (contents == NULL) as I wanted initially, instead of asserting that (length == 0).
For readability, I would prefer to leave the assertion as (padata[i]->pa_type == KRB5_PADATA_S4U_X509_USER) but I realize that (padata[i] != NULL) is technically the same.

Are the changes in ccache_def.rst related?

src/lib/krb5/krb/s4u_creds.c
```diff
@@ -503,15 +521,28 @@ krb5_get_self_cred_from_kdc(krb5_context context,
         krb5_pa_data **enc_padata = NULL;
         krb5_keyblock *subkey = NULL;
 
+        in_padata = calloc(3, sizeof(krb5_pa_data *));
+        if (in_padata == NULL) {
+            code = ENOMEM;
+            goto cleanup;
+        }
+
+        in_padata[0] = malloc(sizeof(krb5_pa_data));
```
frozencemetery: Here too.

greghudson:
The ccache_def.rst change is an artifact of me rebasing onto master at the same time.  (Once I have a HEAD to push that has both present, it's inconvenient to go back and do it in two steps.)  It's not part of the PR commits.

I will change the assertion to what you suggest, although personally I think "we found it" is just as readable as "what we found is what we were looking for" given that the loop is right there above.

src/lib/krb5/krb/s4u_creds.c
```diff
@@ -503,15 +521,28 @@ krb5_get_self_cred_from_kdc(krb5_context context,
         krb5_pa_data **enc_padata = NULL;
         krb5_keyblock *subkey = NULL;
 
+        in_padata = calloc(3, sizeof(krb5_pa_data *));
```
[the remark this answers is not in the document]
greghudson: k5calloc() here, but yes, I'll do that and in the case below.

iboukris:
> The ccache_def.rst change is an artifact of me rebasing onto master at the same time. 

Oh okay, I'd need to consider that when I look at forced-pushed diff.

> although personally I think "we found it" is just as readable as "what we found is what we were looking for" given that the loop is right there above.

Right, please leave it as (padata[i] != NULL). I didn't get it right away, as I forgot it was an iteration on a NULL terminated array.

src/clients/kvno/kvno.c
```diff
@@ -39,14 +42,17 @@ static void
 xusage()
 {
     fprintf(stderr, _("usage: %s [-C] [-u] [-c ccache] [-e etype]\n"), prog);
-    fprintf(stderr, _("\t[-k keytab] [-S sname] [-U for_user [-P]]\n"));
+    fprintf(stderr, _("\t[-k keytab] [-S sname] [-I | -U for_user | "
```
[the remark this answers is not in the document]
greghudson: I'll add braces.  It's already on one line as displayed to the user; it's just split across two lines in the source code.

src/clients/kvno/kvno.c
```diff
@@ -162,13 +181,102 @@ static void extended_com_err_fn(const char *myprog, errcode_t code,
...
+}
+
+/* Read a certificate from file_name in PEM format, placing the DER
+ * representation of the certificate in *der_out. */
+static krb5_error_code
+read_pem_file(char *file_name, krb5_data *der_out)
+{
+    krb5_error_code ret = 0;
+    FILE *fp = NULL;
+    const char *begin_line = "-----BEGIN ", *end_line = "-----END ";
```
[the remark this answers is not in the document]
greghudson: I don't think it's important.  PEM is a defunct standard; its way of storing certificates has managed to survive it (ironically, as one of the least specified parts of the standard).  Any PEM file you encounter in the wild is likely to be a certificate.  So it's not important to give a good error message on the offchance that someone tries to feed in a different sort of PEM file.

src/lib/kdb/kdb5.c
```diff
@@ -324,6 +324,12 @@ copy_vtable(const kdb_vftabl *in, kdb_vftabl *out)
     out->check_allowed_to_delegate = in->check_allowed_to_delegate;
     out->free_principal_e_data = in->free_principal_e_data;
 
+    /* Copy fields for minor version 1 (major version 7). */
+    assert(KRB5_KDB_DAL_MAJOR_VERSION == 7);
+    out->get_s4u_x509_principal = NULL;
+    if (in->min_ver >= 1)
```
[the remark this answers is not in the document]
greghudson: I will add to the commit message, but the assert belongs where it is; the goal is to ensure that the min_ver check is removed when the DAL major version is bumped.

iboukris:
I've added a small fixup, please take a look (sorry).
Otherwise, the PR looks good to me.

greghudson:
Any specific reason for the change?  "process" is a pretty generic verb in a function name, so it seems like a step down from "verify".

iboukris:
It's just that "verify" sounds like it doesn't modify anything, while now we do update the request.

greghudson:
I think it's acceptable, if not great, for the name to imply a restricted contract and then to have a little bleed which is explicitly documented (via the parameter name update_req_user and the function comment).  Adhering to the implied contract (i.e. going back to updating the request in the caller) could be argued as an improvement, although the resultant verbosity in the caller starts to indicate some refactoring there.  Giving the function a more general name and loading it with an additional responsibility does not seem better to me.

iboukris:
Fair enough. Please discard the commit, I'll remove later.

src/clients/kvno/kvno.c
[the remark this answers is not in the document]
kaduk: That feels a little hyperbolic, though is not really wrong:

```
kaduk@prolepsis:~/openssl/openssl$ git grep -h '^-----BEGIN'|sort|uniq -c
     11 -----BEGIN CERTIFICATE REQUEST-----
    307 -----BEGIN CERTIFICATE-----
     10 -----BEGIN DH PARAMETERS-----
      5 -----BEGIN DSA PARAMETERS-----
     28 -----BEGIN DSA PRIVATE KEY-----
    134 -----BEGIN EC PARAMETERS-----
      2 -----BEGIN EC PRIVATE KEY-----
      2 -----BEGIN PKCS7-----
    841 -----BEGIN PRIVATE KEY-----
   1195 -----BEGIN PUBLIC KEY-----
     26 -----BEGIN RSA PRIVATE KEY-----
      1 -----BEGIN SERVERINFO FOR CT-----
      1 -----BEGIN SERVERINFO FOR TACK-----
      1 -----BEGIN SERVERINFOV2 FOR CT-----
      2 -----BEGIN SSL SESSION PARAMETERS-----
     50 -----BEGIN TRUSTED CERTIFICATE-----
      3 -----BEGIN X509 CRL-----
```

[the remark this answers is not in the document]
greghudson: I'll add CERTIFICATE to the begin line check, since it sounds like public and private keys in PEM files might not be uncommon.

PR #888, opened 2019-01-02
let autoconf decide whether particular compier supports 128bit integers

Sashan:
This is yet another take on int128_t. The simple change does right thing for Solaris Studio and gcc.


greghudson:
Yes, closing this one.

PR #889, opened 2019-01-03
k5-platform: Avoid allocating a register for protecting memset()

cryptomilk:
See https://bugs.llvm.org/show_bug.cgi?id=15495

### DIFF
--- a/src/configure.in
+++ b/src/configure.in
@@ -421,7 +421,7 @@ AC_PROG_LEX
 AC_C_CONST
 AC_HEADER_DIRENT
 AC_FUNC_STRERROR_R
-AC_CHECK_FUNCS(strdup setvbuf seteuid setresuid setreuid setegid setresgid setregid setsid flock fchmod chmod strptime geteuid setenv unsetenv getenv gmtime_r localtime_r bswap16 bswap64 mkstemp getusershell access getcwd srand48 srand srandom stat strchr strerror timegm)
+AC_CHECK_FUNCS(strdup setvbuf seteuid setresuid setreuid setegid setresgid setregid setsid flock fchmod chmod strptime geteuid setenv unsetenv getenv gmtime_r localtime_r bswap16 bswap64 mkstemp getusershell access getcwd srand48 srand srandom stat strchr strerror timegm explicit_bzero explicit_memset)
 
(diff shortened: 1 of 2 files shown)


greghudson:
The assembly used to be `asm volatile ("" : : "g" (ptr), "g" (len));`.  Commit c163275f899b201dc2807b3ff2949d5e2ee7d838 changed it to the current form, after I received mail from Zhaomo Yang noting that the previous form still allowed the memset to be optimized out with gcc 5.1.  I chose the current form based on secure_memzero.h from Zhaomo's group.  In hindsight, it looks like adding `: "memory"` to the end of the asm statement was the important part, and changing `"g" (ptr), "g" (len)` to `"r" (ptr)` was unnecessary.

I would like to add explicit_bzero() support (glibc, FreeBSD, OpenBSD) and possibly explicit_memset() support (NetBSD) in the same commit.

greghudson:
I revised this with explicit_bzero() and explicit_memset() as alternatives.

cryptomilk:
Looks fine for me. The question is if it would make sense to rename zap() to explicit_bzero() if it is not available. Then it would be a documented function *most* people on Linux already know.

cryptomilk:
What I want to say, if someone sees `zap()` in the source, the doesn't know what the function does and why it is there. If he sees `explicit_bzero()` he might already know it, he might guess it from the name, but he could do `man explicit_bzero` and find documentation for it (also on the web).

greghudson:
The complication is that we would have to decide what to do about zapfree() and zapfreestr().  zapfree() is used more often than zap() currently.

frozencemetery:
@cryptomilk, you can't do that on any machine I have (arguably a bug in the `explict_bzero()` case because it's present in `bzero(3)`, and I don't have a NetBSD box right now, but nevertheless).

-1 to renaming; that's unnecessary churn (and I don't want to backport around that).  `zap()` is reasonably clear in my opinion, and the definition logic is very readable.  (ctags gets you there easily, for instance, and the definitions are all consecutive in the same file.)

greghudson:
As a matter of trivia, my Ubuntu 18.04 machine gives a man page for bzero/explicit_bzero in response to "man explicit_bzero".  But I think we will leave the name as zap/zapfree/zapfreestr.

PR #890, opened 2019-01-08
Improve S4U2Self realm identification internals

greghudson:
Realm identification for S4U2Self requests ([MS-SFU] 3.1.5.1.1.1) uses
the AS code path with some differences: we might want to include a
subject certificate in pa-data, we want to stop as soon as we get a
reply indicating which realm the client is in, and we want to
communicate that realm to the caller.  The current method of making
these changes is fragile--it uses an optimistic preauth type but does
not actually pre-authenticate, and it assumes that the AS code will
terminate with a predictable error if there is no prompter and a
trivial GAK function.

Instead, add fields to krb5_get_init_creds_context for realm
identification, and support them in the AS state machine, making sure
never to invoke preauth modules.  Add a new library-internal function
k5_identify_realm() to set up an appropriate context, run the state
machine, and copy out the client principal of the last request on
success.

### DIFF
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -227,6 +227,8 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
 #define TRACE_INIT_CREDS_GAK(c, salt, s2kparams)                    \
     TRACE(c, "Getting AS key, salt \"{data}\", params \"{data}\"",  \
           salt, s2kparams)
+#define TRACE_INIT_CREDS_IDENTIFIED_REALM(c, realm)                     \
+    TRACE(c, "Identified realm of client principal as {data}", realm)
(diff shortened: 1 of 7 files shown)


greghudson:
@iboukris please have a look.

src/lib/krb5/krb/gic_opt.c
```diff
@@ -29,6 +29,7 @@ struct extended_options {
     krb5_responder_fn responder;
     void *responder_data;
     int pac_request;            /* -1 unset, 0 false, 1 true */
+    krb5_boolean s4u_x509_user; /* private, used by s4u_creds.c */
```
frozencemetery: I think this is unused?

iboukris:
Thanks! Looks good to me, much cleaner. I'll rebase my work on top of it and run some tests.

PR #891, opened 2019-01-08
enctype logging updates

frozencemetery:
First commit augments the constants with their human-readable names.

Second commit augments that logging with all-caps information that the etype in question is deprecated.  I didn't mark the PKINIT etypes as deprecated because I couldn't find formal docs that they are, but am happy to change that.

### DIFF
--- a/src/clients/klist/klist.c
+++ b/src/clients/klist/klist.c
@@ -571,11 +571,17 @@ static char *
 etype_string(krb5_enctype enctype)
 {
     static char buf[100];
-    krb5_error_code ret;
+    char *bp = buf;
+    size_t deplen, buflen = sizeof(buf);
 
-    ret = krb5_enctype_to_name(enctype, FALSE, buf, sizeof(buf));
-    if (ret)
-        snprintf(buf, sizeof(buf), "etype %d", enctype);
+    if (krb5int_c_deprecated_enctype(enctype)) {
+        deplen = strlcpy(bp, "DEPRECATED:", buflen);
+        buflen -= deplen;
+        bp += deplen;
+    }
+
+    if (krb5_enctype_to_name(enctype, FALSE, bp, buflen))
+        snprintf(bp, buflen, "etype %d", enctype);
     return buf;
 }
 
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2082,6 +2082,7 @@ krb5_get_tgs_ktypes(krb5_context, krb5_const_principal, krb5_enctype **);
 krb5_boolean krb5_is_permitted_enctype(krb5_context, krb5_enctype);
 
 krb5_boolean KRB5_CALLCONV krb5int_c_weak_enctype(krb5_enctype);
+krb5_boolean KRB5_CALLCONV krb5int_c_deprecated_enctype(krb5_enctype);
 krb5_error_code k5_enctype_to_ssf(krb5_enctype enctype, unsigned int *ssf_out);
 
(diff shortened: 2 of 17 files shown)


greghudson:
We have krb5_c_enctype_to_name(), which I would rather use than have a new enctype table outside of libk5crypto.  It doesn't cover the PKINIT types, so we would still need a KDC wrapper to take care of non-3961 enctypes we expect to see in KDC requests.  Per RFC 4556 section 3.1.4, I don't expect that table to grow.

I don't recall any IETF action deprecating any of those CMS types in the context of PKINIT.  Some of them do look inadvisable, like md5WithRSAEncryption and perhaps rc2-cbc.

frozencemetery:
Updated to wrap `krb5_enctype_to_string()` instead, and to flag deprecations there.

greghudson:
I had suggested krb5_c_enctype_to_name(), not krb5_enctype_to_string(); the latter produces rather long output which can't be used as enctype names in input.  Flagging deprecation warnings in krb5_c_enctype_to_name() definitely wouldn't be appropriate, though.

I think the change to the flow control in log_tgs_req() doesn't belong in the same commit as the functional change (which isn't very invasive to the code; it just affects the management of ktypestr and rep_etypestr).  I'm not sure the change is needed at all; there isn't a lot of code in the then-branch of the current if statement, and the else branch isn't an error case.

ktypes2str() and rep_etypes2str() can return NULL.  Using an empty string would be okay, as would just not logging, since allocation errors are rare, but the code shouldn't admit the possibility of passing a NULL corresponding to a `%s` format string (which is undefined, although implementations often render it as `(null)` in practice).

kaduk:
> I don't recall any IETF action deprecating any of those CMS types in the context of PKINIT. Some of them do look inadvisable, like md5WithRSAEncryption and perhaps rc2-cbc.

I had briefly considered doing so as part of what is now RFC 8429, but punted since I wasn't sure at the time how to do the research to confirm that it was safe and correct to do so.  It seems reasonable to start up another document if there's an author that has energy.

frozencemetery:
Updated.   I missed that you didn't want `krb5_enctype_to_name()` to flag deprecations and did it anyway.  I'm leaving it in here because I'm not sure there's value in flagging at all unless it's done there (i.e., I'd just drop the second commit).  For my understanding, why would you consider that inappropriate?  Is it because there's some expectation that it return exact RFC-compliant names?  (i.e., if you don't want this upstream and I apply it downstream, how much of a mess am I setting myself up for?)

greghudson:
krb5_enctype_to_name() was intended as an inverse of krb5_string_to_enctype().  If you add a deprecation marker there, it's no longer an inverse.

frozencemetery:
Thanks, that's helpful.  Revised series to only perform deprecation warnings in `klist(1)` and KDC logs.  Retaining the buffer size bump as a separate commit (second in the series right now).

frozencemetery:
Updated from out-of-band review.
- Renamed `krb5int_enctype_to_name()` wrapper to `enctype_name()` for clarity.
- Solved the rd_req_dec buffer size problem by moving that to use `krb5int_enctype_to_name()`.
- Added a commit to remove the last in-tree call to `krb5_enctype_to_string()` (in kpropd debug logging).
- Added a commit to introduce a flag for deprecated enctypes and a function to check them.
- Expanded deprecated logging to include kadmin interaction and K/M load-in.

greghudson:
Pushed minor changes, hopefully fixing the appveyor build in the process.

PR #893, opened 2019-01-22
Match ccache functions in krb5.h with export list

frozencemetery:
r20743 (part of 1.7) added new ccache functions to krb5.h, but not
either exports list.  Though ad5aa12f13aad7ec4cafcffeec4f2e84e56c9c78
exported krb5_cc_move, the rest remain unexported.

krb5_cc_lock, krb5_cc_unlock, krb5_cccol_lock, and krb5_cccol_unlock
are used internally, so remove them from krb5.h and rename k5_cc_lock,
k5_cc_unlock, k5_cccol_lock, and k5_cccol_unlock, respectively.

krb5_cccol_last_change_time is not used, so remove it.

krb5_cc_last_change_time is also not used, so remove it as well.
Update ccache interface to reflect removal.  Of particular note, this
function didn't behave as documented for MEMORY or KEYRING ccaches at
time of removal.

ticket: 7765

### DIFF
--- a/doc/appdev/h5l_mit_apidiff.rst
+++ b/doc/appdev/h5l_mit_apidiff.rst
@@ -19,10 +19,7 @@ Differences between Heimdal and MIT Kerberos API
   :c:func:`krb5_auth_con_setsendsubkey()`  H5l: Not implemented as of version 1.3.3
   :c:func:`krb5_cc_set_config()`           MIT: Before version 1.10 it was assumed that
                                            the last argument *data* is ALWAYS non-zero.
-  :c:func:`krb5_cccol_last_change_time()`  H5l takes 3 arguments: krb5_context context,
-                                           const char \*type, krb5_timestamp \*change_time
-                                           MIT takes two arguments: krb5_context context,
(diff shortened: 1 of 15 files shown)


frozencemetery:
Oh I have delved too greedily and too deeply, it seems.  What do we actually want to do with `krb5_cc_last_change_time()`?  It looks like removal on the Windows side is... complicated.

greghudson:
Isn't it just a matter of removing krb5_stdccv3_last_change_time() and the entry in krb5_cc_stdcc_ops?  (Also one of the NULLs in the `#else` clause.  cc_mslsa.c also has a NULL entry for lastchange that should be removed.)

frozencemetery:
Yeah, that and a couple other things fix the Windows build.

What persists though is `cc_ccache_get_change_time()` - this is part of the CredentialsCache API, which is versioned.  Since previous versions seem to be still supported, it's not clear to me how we'd go about removing a function here.  (For some reason I thought these two problems were related, which upon reflection they aren't much.)

PR #894, opened 2019-02-11
Always pass tgt-client principal to sign_authdata

iboukris:
to allows a plugin handling s4u2self request, to verify the PAC from the TGT and to generate a new one for the impersonate principal in case the latter is from local realm.

We have discussed it back then, and while relevant section in MS-SFU (3.2.5.x) is ambiguous to me,  Windows KDC seem to require the PAC to exist (which implies that it verifies it), as the following fails with err_c_principal_unknown (works fine if I omit the -U).

`$ echo -n pwd | kinit apache@ACME.COM --no-request-pac && kvno
-Uuser@abc@ACME.COM apache@ACME.COM`

I've tested this prove-of-concept change with Samba plugin patched accordingly, in cross realm trust with WIndows KDC (both ways): https://github.com/iboukris/samba/commit/610afde8cf9efabe24768aa58183f5349e660df5

I've also consulted on samba-technical and got the same feedback that we should verify it.
CC @cryptomilk @abbra

### DIFF
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -1018,9 +1018,9 @@ typedef struct _kdb_vftabl {
      *     requested; also set by the admin interface.  Determines whether the
      *     module should return in-realm aliases.
      *
-     * A module can return in-realm aliases if KRB5_KDB_FLAG_ALIAS_OK is set,
-     * or if search_for->type is KRB5_NT_ENTERPRISE_PRINCIPAL.  To return an
-     * in-realm alias, fill in a different value for entries->princ than the
+     * A module can lookup a principal by alias name if it so desires, but can
+     * only return in-realm aliases if KRB5_KDB_FLAG_ALIAS_OK is set. To return
+     * an in-realm alias, fill in a different value for entries->princ than the
      * one requested.
      *
      * A module can return out-of-realm referrals if KRB5_KDB_FLAG_CANONICALIZE
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -589,9 +589,11 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
      * Note that according to the referrals draft we should
      * always canonicalize enterprise principal names.
      */
-    if (isflagset(state->request->kdc_options, KDC_OPT_CANONICALIZE) ||
-        state->request->client->type == KRB5_NT_ENTERPRISE_PRINCIPAL) {
+    if (isflagset(state->request->kdc_options, KDC_OPT_CANONICALIZE)) {
         setflag(state->c_flags, KRB5_KDB_FLAG_CANONICALIZE);
+        setflag(state->c_flags, KRB5_KDB_FLAG_ALIAS_OK);
+    }
+    if (state->request->client->type == KRB5_NT_ENTERPRISE_PRINCIPAL) {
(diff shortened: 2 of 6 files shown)


greghudson:
I think I see some inconsistency in the ordering of client_princ and for_user_princ in the different changes.

iboukris:
Where? I can't find, it is always client_princ followed by for_user_princ (often followed by client db handle).

greghudson:
Oh, I see my confusion.  Does client_princ differ from client->princ in a way that's important to the DB module?

iboukris:
I think it may, haven't check it thoroughly though.

iboukris:
In the S4U2Self case, client db is of the impersonate principal, if client->princ is the canonical name it may not be the same as the one in padata FOR_USER which we need to sign with (while client_princ is the tgt client we need to verify with).

In regular TGS request if initial credentials were acquired as enterprise name without name canonicalization from Windows KDC, then I think the PAC would be signed with enterprise while client->princ may the canonical name - not sure, will run some tests.

iboukris:
My assumption about client->princ being always canonical was wrong (it's only so when canonicalization is requested or implied).
So I think the only change we need is when the impersonate principal is from local realm as we need to pass 'enc_tkt_req->client' to verify the PAC (which is not the same as 'enc_tkt_reply->client' when the service is local too). To sign the new PAC, the plugin can use 'client->princ', which matches for_user_princ.

I've updated the PR and my poc Samba patch accordingly, and tested again with Windows, see:
https://github.com/iboukris/samba/commit/778e899bc6326883d25ad16fb5bfcfea1442998b

iboukris:
Between client->princ and server->princ, I manage to get most scenarios working by changing only the plugin code, without kdc changes.
One scenario however requires the change, when the impersonate principal is local but not the service as the plugin is not provided with the name of the service so is unable to verify the PAC.

`$ kinit apache@ACME.COM && kvno -I user@MIT.NET apache@ACME.COM`

See printouts just before plugin's sign_authdata() is called, only the last there are currently passed.

```
enc_tkt_req->client: apache@ACME.COM
enc_tkt_reply->client: apache@ACME.COM
for_user_princ: user@MIT.NET

actual_client: user@MIT.NET
client->princ: user@MIT.NET
server->princ: krbtgt/ACME.COM@MIT.NET
```

iboukris:
I added a wip test, cross-realm enterprise gave troubles so I disabled it for now. Also, I didn't realize initially that I was adding an authdata rather than replacing it, so I added the 'last' variable. I'm thinking now, maybe I can create a minimal PAC and use the pac_sign/verify_ex methods instead.

iboukris:
I figured out the failure of the enterprise name case. Since we now treat enterprise as ALIAS_OK we end up returning the canonical name from kdb, which causes a discrepancy in sign_authdata between 'client->princ' and the 'for-user' principal. Back then I only meant to lookup the principal by alias name, not to canonicalize it (sometimes referred to as hard-alias). To allow implementing Windows behavior more closely, perhaps we could always allow the module to do hard-aliases (it can do it anyway), but  still require KRB5_KDB_FLAG_ALIAS_OK for soft ones (where the name is canonicalized).

I added a WIP commit in that spirit, though it needs more thinking and work.

iboukris:
Closing this as I hope to address it as part of #912

PR #900, opened 2019-02-26
Remove unnecessary NULL-check in mk_priv.c

frozencemetery:
encode_krb5_enc_priv_part() will initialize der_encpart on success, so
it can never be NULL at cleanup-time.  Reported by Coverity.


greghudson:
This one is a bit tricky.

For robustness, pointers which get freed in the cleanup label should be initialized to NULL (I missed that in this case, but it's not what Coverity is complaining about and it doesn't create a bug here) and the cleanup code should handle NULL and non-NULL values.  In most cases that conditional is buried within a free function, but here we want to zap the data before freeing it.  I might add a zapfreedata() helper in k5-int.h, which would make Coverity stop complaining.

frozencemetery:
I believe this was superseded by #906.  Thanks!

PR #901, opened 2019-03-01
Memory replay cache for priv/safe/cred messages

greghudson:


greghudson:
I realized the memory replay cache probably wants to be process-global.  That way if a DO_TIME application doesn't use subkeys, but does do most of its work within one process, it still gets protection across auth contexts.  Also, for PR #899, if we fall back to a memory replay cache, we don't want it to be specific to a single rcache handle.

I will revise this PR to use a mutex-locked global hash table and expiration queue.

greghudson:
I realized that if the memory replay cache is process-global, and client and server DO_TIME auth_contexts are located in the same process, they will see each others' messages as replays.  So while I think the fallback memory cache for rd_req should be process-global, we want a separate memory cache for messages.  So this PR should be considered as-is.  Sorry for the confusion.

src/lib/krb5/rcache/memrcache.c
```diff
@@ -0,0 +1,165 @@
...
+    ret = krb5_timeofday(context, &now);
+    if (ret)
+        return ret;
+
+    /* Check if we already have a matching entry. */
+    e = k5_hashtab_get(mrc->hash_table, tag->data, tag->length);
+    if (e != NULL)
+        return KRB5KRB_AP_ERR_REPEAT;
+
+    /* Discard stale entries. */
```
frozencemetery: Shouldn't this happen before access to the hashtable?  It seems better to check for expirations on every call to `k5_memrcache_store()` (not just the ones that actually add entries).

greghudson: I think it's fine either way.  A store which reports a replay doesn't grow the cache, so it isn't important to remove old entries.  I put the replay check first out of a vague sense that it's better not to let a bad actor cause any sort of memory write, but that's not really supported.

frozencemetery: (per out-of-band conversation, this is fine as is - collision chance is very low)

PR #905, opened 2019-03-13
Add option to build without libkeyutils

kellinm:
- There are instances where it is advantageous to be able to
  turn off the keyring ccache feature when compiling on Linux.
  This patch adds --disable-keyring-ccache as an option to
  the configure script.

Signed-off-by: Robert Marshall <rmarshall@gitlab.com>

### DIFF
--- a/doc/build/options2configure.rst
+++ b/doc/build/options2configure.rst
@@ -387,6 +387,10 @@ Optional packages
     ``--without-krb5-config`` to disable the use of krb5-config and
     use the usual built-in defaults.
 
+**-**\ **-without-keyutils**
+    Build without libkeyutils support.  This disables the KEYRING
+    credential cache type.
+
 
 Examples
 --------
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -93,8 +93,6 @@ KRB5_LIB_PARAMS
 KRB5_AC_INITFINI
 KRB5_AC_ENABLE_THREADS
 KRB5_AC_FIND_DLOPEN
-KRB5_AC_KEYRING_CCACHE
-KRB5_AC_PERSISTENT_KEYRING
 ])dnl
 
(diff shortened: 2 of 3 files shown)


frozencemetery:
I'm having trouble coming up with any.  Could you elaborate on why you want to do this, rather than just configuring a different ccache?

kellinm:
Our use case is embedding krb5 into an installer based on Chef's Omnibus project. We do not need or use the libkeyutil features and, as an external library dependency, it causes our healthchecks to fail.

As the libkeyutils is an optional add on only if detected and, according to comments in the source, only on Linux, it seemed reasonable to add a configure flag such that it could be disabled explicitly.

greghudson:
Based on the use case, I would prefer to call this --without-keyutils (using AC_ARG_WITH instead of AC_ARG_ENABLE).

kellinm:
@greghudson - I concur with your assessment and am amenable to changing the name of the configure option.

It seemed to me that a downstream packager would also expect this flag to disable the persistent keyring functionality.  I made that change as well for consistency when updating the pull request.

greghudson:
I made some changes:

* aclocal.m4 is a relic of when we used to have multiple configure.in scripts, and we've been slowly moving away from it.  Since we're making non-trivial changes to the keyutils check, I moved it to configure.ac.

* If --with-keyutils is explicitly given, we should error out if we can't find libkeyutils, similarly to other --with options.

* Persistent keyring support is a sub-feature of the KEYRING ccache.  For clarity, the check for it should be nested under successfully finding libkeyutils, so I moved it there.

kellinm:
I have no objection to those changes; thank you for pointing out the coming change in location from aclocal.m4 to configure.ac and moving the change over.

I'm comfortable with the changes and the patch as it stands, what's our next step to move toward merging?

PR #911, opened 2019-04-01
Implement krb5_cc_remove_cred for remaining types

frozencemetery:
Previously, only KCM and MSLA implemented credential removal.  Add
support for MEMORY and FILE (and therefore DIR), matching Heimdal's
implementations.  Also add a new implementation for KEYRING.

### DIFF
--- a/src/lib/krb5/ccache/cc_file.c
+++ b/src/lib/krb5/ccache/cc_file.c
@@ -744,6 +744,14 @@ cleanup:
     return set_errmsg_filename(context, ret, data->filename);
 }
 
+/* Return true if cred is a removed entry (assuming that no legitimate cred
+ * entries will have authtime=-1 and endtime=0). */
+static inline krb5_boolean
+cred_removed(krb5_creds *c)
+{
+    return c->times.endtime == 0 && c->times.authtime == -1;
+}
+
 /* Get the next credential from the cache file. */
 static krb5_error_code KRB5_CALLCONV
 fcc_next_cred(krb5_context context, krb5_ccache id, krb5_cc_cursor *cursor,
@@ -765,19 +773,30 @@ fcc_next_cred(krb5_context context, krb5_ccache id, krb5_cc_cursor *cursor,
         goto cleanup;
     file_locked = TRUE;
 
-    /* Load a marshalled cred into memory. */
-    ret = get_size(context, fcursor->fp, &maxsize);
-    if (ret)
-        goto cleanup;
-    ret = load_cred(context, fcursor->fp, fcursor->version, maxsize, &buf);
-    if (ret)
-        goto cleanup;
-    ret = k5_buf_status(&buf);
-    if (ret)
-        goto cleanup;
+    for (;;) {
+        /* Load a marshalled cred into memory. */
+        ret = get_size(context, fcursor->fp, &maxsize);
+        if (ret)
+            goto cleanup;
+        ret = load_cred(context, fcursor->fp, fcursor->version, maxsize, &buf);
+        if (ret)
+            goto cleanup;
+        ret = k5_buf_status(&buf);
+        if (ret)
+            goto cleanup;
 
-    /* Unmarshal it from buf into creds. */
-    ret = k5_unmarshal_cred(buf.data, buf.len, fcursor->version, creds);
+        /* Unmarshal it from buf into creds. */
+        ret = k5_unmarshal_cred(buf.data, buf.len, fcursor->version, creds);
+        if (ret)
+            goto cleanup;
+
+        /* Keep going if this entry has been removed; otherwise stop. */
+        if (!cred_removed(creds))
+            break;
+
+        k5_buf_truncate(&buf, 0);
+        krb5_free_cred_contents(context, creds);
+    }
 
 cleanup:
     if (file_locked)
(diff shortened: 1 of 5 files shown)


greghudson:
The mcc cursor structure holds a pointer to the next credential in the linked list.  If that credential is removed (potentially through a different cache handle in a different thread), the iterator's pointer is invalidated.  We can talk about design options for addressing that on the call tomorrow.

I haven't checked if there's a similar problem for keyring.

(Also to check: does Heimdal have a means to skip removed creds when iterating?  We don't want them to be visible in klist.)

frozencemetery:
Future calls with the same key_serial_t will indicate that the key is no longer in the keyring (ENOKEY if this removes all links to it).

greghudson:
Summarizing the call:

* Robbie notes that krb5_cc_start_seq_get's documentation says "If cache is modified between the time of the call to this function and the time of the final krb5_cc_end_seq_get(), the results are undefined."  This is kind of a ridiculous statement; the caller has no control over whether another process modifies the cache during an iteration, using a different handle opened to the same name.  Even for memory ccaches, a non-trivial application can't usually ensure that one thread doesn't open the memory ccache by name and modify it while another thread is iterating through it.

* The simplest thing to do is leave behind empty nodes (krb5_mcc_link structures with creds=NULL) instead of removing them from the list, and skip those nodes during iteration.  This obviously doesn't perform as well as we'd hope if a caller is doing many adds and removals, but that's not really a problem.  It's much more likely that we will need to address the many-adds case (probably introducing a hash table on the service principal name) than the many-adds-many-removes case.

* Should we need a more performant solution for that case, some ideas:
  * Assign each entry an index (increasing by one every time a cred is added).  In addition to the linked list, there is a hash table mapping index numbers to entries.  At each iteration step we store the index of the next entry.  If during iteration the next index isn't present in the table, we keep looking up successive numbers until we find an entry or reach the maximum index.
  * Assign an index to each entry as above, and organize the nodes into a balanced binary tree sorted by index.  If during iteration the next index isn't present in the tree, we can easily find the node with the next-higher index.
  * In either of the above schemes, the iterator state could also store a pointer to the next list node, and use a sub-generation counter to detect if that pointer might be invalid.  The sub-generation counter would increase whenever a removal operation happens.  This way we retain a low constant factor for the common case.
  * This has to be a frequently-solved problem since many languages have to implement ordered lists with non-ephemeral iterators.  It's hard to find literature via google search because the results are filled with tutorials on how to use language-specific containers rather than how to implement them, but we could look at language implementations for ideas.

* For the KEYRING ccache type, the iterator needs to detect ENOKEY errors and skip to the next key ID in the list.

frozencemetery:
I think in the Linux world we usually solve it with RCU, but I agree that we should only worry about making that more performant when needed.

- Updated comment in krb5.hin
- Heimdal doesn't have a way to skip FILE creds during iteration that I can see.  Setting the server realm isn't a general solution either (since it could be shorter, and we need the overwrite object's size to match).  gssproxy-encrypted creds are stored with endtime 0, so we can't just skip based on that, but they also have starttime 0, which suggests an alternate approach I've included here.
- Adjust test suite for above.
- Change MEMORY to preserve empty cred objects in the link.
- Updated KEYRING iterator to handle removed key
- Added test support for iteration-with-removal

src/lib/krb5/ccache/cc_file.c
```diff
@@ -765,19 +765,35 @@ fcc_next_cred(krb5_context context, krb5_ccache id, krb5_cc_cursor *cursor,
...
+        ret = k5_unmarshal_cred(buf.data, buf.len, fcursor->version, creds);
+        if (ret)
+            goto cleanup;
 
-    /* Unmarshal it from buf into creds. */
-    ret = k5_unmarshal_cred(buf.data, buf.len, fcursor->version, creds);
+        /*
+         * Skip creds that have been removed by fcc_remove_cred.  Avoid skipping
+         * creds that were never valid (such as gssproxy's), which need to appear
+         * in listings.
```
greghudson: I think endtime == authtime == 0 also applies to config entries.

The reversed nature of the conditional makes this code a little hard to read, so we might want a trivial predicate function for "cred is removed".

src/lib/krb5/ccache/cc_file.c
```diff
@@ -1002,12 +1018,144 @@ fcc_store(krb5_context context, krb5_ccache id, krb5_creds *creds)
...
+
+    /* Mark the cred expired so that it will hopefully be skipped over by any
+     * future match checks. */
+    cred->times.endtime = 0;
+
+    /* For config creds, endtimes aren't checked, so change the realm.
+     * Heimdal uses X-RMED-CONF here, so we do too. */
+    if (cred->server != NULL && cred->server->realm.length > 0 &&
+        strcmp(cred->server->realm.data, "X-CACHECONF:") == 0)
+        strcpy(cred->server->realm.data, "X-RMED-CONF:");
```
greghudson: I think any use of strcpy() will create a Coverity defect; you might need to use memcpy() instead.

src/lib/krb5/ccache/cc_file.c
```diff
@@ -1002,12 +1018,144 @@ fcc_store(krb5_context context, krb5_ccache id, krb5_creds *creds)
...
+    if (ret)
+        goto cleanup;
+
+    if (expected.len != overwrite.len) {
+        ret = KRB5_CC_FORMAT;
+        goto cleanup;
+    }
+
+    /* Get a non-O_APPEND handle to the raw file. */
+    fd = open(fdata->filename, O_RDWR);
```
greghudson: Add O_BINARY here for Windows, and O_CLOEXEC.

src/lib/krb5/ccache/cc_file.c
```diff
@@ -1002,12 +1018,144 @@ fcc_store(krb5_context context, krb5_ccache id, krb5_creds *creds)
...
+    if (fd == -1) {
+        ret = interpret_errno(context, errno);
+        goto cleanup;
+    }
+
+    ret = krb5_lock_file(context, fd, KRB5_LOCKMODE_EXCLUSIVE);
```
greghudson: I don't think we need to lock; the worst that happens is we race with another thread or process trying to remove the same credential, performing the same write operation.

src/lib/krb5/ccache/cc_file.c
```diff
@@ -1002,12 +1018,144 @@ fcc_store(krb5_context context, krb5_ccache id, krb5_creds *creds)
...
+    if (ret != 0) {
+        close(fd);
+        fd = -1;
+        goto cleanup;
+    }
+
+    start_offset = ftell(fcursor->fp) - expected.len;
```
greghudson: ftell() can in principal return -1 and set errno; although this is unlikely, static analyzers might ding us for not checking.

src/lib/krb5/ccache/cc_keyring.c
```diff
@@ -1028,35 +1028,45 @@ krcc_next_cred(krb5_context context, krb5_ccache id, krb5_cc_cursor *cursor,
...
-    if (krcursor->currkey >= krcursor->numkeys)
-        return KRB5_CC_END;
-
-    /* If we're pointing at the entry with the principal, or at the key
-     * with the time offsets, skip it. */
-    while (krcursor->keys[krcursor->currkey] == krcursor->princ_id ||
-           krcursor->keys[krcursor->currkey] == krcursor->offsets_id) {
-        krcursor->currkey++;
-        /* Check if we have now reached the end */
+    while (1) {
```
greghudson: Perhaps `while (krcursor->currkey >= krcursor->numkeys)` and return KRB5_CC_END if the loop terminates?

src/lib/krb5/ccache/cc_memory.c
```diff
@@ -583,16 +592,33 @@ krb5_mcc_retrieve(krb5_context context, krb5_ccache id, krb5_flags whichfields,
...
+    krb5_mcc_data *data = (krb5_mcc_data *)cache->data;
+    krb5_mcc_link *l;
+    krb5_creds *found;
+
+    k5_cc_mutex_lock(context, &data->lock);
+
+    for (l = data->link; l != NULL; l = l->next) {
+        if (l->creds != NULL &&
+            krb5int_cc_creds_match_request(context, flags, creds, l->creds)) {
+            found = l->creds;
```
greghudson: We have the cache locked, so we can just free the cred via l->creds and then set it to NULL, without the intermediary variable `found`.

src/lib/krb5/ccache/cc_file.c
frozencemetery: gssproxy stores an encrypted copy of the actual credentials in the ccache; this prevents the applications from actually using them directly.  I'm unsure if there's a reason the times being 0 is important; maybe @simo5 can help with that?

frozencemetery:
Updated, hopefully addressing all comments except the skipcheck, which hopefully @simo5 can help with.

src/lib/krb5/ccache/cc_file.c
simo5: I do not recall if the times were important, sorry.

greghudson: It's a step beyond Heimdal, but we could set authtime=~0 and endtime=0 when deleting creds.  That combination should be safe to skip, since psuedo-creds such as config entries should generally have authtime=0 and real creds should have endtime!=0.

src/lib/krb5/ccache/cc_file.c
greghudson: On further reflection, the situation here is muddier than I thought, because of the sanity check.  Two processes racing to delete the same cred will be trying to perform the same write, but one of them might not see the expected contents when it reads what was previously there.  This won't corrupt the ccache, but will produce a spurious KRB5_CC_FORMAT error.

However, locking here does not solve that problem, because the lock comes too late--we read in the previous cred entry before obtaining the lock, so another process could get in and remove the cred before our process obtains the write lock.

(Also, I'm surprised that obtaining a write lock here doesn't deadlock against the read lock obtained for iteration, on kernels recent enough to have OFD locking.)

Options I can see include:

1. Remove the sanity check, trusting that it wouldn't fail in practice anyway.  If we do this, I would like to pass in the start_offset from fcc_remove_cred() (which can do an ftell() on the file handle before each call to fcc_next_cred()) and sanity-check the difference in file offsets against the length of the new marhshalled cred.

2. Instead of comparing byte for byte, unmarshal the old byte values and check it... or don't check it, because if it unmarshals successfully, we at least skipped back to the beginning of a cred entry.

3. Arrange to obtain the write lock when fcc_remove_cred() starts iterating.  Not my preferred option.

frozencemetery:
Updated, now setting authtime to -1.

src/lib/krb5/ccache/cc_file.c
frozencemetery: The file descriptor isn't kept locked between iteration calls, so there's no contention and therefore no deadlock.

Heimdal's approach is best-effort to the point that `cred_delete()` can't pass an error back up.  If we're not getting a lock from `fcc_remove_cred()`, it seems like the best option is to make any failure here graceful.  How would you feel about:

4. Make the `memcmp()` failure a non-failure - just return 0 there instead of `KRB5_CC_FORMAT`.

greghudson: I think that would be okay too.  If the byte comparison doesn't succeed, either another process removed the same cred before we got to it, or the cache was reinitialized.  Either way, the cred is no longer present in the cache.  (I hadn't thought about the possibility of the cache being reinitialized, or I wouldn't have listed option 1 and would have phrased option 2 differently.)

(Well, unless our marshalling functions aren't perfect reversals of each other.  If we had a good way to save the result of load_cred() while iterating that wouldn't be a concern.  But the marshalling functions should round-trip correctly.)

frozencemetery:
Updated to make the `memcmp()` failure non-fatal.  Also renamed `cred_delete()` to `delete_cred()`.

PR #914, opened 2019-04-11
Remove some likely-unused code

frozencemetery:
The approach to ovsec dump removal was previously discussed in #913.

I believe that ovsec dump and sam2 are both unusable in their current form.

confvalidator has not been kept updated, and doesn't seem to be finished.  While such a tool might be useful, parsing krb5.conf using something other than krb5 itself (e.g., through the profile interface) is a fraught endeavor (as we've learned the hard way in freeipa and crypto-policies).


greghudson:
We use sam2 preauth at MIT, so it's definitely usable and we can't get rid of it.

I'm okay with removing dump -ov (as Nico points out, if someone really needs to upgrade from a prehistoric version, they can always use an intermediate version), but I'm not sure what would render it unusable.

frozencemetery:
Dropped SAM changes since they're no longer in scope; will account for that feedback in a future PR.  Per voice conversation, reverted the ovsec commit to a full removal and fixed up its commit message a bit.

PR #921, opened 2019-04-24
Secure getenv

greghudson:
The first commit is a placeholder for now.  I'm submitting this early to have a place to discuss implementation details (or at least record my own thoughts on them).

The goal here is to automatically ignore environment variables in our libraries when the process has higher privileges than its environment.  Currently we have a programmatic way to create "secure" krb5_contexts which ignore most environment variables; we have no story for GSS.

On glibc 2.17+, secure_getenv() solves the problem.  Its semantics seem solid as far as I can tell.  None of the other platforms I looked at have secure_getenv(), and I don't think other libc implementations for Linux have it.

On Windows there is, as far as I know, never a need to ignore the process environment.

An imperfect but simple fallback to detect elevated privileges is "getuid() != geteuid() || getgid() != getegid()", as libev and probably a bunch of other libraries do.  This doesn't detect exotic system features (such as executables having elevated privileges other than running as a different uid or gid), and it doesn't detect the case where an executable started off with elevated privileges, acquired a privileged resource or read in privileged data, then dropped privileges.

The BSDs have issetugid(), but unfortunately FreeBSD (and NetBSD according to Nico) messed it up, returning true if a process run as root lowered its privilege level using setuid() or similar.  We don't want to ignore environment variables in that case because the process environment is more privileged than the process, not less.  Ignoring the process environment for programs like Apache httpd would be a large inconvenience.

Nico did a bunch of research on this for Heimdal and came up with libroken/{secure_getenv.c,issuid.c,getauxval.c}.  The sum of these is more code than I'm really comfortable with.  The full logic is:

1. If getauxval(AT_SECURE) is available and succeeds, use that answer authoritatively.
2. If getauxval(AT_EUID), getauxval(AT_UID), getauxval(AT_EGID), and getauxval(AT_GID) all succeed and the uid/euid and gid/egid match, decide there is no elevated privilege.  If a uid/euid or gid/egid mismatch is detected, decide there is elevated privilege.  I believe these are the values from process start time, so this check won't miss the temporary-setuid case. 
3. If issetugid() is available and says 0, decide there is no elevated privilege.  But because of FreeBSD
4. If getauxval(AT_EXECPATH) succeeds and can be stat()d, answer authoritatively based on whether S_ISUID or S_ISGID is present in the mode.  (The file there isn't guaranteed to be the one we're executing, but since setuid/setgid programs are installed that way by root, there's no realistic attack scenario.)
5. Detect runtime uid/euid/saved-uid and gid/egid/saved-gid mismatches using getresuid() and getresgid() if possible.  Otherwise check using getuid()/geteuid()/getgid()/getegid().

Rather than use the system getauxval(), libroken uses its own implementation which reads /proc/self/auxv and parses it.  getauxval() itself seems to only be present in glibc 2.16+, and appears to have been broken (doesn't set errno on failure) before glibc 2.19.  The libroken implementation may also work on Solaris/Illumos, and on Linux systems with older glibc versions.

There is also the question of memoization.  Heimdal's implementation memoizes inside issuid() using a static integer with no locking, beginning with the value -1 and changing to 0 or 1.  This probably works okay in a practical threaded environment, but might run afoul of tools which detect thread-safety violations.  We could do more careful memoization using a library initializer, or an explicit pthread_once() in secure_getenv().

### DIFF
--- a/src/util/support/secure_getenv.c
+++ b/src/util/support/secure_getenv.c
@@ -1,0 +1,111 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/* util/support/secure_getenv.c - secure_getenv() portability support */
+/*
+ * Copyright (C) 2019 by the Massachusetts Institute of Technology.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * This file contains the fallback implementation for secure_getenv(), which is
+ * currently only provided by glibc 2.17 and later.  The goal is to ignore the
+ * environment if this process is (or previously was) running at elevated
+ * privilege compared to the calling process.
+ *
+ * In this fallback version we compare the real and effective uid/gid, and also
+ * compare the saved uid/gid if possible.  These comparisons detect a setuid or
+ * setgid process which is still running with elevated privilege; if we can
+ * fetch the saved uid/gid, we also detect a process which has temporarily
+ * dropped privilege with seteuid() or setegid().  These comparisons do not
+ * detect the case where a setuid or setgid process has permanently dropped
+ * privilege before the library initializer ran; this is not ideal because such
+ * a process may possess a privileged resource or have privileged information
+ * in its address space.
+ *
+ * Heimdal also looks at the ELF aux vector in /proc/self/auxv to determine the
+ * starting uid/euid/gid/euid on Solaris/Illumos and NetBSD.  On FreeBSD this
+ * approach can determine the executable path to do a stat() check.  We do not
+ * go to this length due to the amount of code required.
+ *
+ * The BSDs and Solaris provide issetugid(), but the FreeBSD and NetBSD
+ * versions are not useful; they return true if a non-setuid/setgid executable
+ * is run by root and drops privilege, such as Apache httpd.  We do not want to
+ * ignore the process environment in this case.
+ *
+ * On some platforms a process may have elevated privilege via mechanisms other
+ * than setuid/setgid.  glibc's secure_getenv() should detect these cases on
+ * Linux; we do not detect them in this fallback version.
+ */
+
+#include "k5-platform.h"
+
+static int elevated_privilege = 0;
+
+MAKE_INIT_FUNCTION(k5_secure_getenv_init);
+
+int
+k5_secure_getenv_init()
+{
+    int saved_errno = errno;
+
+#ifdef HAVE_GETRESUID
+    {
+        uid_t r, e, s;
+        if (getresuid(&r, &e, &s) == 0) {
+            if (r != e || r != s)
+                elevated_privilege = 1;
+        }
+    }
+#else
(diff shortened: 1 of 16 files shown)


greghudson:
Going over the Heimdal measures by benefit:

getuid/geteuid/getgid/getegid: all POSIX platforms; detects a setuid/setgid process which is still running at the elevated euid or egid.

getresuid/getresgid: Linux and BSDs (and HPUX); detects a setuid/setgid process which has temporarily dropped privilege with seteuid() or setegid() (but not one which has permanently dropped privilege with setuid() or setgid()).

secure_getenv(): Linux with glibc 2.17+; detects all elevated privilege scenarios as far as we know.

issetugid(): BSDs; unless restricted to OpenBSD, of little value because it can only be used as a negative signal.  Could save a system call.

System getauxval (not used by Heimdal): no additional benefit.  Would benefit glibc 2.16 but the version there can't be safely used.

Hand-coded getauxval: Solaris/Illumos and NetBSD for start-time uid/euid/gid/euid checks.  On FreeBSD the executable path can be determined, for a stat() check, but the uid/gid information doesn't seem to be included.  OpenBSD doesn't seem to have /proc/self/auxv.

My inclination is to do getresuid/getresgid with getuid/geteuid/getgid/getegid as fallback, and not go as far as the hand-coded getauxval because it's too much code for the benefit.  (And of course we're doing secure_getenv() implicitly, because we'll use the API name secure_getenv() and only implement it ourselves when it isn't available.)

frozencemetery:
Memoization: do we actually need this?  If unsure, we could always leave it out for now and add it later if this introduces a performance issue.

/proc: FreeBSD and OpenBSD don't seem to support this at all (and NetBSD requires some additional setup for Linux emulation).  It doesn't seem like having this custom getauxval buys us much, other than the glibc concerns (i.e., below 2.16 or 2.17, depending, which don't matter to any LTS distro even).

So I think I agree with your inclination.

src/util/support/secure_getenv.c
```diff
@@ -0,0 +1,111 @@
...
+k5_secure_getenv_init()
+{
+    int save_errno = errno;
+
+#ifdef HAVE_GETRESUID
+    {
+        uid_t r, e, s;
+        if (getresuid(&r, &e, &s) == 0) {
+            if (r != e || r != s)
+                elevated_privilege = 1;
```
frozencemetery: A minor optimization here would be to have `k5_secure_getenv_init()` exit after setting `elevated_privilege`.  It simplifies some checks further down, too.

src/util/support/secure_getenv.c
```diff
@@ -0,0 +1,111 @@
...
+    errno = saved_errno;
+    return 0;
+}
+
+char *
+k5_secure_getenv(const char *name)
+{
+    if (CALL_INIT_FUNCTION(k5_secure_getenv_init) != 0)
+        return NULL;
+    return elevated_privilege ? NULL : getenv(name);
```
frozencemetery: If we're okay to assume that `getenv` exists, can ksu lose its custom implementation?

greghudson:
I looked at the FreeBSD kernel code and I think it does support /proc/self/auxv, though of the values we care about only AT_EXECPATH is supported.  I don't have a FreeBSD machine to check, though.

An advantage of using a library initializer (beyond memoization) is that in some situations, the initializer will run early in the process's lifetime, before the process has a chance to drop privileges.  This isn't guaranteed--libkrb5support might not be loaded until later in the process lifetime, and on some platforms library initializers get run lazily via the CALL_INIT_FUNCTION macro.

src/util/support/secure_getenv.c
greghudson: It would simplify the checks further down at the cost of having to reset errno to saved_errno in each exit case.  It wouldn't save any system calls (which is the point of checking elevated_privilege further down).

src/util/support/secure_getenv.c
greghudson: Yes, although the majority of the code there is for setenv().  We can lose that too, though, as it's in POSIX.1-2001.

martinpitt:
This broke GSS authentication in [Cockpit](https://cockpit-project.org/). It uses a [suid root authentictaion helper cockpit-session](https://github.com/cockpit-project/cockpit/blob/master/src/ws/session.c#L410) which can be called from the unprivileged web server only, so that the web server itself does not need to run as root. This [sets the keytab for the server](https://github.com/cockpit-project/cockpit/blob/master/src/ws/session.c#L410) , which lives in /etc/cockpit/krb5.keytab:

      setenv ("KRB5_KTNAME", COCKPIT_KTAB, 1);

But with this change this is not respected any more, as `secure_getenv  ("KRB5_KTNAME") == NULL`. Note that this is not really insecure -- it's the privileged root process *itself* that sets this value, it's not inherited from the environment. A simple `setuid (geteuid ())` and `setgid (getegid ())` is not enough to fool `secure_getenv()`, as that looks at `AT_SECURE` auxval. So there's not much that cockpit-session can do to work around that (by design of secure_getenv).

There is unfortunately [no other way to set the keytab](https://web.mit.edu/kerberos/krb5-latest/doc/basic/keytab_def.html) than through the environment, as changing the global configuration file is out of the question for cockpit-session -- this is a service specific keytab. Is there some library API to set the keytab name that the docs failed to mention? As the env variable now became unavailable, I see no way how to do that now.

I was thinking of running cockpit-session in an unshared namespace and bind-mounting the custom keytab over the default one, but this is not practical - that process starts a user session which should be in the default namespace,  it would actually break cockpit's storage administration functions to do that. And temporarily bind-mounting it in the default namespace would introduce a race condition for other krb5 clients.

greghudson:
> Is there some library API to set the keytab name that the docs failed to mention?

Since 1.11, you can use gss_acquire_cred_from().  I thought this was documented in https://web.mit.edu/kerberos/krb5-latest/doc/appdev/gssapi.html but unfortunately it appears not to be.  See https://k5wiki.kerberos.org/wiki/Projects/Credential_Store_extensions for now.  A "keytab" entry in the cred_store key/value set will determine the keytab name.

There are some older extensions which can do this as well: gss_krb5_import_cred() and krb5_gss_register_acceptor_identity().

martinpitt:
@greghudson : Awesome, thanks for your super-fast reply! This works fine indeed. There doesn't seem to be a way of using a `NULL` store or a store with zero entries to make it just use the default location, so that one needs a structure like
```
if (have_custom_keytab)
    gss_acquire_cred_from(...);
else
    gss_acquire_cred (...);
```

but this is only a minor annoyance. This works well enough now, thank you!

greghudson:
> There doesn't seem to be a way of using a NULL store or a store with zero entries to make it just use the default location

I think either of those should work.  Absent documentation or a spec, it's not clear whether a NULL cred_store is allowed, but GSS functions generally allow null defaults wherever possible, and the code appears to allow it (in both MIT krb5 and Heimdal).  A zero-element cred store is definitely allowed and should definitely use the default keytab.

PR #922, opened 2019-05-02
Static analysis update 2019-05-02

frozencemetery:
Here are (some) things flagged by various analyzers in a pass today.  If there's a better way I can organize any of this please let me know.

### DIFF
--- a/src/clients/klist/klist.c
+++ b/src/clients/klist/klist.c
@@ -720,11 +720,6 @@ show_credential(krb5_creds *cred)
         extra_field += 2;
     }
 
-    if (extra_field > 3) {
-        fputs("\n", stdout);
-        extra_field = 0;
-    }
-
     if (show_flags) {
         flags = flags_string(cred);
         if (flags && *flags) {
--- a/src/lib/kdb/kdb_convert.c
+++ b/src/lib/kdb/kdb_convert.c
@@ -305,8 +305,6 @@ ulog_conv_2logentry(krb5_context context, krb5_db_entry *entry,
     krb5_error_code ret;
     kdbe_attr_type_t *attr_types;
     int kadm_data_yes;
-    /* always exclude non-replicated attributes, for now */
-    krb5_boolean exclude_nra = TRUE;
 
     nattrs = tmpint = 0;
     final = -1;
@@ -356,7 +354,8 @@ ulog_conv_2logentry(krb5_context context, krb5_db_entry *entry,
             nattrs++;
         }
     } else {
-        find_changed_attrs(curr, entry, exclude_nra, attr_types, &nattrs);
+        /* Always exclude non-replicated attributes for now. */
+        find_changed_attrs(curr, entry, TRUE, attr_types, &nattrs);
         krb5_db_free_principal(context, curr);
     }
 
@@ -399,31 +398,6 @@ ulog_conv_2logentry(krb5_context context, krb5_db_entry *entry,
                 ULOG_ENTRY_TYPE(update, ++final).av_type = AT_PW_EXP;
                 ULOG_ENTRY(update, final).av_pw_exp =
                     (uint32_t)entry->pw_expiration;
-            }
-            break;
-
-        case AT_LAST_SUCCESS:
-            if (!exclude_nra && entry->last_success >= 0) {
-                ULOG_ENTRY_TYPE(update, ++final).av_type = AT_LAST_SUCCESS;
-                ULOG_ENTRY(update, final).av_last_success =
-                    (uint32_t)entry->last_success;
-            }
-            break;
-
-        case AT_LAST_FAILED:
-            if (!exclude_nra && entry->last_failed >= 0) {
-                ULOG_ENTRY_TYPE(update, ++final).av_type = AT_LAST_FAILED;
-                ULOG_ENTRY(update, final).av_last_failed =
-                    (uint32_t)entry->last_failed;
-            }
-            break;
-
-        case AT_FAIL_AUTH_COUNT:
-            if (!exclude_nra) {
-                ULOG_ENTRY_TYPE(update, ++final).av_type =
-                    AT_FAIL_AUTH_COUNT;
-                ULOG_ENTRY(update, final).av_fail_auth_count =
-                    (uint32_t)entry->fail_auth_count;
             }
             break;
 
@@ -552,10 +526,8 @@ ulog_conv_2logentry(krb5_context context, krb5_db_entry *entry,
 /* END CSTYLED */
 
         case AT_LEN:
-            if (entry->len >= 0) {
-                ULOG_ENTRY_TYPE(update, ++final).av_type = AT_LEN;
-                ULOG_ENTRY(update, final).av_len = (int16_t)entry->len;
-            }
+            ULOG_ENTRY_TYPE(update, ++final).av_type = AT_LEN;
+            ULOG_ENTRY(update, final).av_len = (int16_t)entry->len;
             break;
 
         default:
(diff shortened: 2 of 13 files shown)


frozencemetery:
I'm amused by the static analysis PR failing the CI static analysis checks.  Updated.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -5932,7 +5932,7 @@ pkcs7_dataDecode(krb5_context context,
     }
     if (ek != NULL) {
         OPENSSL_cleanse(ek, eklen);
-        OPENSSL_free(ek);
+        free(ek);
```
greghudson: There is a (minor) problem here, but the fix isn't so simple, because ek could be replaced with tkey, which is allocated with OPENSSL_malloc().  I think there's also a memory leak at line 5903.

We can change tkey to be allocated normally, and use zapfree() instead of OPENSSL_cleanse() and OPENSSL_free() for both values.

Out of curiosity, what tool flagged this?

frozencemetery: I believe it is Coverity - from the ALLOC_FREE_MISMATCH checker.

greghudson:
I squashed five of the unused-return-code patches and pushed those, leaving seven commits.

src/lib/krb5/krb/preauth_sam2.c
```diff
@@ -150,7 +150,6 @@ sam2_process(krb5_context context, krb5_clpreauth_moddata moddata,
         /* Go ahead and get it now, preserving the ordering of  */
         /* prompts for the user.                                */
 
-        salt = ctx->default_salt ? NULL : &ctx->salt;
```
greghudson: This one is weird, and probably requires a closer look.

The gak_fct() call here should pass salt instead of &ctx->salt, so that krb5_get_as_key_password() computes and uses the default salt instead of using an empty salt, when the KDC didn't send an explicit salt.  That's a bug in commit 787cc1e2411 or thereabouts, and the unused value of salt is correctly flagged by Coverity.

However, the result of this gak_fct() call is ignored, because at line 224 we explicitly call krb5_c_string_to_key() on the password with the salt we correctly determined at lines 200-211.  That apparently redundant operation dates back much longer.

I would like to fix the gak_fct() call and then eliminate the krb5_c_string_to_key() call at line 224.  I will need to do some testing.

We unfortunately still need lines 200-211 as long as we support either of the non-SEND_ENCRYPTED_SAD cases.

frozencemetery: I believe you've superseded this with https://github.com/krb5/krb5/pull/924 so dropped from this series.

frozencemetery:
Folded the mismatch-removal into the top commit.  Dropped the tautological comparisons commit.  Redid the `pkcs7_dataDecode()` changes into a single modernization commit (largely because I couldn't read it as it was; something about whitespace), fixing the memory handling in the process.

greghudson:
I made some changes to the first commit, but I don't think it's quite right; ek and tkey are leaked on the failure exit path.

This is the tricky case for doing memory management with an exit label.  We have one allocated object (etmp) whose ownership is transferred to the caller on success, but should be freed on failure.  We have other allocated objects which should be freed all the time.  In a typical modern krb5 function we handle this with an explicit ownership transfer on the accept path:

        *something_out = ptr;
        ptr = NULL;
    
    cleanup:
        free(ptr);
        ....

But in this case we're returning etmp rather than putting it into an output parameter.  So we would need to create a new BIO * variable for the return value.

What I would really like to do is fold this function into its caller and use EVP_CipherUpdate() instead of the meaningless BIO manipulations.  (These extra manipulations arise because pkcs7_dataDecode() is derived from OpenSSL's PKCS7_datadecode(), and we didn't fully adapt it to its local use.)  Even better would be a one-shot decrypt call--you'd think OpenSSL would have one of those for the common case where the caller has all of the ciphertext in a single memory region, but as far as I know it does not.

frozencemetery:
(Per voice conversation, will wait to see your version since you've looked into the problem further at this point.)

greghudson:
Pushed a new first commit (and removed the pkinit_crypto_openssl.c change from the last commit since it no longer applies).

I decided to remove the "Some S/MIME clients don't use the same key and effective key length" workaround, since this code is not for S/MIME.  This code path is also only used for PKINIT RSA mode, which I believe is basically never used (and which I wish had never been specified), so the change is unlikely to bite us.

greghudson:
I pushed an updated second commit (with a bit of git fumbling).  There is some ambiguity as to whether `cpw -pw a` reflects a missing password or a missing principal name; I opted for the latter because I think it makes each argc check make more sense in context.

greghudson:
"Always free creds in krb5_ccache_copy"": isn't this one of like 17 memory leaks defects in ksu, none of which are important since it's a short-lived process?  My position on these is that it's okay to fix them all at once, but that fixing them one at a time is too much noise in the commit history.

I pushed an updated version of it but I think I would prefer to drop it for now.

src/clients/klist/klist.c
```diff
@@ -720,11 +720,6 @@ show_credential(krb5_creds *cred)
         extra_field += 2;
     }
 
-    if (extra_field > 3) {
-        fputs("\n", stdout);
-        extra_field = 0;
-    }
-
```
greghudson: This one seemed interesting, so I looked into it further.

The dead code was added in commit 3f7b5fca80c0aaeffcd9348c70613af5b63a61f1, which was scoped to adding the klist -e option.  Prior to that commit, instead of `extra_field` we had `first`, and it was just there to track whether to use a tab or ", " prefix and whether to output a newline at the end; we never stuck a newline in the middle.

We can output extra information about (1) a non-default client principal, (2) the renew-till time, (3) the flags, and (4) the etype information (after the commit).  The inoperable code would add a newline between (2) and (3) if it ever fired.

There is also a non-dead check which adds a newline between (3) and (4) if extra_field > 2.  For whatever reason, the code increments extra_field by 2 for (2) and by 1 for everything else.  (It is true that the flags display is pretty short compared to a renew-till time, but a client principal name is not.)  So that conditional triggers if (2) and at least one of (1) and (3) are present.  The net result is a fairly nice display if (2), (3), and (4) are all present:

	renew until 05/10/2019 11:02:55, Flags: FfPRAT
	Etype (skey, tkt): aes256-cts-hmac-sha1-96, aes256-cts-hmac-sha1-96 

but perhaps not so nice if only (2) or (3) are present:

	renew until 05/10/2019 11:02:55, Etype (skey, tkt): aes256-cts-hmac-sha1-96, aes256-cts-hmac-sha1-96 

	Flags: I, Etype (skey, tkt): aes256-cts-hmac-sha1-96, aes256-cts-hmac-sha1-96 

I think adding a newline between (2) and (3) was out of scope for the commit, and should generally not be necessary (non-default client principals are rare), so I will take the change as is.

src/lib/kdb/kdb_convert.c
```diff
@@ -411,20 +411,7 @@ ulog_conv_2logentry(krb5_context context, krb5_db_entry *entry,
...
-            }
-            break;
-
         case AT_FAIL_AUTH_COUNT:
-            if (!exclude_nra) {
-                ULOG_ENTRY_TYPE(update, ++final).av_type =
-                    AT_FAIL_AUTH_COUNT;
-                ULOG_ENTRY(update, final).av_fail_auth_count =
-                    (uint32_t)entry->fail_auth_count;
-            }
```
greghudson: Here we're removing some speculative code, based on the comment "always exclude non-replicated attributes, for now".

This is probably fine; I don't see any circumstance where we'd want to transmit these fields over iprop.  But at the same time we should hit AT_LAST_SUCCESS, remove the case labels (there's already a default-break), and remove the exclude_nra variable and just pass TRUE to find_changed_attrs().

frozencemetery:
You're right - that shouldn't have snuck in.  I got bit by the function naming violation.  Will remove.

frozencemetery:
Updated, dropping commit and adjusting kdb_convert.c.

greghudson:
> Updated, dropping commit and adjusting kdb_convert.c.

I'm not seeing an update to the kdb_convert.c changes.

frozencemetery:
Forgot `-a`.  Should be there now.

PR #923, opened 2019-05-02
Remove single-DES support

frozencemetery:
This is technically part of a larger series at https://github.com/krb5/krb5/compare/master...frozencemetery:die-des which removes 3des support as well.  From previous conversations, it's my understanding that upstream doesn't want the full series yet and would prefer a more gradual deprecation path.  I do plan to deploy it in Fedora soon however, so any extra eyes are appreciated.

v4 salttype removal isn't technically necessary, but afs3 is.  (Use of afs3 salttypes is limited to single-DES principals.)

Attempting to kinit a prinicipal with a modern enctype and a v4 salt using a version of krb5 with these patches applied produces a line like this in krb5kdc.log:

```
May 02 22:28:14 localhost.localdomain krb5kdc[13933](info): AS_REQ (8 etypes {aes256-cts-hmac-sha1-96(18), aes128-cts-hmac-sha1-96(17), aes256-cts-hmac-sha384-192(20), aes128-cts-hmac-sha256-128(19), DEPRECATED:des3-cbc-sha1(16), DEPRECATED:arcfour-hmac(23), camellia128-cts-cmac(25), camellia256-cts-cmac(26)}) ::1: KDC_RETURN_PADATA: v4@EXAMPLE.COM for krbtgt/EXAMPLE.COM@EXAMPLE.COM, Unsupported salt type
```

though this error only appears as generic error to the client kinit.

The only other not-strictly-necessary change in here should be the doc/kadm5 tex files.  I can move that to a separate PR if that would be preferable, or delete them entirely.

Finally, I know this is a large change, so my thanks and appreciation in advance for reviewing it.


src/lib/kdb/kdb_cpw.c
```diff
@@ -316,18 +316,6 @@ add_key_pwd(context, master_key, ks_tuple, ks_tuple_count, passwd,
...
-            key_salt.data.data = 0;
-            break;
-        case KRB5_KDB_SALTTYPE_AFS3:
-            retval = krb5int_copy_data_contents(context,
-                                                &db_entry->princ->realm,
-                                                &key_salt.data);
-            if (retval)
-                return retval;
-            s2k_params = &afs_params;
-            break;
```
greghudson: This is the only code to set s2k_params, so the variable should be eliminated and we should call krb5_c_string_to_key_with_params() with NULL.

frozencemetery:
Nice, updated.

greghudson:
Initial impressions:

* kdc_req_sumtype, default_ap_req_sumtype, and default_safe_sumtype all seem independent of single-DES.  Perhaps we don't need them any more, but removing them isn't part of removing DES support.  kdc_req_sumtype isn't used for modern enctypes but is used for RC4.

* def_kslist isn't actually used.  In kdb5_ldap_realm.c it's completely dead; in kdb5_create.c it's used in the initializer for rblock, but rblock is completely reinitialized in kdb5_create(), so the whole initializer can be dropped.

* The change to cc_mslsa.c in the final commit is wrong; we still need to call krb5_get_tgs_ktypes().  If it errors we can error out instead of defaulting to DES.

* The changes to weak enctype documentation seem a bit off, and the placement of this commit is awkward because it precedes the removal of single-DES.  Probably it can be folded into the final commit.  For the three variables that default to `|defetypes|`, since there will no longer be a weak enctype in the default set, the whole "but..." clause can be removed.  The documentation for `allow_weak_crypto` should probably just say that it enables single-DES support in 1.8-1.17 and doesn't do anything in 1.18+.  In the session key selection section, the final sentence of the first paragraph should be removed, since we will no longer assume services support any enctypes other than the ones they have keys for.  (That documentation also wasn't updated for des_crc_session_supported, but that's moot now.)

* In t_invalid.c, test_bogus_1964_token() should use different sign and seal algs, but shouldn't be removed.  test_short_checksum() doesn't have any relationship to single-DES and should remain.

* It is possible that we should remember the enctype names but refuse to encrypt or decrypt with them, instead of forgetting the names.  I will do some testing and form a more solid opinion about that.  I am  concerned about a kadmin client being unable to display DES key metadata on an old server, and am a little concerned that "-des" or "-des-cbc-crc" in an enctype string will produce a trace message warning.  (The checksum types are less of a concern.)

frozencemetery:
- `default_ap_req_sumtype`, `default_safe_sumtype`, and `kdc_req_sumtype` are used for ap_req_checksum_type, safe_checksum_type, and kdc_req_checksum_type, respectively.  Per documentation in krb5.conf, those fields are only used for DES keys.  I will look into that further and see what I missed.

- The problem is that allow_weak_crypto still does things in 1.18+: it brings back export-grade RC4.  I will clarify that.  (This is why I've been against adding new things to the `allow_weak_crypto` set - this result is surprising.)  If you wanted me to do an export-RC4 removal for the 1.18-timeframe I'm happy to do that, though single-DES removal is a higher priority for me.

Will update with the rest of the changes.

frozencemetery:
Yeah, these uses contradict the docs: kdc_req_sumtype, default_safe_sumtype, and default_ap_req_sumtype all can manipulate RC4.  Which direction would you like this resolved in?  (And if it's a doc update keeping these handles around, does that mean you want CRC-32 kept because it can be configure with RC4?)

greghudson:
> Yeah, these uses contradict the docs: kdc_req_sumtype, default_safe_sumtype, and default_ap_req_sumtype all can manipulate RC4.

I may be missing something.  kdc_req_sumtype is documented as only applying to DES keys, and in reality applies to DES and RC4 keys.  The other two types don't seem to be documented that way.

default_safe_sumtype is restricted by krb5_c_keyed_checksum_types() for the enctype, which is somewhat limiting but still allows hmac-md5-rc4 (which does not actually use RC4) for any enctype.

default_ap_req_sumtype doesn't appear to be limited at all.

greghudson:
To clarify: I do think it's a good idea to remove the three checksum type profile variables, and it's not wholly unrelated to removing single-DES (in that every enctype after des-cbc-crc has a closely associated keyed checksum type, so the need for configuration of checksum types is effectively a throwback to the original enctype).  But it should be a separate commit with a separate ticket number.

frozencemetery:
I see.  Since safe_checksum_type and ap_req_checksum_type don't mention encryption types but do mention kdc_req_checksum_type, I read them is inheriting the restrictions from there.

I'll open a separate ticket.

frozencemetery:
Opened #928.  This PR is based off of that branch; I'm happy to move things around if you have a different preferred merge order.

greghudson:
"Update default krb5kdc mkey enctype for manual entry": this should use DEFAULT_KDC_ENCTYPE for consistency with kadmind and kdb5_util.  (Having to configure this is a bit silly, since at least in the common case K/M contains enough information to make a better guess than some hardcoded default.  But that kind of improvement is out of scope for this PR.)

frozencemetery:
Makes sense, updated.

greghudson:
I merged all of the side-issue commits, leaving just three commits to remove the v4/afs3 salttypes and the single-DES enctypes and cksumtypes.

For removing the v4 salt type, I got the following testing results:

* Dump and load will successfully round-trip unknown salt types.
* kadmin getprinc displays the unknown salt type as "<Salt type 0x1>".
* kadmind can present the unknown salt type to the client, which will display it as either "v4" or as above depending on whether it still knows about the v4 salt type.
* kinit displays a generic error (as Robbie noted), with a reasonable line in kdc.log.  If a principal has aes256-cts:v4,aes128-cts:normal keys, the KDC will not fall back to the aes128 key; it selects the aes256-cts key according to enctype preferences and then errors out trying to generate etype-info.

For removing the des-cbc-crc enctype:

* Dump and load will successfully round-trip unknown enctypes.
* kadmin getprinc displays the unknown enctype as "<Encryption type 0x1>".
* kadmind can present the unknown enctype to the client.
* kinit displays "KDC has no support for encryption type" if a principal only has a des-cbc-crc key.  This applies if either the client is new (because it doesn't include des-cbc-crc in its request enctypes, so the KDC can't select a reply key) or if the KDC is new.

I think the v4 salttype issues aren't important (even the failure to fall back; that case is probably very rare).  So the big question is whether we should have dummy enctype entries for the single-DES enctype names for the sake of kadmin getprinc.

greghudson:
I pushed a fourth commit implementing a kind of compromise: krb5_enctype_to_name() can translate the names of no-longer-supported enctypes, but they don't show up as valid enctypes to krb5_c_valid_enctype() and we display them with an "UNSUPPORTED:" tag in kadmin.

frozencemetery:
Nice, I like that a lot.

greghudson:
Made some revisions.  Notably:

* Since |defetypes| no longer includes any weak enctypes (it does not include rc4-hmac-exp), we don't need to talk about removing weak enctypes from that list in the variables which use it by default.

* enc_old.c is no longer used after the removal of the single-DES enctype entries, and can be removed.

* There were some out-of-scope changes to default.exp related to BINSH; I undid them.

The changes to mk_safe.c seem out of scope.  As with the profile variables, I don't know of any reason we still need krb5_auth_con_set_safe_cksumtype() to work, but I don't think it has anything to do with single-DES support.

frozencemetery:
Ah, good catch.  That shouldn't have been in here.  I'm not particularly concerned with those handles right now, so I just pulled it out.

kaduk:
> The changes to weak enctype documentation seem a bit off, and the placement of this commit is awkward because it precedes the removal of single-DES. Probably it can be folded into the final commit. For the three variables that default to |defetypes|, since there will no longer be a weak enctype in the default set, the whole "but..." clause can be removed. The documentation for allow_weak_crypto should probably just say that it enables single-DES support in 1.8-1.17 and doesn't do anything in 1.18+. 

Do we want to leave some rope for "further enctypes may be brought under the control of allow_weak_crypto in subsequent releases"?

greghudson:
> Do we want to leave some rope for "further enctypes may be brought under the control of allow_weak_crypto in subsequent releases"?

(rc4-hmac-exp is still affected by allow_weak_crypto as Robbie pointed out, but I think the question still stands.)

At this time I don't think we will add anything to allow_weak_crypto, because I believe a single setting to re-enable all deprecated algorithms isn't great, and because the standard we set for "weak" when we deprecated single-DES is a pretty high bar--des3 and rc4 have problems, but as far as we know, nothing nearly so bad as being able to recover a random key in under a day.

I suggested one boolean variable per deprecated family, so allow_des3 for triple-DES and allow_rc4 for RC4.  Simo prefers "allow_weak_enctype = des3-cbc-sha1" (or perhaps allow_deprecated_enctype).  We'll hash that out when we get there.

PR #926, opened 2019-05-06
Rename hmac() function.

0-wiz-0:
NetBSD's libc includes a hmac() function as well; this avoids a
build problem on NetBSD.


greghudson:
It looks like NetBSD added an hmac() declaration to stdlib.h in 2016.  It's protected by _NETBSD_SOURCE, but that feature test is assumed unless you ask for ANSI or POSIX or XOPEN (which I agree with; I never liked how _GNU_SOURCE had to be explicitly asked for).  In 2017 they discussed a plan to move the declaration out of stdlib.h due to conflicts with multiple code bases, but it doesn't appear to have been implemented.

I will probably pick a different name (the krb5_ prefix is used for public API functions) and add a comment, and take this reluctantly.

PR #927, opened 2019-05-06
Add a Travis job for building with OpenSSL crypto exclusively

frozencemetery:
This matches closely the job that OpenSSL uses ([krb5.sh](https://github.com/openssl/openssl/blob/master/test/recipes/95-test_external_krb5_data/krb5.sh)) and is part of the one I use in Fedora ([krb5.spec](https://github.com/frozencemetery/krb5.fedora/blob/master/krb5.spec#L299)).

I considered making this a full "non-builtin" job, but that didn't seem especially useful given what Travis is running.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,6 +1,6 @@
 cd src
 autoreconf
-./configure --enable-maintainer-mode --with-ldap
+./configure --enable-maintainer-mode --with-ldap $CONFIGURE_OPTS
 make $MAKEVARS
 make check
 make distclean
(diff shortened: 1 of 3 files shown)


frozencemetery:
The alignment behavior isn't great, but I don't see a way to correct it without redesigning the interface.

greghudson:
> I considered making this a full "non-builtin" job, but that didn't seem especially useful given what Travis is running.

Can you expand on that?  "builtin" only applies to the k5crypto backend as far as I know.

I made a few changes.  --with-tls-impl=openssl is the default so I dropped that from the extra configure options.

frozencemetery:
Sure.  I was referring to things like --with-system-et, --with-system-ss, --with-system-verto.

PR #930, opened 2019-05-14
Add NULL-checks to realm and principal comparisons

frozencemetery:
Neither krb5_realm_compare() nor krb5_principal_compare_flags()
specify behavior on NULL entries, so define them non-matching rather
than crashing.

Also-authored-by: Nalin Dahyabhai <nalin@redhat.com>

(Patch carried downstream in Fedora since 2009; originally against 1.7.)

### DIFF
--- a/src/lib/krb5/krb/princ_comp.c
+++ b/src/lib/krb5/krb/princ_comp.c
@@ -36,6 +36,10 @@ realm_compare_flags(krb5_context context,
     const krb5_data *realm1 = &princ1->realm;
     const krb5_data *realm2 = &princ2->realm;
 
+    if (princ1 == NULL || princ2 == NULL)
+        return FALSE;
+    if (realm1 == NULL || realm2 == NULL)
+        return FALSE;
     if (realm1->length != realm2->length)
         return FALSE;
     if (realm1->length == 0)
@@ -87,6 +91,9 @@ krb5_principal_compare_flags(krb5_context context,
     krb5_principal upn1 = NULL;
     krb5_principal upn2 = NULL;
     krb5_boolean ret = FALSE;
+
+    if (princ1 == NULL || princ2 == NULL)
+        return FALSE;
 
     if (flags & KRB5_PRINCIPAL_COMPARE_ENTERPRISE) {
         /* Treat UPNs as if they were real principals */
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -3641,7 +3641,7 @@ krb5_address_order(krb5_context context, const krb5_address *addr1,
  * @param [in] princ2           Second principal
  *
  * @retval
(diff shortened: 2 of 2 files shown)


greghudson:
This is a weird one; libkrb5 functions which accept pointers don't have a blanket guarantee of not crashing when passed null pointers, so a downstream patch to add defined behavior on null to two particular functions is confusing.

I found https://bugzilla.redhat.com/show_bug.cgi?id=161471 and https://access.redhat.com/errata/RHSA-2005:567.  The bugzilla description doesn't delve beyond the immediate cause of whatever crashes were reported (that is, why one of the comparison functions was being passed NULL) and the errata description doesn't mention any null dereferences, even though the change is bundled into the fixes.  RT issue 3907 shows a submission of the patch from a SuSE email address, with no explanation of the motivation.

Adding defined behavior on null pointers for object comparison functions also raises the eternal question of whether null is equal to null.

My inclination is to skip this one unless we uncover more information.

src/lib/krb5/krb/princ_comp.c
```diff
@@ -36,6 +36,10 @@ realm_compare_flags(krb5_context context,
     const krb5_data *realm1 = &princ1->realm;
     const krb5_data *realm2 = &princ2->realm;
 
+    if (princ1 == NULL || princ2 == NULL)
+        return FALSE;
+    if (realm1 == NULL || realm2 == NULL)
```
greghudson: Probably moot, but realm1 and realm2 cannot be null, as they are assigned the addresses of structure fields just above.

cipherboy: Wouldn't the `princ1` / `princ2` checks on line 39 also be moot, since we've already dereference them on line 36/37 (and, would have presumably crashed if we did dereference a `NULL` pointer there)? I'm not sure if the behavior is modified by `realm1` / `realm2` being declared `const` though.

greghudson: No, simply computing &princ1->realm and &princ2->realm, doesn't dereference the princ1 and princ2 pointers; it just adds the offset to them.  (At least in practice; at a language definition level I'm not certain.)  The const qualifier is irrelevant.

cipherboy: Ah, thanks for the explanation. :)

(Still, kinda seems weird to refer to the address of something in the struct before checking that the pointer to the struct isn't NULL, but I guess that's a side effect of enforcing C89's variables-before-code).

frozencemetery:
Agreed - I'll drop it downstream as well and re-open if it breaks something.

src/lib/krb5/krb/princ_comp.c
kaduk: I think at the language definition level computing the pointer is still undefined behavior.  Consider the case of a segmented architecture, where you end up trying to load in a segment that doesn't exist.

PR #932, opened 2019-05-14
Preserve intermediate yacc files for debuginfo generation

frozencemetery:
Keep y.tab.c files around because the debuginfo points to them.

Also-authored-by: Nalin Dahyabhai <nalin@redhat.com>

(Originally part of fixes for [rhbz#729044](https://bugzilla.redhat.com/show_bug.cgi?id=729044); carried downstream in Fedora since 2011, originally against 1.9.1.)


frozencemetery:
Ah, I see.  Fair enough.

greghudson:
> Ah, I see. Fair enough.

What is this in reference to?  I see the CI failed, but I think that just means y.tab.c has to be removed in make clean.

frozencemetery:
I need to look into whether adding this to `make clean` still solves the problem on my end or not, and don't plan to do that right now.

PR #933, opened 2019-05-20
Avoid show internal window (including in code) for change password when it's expired

Joshuacosta:
We are developed a new API at level **leash** to avoid show internal window (including in code) for change password when is expired and return without any actuation, only the error code.

This new API (it's name **Leash_set_show_password_prompt**) is called before Leash_kinit and allow in a situation of expired password, do not show the "internal" original window and return with the error code that marks this state, so we can show our "own" window to change password.

We have done this modification in our installation but we create a "draft" request to validate it. The changes affect 10 programs.


greghudson:
It doesn't look like this pull request went through properly.  We have some notes on maintaining a github for pull requests here: https://k5wiki.kerberos.org/wiki/Maintaining_a_github_fork

All topic branches for pull requests should be prepared against the master branch, not against release branches.

Joshuacosta:
Sorry.
I will in the correct way.

Regards

greghudson:
I'm closing this for now.  Please feel free to reopen it with the correct topic branch whenever you're ready.

PR #912, opened 2019-04-02
Implement resource-based-constrained-delegation support

iboukris:
This is work-in-progress, following the discussion on krbdev:
http://mailman.mit.edu/pipermail/krbdev/2019-March/013083.html

I'm still not happy with the chase-referral commit, which is a cleaned-up version of the prove-of-concept code. I wanted to move more logic to s4u_creds.c, implement a chase-referral loop that calls krb5_get_cred_via_tkt_ext() in each iteration, and call this loop two or three time to get the job done. It also depends whether or not we want krb5_get_credentials() to work with RBCD or the caller code can be adapted. Anyway, for now I'll start working on the KDC side (and tests) to get a better picture.

PR #935, opened 2019-05-22
krb5_cc_store_cred: avoid calling krb5_cc_remove_cred()

iboukris:
Follow up to the discussion at #912 
If I understood your comment correctly and it may still be worth fixing (even though on master it doesn't exhibit it self), then this patch fixes it for my test case (no duplicates shown before rebase).

### DIFF
--- a/src/lib/krb5/ccache/ccfns.c
+++ b/src/lib/krb5/ccache/ccfns.c
@@ -82,6 +82,7 @@ krb5_cc_store_cred(krb5_context context, krb5_ccache cache,
 {
     krb5_error_code ret;
     krb5_ticket *tkt;
+    krb5_creds mcreds, old_creds;
     krb5_principal s1, s2;
 
     TRACE_CC_STORE(context, cache, creds);
@@ -98,12 +99,15 @@ krb5_cc_store_cred(krb5_context context, krb5_ccache cache,
     if (ret) return 0;
     s2 = tkt->server;
     if (!krb5_principal_compare(context, s1, s2)) {
-        creds->server = s2;
-        TRACE_CC_STORE_TKT(context, cache, creds);
-        /* remove any dups */
-        krb5_cc_remove_cred(context, cache, KRB5_TC_MATCH_AUTHDATA, creds);
-        ret = cache->ops->store(context, cache, creds);
-        creds->server = s1;
+        mcreds = *creds;
+        mcreds.server = s2;
+        TRACE_CC_STORE_TKT(context, cache, &mcreds);
+        ret = krb5_cc_retrieve_cred(context, cache, KRB5_TC_MATCH_AUTHDATA,
+                                    &mcreds, &old_creds);
+        if (ret == 0)
+            krb5_free_cred_contents(context, &old_creds);
+        else if (ret == KRB5_CC_NOTFOUND)
+            ret = cache->ops->store(context, cache, &mcreds);
     }
     krb5_free_ticket(context, tkt);
     return ret;


iboukris:
I wonder why there is no CI, I'll try to rebase to master and re-push (and fix the commit message).

iboukris:
Still no CI, however it does pass basic make-check locally.

src/lib/krb5/ccache/ccfns.c
```diff
@@ -100,9 +101,14 @@ krb5_cc_store_cred(krb5_context context, krb5_ccache cache,
...
         TRACE_CC_STORE_TKT(context, cache, creds);
-        /* remove any dups */
-        krb5_cc_remove_cred(context, cache, KRB5_TC_MATCH_AUTHDATA, creds);
-        ret = cache->ops->store(context, cache, creds);
+        ret = krb5_cc_retrieve_cred(context, cache, KRB5_TC_MATCH_AUTHDATA,
+                                    creds, &old_creds);
+        if (ret != KRB5_CC_NOTFOUND)
+            return ret;
+        if (ret == 0)
+            krb5_free_cred_contents(context, &old_creds);
```
frozencemetery: This line is unreachable.

src/lib/krb5/ccache/ccfns.c
```diff
@@ -100,9 +101,14 @@ krb5_cc_store_cred(krb5_context context, krb5_ccache cache,
     if (!krb5_principal_compare(context, s1, s2)) {
         creds->server = s2;
         TRACE_CC_STORE_TKT(context, cache, creds);
-        /* remove any dups */
-        krb5_cc_remove_cred(context, cache, KRB5_TC_MATCH_AUTHDATA, creds);
-        ret = cache->ops->store(context, cache, creds);
+        ret = krb5_cc_retrieve_cred(context, cache, KRB5_TC_MATCH_AUTHDATA,
+                                    creds, &old_creds);
+        if (ret != KRB5_CC_NOTFOUND)
+            return ret;
```
frozencemetery: This leaks tkt.

greghudson:
I am actually thinking of implementing option 4 from http://krbdev.mit.edu/rt/Ticket/Display.html?id=8764 and then (as I suggested there) modifying klist to display the ticket server name if it differs from the cred server.  Everything is a lot more predictable if krb5_cc_store_cred() goes back to being a simple dispatch to the per-type store method.

greghudson:
Regarding CI: apparently github webhook deliveries were paused for a bit today, and Travis is catching up with the backlog after they resumed.

src/lib/krb5/ccache/ccfns.c
iboukris: Right, fixed it.

src/lib/krb5/ccache/ccfns.c
iboukris: Which line? I don't see, please elaborate.

iboukris:
I think I'm okay with option (4), if the application ask again for the same server but with the realm (after a referral request) it will cost one more request, and it will be cached.

RBCD only works with referrals, currently, after a referral request a non-referral with the correct realm would work too, but I think it's okay if it won't.

src/lib/krb5/ccache/ccfns.c
frozencemetery: If `ret == 0`, it would also be the case that `ret != KRB5_CC_NOTFOUND` (line 106) and be returned (line 107) before getting to the check (line 108).  This makes line 109 unreachable.

iboukris: Ah sorry, I see now.. I fixed it as it was meant to be.

iboukris:
I have simplified the commit a little, just in case we are still interested.

greghudson:
I still plan to implement option 4.

I thought of a reason not to like this approach (apart from a general dislike of implementing logic at the dispatch layer): we could still create duplicates during ccache copy operations if the copy stores the cred with the referral server realm before the cred with the canonical server realm.  In fact, I think the only reason this approach suppresses the extra creds for the copy in PR #912 is because cc_memory.c reverses the order of creds when iterating.

iboukris:
I suspected the order might matter but I didn't debug it, was just satisfied that it fixed the issue at hand. 
Closing this - thanks!

PR #936, opened 2019-05-26
Move links to HTTPS

mmattioli:
HTTPS, by nature, is more secure than HTTP. While the information behind the links may not necessarily contain sensitive information, moving to HTTPS validates the authenticity and source of the information being retrieved.

Some links are also dead or have been moved; they have been updated accordingly.


greghudson:
I suspect the whitespace changes are a product of an editor.  I can remove them before merging.

mmattioli:
* The whitespace changes are from the editor (as @greghudson suggested).
* Any links that have a valid archive.org link were put back; the one link that truly is dead/broken and has no archive.org link has been marked as such.
* I rebased and squashed all three into one.
* I am not a layer either but the content behind those links nor the location of the information behind them or what is being conveyed/communicated is not being charged in any way.

frozencemetery:
Much better, thanks!

mmattioli:
@greghudson @frozencemetery I also just fixed a broken link as reported in [8712](http://krbdev.mit.edu/rt/Ticket/Display.html?id=8712); this would fix/close that bug.

greghudson:
I have somewhat differing opinions on this one.  I think it's okay to have a single "update HTTP links" commit which migrates links to https and deletes or updates broken links.  I would rather delete broken links than annotate them or use archive.org links.

For licenses, if there is an upstream version of the license which uses an https URL, I think we can safely update our copy of the license text, but otherwise we should leave it alone.  I can try to track down the specific changes.

mmattioli:
@greghudson I felt the same way but adjusted as requested by @frozencemetery.

For the license, the text that is being adjusted is actually _not_ a license (or the license itself), it's text that basically says "please refer to the license located here". As I mentioned in a previous comment, the content behind the link itself is not changing in any form/shape/way. If you look at it in a slightly different manner, consider a license which is located as a text file within a repository or explicitly linked within the header comment of another file (e.g. "please refer to the license here"); whether it is in its own file embedded within the repository or linked within a header comment the content of the license is the same and so the meaning and message of the license is not changed.

greghudson:
* I edited out the whitespace changes.
* I removed the changes to ev.c, which is edited as little as possible from upstream source.
* Per my preference, I removed some outdated links, as well as the link in the German translation comment.

All of the license-adjacent changes seem okay.

PR #937, opened 2019-05-28
Stop double-caching referral tickets

greghudson:
Here is the ticket 8764 option 4 implementation.

### DIFF
--- a/src/clients/klist/klist.c
+++ b/src/clients/klist/klist.c
@@ -662,25 +662,27 @@ static void
 show_credential(krb5_creds *cred)
 {
     krb5_error_code ret;
-    krb5_ticket *tkt;
-    char *name, *sname, *flags;
+    krb5_ticket *tkt = NULL;
+    char *name = NULL, *sname = NULL, *tktsname, *flags;
     int extra_field = 0, ccol = 0, i;
+    krb5_boolean is_config = krb5_is_config_principal(context, cred->server);
 
     ret = krb5_unparse_name(context, cred->client, &name);
     if (ret) {
         com_err(progname, ret, _("while unparsing client name"));
-        return;
+        goto cleanup;
     }
     ret = krb5_unparse_name(context, cred->server, &sname);
     if (ret) {
         com_err(progname, ret, _("while unparsing server name"));
-        krb5_free_unparsed_name(context, name);
-        return;
+        goto cleanup;
     }
+    if (!is_config)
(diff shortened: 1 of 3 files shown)


src/clients/klist/klist.c
```diff
@@ -662,25 +662,27 @@ static void
...
     }
     ret = krb5_unparse_name(context, cred->server, &sname);
     if (ret) {
         com_err(progname, ret, _("while unparsing server name"));
-        krb5_free_unparsed_name(context, name);
-        return;
+        goto cleanup;
     }
+    if (!is_config)
+        krb5_decode_ticket(&cred->ticket, &tkt);
```
frozencemetery: Shouldn't failure here be handled?  Or will that always result in NULL tkt?  (Even then, isn't failure to decode cred->ticket kind of a problem?)

greghudson: Oops, I meant to have a void cast there.

Java's krb5 implementation used to have problems with invalid ticket encodings in the ccache, and it broke when MIT and Heimdal added ccache config entries.  That was kind of our fault for adding config entries the way we did, but it makes me think that we should behave gracefully (not bailing out) if we run into a cred entry with a non-decodable ticket, even if we don't recognize it as a config entry.  I'm also preserving the behavior that klist -e just doesn't display enctypes if we can't decode the ticket.

src/clients/klist/klist.c
```diff
@@ -662,25 +662,27 @@ static void
...
+    krb5_ticket *tkt = NULL;
+    char *name = NULL, *sname = NULL, *tktsname, *flags;
     int extra_field = 0, ccol = 0, i;
+    krb5_boolean is_config = krb5_is_config_principal(context, cred->server);
 
     ret = krb5_unparse_name(context, cred->client, &name);
     if (ret) {
         com_err(progname, ret, _("while unparsing client name"));
-        return;
+        goto cleanup;
```
iboukris: Technically, here we could just return.

iboukris:
Looks good to me, and if I borrow this change to #912, I don't need the workaround to skip tickets with referral realm in check_ticket_count() and the test pass normally.

src/clients/klist/klist.c
greghudson: We could, but if another allocation was added before this one, the `return` would have to be changed to `goto cleanup`.  For this reason I tend to use `goto cleanup` for the first of several allocations.  (I still use `return` when validating arguments prior to any allocations.)

PR #938, opened 2019-05-28
Remove key combination support

frozencemetery:
(The other approach would be to extend key combination, but I don't see that as valuable given unencrypted SAD support is the only user.)

This is part of a larger series in preparation for 3DES removal, which can be viewed at https://github.com/krb5/krb5/compare/master...frozencemetery:die-des .  Since I do intend to remove 3DES downstream before upstream krb5 is ready to do so, I want to be sure not to break SAM support when I do so.  I don't have a SAM environment for testing, though.


greghudson:
I will test this against MIT's environment soon.

greghudson:
I tested this successfully against MIT's environment.

However, I think the last hunk of the first commit should be reverted.  There are three modes SAM-2 can be used in:

1. use-sad-as-key set: encrypt response in string-to-key(SAD)
2. send-encrypted-sad set: encrypt response in string-to-key(password), include SAD in response
3. neither flag set: encrypt response in combine-key(string-to-key(password), string-to-key(SAD))

MIT uses case 2.  Both flags being set is a protocol violation, and must-pk-encrypt-sad being set is effectively a protocol violation because it was a future-use flag that never became defined.

The code for computing the reply key is structured as:

    if (not case 1) {
      if (case 3) compute string-to-key(SAD) and combine it with current reply key
      (for case 2 do nothing, reply key is already string-to-key(password))
    } else {
      replace reply key with string-to-key(SAD) for case 1
    }

The code for setting the sam_sad field includes the SAD for case 2, and not for case 1 or 3.

The first hunk removes case 3 support, as is intended by the PR.  Case 1 and 2 support remain.

The second hunk removes case 1 and case 3 support.  Case 2 continues to function, but removing case 1 support isn't in scope for the PR.

frozencemetery:
Thanks, I think I understand now.  Updated.

PR #939, opened 2019-06-01
Simplify and modernize replay cache interface

greghudson:
Remove all of the replay cache methods except for resolve, close, and
store.  Rename the dispatch functions to use a k5_ prefix to indicate
that they are not part of the libkrb5 API.  Remove the unused code for
registering replay cache types at runtime.

Remove the krb5_donot_replay structure, as the only concrete replay
cache type just needs the authenticator checksum.  Identify the
checksum part of the authenticator ciphertext at the dispatch layer.
Handle container allocation at the dispatch layer so that each replay
cache type only needs to work with its data pointer.  Remember the
full name at the dispatch layer to implement a get_name operation.
Consolidate what remains of rcfns.c and rc_conv.c into rc_base.c.

Use k5_rc_default() in preference to krb5_get_server_rcache() since
they now do the same thing.  Remove the sim_client code to create a
replay cache, as the per-message functions use a memory reply cache
now.


src/lib/krb5/rcache/rc_base.c
```diff
@@ -177,34 +56,107 @@ krb5_rc_default(krb5_context context, krb5_rcache *idptr)
...
+    for (t = typehead; t != NULL; t = t->next) {
+        if (strncmp(t->ops->type, name, len) == 0 && t->ops->type[len] == '\0')
+            break;
+    }
+    if (t == NULL)
+        return KRB5_RC_TYPE_NOTFOUND;
 
-    ret = resolve_type_and_residual(context, type, sep + 1, idptr);
-    free(type);
+    rc = k5alloc(sizeof(*rc), &ret);
```
frozencemetery: Though they're both shims for it anyway, this should probably be `k5calloc()` for clarity.

greghudson: Clarity that the memory is zeroed, you mean?  k5alloc() was first added without a k5calloc() to go with it, and has always allocated zeroed memory (its documenting comment says so).  k5calloc() was later added to avoid having to worry about multiplication overflow.

frozencemetery: You're right - I misremembered and thought you preferred k5calloc for when zero-initialization was important.  Nevermind!

PR #940, opened 2019-06-01
Address Coverity defect in new replay cache code

greghudson:
Coverity flags the following code in rc_file2.c as a NEGATIVE_RETURNS defect:

        if (avail_offset == -1) {
            if (nread == 0 || ts_after(now, ts_incr(rec1_stamp, skew)))
                avail_offset = record_offset;
            else if (nread == 1 || ts_after(now, ts_incr(rec2_stamp, skew)))
                avail_offset = record_offset + RECORD_LEN;
        }
        
        if (nread < 2 || rec1_stamp == 0 || rec2_stamp == 0)
            return write_record(fd, avail_offset, tag, now);

on the grounds that it might pass avail_offset=-1 to write_record(), and from there to lseek().

The code here is implementing the following logic:

* If a slot is beyond the end of the file, or empty, or expired, it is available for writing.  We want to track the first available offset.
* If a slot is beyond the end of the file or empty, it terminates the search, because a previous store of the same record would have been written to that slot or an earlier slot.
* If a slot is expired, it does *not* terminate the search, because it might not have been expired during a previous store operation.

(Note that the slot that terminates the search isn't necessarily the slot that we write to.)

When the code checks if a slot is available, it doesn't explicitly check that it is empty, because we "know" that now > skew in any realistic scenario, and therefore a rec1_stamp or rec2_stamp of 0 will appear as expired.  Coverity does not know that now > skew.  It is possible that Coverity would flag the code even if  we did explicitly check for rec[12]_stamp == 0, depending on how sophisticated its analysis is.

A surefire way to shut up Coverity is to use 0 as the sentinel for avail_offset, since the checker only cares about negative values.  0 is never a candidate slot (because we store the hash seed there), so the code would remain correct.  That's what I have put in the commit initially.  But I want to think about other solutions, on the principle that if my code confuses a static analyzer, it will probably confuse some fraction of humans as well.

### DIFF
--- a/src/lib/krb5/rcache/rc_file2.c
+++ b/src/lib/krb5/rcache/rc_file2.c
@@ -123,6 +123,14 @@ write_record(int fd, off_t offset, const uint8_t tag[TAG_LEN],
     return 0;
 }
 
+/* Return true if timestamp is expired, for the current timestamp (now) and
+ * allowable clock skew. */
+static inline krb5_boolean
+expired(uint32_t timestamp, uint32_t now, uint32_t skew)
+{
+    return ts_after(now, ts_incr(timestamp, skew));
+}
+
 /* Check and store a record into an open and locked file.  fd is assumed to be
  * at offset 0. */
 static krb5_error_code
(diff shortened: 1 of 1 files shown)


frozencemetery:
0 is epsilon less clear as a sentinel value to me, but otherwise it seems fine, and I didn't have any problems with this the first time around so :)

greghudson:
> It is possible that Coverity would flag the code even if we did explicitly check for rec[12]_stamp == 0, depending on how sophisticated its analysis is.

I manually submitted a build with checks for rec1_stamp == 0 and rec2_stamp == 0 (without changing the sentinel value from -1 to 0), and the defect went away.  So I might go that route.

greghudson:
Updated to explicitly check the record timestamps.

Having the conditionals spanned two lines bugged me, so I made some additional changes to keep them on one line.  I also added some comments.

PR #942, opened 2019-06-06
Improve logging documentation

greghudson:
The default severity was removed by commit
6ce8fd4cfa2e9b1e92debd204a5b2ddf053cca55 (ticket 8630) but the example
still talks about it; remove that text.  Add a note about the default
being syslog if nothing else is specified, and a note on how to
disable logging.


greghudson:
Possible nitpick: the example still says `kdc = SYSLOG:INFO:DAEMON` which includes a default severity, though the text no longer discusses it.  I could change it to `kdc = SYSLOG::DAEMON` which would work with 1.17+ but would cause an error in previous releases; I decided not to change it.

PR #945, opened 2019-06-10
Simplify pac verify logic based on Heimdal implementation

iboukris:
Use strcmp instead of trying to parse pac principal, in
order to avoid relying on the provided name type. Windows
seem to be using KRB5_NT_MS_PRINCIPAL for enterprise name
type too, and this change allows us to verify it correctly.

This is an alternative fix for #744

### DIFF
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -408,12 +408,11 @@ k5_pac_validate_client(krb5_context context,
 {
     krb5_error_code ret;
     krb5_data client_info;
-    char *pac_princname;
+    char *pac_princname, *princname;
     unsigned char *p;
     krb5_timestamp pac_authtime;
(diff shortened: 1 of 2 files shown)


iboukris:
I referenced the wrong PR, meant #744

greghudson:
The code change seems fine at a glance, but I am curious under what circumstances Microsoft uses KRB5_NT_MS_PRINCIPAL with an enterprise principal name.

iboukris:
In cross-realm RBCD, if an enterprise name is used for impersonated client (windows often uses enterprise), then the name type gets lost (because it is retrieved from PAC signature where the name type is not saved), so the final ticket always ends up KRB5_NT_MS_PRINCIPAL, even if it is a UPN including an at sign in the name component.
You can take a look at the packet capture I referenced in PR #912 (or in the tshark output).

PR #947, opened 2019-06-13
New leash API to prevent show change password window using LEASHDLL

Joshuacosta:
Hello,
We created a new leash API to prevent show the "change password window" in LEASH_KINIT and return control to call program. This new API is called:
Leash_set_show_password_prompt(DWORD onoff).
For implement this solution we use the unused existing API:
krb5_get_init_creds opt_set change_password prompt.

### DIFF
--- a/src/windows/include/leashwin.h
+++ b/src/windows/include/leashwin.h
@@ -205,6 +205,9 @@ DWORD Leash_reset_default_mslsa_import();
 DWORD Leash_get_default_preserve_kinit_settings();
 DWORD Leash_set_default_preserve_kinit_settings(DWORD onoff);
 DWORD Leash_reset_default_preserve_kinit_settings();
+DWORD Leash_get_show_password_prompt();
+DWORD Leash_set_show_password_prompt(DWORD onoff);
+DWORD Leash_reset_show_password_prompt();
 #ifdef __cplusplus
 }
 #endif
--- a/src/windows/include/loadfuncs-leash.h
+++ b/src/windows/include/loadfuncs-leash.h
@@ -372,6 +372,24 @@ TYPEDEF_FUNC(
     Leash_reset_defaults,
     (void)
     );
+TYPEDEF_FUNC(
+    DWORD,
+    CALLCONV_C,
+    Leash_reset_show_password_prompt,
+    (void)
+    );
+TYPEDEF_FUNC(
+    DWORD,
+    CALLCONV_C,
+    Leash_set_show_password_prompt,
+    (DWORD)
+    );
+TYPEDEF_FUNC(
+    DWORD,
+    CALLCONV_C,
+    Leash_get_show_password_prompt,
+    (void)
+    );		
 /* They are not yet all here... */
 
 #endif /* __LOADFUNCS_LEASH_H__ */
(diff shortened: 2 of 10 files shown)


greghudson:
I have two questions:

1. Are you passing a password to Leash_kinit(), or letting it prompt for the current password?  If you are supplying the password, you could consider bypassing Leash_kinit() and calling krb5_get_init_creds_password() from the krb5 library.

2. Do you need a registry key to control this setting in a persistent fashion, or was that just the only obvious way to communicate this information to Leash_kinit() for a specific application?

Joshuacosta:
1) This behavior is for implement the solution when the password is expired, to avoid yours "change's password window" (in english) and return control -with the Kerberos warning code ocurred: KRB5KDC_ERR_KEY_EXP- to our code, and show ours window (in catalan, the user's language).

2) Is the second sentence. We don't want to define a new function or another parameter to the call.

greghudson:
> This behavior is for implement the solution when the password is expired, to avoid yours "change's password window" (in english) and return control -with the Kerberos warning code ocurred: KRB5KDC_ERR_KEY_EXP- to our code, and show ours window (in catalan, the user's language).

I understand.  My question is: is your application having Leash_kinit() prompt for the password in the first place, or does your application already have the password and is passing it to Leash_kinit()?

If you don't need Leash to prompt for the password, you don't need to use Leash_kinit() and could use krb5_get_init_creds_password() instead.

Joshuacosta:
Hi Greg,

We would not like use "apis krb5" in our code. We use the leashdll like a simple level to obtain a TGT and use it to obtain TGS with GSSAPI functions.
Our code looks like:

Leash_set_show_password_prompt(0);
resultat = Leash_kinit(...);
if (resultat == 0){
   // GSSAPI Level
   ... 
  gss_init_sec_context
}

If we use the krb5 apis its become more complex and we think that this new LEASH api simplified the solution and could be reused by another programmers.

Thanks for you attention.
Joshua

greghudson:
Thanks for clarifying.  For a couple of reasons, I don't think I will be merging this pull request.

First and foremost, using registry state to communicate behavior for a particular application call can have undesirable effects.  One application wanting to suppress graphical prompting for one call to Leash_kinit() should not change the behavior of all other applications (including Leash itself) run by the same user.

Second, the purpose of Leash_kinit() as I understand it is to make an initial ticket request with graphical prompts; it is not there to be a simplified convenience wrapper.  Although it may not be as simple to use, we have an existing API which allows the application to control or suppress interactive behavior.

For the second reason, I also don't favor creating a new variant of Leash_kinit() with an additional argument.

Joshuacosta:
Good morning Greg,

We can understand your doubts but:

- We use the registry like other Leash's functions does. I know that this
behaviour will be for all calls but this is that we want for us (maybe in
other situations this is bad).

- And the fact that leash_kinit have a graphic behaviour is a limitation to
simplify the code when its used from an application that have graphic
behaviour too. The problem or limitation is that the existent API that
change Leash's behavior can't be called from Leash level.

So, the solution, the solution that we should develope, will should be to
make an own Leash_krb5_kinit.

Thanks in advance.
Joshua

PR #948, opened 2019-06-18
Remove PKINIT draft 9 support

greghudson:


greghudson:
I added a commit to make it easier to run the PKCS11 tests, since soft-pkcs11 isn't commonly packaged and therefore tends to live outside the runtime linker path.

I happen to have a soft-pkcs11.so lying around (linked against OpenSSL 1.0), and the 1.8 release.  I found https://github.com/kinnalru/soft-pkcs11 but I don't know how different it is.  I might upload Love's code to github and modify it to work with OpenSSL 1.1.  Of course if there is a commonly packaged PKCS11 soft-token library that we can easily use in t_pkinit.py that would be a superior state to be in.

greghudson:
I removed the third commit in favor of PR #949.

PR #949, opened 2019-06-19
Convert OTP and kdcproxy tests to Python 3

greghudson:
[The KDC proxy tests are still broken for me after this.  If I "pip3 install paste", I get a Paste installation which doesn't work for SSL, as it uses socket._fileobject() which is no longer present in Python 3.  Paste appears to be lightly or not maintained, so we should look for alternatives.]

Commit e23d24beacb73581bbf4351250f3955e6fd44361 did not convert
t_otp.py or paste-kdcproxy.py.  Convert them now.


greghudson:
I converted from Paste to wsgiref (which is in the standard Python library) and got the proxy tests to work again.

greghudson:
I added a commit to improve the Travis coverage.  The only skips now are t_discover_uri.py because resolv_wrapper isn't in the package repository until 18.04, and the PKINIT pkcs11 tests because soft-pkcs11 isn't packaged.  I found SoftHSM as a possible mainstream candidate for soft-pkcs11, and will look into it.

(Also, we can't use the name kdcproxy.py as that conflicts with the kdcproxy package, so I renamed paste-kdcproxy.py to wsgiref-kdcproxy.py instead.)

greghudson:
I decided it would make the most sense to import soft-pkcs11 into the source tree so that it's always available for tests, as it doesn't have an upstream.  I added two commits to do that.

PR #951, opened 2019-06-30
Fix typos

Coeur:


greghudson:
I reverted changes to three upstream files where I don't want to be in the business of fixing typos (particularly libev), and added a hyphen in ldap_misc.c.  (The first two force-pushes were mistakes and should cancel each other out.)

PR #952, opened 2019-07-10
Don't require all enctypes in a keytab to be valid

frozencemetery:
When acquiring credentials using a keytab, krb5 examines each entry in
the keytab to check for a better match.  Relax the check on enctypes so
that the presence of removed enctypes (like DES) in the keytab doesn't
cause failure in kinit.

### DIFF
--- a/src/lib/krb5/keytab/kt_file.c
+++ b/src/lib/krb5/keytab/kt_file.c
@@ -289,7 +289,6 @@ krb5_ktfile_get_entry(krb5_context context, krb5_keytab id,
     krb5_keytab_entry cur_entry, new_entry;
     krb5_error_code kerror = 0;
     int found_wrong_kvno = 0;
-    krb5_boolean similar;
     int was_open;
     char *princname;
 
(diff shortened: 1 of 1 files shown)


frozencemetery:
(I think this was overlooked because an intermediate design had the enctypes still present in the list, but with no operator functions.  Hard to know, though.)

greghudson:
We only needed krb5_c_enctype_compare() for single-DES enctypes, where we wanted to share key entries between des-cbc-crc, des-cbc-md4, and des-cbc-md5.  Today even when we introduce a new enctype with the same encryption algorithm as an old one (like the aes-sha2 enctypes) we tweak the string-to-key so that we don't share keys.

So this fix can just simplify the code to do an equality test on the enctype, and remove the coercion.  As the PR currently stands, I think it could leave kerror as non-zero if there is a DES key in the final entry of the keytab.

frozencemetery:
Ah, good catch.  Updated.

greghudson:
I was suggesting to use the `==` operator instead of krb5_c_enctype_compare().

frozencemetery:
Ah, thanks, got confused by the mention of `kerror`.

PR #953, opened 2019-07-14
Fix doc for krb5.conf file

mmattioli:
The documentation for krb5.conf explaining how tags are marked as final
is incorrect.

http://mailman.mit.edu/pipermail/krbdev/2019-July/013131.html

### DIFF
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -35,17 +35,11 @@ or::
         baz = quux
     }
 
-Placing a '\*' at the end of a line indicates that this is the *final*
-value for the tag.  This means that neither the remainder of this
-configuration file nor any other configuration file will be checked
-for any other values for this tag.
-
-For example, if you have the following lines::
-
-    foo = bar*
-    foo = baz
-
-then the second value of ``foo`` (``baz``) would never be read.
+Placing a '\*' after the closing bracket of a section name indicates
+that the section is *final*, meaning that if the same section appears
+within a later file specified in **KRB5_CONFIG**, it will be ignored.
+A subsection can be marked as final by placing a '\*' after either the
+tag name or the closing brace.
 
 The krb5.conf file can include other files using either of the
 following directives at the beginning of a line::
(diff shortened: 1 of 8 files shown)


greghudson:
Sections and subsections can also be marked as final, and I was planning to document that in the fix.  (For subsections, it looks like either adding a * to the tag name or to the closing brace might work; I need to test that.)

I was also considering adding tests at the same time.  test_profile.c seems like a suitable harness (it can load multiple profile files and its "query" subcommand will output all values of a given relation), but I hadn't yet decided among the options for scripting the tests.

mmattioli:
Let me know what you decide and what you'd like me to help with or adjust.

greghudson:
I did some testing by hand.  While an individual relation can be set to final, doing so has no effect.  profile_node_iterator() only looks at the final flag when iterating through sections and subsections (prof_tree.c:540), not in the final loop (prof_tree.c:557).

I did confirm that subsections can be made final either via the tag name or the closing brace.

It's hard to decide the best thing to do about these half-finished, most-likely-unused features from decades ago.  We could change the code to match its apparent intent (by adding a p->final check to the final loop in profile_node_iterator()), but absent any known users of this feature, I think I would rather be conservative and just document what works (making sections and subsections final).

greghudson:
I added a commit with tests.  If you would like to take a crack at documenting the current behavior, please go ahead; otherwise I can do it.

frozencemetery:
My inclination would be to remove it since it's never worked, but that's kind of a theme for me at this point :)

mmattioli:
@greghudson let me know what you think.

greghudson:
Individual tags cannot be made final, so that clause in the documentation needs to be removed, not corrected.

Subsections are begun with ``subsection = {``, not ``subsection {``.

I don't think it would be great to devote a lot of space near the front of the krb5.conf documentation talking about a feature nobody uses.  So I'm leaning towards not having examples.

mmattioli:
@greghudson fixed. Sorry, I pushed the wrong branch (was late late night when I committed). Lemme know?

greghudson:
It turns out that the old text "neither the remainder of this configuration file nor any other configuration file" was also inaccurate.  The final flag causes profile_node_iterator() to stop before going on to the next file, but it has no effect on the parser, so the same file (as well as any files included by that file) can still merge in relations to the section or subsection.

I pushed a revision with some rewording and with the removal of the examples.

mmattioli:
@greghudson lgtm

PR #955, opened 2019-07-20
Fix Coverity defects in soft-pkcs11 test code

greghudson:
Coverity claimed to find twelve defects in the softpkcs11 code.  None of these are important because the code is only used for testing, but we don't like the defect count to go up.  Nine of the defects are addressed here so far:

* read_conf_file() leaks its file handle.
* OPENSSL_ASN1_MALLOC_ENCODE() can leave behind its B variable as a freed pointer, causing a double-free in the exit handler of add_certificate().  This case probably can't happen because it would require `i2d_##T` to succeed the first time but fail the second time.  (Four defects.)
* add_st_object() uses the wrong test for an empty table slot (it tests whether the table is NULL, not the slot).
* It doesn't check the return value of asprintf() in two places.  (Two defects.)
* sprintf_fill() doesn't initialize its ap variable with va_start().

I'm still considering what to do about the other three:

* Coverity complains about any use of random(), because it is not a CSPRNG.  We of course don't need a CSPRNG for session and token handles.  We can perhaps use a global counter.  (Two defects.)

* add_st_object() allocates and returns a `struct st_object` pointer, and also remembers the pointer in a table slot in the global soft_token table.  Coverity thinks it finds a path through add_st_object() which does not alias the object, and therefore finds multiple leaks in add_certificate() because the object isn't freed.  Coverity is being uncharacteristically dumb here, because in its hypothetical:
  * the loop executes twice, so `i < soft_token.object.num_objs` is true for two iterations
  * the third time, `i < soft_token.object.num_objs` is false
  * the conditional after the loop, `i == soft_token.object.num_objs` is false

### DIFF
--- a/src/tests/softpkcs11/main.c
+++ b/src/tests/softpkcs11/main.c
@@ -78,6 +78,7 @@ compat_rsa_get0_key(const RSA *rsa, const BIGNUM **n, const BIGNUM **e,
                 (BL) = i2d_##T((S), &p);                \
                 if ((BL) <= 0) {                        \
                     free((B));                          \
+                    (B) = NULL;                         \
                     (R) = EINVAL;                       \
                 }                                       \
             }                                           \
@@ -149,6 +150,7 @@ static struct soft_token {
     } state[10];
 #define MAX_NUM_SESSION (sizeof(soft_token.state)/sizeof(soft_token.state[0]))
     FILE *logfile;
+    CK_SESSION_HANDLE next_session_handle;
 } soft_token;
 
 static void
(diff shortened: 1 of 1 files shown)


greghudson:
I pushed a revision addressing the remaining defects.

greghudson:
I manually submitted a build to Coverity Scan with these changes and verified that there were no defects found in the softpkcs11 source.  (I did that a few days ago, so the build has since been replaced by one without the changes.)

frozencemetery:
Ah, that explains why all the links weren't working.  Thanks!

PR #956, opened 2019-07-26
Fix test suite issues affecting Travis images

greghudson:
In the utilities used by the dejagnu test suites, use
getaddrinfo()/getnameinfo() instead of
gethostbyname()/gethostbyaddr(), as the results can vary when the
local hostname appears in multiple lines in /etc/hosts.

In t_ccselect.py, don't cause an error if the canonicalized local
hostname is "localhost".  The tests will continue to run in this case,
as long as we don't try to create duplicate principals.

In sim_server.c, bind to the wildcard address instead of the resolved
local hostname, to resolve a mysterious problem observed in Travis
where the second of three sim_client send() operations fails with
ECONNREFUSED.

### DIFF
--- a/src/appl/simple/server/sim_server.c
+++ b/src/appl/simple/server/sim_server.c
@@ -69,10 +69,8 @@ main(int argc, char *argv[])
     int flags = 0;                      /* for recvfrom() */
     int on = 1;
     struct servent *serv;
-    struct hostent *host;
     struct sockaddr_in s_sock;          /* server's address */
     struct sockaddr_in c_sock;          /* client's address */
-    char full_hname[MAXHOSTNAMELEN];
     char *cp;
     extern char * optarg;
     int ch;
@@ -133,6 +131,7 @@ main(int argc, char *argv[])
     /* Set up server address */
     memset(&s_sock, 0, sizeof(s_sock));
     s_sock.sin_family = AF_INET;
+    s_sock.sin_addr.s_addr = INADDR_ANY;
 
     if (port == 0) {
         /* Look up service */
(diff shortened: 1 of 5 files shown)


greghudson:
Background: some time recently, Travis appears to have changed the /etc/hosts configuration on their images.  On a sample image with the hostname `travis-job-5e491971-54a5-47aa-ac32-367989b4cd73`, /etc/hosts contained:

    ## Managed by Chef on packer-5ce6b5d9-49d8-d1ca-bd09-7360ee07baea.c.eco-emissary-99515.internal :heart_eyes_cat:
    ## cookbook:: travis_build_environment
    ##     file:: templates/default/etc/cloud/templates/hosts.tmpl.erb
    127.0.1.1 travis-job-5e491971-54a5-47aa-ac32-367989b4cd73 travis-job-5e491971-54a5-47aa-ac32-367989b4cd73 ip4-loopback xenial64
    127.0.0.1 localhost nettuno travis vagrant travis-job-5e491971-54a5-47aa-ac32-367989b4cd73

With the hostname appearing on both the 127.0.1.1 and 127.0.0.1 lines, gethostname->getaddrinfo->getnameinfo yields "localhost" while gethostname->gethostbyname->gethostbyaddr yields the hostname.  The Python test suite was fine, but the various dejagnu test suites were still using gethostbyname.

I have no idea what was causing issues with sim_server.  The krb.log output from sim_client in the failure cases was:

    Got credentials for sample.
    Sent authentication data: 542 bytes
    /home/travis/build/greghudson/krb5/src/tests/dejagnu/../../appl/simple/client/sim_client: Connection refused while sending SAFE message
    Sent checksummed message: -1 bytes
    Sent encrypted message: 137 bytes

I verified that the KRB-SAFE message had a reasonable length (126), but I can't imagine what kind of address misconfiguration might cause only the second of three UDP messages to be consistently rejected with ECONNREFUSED on one datagram socket.  I don't like fixing issues without identifying them, but I don't know how to investigate this one further, and the fix is a desirable simplification anyway.  (Both the appl/sample and appl/gss-sample servers bind to the wildcard address.)

mmattioli:
It's a shot in the dark but have you tried using 18.04?

```
dist: bionic
```

greghudson:
We didn't have much luck using the 16.04 image before Travis was ready to make it the default (builds were slow, and apt would randomly fail due to locking problems when run from the build script), so I don't want to do that with 18.04.  We want the test suite to be resilient to weird hostname configurations like this, so these fixes are needed either way.

mmattioli:
I hear you. I'm just a little wary as this appears to be related and specific to Travis.

PR #957, opened 2019-07-28
Remove Kerberos for macOS support

mmattioli:
Remove all code for supporting Kerberos for macOS.


greghudson:
It seems likely that we can just remove this code; it looks like leftover KfM stuff.  We'd probably want to remove ccapi/lib/mac, ccapi/common/mac, and ccapi/server/mac at the same time.

mmattioli:
@greghudson I was just doing some more edits and then saw your latest comment on removing it altogether 😅 

Just force-pushed a single commit which removes support for Apple macOS. Lemme know wutcha think.

greghudson:
I think the current commit in this PR just removes the old KfM files, so it looks fine even if it's framed improperly as removing macOS support.  (src/ccapi was used in KfM and is used in KfW, but is not used in the main Unix build.  The Heimdal-based OS X native Kerberos implementation uses the KCM protocol rather than CCAPI protocol, and we have KCM client support in the main Unix build to interoperate with it.)

I'm conditionally in favor of cleaning up old code, but I'd generally like to hit any given area of code as few times as possible to minimize the merge headaches.  Sometimes there are so many things to clean up about a given piece of code that cleaning it up all at once can be difficult (example: spnego_mech.c has the wrong style, lots of forward declarations which could be minimized by reordering functions, mixes up protocol logic with DER encoding, and we should probably factor out DER encoding anyway).  On the other hand, if there are only a few small things wrong with a piece of code that don't really get in the way of maintaining it, then it's often simplest to leave it alone.

I will merge #956 soon and can rebase this PR onto master afterwards.

mmattioli:
Passing builds! 😅

PR #959, opened 2019-07-31
Eliminate qualname.plin

greghudson:
Rewriting the qualname Perl script to use getaddrinfo created an
unchecked dependency on Perl 5.14.  Instead, remove the script and use
the C program in tests/resolve for the kadmin and gssrpc test suites.


greghudson:
The additional background here is that my Solaris buildbot worker machine uses OpenCSW to piece together a reasonable build environment, and OpenCSW's perl package tops out at 5.10.

PR #962, opened 2019-08-03
Display daemon output in Python test scripts

greghudson:
In k5test.py, if a daemon process exits before we terminate it,
display the exit status.  If a daemon process generates output beyond
the sentinel, display the output before terminating the process.

[This change is particularly useful when investigating memory bugs in daemon processes using asan, as it displays the asan output which otherwise disappears into the k5test pipe.]

### DIFF
--- a/src/util/k5test.py
+++ b/src/util/k5test.py
@@ -371,6 +371,7 @@ command-line flags.  These are documented in the --help output.
 """
 
 import atexit
+import fcntl
 import optparse
 import os
 import shlex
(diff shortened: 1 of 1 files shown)


src/util/k5test.py
```diff
@@ -812,7 +814,26 @@ def _start_daemon(args, env, sentinel):
     return proc
 
 
+# Check a daemon's status prior to terminating it.  Display its return
```
frozencemetery: I don't know if our python style cares about declaration-before-use, but this is used on line 475 before being declared on 819.

greghudson: It's fine; since Python doesn't have forward declarations, you couldn't do mutually recursive functions if you couldn't make a forward reference.  I think it makes more sense to have it here with the other two daemon functions than before its use in _onexit().

PR #963, opened 2019-08-04
Skip URI tests when using asan

greghudson:
resolve_wrapper uses RTLD_DEEPBIND to load libresolv, triggering a
failure in the asan runtime.


greghudson:
While working on this, I noticed that I don't get a clean result from the test suite under asan now that the PKCS11 tests are always run, because soft_pkcs11 doesn't fully clean up its state in C_Finalize().  I added a second commit to fix those leaks.

PR #965, opened 2019-08-07
Allow the KDB to see and modify auth indicators

greghudson:
Amend the sign_authdata method signature to include a modifiable
auth_indicators array.  Bump the DAL major version and the libkdb5
soname.

[This overlaps with PR #912 which also adds a parameter to sign_authdata, but the merge should be straightforward.  I will likely want to add a test case by having the test KDB module do something to the auth indicators in its sign_authdata method before merging, but this should be enough to get feedback from Samba developers.]

### DIFF
--- a/src/kdc/kdc_authdata.c
+++ b/src/kdc/kdc_authdata.c
@@ -320,7 +320,8 @@ fetch_kdb_authdata(krb5_context context, unsigned int flags,
                    krb5_keyblock *server_key, krb5_keyblock *header_key,
                    krb5_kdc_req *req, krb5_const_principal for_user_princ,
                    krb5_enc_tkt_part *enc_tkt_req,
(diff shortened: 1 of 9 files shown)


iboukris:
I guess kdb_test plugin minor version should be set to 0.

greghudson:
I added a test case and corrected the test KDB module's minor version.

Alexander said he would attempt an experiment using FreeIPA and Samba, so I will likely wait for that before merging.

src/kdc/kdc_authdata.c
```diff
@@ -892,5 +895,6 @@ handle_authdata(krb5_context context, unsigned int flags,
             return ret;
     }
 
+
```
frozencemetery: Extra whitespace.

greghudson: Fixed, thanks.

PR #966, opened 2019-08-09
Improve daemon checking in Python test scripts

greghudson:
[I ran into a daemon issue in a recent PR and PR #962 was useful, but the test script threw a Python exception trying to kill the KDC process which had already been reaped.  There is also a possibility that a daemon might crash in a way that isn't visible to the client side of the tests, in which case the script might exit successfully even though we noticed and logged an early exit.  This PR corrects both problems.]

If a daemon exits early and we detect it with check_daemon(), avoid
trying to terminate it again as the process entry will have been
reaped.  Check all daemons on successful exit and exit with an error
if any daemons exited early.

Also remove a piece of Python 2.5 compatibility code which is no
longer relevant with Python 3.

### DIFF
--- a/src/util/k5test.py
+++ b/src/util/k5test.py
@@ -407,6 +407,7 @@ def fail(msg):
 
 def success(msg):
     global _success
+    _check_daemons()
     output('*** Success: %s\n' % msg)
     _success = True
 
@@ -427,6 +428,7 @@ def skipped(whatmsg, whymsg):
 def skip_rest(whatmsg, whymsg):
     global _success
     skipped(whatmsg, whymsg)
+    _check_daemons()
     _success = True
     sys.exit(0)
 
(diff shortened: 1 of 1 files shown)


greghudson:
The part where we check daemon exit statuses exposes some bad wiring behind the drywall:

* uuserver (from appl/user_user) serves one connection and exits, if it is compiled with -DDEBUG.  Otherwise it assumes it is started from inetd.
* It is built with -DDEBUG by default.
* It is installed as built that way.  (No man page, though.)
* We use it in Python test scripts in two places, in appl/user_user/t_user2user.py and in tests/t_audit.py.
* The appl/sample server also serves one connection and exits, or (if no port number is given) assumes it is run out of inetd.  We have a dejagnu test using that server but no Python test.
* The appl/simple server binds a UDP socket and exits after one request.  We don't have any Python tests using that server.  We have a dejagnu test using that server but no Python test.
* We have a way to wrap servers in an inetd-like wrapper (tests/dejagu/t_inetd.c) but it makes it harder to attach to the server process in gdb if there is a server-side problem to debug.  Although there is a k5test.py function to use t_inetd, I think right now only the dejagnu sample.exp script uses it for anything.

For now I will probably make this PR tolerate early exits with status 0.

frozencemetery:
uuserver is one of a few binaries that lack man pages and therefore I don't ship in more recent builds (the full skip list is krb5-send-pr, sim_server, gss-server, uuserver, sim_client, gss-client, and uuclient).

Are you planning a follow-up PR to address some of these problems?  If so, +1 to this PR from me.

greghudson:
I'm not certain how to scope cleanup work here.  I could sink a lot of time into the sample apps (separating out testing code which has crept into gss-sample, trying to make them more readable, maybe getting rid of appl/simple altogether) but it doesn't seem like a high priority.

PR #969, opened 2019-08-14
KCM: fix k5_input status check in get_kdc_offset()

frozencemetery:
An inverted status check in get_kdc_offset() would cause querying the
offset time from the ccache to always fail (silently) on KCM.  Fix the
status check so that KCM can properly handle desync.


abbra:
LGTM. The other place where an explicit check for the reply status is done,  `kcmio_call()`, confirms that a non-zero status is treated as an error condition.

PR #970, opened 2019-08-17
Add new cross-realm KDB flags and adjust S4U2Self handling

iboukris:
Following the discussion in #912 about KRB5_KDB_FLAG_CROSS_REALM, I went over the work we did on adding constraints to S4U2Self handling, and realized the logic could be adjusted slightly to better address these cases where header ticket server appearances are misleading. I'm working on documentation.


iboukris:
I thought about it some more, it looks like we can do okay using only LOCAL_TGT again, so if we are okay with changing CROSS_REALM definition to be as LOCAL_TGT as in #912, then we don't need an extra flag. It'd definitely make it easier to rebase RBCD PR.

greghudson:
Because of renewal and validation requests, the header ticket server isn't necessarily a TGT.  It could even be a cross TGT issued by the local realm (krbtgt/FOREIGN@LOCAL, being renewed or validated).

iboukris:
Hmmm, could you elaborate, it looks like s4u2self doesn't go with renewal, see: https://github.com/krb5/krb5/blob/154551ad22e90d2e5f60103059fbaaadac017420/src/kdc/kdc_util.c#L1551

greghudson:
Oh, I didn't get the impression that the new LOCAL_TGT flag was scoped specifically to S4U2Self requests.

iboukris:
It isn't, in fact I rather change CROSS_REALM to it as mentioned, I just noticed that comment.
But I'm unclear what would go wrong with renew/validate (I'm not familiar with it).

iboukris:
Oh, you mean the name LOCAL_TGT, we could find an alternative if we don't change CROSS_REALM. As far as I can tell the new definition would make sense in all the cases where CROSS_REALM is currently being used (transit check, mandating authdata and S4U).

iboukris:
I spoke too soon, once I update the code and changed LOCAL_TGT to CROSS_REALM, final cross-realm s4u2self request fails in tests, it's strange as I have the same definition in #912 which doesn't fail the tests, don't know what's missing, will check tomorrow.

greghudson:
In PR #912 you also change the code to set enc_tkt_reply.client so that it no longer cares about the CROSS_REALM flag.  In this PR you did not, and the change in definition of CROSS_REALM causes enc_tkt_reply.client to be set to subject_tkt->client instead of to s4u_x509_user->user_id.user as it is on master (or to altcprinc as it is in PR #912).  As a result, the KDC decides to add UREALM to the transited list, and then fails the transited list check (because the client and server realms are both SREALM and it doesn't like seeing anything in between).

greghudson:
To see if we can redefine CROSS_REALM safely, I looked at the existing uses.  They are:

* Deciding whether to set the issued ticket client to the S4U2Self requested user.  That case can be addressed by this PR.

* Deciding whether to call validate_transit_path(), which is a weird function that doesn't actually look at the transit path but instead enforces the XREALM_NON_TRANSITIVE flag, which isn't documented and doesn't have any kadmin support code.  This code came in with 1.7 as part of a big Novell merge, and should probably be removed.

* In fetch_kdb_authdata() to decide whether to call krb5_db_sign_authdata() for a non-S4U request, if there was no authdata in the header ticket.  I think the theory is that if a locally issued ticket has no authdata, the KDB module already had its chance to put authdata in there during the AS-REQ, so there's no need to give it a second go.  The new definition of CROSS_REALM seems like it should also work for this check.

The string "CROSS_REALM" does not appear in FreeIPA or Samba so we probably don't have to worry about existing KDB modules.

iboukris:
> In PR #912 you also change the code to set enc_tkt_reply.client so that it no longer cares about the CROSS_REALM flag.

Right, forgot about that bit, fixed now. I'll start rebasing RBCD PR.

greghudson:
> Deciding whether to call validate_transit_path(), which is a weird function that doesn't actually look at the transit path but instead enforces the XREALM_NON_TRANSITIVE flag, which isn't documented and doesn't have any kadmin support code. This code came in with 1.7 as part of a big Novell merge, and should probably be removed.

The string XREALM_NON_TRANSITIVE does not appear in FreeIPA or Samba, so removing this code would probably have no short-term negative impact, but I'm looking closer since Windows has the concept of non-transitive trusts.

validate_transit_path() applies these rules to cross-realm requests:

1. If the request server has the XREALM_NON_TRANSITIVE bit set, the request is rejected.  This rule is labeled "Incoming".  So service principals (or cross TGTs to other realms) with the non-transitive bit can't be accessed from foreign realms.

2. If the header ticket server has the XREALM_NON_TRANSITIVE bit set, then (a) the request server must be the same as the header ticket server, and (b) the client must be in the same realm as the header ticket server.  This rule is labeled "outgoing".  This rule seems too strict to be useful.  (b) alone would mean that the foreign realm can only issue cross-TGTs for its own clients, but the addition of (a) means that the holder of a cross-TGT can't do anything except get more of the same cross-TGTs.

I couldn't find any relevant material on the web about this bit in the context of eDirectory (where the code came from).  It seems like it might have been an unfinished feature in the upstream code.  I still think we can remove it.

iboukris:
> but the addition of (a) means that the holder of a cross-TGT can't do anything except get more of the same cross-TGTs.

Ought to be highly addictive.

I think non-transitive trust policy can be implemented via kdb module so I guess we can get rid of it. The other private flag also seem unused.

greghudson:
I added a preceding commit to remove validate_transit_path(), and edited the main commit.  I expect to merge this tomorrow if nothing comes up.

iboukris:
LGTM, why do we keep the symbol though?

iboukris:
> If the header ticket server has the XREALM_NON_TRANSITIVE bit set, then (a) the request server must be the same as the header ticket server, and (b) the client must be in the same realm as the header ticket server.

Wait, isn't that actually an OR between (a) and (b) ? It also makes more sense.

I realize that code actually wanted the old definition of CROSS_REALM.

iboukris:
Another late thought, in the new definition of CROSS_REALM, wouldn't it be better to use 'header_ticket->server' instead of 'header_server->princ', the latter being filled by kdb module.

greghudson:
> Wait, isn't that actually an OR between (a) and (b) ? It also makes more sense.

That was my first reading.  But these are the conditions that will cause us to reject the request, not the ones that will cause us to accept it.  So both conditions must be false for the request to continue.

> Another late thought, in the new definition of CROSS_REALM, wouldn't it be better to use 'header_ticket->server' instead of 'header_server->princ', the latter being filled by kdb module.

Yes, probably.

iboukris:
> That was my first reading. But these are the conditions that will cause us to reject the request, not the ones that will cause us to accept it. So both conditions must be false for the request to continue.

Oh right ... I bet it's a bug and was meant to be an '&&' so either will allow the request to continue.

>>Another late thought, in the new definition of CROSS_REALM, wouldn't it be better to use 'header_ticket->server' instead of 'header_server->princ', the latter being filled by kdb module.
>
>Yes, probably.

Okay, I'll also move it a bit upward so it will be available earlier.

iboukris:
Hmm, I think the new definition CROSS_REALM isn't actually what we want to check for in the new check_rbcd_policy() call, I'm trying to think what to change, sorry..

greghudson:
I'm not sure what the issue is with check_rbcd_policy(), but does defining cross-realm as "either the header client or server is foreign" work for that case?  It should be fine for the other cases.

iboukris:
In check_rbcd_policy() we check cross_realm to know if we should expect a 2nd ticket or a 2nd proxy TGT. Either the client or server is foreign would work ok for this, I'm reviewing the other cases.

greghudson:
> In check_rbcd_policy() we check cross_realm to know if we should expect a 2nd ticket or a 2nd proxy TGT.

Hm, I'm not seeing the problem.  When the evidence ticket is normal, the header ticket client and server should both be local.  When the evidence ticket is a proxy TGT, the header ticket client and server should both be foreign.  So any definition of CROSS_REALM we've considered (client foreign, server foreign, either of them foreign) seems like it should work to recognize the proxy TGT.

greghudson:
> LGTM, why do we keep the symbol though?

I changed the predecessor commit to remove the private flag symbols.  I didn't change the main commit.

iboukris:
> I'm not seeing the problem. When the evidence ticket is normal, the header ticket client and server should both be local. When the evidence ticket is a proxy TGT, the header ticket client and server should both be foreign. 

My concern was a foreign principal with local tgt, with a normal 2nd ticket for self, but I guess this won't pass the server lookup of the 2nd ticket.

iboukris:
In RBCD kdb_test code, in the case of a foreign principal using a local tgt.
If we use cross_realm_either:  we won't verify the KDC signature as it will be considered cross_realm.
If we use cross_realm_server:  we won't consider it cross_realm which impacts the expectations of PAC for RBCD.

greghudson:
For S4U2Proxy it's fine to break a foreign principal using a local TGT; the client should only be following referrals.  The in-practice case I know of where a foreign principal uses a local TGT is for traditional forwarded-ticket delegation.

From what's been said so far I prefer cross-realm to be defined as cross-realm-server.

iboukris:
So yea, if we know this case fails there is no need to worry about sign_authdata. I reviewed again and I'm okay with cross-realm-server so the second commit looks good to me, note that it changed since you last reviewed.

greghudson:
> I'm okay with cross-realm-server so the second commit looks good to me, note that it changed since you last reviewed.

Moving the setting of CROSS_REALM before search_sprinc() doesn't cause it to be set for the server principal lookup, because search_sprinc() takes s_flags and CROSS_REALM is only set in C_FLAGS.  Do we need it set for the server lookup?

iboukris:
No it doesn't, i noticed s_flags and then forgot about it, I'll fix..

PR #971, opened 2019-08-22
Track first local TGT key in KDC code

greghudson:
[One of the commits in PR #912 wants to provide the first local TGT key to the KDB sign_authdata method for PAC signing, adding a fourth consumer of this key to the existing three.  This commit extracts that key at a higher level so that we aren't doing it multiple times per request.]

Decrypt the first local TGT key in get_local_tgt() and save it in the
AS and TGS processing functions.  (As we now sort key data by
descending kvno, this is guaranteed to be the most recent key.)  Pass
this key to the authdata and FAST cookie functions to simplify cookie
encryption and authdata signing.  Decryption and verification
functions must still sometimes decrypt earlier keys to process tickets
predating the last local TGT key rollover.

### DIFF
--- a/src/kdc/fast_util.c
+++ b/src/kdc/fast_util.c
@@ -463,37 +463,19 @@ make_padata(krb5_preauthtype pa_type, const void *contents, size_t len,
 }
 
 /*
- * Construct the secure cookie encryption key for the given local-realm TGT
- * entry, kvno, and client principal.  The cookie key is derived from the first
- * TGT key for the given kvno, using the concatenation of "COOKIE" and the
- * unparsed client principal name as input.  If kvno is 0, the highest current
- * kvno of the TGT is used.  If kvno_out is not null, *kvno_out is set to the
- * kvno used.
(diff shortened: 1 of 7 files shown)


src/kdc/fast_util.c
```diff
@@ -504,18 +486,47 @@ get_cookie_key(krb5_context context, krb5_db_entry *tgt, krb5_kvno kvno,
...
+    krb5_key_data *kd;
+
+    *key_out = NULL;
+    memset(&storage, 0, sizeof(storage));
+
+    if (kvno == tgt->key_data[0].key_data_kvno) {
+        /* We can use the already-decrypted first key of the local TGT. */
+        key = tgt_key;
+    } else {
+        /* The cookie used an oldder TGT key; find and decrypt it. */
```
frozencemetery: "older"

greghudson:
I just noticed that the KDC CAMMAC code also uses the local TGT key, so I need to revise this PR to touch that code as well.

greghudson:
Okay, I revised the PR to also pass the local TGT key to the CAMMAC functions.

PR #972, opened 2019-08-22
Simplify krb5_dbe_def_search_enctype()

greghudson:
Key data is now sorted in descending kvno order (since commit
44ad57d8d38efc944f64536354435f5b721c0ee0) and key enctypes can be
compared with a simple equality test (since single-DES support was
removed in commit fb2dada5eb89c4cd4e39dedd6dbb7dbd5e94f8b8).  Use
these assumptions to simplify krb5_dbe_def_search_enctype().

The rewrite contains one probably-unnoticeable bugfix: if enctype,
salttype, and kvno are all given as -1 in a repeated search, yield all
key entries of permitted enctype, not just entries of the maximum
kvno.

### DIFF
--- a/src/lib/kdb/kdb_default.c
+++ b/src/lib/kdb/kdb_default.c
@@ -37,94 +37,67 @@
 
 
 /*
- * Given a particular enctype and optional salttype and kvno, find the
- * most appropriate krb5_key_data entry of the database entry.
- *
- * If stype or kvno is negative, it is ignored.
- * If kvno is 0 get the key which is maxkvno for the princ and matches
- * the other attributes.
+ * Set *kd_out to the key data entry matching kvno, enctype, and salttype.  If
+ * any of those three parameters are -1, ignore them.  If kvno is 0, match only
+ * the highest kvno.  Begin searching at the index *start and set *start to the
+ * index after the match.  Do not return keys of non-permitted enctypes; return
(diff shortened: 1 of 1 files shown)


greghudson:
I'm mulling over whether to go further and change the interface, since we're bumping the libkdb5 soname anyway:

* There are only two in-tree consumers of krb5_dbe_search_enctype() (as opposed to _find_enctype()).  One is already slated in PR #971 to use _find_enctype() instead, and the other one (preauth_ents.c) could use the client_keys() callback instead.  I think it's unlikely that out-of-tree programs are using the search API.  So we could get rid of the search API and simplify the find function.

* KDB modules don't really need to be able to override this function, so we could remove it from the DAL function table and implement it directly.

* No one wants to be able to search by salt type, so we could remove that parameter.  Unfortunately, this is only 99% true; we have an libkadm5srv API kadm5_decrypt_key() which takes an optional salt parameter and searches with it.  We could change that API as well, or even disable it; I think it was only ever used by krb4-related stuff like krb524 and OpenAFS fakeka.

* The enctype parameter should be of type krb5_enctype, and ENCTYPE_NULL (0) should be used as the "don't care" value instead of -1, which is theoretically a valid local-use enctype.

But there's also something to be said for not pulling on the thread for too long.

PR #973, opened 2019-08-30
Static analysis reports (2019-08-30)

frozencemetery:
Fix three pedantic bugs.  hashtable one is new; the rest are newly detected as a result of tooling updates on our end.

### DIFF
--- a/src/lib/gssapi/krb5/duplicate_name.c
+++ b/src/lib/gssapi/krb5/duplicate_name.c
@@ -34,8 +34,7 @@ krb5_gss_duplicate_name(OM_uint32 *minor_status, const gss_name_t input_name,
     krb5_error_code code;
     krb5_gss_name_t princ, outprinc;
 
-    if (minor_status)
-        *minor_status = 0;
+    *minor_status = 0;
 
     code = krb5_gss_init_context(&context);
     if (code) {
--- a/src/lib/rpc/clnt_tcp.c
+++ b/src/lib/rpc/clnt_tcp.c
@@ -168,9 +168,9 @@ clnttcp_create(
 	if (*sockp < 0) {
 		*sockp = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
 		(void)bindresvport_sa(*sockp, NULL);
-		if ((*sockp < 0)
-		    || (connect(*sockp, (struct sockaddr *)raddr,
-		    sizeof(*raddr)) < 0)) {
+		if (*sockp < 0 || raddr == NULL ||
+		    connect(*sockp, (struct sockaddr *)raddr,
+			    sizeof(*raddr)) < 0) {
 			rpc_createerr.cf_stat = RPC_SYSTEMERROR;
 			rpc_createerr.cf_error.re_errno = errno;
                         (void)closesocket(*sockp);


greghudson:
I don't think uint64_t % size_t is undefined.  Can you say more about the static analysis report there?

greghudson:
"Fix explicit NULL dereference in krb5_gss_duplicate_name()": I took a look at other gss-krb5 functions and at the mechglue, and I think we want to remove the conditional elsewhere in the function rather than add a new one.

The minor_status parameter to GSS functions is not optional.  In the mechglue, there is a great deal of argument validation, and the practice is to return GSS_S_CALL_INACCESSIBLE_WRITE if minor_status is NULL.  In the krb5 mech, the overwhelmingly common practice is to just assume that minor_status is non-null, since the mechglue will have validated it already.

frozencemetery:
Perhaps I misunderstood what it was complaining about?  Here it is in full:

```
#     i = siphash24(key, klen, ht->k0, ht->k1) % ht->nbuckets;
#         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
4. krb5-1.17/src/util/support/hashtab.c:194:9: note: Assuming the condition is true
#     if (ht->nentries == ht->nbuckets) {
#         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
7. krb5-1.17/src/util/support/hashtab.c:194:5: note: Taking true branch
#     if (ht->nentries == ht->nbuckets) {
#     ^
10. krb5-1.17/src/util/support/hashtab.c:195:9: note: Taking false branch
#         if (resize_table(ht) != 0)
#         ^
13. krb5-1.17/src/util/support/hashtab.c:200:9: note: Assuming 'ent' is not equal to NULL
#     if (ent == NULL)
#         ^~~~~~~~~~~
16. krb5-1.17/src/util/support/hashtab.c:200:5: note: Taking false branch
#     if (ent == NULL)
#     ^
19. krb5-1.17/src/util/support/hashtab.c:206:46: note: The result of the '%' expression is undefined
#     i = siphash24(key, klen, ht->k0, ht->k1) % ht->nbuckets;
#         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
#   204|       ent->val = val;
#   205|   
#   206|->     i = siphash24(key, klen, ht->k0, ht->k1) % ht->nbuckets;
#   207|       K5_SLIST_INSERT_HEAD(&ht->buckets[i], ent, next);
#   208|   
```

I've dropped it from the series for now.

frozencemetery:
Updated the ` krb5_gss_duplicate_name()` commit to remove the NULL check entirely.

greghudson:
Hm.  The C99 standard says of the modulus operator:

"The usual arithmetic conversions are performed on the operands."

"The result of the / operator is the quotient from the division of the first operand by the
second; the result of the % operator is the remainder. In both operations, if the value of
the second operand is zero, the behavior is undefined."

The section on "the usual arithmetic conversions" (6.3.1.8) doesn't have any undefined behavior.  So unless clang-analyzer is way off base here, it thinks the second operand is 0 in some case.  There's a resize_table() call in the hypothetical, but nothing else to support the notion that ht->nbuckets could be 0.

greghudson:
"Fix explicit NULL deref in gssrpc's TCP client": this is a false positive arising from to clnttcp_create()'s complicated contract and a lack of argument validation.  It is a caller error if *sockp < 0 and raddr == NULL, but it is valid for raddr to be NULL if *sockp is already set.  The analyzer sees a check for raddr != NULL, followed by a connect() conditional on *sockp < 0, and treats that as a forward null defect, since it does not know anything about the relationship between *sockp and raddr.

This contract is actually my creation from commit 0080380b3b3e945c5eb84504771d9d01ee76a611 which added minimal IPv6 support.  Prior to that (looking at the version 1.8 source), raddr had to be non-NULL in all cases, although the connect() was still conditional on *sockp being unset.

In both the prior and modified contracts, the code permits the awkward possibility that the portmapper is consulted to fill in raddr->sin_port (which remains set for the caller), but the address is ignored.

libtirpc's clnttcp_create() requires raddr must always be non-null, similar to the 1.8 contract.  It also appears to handle the case where *sockp is set but unconnected (as determined by a getpeername() call), in which case libtirpc connects it to raddr.  If  a connected socket is passed in, the same awkward possibility exists of contacting the portmapper and ignoring the result.

The proposed change is to return RPC_SYSTEMERROR if raddr is NULL at the time we want to use it for connect().  I think that's a reasonable minimal change, although I'll want to rewrite the commit message.  A more invasive change would be to restructure the code so that all uses of raddr  are conditional on *sockp < 0.  But we try not to churn the RPC code more than necessary, so going with the minimal change makes sense.

PR #974, opened 2019-09-10
Update LDAP KDB module documentation

greghudson:
[I'm considering getting rid of ldapbackend.rst instead of updating it.  It is weirdly filed under "advanced", when it is actually a simpler set of instructions for minimally setting up the LDAP KDB module cohosted with the KDC on Ubuntu using the administrative DN for binding.  There is some value in that, but it is largely redundant with conf_ldap.rst, and it's not clear that this kind of setup is really valuable outside of testing.  Perhaps it could be moved to a new section in https://k5wiki.kerberos.org/wiki/Manual_Testing .]

The instructions for setting up an index won't work with cn=config, so I changed them to say less.  Unfortunately the OpenLDAP documentation also doesn't say how to add an index with cn=config, but that isn't really our job.]

Simplify ldapbackend.rst, using kerberos.openldap.ldif (added by
ticket 8529) and removing unnecessary command arguments.

Simplify conf_ldap.rst in the same way.  Mention the possibility of
using SASL authentication (added by ticket 7944) as an alternative to
binding with DN and password.  Remove unnecessary access rights.

In kdc_conf.rst, remove ldap_servers from the list of relations read
from [dbdefaults], as it is only read from the realm's database
configuration section.

ticket: 8831 (new)
tags: pullup
target_version: 1.17-next


dilyanpalauzov:
> Unfortunately the OpenLDAP documentation also doesn't say how to add an index with cn=config.

What exactly shall be indexed?  My undersnading is that OpenLDAP contains many databases and the index has to be added in the database containing the specific ldap tree.  E.g. in the database for kerberos5-ldap-container. The cn=config is a separate database in ldif format and this format probaby does not support indices.

dilyanpalauzov:
You can add in the documentation for kdb5_ldap_util list, that if the -H parameter is not provided, then the ldap connection setrings for the default realm are used.  If the default realm is not linked to LDAP backend, then -H must be used.

greghudson:
> What exactly shall be indexed?

I think there is a global list of attributes to be indexed, and we want the admin to add krbPrincipalName to that list so that lookups by principal name can be fast.  The old instructions were:

    Add ``krbPrincipalName`` to the indexes in slapd.conf to speed up the access.

In slapd.conf, this is controlled by the "index" directive.  But OpenLDAP has deprecated slapd.conf in favor of LDAP attributes under cn=config which can be updated live, unfortunately without updating the documentation or creating easy-to-use tools.  There are instructions here: https://miek.nl/2009/april/15/openldap-2.4-cnconfig/
but I don't want to link to a blog post from our docs.

dilyanpalauzov:
I do not think that the Kerberos manual shall copy the OpenLDAP manual in every detail.  It is sufficient to recommend creating an index over `krbPrincipalName eq,pres,sub`.  The documentation of OpenLDAP itself is not completely converted to using the ldif format.

On my system I have in `etc/openldap/cn=config/olcDatabase\=\{2\}mdb.ldif`:
```
dn: olcDatabase={2}mdb
objectClass: olcDatabaseConfig
objectClass: olcMdbConfig
olcDatabase: mdb
olcDbDirectory: /var/openldap-data/krbContainer
olcSuffix: cn=krbContainer
olcRootDN: uid=admin_kdc,cn=krbContainer
olcRootPW:: ABCD
olcDbIndex: krbPrincipalName eq,pres,sub
olcDbMaxSize: 31457280
structuralObjectClass: olcMdbConfig
entryUUID: 52e1a35c-a1a2-43a4-bc51-cc3cfba75953
creatorsName: gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth
createTimestamp: 20190830190418Z
olcAccess: {0}to attrs=userPassword,shadowLastChange,krbPrincipalKey by dn="
 uid=admin_kdc,cn=krbContainer" write by anonymous auth by self write by * d
 isclose
olcAccess: {1}to attrs=shadowLastChange by dn="uid=admin_kdc,cn=krbContainer
 " write by self write by * read
olcAccess: {2}to dn.subtree="cn=krbContainer" by dn.exact="uid=admin_kdc,cn=
 krbContainer" write by dn.exact="gidNumber=0+uidNumber=0,cn=peercred,cn=ext
 ernal,cn=auth" read by * disclose
```

Just assume the user knows how to insert indices, here `olcDbIndex: krbPrincipalName eq,pres,sub`.

`9. Add an index for ``krbPrincipalName`` to speed up access. ` says it.

I have `olcDbIndex: krbPrincipalName eq,pres,sub` which is taken from https://help.ubuntu.com/lts/serverguide/kerberos-ldap.html.  Shall the index be really on `pres,eq,sub` searches?

greghudson:
After looking at the index documentation and the LDAP KDB code, I think only `eq` is necessary.

dilyanpalauzov:
Okay, I filed https://bugs.launchpad.net/serverguide/+bug/1843504 .

`9. Add an index for ``krbPrincipalName`` to speed up access.` should be changed to `9. Add an ‘eq’ index for ``krbPrincipalName`` to speed up access.`

greghudson:
I added some kdb5_ldap_util doc revisions after some research:

* The -r option is implemented as a global option.  The documentation pretended that it only applied to certain commands, but the choice of realm affects every command in some way.  stashsrvpw  comes the closest to not using the realm, but even then the realm configuration still affects the default service password file.

* The options -P, -k, -kv, -M, -m, and -sf are also implemented as global options.  (These all affect the master key and are similar to kdb5_util options.)  -M wasn't documented at all; the others were documented under "create".  I think create is the only kdb5_ldap_util command which does anything with principle key data, so this is actually correct.  I added documentation for -M.

As requested, I also documented that "kdb5_ldap_util list" uses realm configuration by default.

greghudson:
Updated to remove ldapbackend.rst instead of updating it.  I created https://k5wiki.kerberos.org/wiki/Manual_Testing#LDAP_test_KDB_setup .

doc/admin/conf_ldap.rst
```diff
@@ -85,33 +73,23 @@ Configuring Kerberos with OpenLDAP back-end
...
-           ldap_kdc_dn
-           ldap_kadmind_dn
-           ldap_service_password_file
-           ldap_servers
-           ldap_conns_per_server
+       db_library (set to ``kldap``)
+       ldap_kdc_dn
+       ldap_kadmind_dn
+       ldap_service_password_file
+       ldap_servers
```
dilyanpalauzov: Isn’t **ldap_kerberos_container_dn** also mandatory to set?

greghudson: Yes.  I have updated the PR to add that to the list.

doc/admin/admin_commands/kdb5_ldap_util.rst
```diff
@@ -252,7 +250,9 @@ list
 
     **list**
 
-Lists the name of realms.
+Lists the name of realms under the container.  If not overridden by
+command-line options, the LDAP server information for the default
+realm is used to search the container.
```
dilyanpalauzov: The synopsis of kdb5_ldap_util was updated, so that -r REALM can be passed to all commands and the lack of it implies the default realm.  This means, that “kdb5_ldap_util list -r X.NET” will use the LDAP settings for the X.NET realm and “kdb5_ldap_util list” will use the LDAP settings of the default realm.  Before definigng a global `-r` this was not so obvious, and ignoring the ldap_servers in [dbdefaults] made things even more confusing.  In my experiment I had ldap_servers in [dbdefault] and the default realm has not used LDAP backend.

I did ask to document for `list` where it obtains its settings for the LDAP connection from.  Provided that a missing `-r` implies settings for the default realm, I think now that `If not overridden by command-line options, the LDAP server information for the default realm is used to search the container.` is redundant.

greghudson:
My test slapd installation seems to have these indexes by default:

    olcDbIndex: objectClass eq
    olcDbIndex: cn,uid eq
    olcDbIndex: uidNumber,gidNumber eq
    olcDbIndex: member,memberUid eq

Also, I can't tell from the logs whether this is what's happening, but it seems like if as long as there is an index on krbPrincipalName, slapd could narrow down the candidates to just one, and then sequentially verify the objectClass on that one candidate.

doc/admin/admin_commands/kdb5_ldap_util.rst
greghudson: Okay, I'll take it out.

dilyanpalauzov:
I have compiled openldap and krb5 from source and have written the configuration files from scratch.  i do not have any default indices in OpenLDAP. Since I had no `olcDbIndex: objectClass eq`, but only index on `krbPrincipalName eq,pres,sub`, this query `SRCH attr=krbprincipalname krbcanonicalname objectclass krbprincipalkey krbmaxrenewable age krbmaxticketlife krbticketflags krbprincipalexpiration krbticketpolicyreference krbUpEnabled krbpwdpolicyreference krbpasswordexpiration krbLastFailedAuth krbLoginFailedCount krbLastSuccessfulAuth krbLastPwdChange krbLastAdminUnlock krbPrincipalAuthInd krbExtraData krbObjectReferences krbAllowedToDelegateTo krbPwdHistory` produced in debugging mode the output `<= mdb_equality_candidates: (objectClass) not indexed`.

I do not see why shall one assume that an index of (objectClass) does exist.  Moreover, I do not know now, if a search can use a single index and thus an index over objectClass means than an index over krbprincipalname will not be used, or if "index objectClass eq; index krbprincipalname eq" are two indices, while "index objectClass, krbprincipalname eq" is a single index.

About this I filled  https://www.openldap.org/its/index.cgi/Incoming?id=9487;selectid=9487 .

dilyanpalauzov:
You can delete the index on objectClass , turn on debugging and do some queries over krb5kdc to see then that an index over objectClass is missing.

greghudson:
Sure, but I can't tell if that debug message means the search is actually inefficient.

dilyanpalauzov:
My reading of https://www.openldap.org/doc/admin24/tuning.html , Section 21.3.2. “What to watch out for”, is that the output “mdb_equality_candidates” means inefficient searches.

greghudson:
That advice may be targeted at the simple case, where the search filter is just one equality test, and where an index could significantly reduce the number of candidates to be loaded and checked.  For a krb5 LDAP KDB, the filter expression is `(&(|(objectclass=krbprincipalaux)(objectclass=krbprincipal))(\
krbprincipalname=...))`, so the objectClass equality test is intersected with a krbPrincipalName equality test.

If the KDB is large and isn't shared with a lot of other LDAP data, almost every object in the database will match one of the objectClass equality tests, but only one will match the krbPrincipalName test.  In that scenario, an objectClass index is useless, as it doesn't reduce the number of candidates at all, while the krbPrincipalName index is useful.  In fact, I could easily imagine it slowing down the search, relative to having only a krbPrincipalName index--slapd might have to read in a large list of krbprincipal objects, then intersect that with a one-element list of objects matching the krbPrincipalName.  While that would be much faster than loading and checking every principal object, it would still be O(n) in space and time rather than O(1).

I unfortunately can't find much information about the data model for slapd indexes or how slapd uses indexes for AND and OR expressions, and the code is too complicated for me to easily understand.

greghudson:
I'm still far from an expert on slapd filters, but after a bit more research, it seems like:

* an objectClass index might not provide much value, but it probably wouldn't slow down the search to O(n), because filters will fall back to using a range rather than listing a whole bunch of objects.

* slapd can intersect and union index results for AND and OR filter expressions.  It looks like a get_principal search should still be fast when there is no objectClass index, even though the lack of the index is logged.

dilyanpalauzov:
I see your point on the objectClass index, but cannot say if it corresponds to the reality.

The bug reports was closed and I was told to post to the [openldap-technical@openldap.org](http://www.openldap.org/lists/openldap-technical/) mailing list.  I did so on Friday, but the message has not yet appeared in the web archives.  Maybe it will be approved for distribution on Monday.

I think it will be better to continue the discussion on openldap indices over that mailing list, as neither you nor me understand the openldap internals to conclude some reasonable recommendations for the KRB documentation here.

dilyanpalauzov:
I asked https://www.openldap.org/lists/openldap-technical/201909/msg00053.html and the answer was https://www.openldap.org/lists/openldap-technical/201909/msg00055.html .  Maybe you can reword the question over the mailing list, so that the essence is touched?

greghudson:
I haven't had a lot of luck engaging with the OpenLDAP community in the past.

The research I did was looking at: https://github.com/openldap/openldap/blob/master/servers/slapd/back-mdb/filterindex.c

Specifically:

* At line 144 we see that equality_candidates() is used for equality filters.
* At line 193 we see that list_candidates() is used for AND or OR filters.
* equality_candidates(), if it doesn't find an index, logs the message you saw and returns success with the output ID list set to "all" (via the call to MDB_IDL_ALL at line 731).
* list_candidates(), for AND filters, computes the intersection of the output list from each of the sub-expressions.  For OR filters, it computes the union.

Looking at idl.h and idl.c, ID lists are represented as an array of ID values, and can contain either a range (if the first value is NOID) or a list of IDs.  "all" is represented as a range from 1 to infinity.

So for the get_principal query (which is `(&(|(objectclass=krbprincipalaux)(objectclass=krbprincipal))(krbprincipalname=...))`), if there is a krbPrincipalName index but no objectClass index, the code will compute the union of "all" and "all" for the OR subexpression (the result of which is of course still "all"), and will then intersect that with the krbPrincipalName index result, which will just be the one item matching the krbPrincipalName.

dilyanpalauzov:
Update: http://www.openldap.org/lists/openldap-technical/201909/msg00059.html.

PR #977, opened 2019-09-22
doc/admin/database.rst: typo

dilyanpalauzov:


doc/admin/database.rst
```diff
@@ -761,7 +761,7 @@ in the krbtgt keys is used by default to determine the session key
 types supported by the krbtgt service (see
 :ref:`session_key_selection`).  Because non-MIT Kerberos clients
 sometimes send a limited set of encryption types when making AS
-requests, it can be important to for the krbtgt service to support
+requests, it can be important for the krbtgt service to support
```
frozencemetery: It could also have been corrected to "too", but I'm not sure what the original intent was.  This seems fine as you've fixed it.

dilyanpalauzov: The current form needs correction, I do not mind how it is corrected.

PR #980, opened 2019-09-24
NegoEx implementation

greghudson:
Due to a github permissions issue, I don't appear able to update PR #883, so I'm creating a new branch and PR.  (I'll try to give Luke access to it if I can figure out how.)

PR #883 had two commits but the first (small) commit has already been merged.  The initial state of this PR branch is the second commit rebased against the current master branch.


greghudson:
Update #1: removed the draft, as discussed at https://github.com/krb5/krb5/pull/883#discussion_r250396835 .

Update #2: dependencies had been incorrectly generated, causing the build to fail when the build directory is separate from the source directory.   (Our Travis CI script builds in the source directory, so this is independent of the Travis build failure.)

greghudson:
Update #3: passes tests.  A stray debugging printf was interfering with the RPC tests, and t_spnego.c needed an update for NegoEx appearing in the NegHints response.

greghudson:
Update 4: in g_negoex.c, replaced the old license comment with the 2-clause BSD one (which we switched to in 2012).

greghudson:
Update 5: whitespace fixes and the addition of mode lines.  For this update I limited myself to the stuff flagged by git diff --check and the style checker, and I didn't fix everything.

greghudson:
Update 6: the negoex directory does not need the SPNEGO internal header, so remove that include and regenerate dependencies.

greghudson:
Update 7: use C99 integer types (like uint32_t) instead of Windows types (like ULONG).

greghudson:
@lhoward: I'm working on tests, and am starting with the simplest possible mechanism behavior.  If the initiator mech completes in one step, I'm not seeing the state machine advance from INITIAL to NEGOTIATE, so the second gss_init_sec_context() fails with NEGOEX_INVALID_MESSAGE_TYPE.  Is that an oversight?

(It also looks like this would happen if the initiator doesn't complete in one step, but does produce a checksum key after the first step.)

lhoward:
@greghudson: quite likely, I never tested with a single leg mechanism. Does this happen without optimistic tokens? You can set the environment variable `NEGOEX_NO_OPTIMISTIC_TOKEN` to disable sending optimistic tokens.

lhoward:
If I read the code correctly, the initiator state advances to NEGOTIATE only in `negoex_sm_verify()`. Is that function being called?

greghudson:
The function is being called, but it only advances the initiator state from INITIAL if no checksum is generated.

lhoward:
Investigating.

lhoward:
OK, this is complicated, clearly I hadn't considered this case. Suggestions below are based on code inspection alone, I don't have the time right now to test these unfortunately (apologies for creating more work for you). Does this work?

[made_checksum.txt](https://github.com/krb5/krb5/files/3688769/made_checksum.txt)

lhoward:
BTW, just looking at the code, `negoex_{enter,leave}_call()` and `negoex_in_call_p()` could be useful for the re-entrancy issue we discussed in the original PR.

lhoward:
(But, maybe we need that for SPNEGO instead.)

greghudson:
That seems to fix the immediate problem.

Update 8: rebase to master, apply Luke's patch, and adjust comments in light of the patch.

greghudson:
@lhoward: do you have any knowledge of NegoEx alerts beyond what's visible in the code?  draft-zhu-negoex defines the alert structure and ALERT_TYPE_PULSE/ALERT_VERIFY_NO_KEY, but contains no narrative text about them.  The code contains an `#if 0` block to send a no-key alert if no checksum can be made yet.  There is also code to process alerts and set the VERIFY_NO_KEY context flag, but that flag isn't checked.  As things stand, I think everything about alerts could be ripped out of the code without changing its behavior.

lhoward:
Shall we check with Larry and possibly improve the code, before ripping it out?

greghudson:
I think Larry moved on from Kerberos work years ago, but I'll try dochelp.

lhoward:
I need to have a good read of the draft abc code, to check that we correctly generate and verify checksums if a key is available before the context is complete. I think we probably just generate it once, perhaps we need to generate and verify multiple times...

lhoward:
Thinking over the implementation, it’s going to break I think if a checksum key is available before the last round trip. But we can’t delay the checksum generation on the initiator side until the context is complete as in an even numbered leg mechanism, that would cost us an additional round-trip. It may be that as-is is good enough as long as mechanisms know not to make NegoEx keys available until the last round trip.

greghudson:
> Thinking over the implementation, it’s going to break I think if a checksum key is available before the last round trip.

How will it break?

lhoward:
If `negoex_checksum()` succeeds before the last round trip, then the VERIFY message is sent too early, and doesn't protect subsequent messages (if I read the code right). So we rely on the mechanism to behave correctly here. See my previous comment, too.

greghudson:
Another apparent state machine bug:

* The underlying mech returns GSS_S_COMPLETE on the first gss_init_sec_context(), generates a token, and makes negoex keys available.  The initiator state machine generates a verify message and returns GSS_S_COMPLETE to SPNEGO.

* The acceptor state machine runs up through negoex_sm_query_meta_data(), which advances the state to AUTHENTICATE at the end.  negoex_sm_step() sees the state change and doesn't call negoex_sm_verify(), instead returning GSS_S_CONTINUE_NEEDED to SPNEGO and asking for another token.

* The SPNEGO initiator can't figure out anything to do and produces a SPNEGO token with nothing in it.  (Perhaps it should error out instead.)

* The SPNEGO acceptor sees a token with nothing in it and returns GSS_S_DEFECTIVE_TOKEN (spnego_mech.c:1460).

I feel like terminating the state machine on state change is creating some awkwardness (like how we delay advancing the initiator state from INITIAL until negoex_sm_verify()).  But simply taking that out would create numerous other bugs, I think.

lhoward:
If the acceptor's `gss_query_meta_data()` returned nothing, won't it continue the state machine because cOutMessages is zero?

greghudson:
> If the acceptor's gss_query_meta_data() returned nothing, won't it continue the state machine because cOutMessages is zero?

negoex_sm_accept_nego() produces one MESSAGE_TYPE_ACCEPTOR_NEGO so cOutMessages is 1.

lhoward:
You're right. I guess the right solution is to allow multiple state changes within a single leg. But let me think if I can think of a not too intrusive workaround to the issue you identified, I suppose if that's possible it might be simpler to leave as-is.

In any case, once you've made your changes, I will of course re-test against the Windows NegoEx implementation to ensure we haven't regressed.

lhoward:
Hmm. I think the right fix is probably going to replace 
```c
        if (ctx->State != oldState && cOutMessages)
            break;
```
with a better test of whether we're ready to emit a token. I'm not sure what that test is yet, though. 

For now, how about this patch.

[state.diff.txt](https://github.com/krb5/krb5/files/3695609/state.diff.txt)

lhoward:
Try this

[state2.diff.txt](https://github.com/krb5/krb5/files/3695646/state2.diff.txt)

greghudson:
Update 9: apply state2.diff.txt and rebase.  I can now successfully negotiate with a minimal testing mech.

greghudson:
If I induce a failure in the first gss_init_sec_context() operation for my test mechanism, there is no fallback to other mechanisms.  (The first mechanism will be tried twice, since a failure of the optimistic token doesn't count, but no further mechanisms will be tried.)  This is different from SPNEGO, which (since commit e946963e0f8488a8d47c527cca31d4666c32e063) will fall back if a failure happens early enough.  I guess the negotiable mechanism set can only be pruned by acquire_cred and query_meta_data/exchange_meta_data?  (Or query_mechanism_info, to weed out non-NegoEx mechs initially.)

greghudson:
The trace logging uses a `NEGOEX<` prefix for outgoing messages and a `NEGOEX>` prefix for incoming messages.  Is that intentional?  My intuition would have those reversed.

greghudson:
Update 10: fix a bug in negoex_sm_query_meta_data() where if gss_query_meta_data() on a mech failed, the next mech in the list would be skipped.  Fix a bug in negoex_sm_verify() where if the selected mech never makes NegoEx keys available, the error is lost and a packet is sent with no verify message.

I uncovered a more difficult problem: if gss_query_meta_data() fails for a candidate mech on the first negoex_gss_init_sec_context(), that mech is removed from ctx->AuthMechs.  But negoex_sm_init_nego() has already run at this point (it is the second entry in the initiator state machines), so the mechanism is proposed anyway.  Likewise, in the acceptor, negoex_sm_accept_nego() runs before negoex_sm_exchange_meta_data(), so failure to accept a mech's metadata doesn't affect the mechanisms proposed by the acceptor.

greghudson:
Also, if the first candidate mech completes on the optimistic token, `ctx->Flags & NEGOEX_CTX_FLAG_MECH_COMPLETE` prevents negoex_sm_ap_request() from doing anything on the next initiator step.  So even if negoex_sm_init_nego2() selected a different mech, the initiator won't begin a context with the negotiated mech.

lhoward:
> The trace logging uses a `NEGOEX<` prefix for outgoing messages and a `NEGOEX>` prefix for incoming messages. Is that intentional? My intuition would have those reversed.

It sort of makes sense either way to me. Feel free to reverse it (although it's possible I based it on Windows' debugging, in which case it would make sense to keep it consistent; I'll check this when I do my final round of interop testing, unfortunately booting up my Windows development system is a bit complicated so can't check easily).

greghudson:
Update 11:

* In negoex_ctx.c, I separated the processing of input tokens from the generation of output tokens, to fix some of the flaws described above.  I stored outgoing metadata values in auth_mech records so they can be gathered during output token generation.  I removed the function table for init/accept processing, instead just calling the processing functions explicitly.

* Instead of tracking states, I just tracked what step we are on.  The step number clearly indicates what negotiation steps happen (including whether the ap-req/challenge token is optimistic).

* I switched to tracking mech completion and verification in auth_mech records, so that they automatically reset if the optimistic mech is removed.

* At initiator step 2, we now reorder the mech list according to the ACCEPTOR_NEGO preference order.  draft-zhu-negoex says "If more than one security mechanism is available, unless otherwise specified, the highest one in the acceptor's preference order SHOULD be selected".  I also made the code ignore the challenge token if it doesn't match the first mech entry; this can happen if the acceptor processes and responds to the optimistic token but also sends metadata which causes the optimistic mech to be discarded.

* I changed "selection" of a mech to discard all other mech entries.  We no longer select the optimistic mech in acceptor step 1, because even if the first init_sec_context and accept_sec_context steps succeed, the mech could be discarded by the initiator due to metadata exchange.  But we do select the optimistic mech in initiator step 2 if the acceptor processed the token (as indicated by a challenge or verify message) and the initiator didn't discard the mech.

* In the initiator, I had the code iterate over the mechs for gss_init_sec_context() until one works, deleting mech entries on failure.  This will only happen in step 1, or in step 2 if the optimistic mech isn't used, because we will only have a single mech entry later on.

* At acceptor step 1, we now ignore the optimistic ap-req if it isn't for the acceptor's highest-preference mech, instead of selecting it (or erroring if we don't have an entry for it).  draft-zhu-negoex says "If the target preferred mechanism matches the initiator's preferred mechanism, and when the NEGOEX negotiation protection messages are included along with the mechanism token, no additional round trips are incurred by using the NEGOEX protocol with SPNEGO."  This is also how SPNEGO optimistic tokens work.  Acceptor step 2 still allows any mech to be selected by the initiator.

* We now ignore verify tokens for unknown mechs, instead of erroring out.  At acceptor step 1, this can happen if the optimistic mech isn't supported by the acceptor or was removed in metadata exchange.  At initiator step 2, this can happen if the acceptor liked the optimistic mech but the initiator removed it in metadata exchange.

[There is a screw case here where I think the NegoEx protocol is broken as described.  If the optimistic mech is one-leg or two-leg and makes verification keys available after the first leg, the acceptor could think everything is done and return GSS_S_COMPLETE to its caller after one call to gss_accept_sec_context(), but then the initiator could discard the optimistic mech due to the acceptor's provided metadata and want to continue with a different mech.]

* Challenge messages now appear after metadata messages in output tokens, instead of before.  This is probably unimportant.

lhoward:
Great work, Greg! Is it ready to test against Windows?

greghudson:
> Great work, Greg! Is it ready to test against Windows?

Not yet.  I want to finish the automated tests, rewrite the marshalling, refactor the checksum code, rename a lot of identifiers, and see whether it could make sense to relocate NegoEx as a SPNEGO extension rather than an independent mechanism as discussed previously.

The screw case I discussed above is also pretty worrisome, so I need to think about how to handle it, and maybe ask more questions to dochelp.

greghudson:
I sent some additional questions about verify tokens and optimistic ap-reqs to dochelp.

greghudson:
Update 12: fix two bugs discovered while fleshing out the tests.  `negoex_delete_auth_mech(ctx, &ctx->AuthMechs)` turns out not to work, and the assertion after the call to mech_init() is invalid when the optimistic token is suppressed via $NEGOEX_NO_OPTIMISTIC_TOKEN.

greghudson:
Update 13: more bugfixes, mostly for memory leaks.  negoex_get_session_keys() can be called multiple times on the same auth_mech record, so needs to check if it already has keys.  negoex_free_messages() didn't free the outer container.  negoex_make_token() could allocate zero bytes if the message list is empty; aside from spuriously failing if malloc(0) returns NULL, this would leak memory on typical systems because gss_release_buffer() does nothing if the buffer length is zero.  (Not sure if that's really correct, but changing it is too risky.)  negoex_negotiate_mechs() used the mechglue gss_release_cred() on a non-mechglue cred; the loopback check prevented untoward memory behavior but the negoex cred was leaked.

Also, I failed to preserve the mech_type/ret_flags/time_rec output code in negoex_gss_init_sec_context(), so I added that back in.

greghudson:
Automated tests added.

Update 14: explicit serialization (but not deserialization).  The put16/put32/put64 functions will likely be replaced with k5-buf.h helpers at some point, since we have similar functions elsewhere.

greghudson:
The tests are failing on Travis due to https://bugzilla.redhat.com/show_bug.cgi?id=1398370 .  I'll try to figure out the best option to address that.  They pass on my Ubuntu 18.04 machine since it has the glibc fix.

greghudson:
About the screw case (where the acceptor completes on the optimistic token but the initiator fails that mech's metadata), Microsoft said:
> If after receiving accept-completed, the initiator fails on the token, the security context should abort. This should also be the case if checksum validation in the VERIFY message fails. Once the mechanism is selected, it cannot be changed. This would be a problematic case, for example a man-int-the-middle tampering with the token, or a bug in one of the sides either acceptor or initiator. The initiator did suggest the preferred mech and sent a token at the first place. It ought to be able to process the related tokens.
We can't detect within NegoEx whether SPNEGO saw accept-complete (we could if we were using a private interface between SPNEGO and NegoEx).  But I guess we can select the optimistic mech in the initiator if the acceptor gives back a CHALLENGE or VERIFY message, and then fail the authentication if there is a metadata exchange failure.

I asked whether early verify key availability makes an exchange look like AP_REQUEST/VERIFY, CHALLENGE/VERIFY, AP_REQUEST/VERIFY, CHALLENGE/VERIFY, or just AP_REQUEST/VERIFY, CHALLENGE/VERIFY, AP_REQUEST, CHALLENGE.  The response was "the exchange ends with AP_REQUEST/VERIFY, CHALLENGE/VERIFY."  So I guess we should keep sending verify messages rather than stopping after the first.  We probably don't need to insist on more than one verify message (for the selected mech) since the nego messages are the important ones to protect.

greghudson:
Update 15: made both of the aforementioned changes.

greghudson:
Update 16:

* Marshalling is now explicit in both directions.

* Incoming messages are represented in a form more friendly to the negoex_ctx.c code (no need to retrieve byte vectors when we need to access them, for instance).  Fields we don't need to remember are no longer part of the representation.  The new representation uses a union rather than casts, which allows us to use an array of structures rather than an array of pointers.

* Since we can't necessarily marshal the new representation of messages back to its original byte sequence for checksums, the transcript is now kept as a raw byte buffer.  We remember the offset of VERIFY messages within the input token so we can checksum over the preceding bytes of the token as well as the prior transcript.

* Outgoing messages aren't represented as data structures at all; we just have three functions to marshal outgoing messages into the transcript given the fields of those messages, with a common helper for the message header.  The tracing APIs are modified accordingly, and negoex_util.c takes care of calling the tracing APIs when messages are parsed or emitted.

* We now check that VERIFY messages use a keyed checksum.  There are now completely separate functions for making and verifying checksums, rather than a unified function with a lot of conditionals on the makeChecksum flag.

* The bounds check for byte vectors is no longer subject to integer overflow attacks.

* The alert code isn't updated, and is `#if 0`d out for now, while I wait for an answer from dochelp about whether they are actually part of the protocol.

lhoward:
If you've the calculated the message lengths explicitly, you can remove the NEGOEX_PADDING and NEGOEX_PAD macros.

greghudson:
Update 17 (after a rebase with no changes): removed some unused code, including the padding macros.  Replaced the context `Flags` field with an `initiator` boolean; if it turns out we need the VERIFY_NO_KEY flag, it can be a second boolean.  Removed the C++ guards around gssapiP_negoex.h since it's a private header only used by C code (these are present in gssapiP_spnego.h and that's probably why they showed up here; they aren't needed there either).

greghudson:
Update 18: identifier renames and some argument bin-packing.  No functional changes.

greghudson:
I got an answer from dochelp about alerts.  It sounds like a VERIFY_NO_KEY alert is used in a rare circumstance: when one party cannot verify the other party's checksum because its side of the mech has not produced a verify key.  Under ordinary circumstances this doesn't happen, but it could in the case of correctable errors (things like having a cached Kerberos ticket with the wrong kvno, although in Kerberos).

The confusing bit is that it sounds like endpoints continue to generate verify messages with each mech token (if they can) even without an alert.  So it seems like the alert should only be necessary when correcting the error doesn't result in a new mech token, like this:

  initiator: INITIATOR_NEGO, INITIATOR_META_DATA, AP_REQUEST, VERIFY
  acceptor: ACCEPTOR_NEGO, ACCEPTOR_META_DATA, CHALLENGE, ALERT, VERIFY
  initiator: VERIFY

In this case, the initiator wouldn't know to send the final VERIFY message without the ALERT; it would think it is done.

Assuming I understand correctly, I need to make the following changes:

* Add marshalling code for alerts
* Generate a VERIFY_NO_KEY alert if VERIFY_NO_CHECKSUM finds a verify message for the active mech but doesn't have a verify key.
* If a VERIFY_NO_KEY alert is received, clear the sent_checksum flag on the active mech.
* In negoex_get_session_keys(), query for a new mech->key and mech->verify_key each time (destroying the old one), since it sounds like the keys can change.
* Contrive a test case.

lhoward:
Building the GSS EAP mechanism against your branch (modifying it to set `GSS_C_MA_NEGOEX_ONLY`), it appears to negotiate successfully against itself using NegoEx.

When you are ready, I will try against Windows to check for any regressions.

lhoward:
I do see more VERIFY messages than I would expect, though:

Initiator
```
GSS-EAP: state transition INITIAL->AUTHENTICATE
[6987] 1571900013.822898: NEGOEXTS:             Proposed[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[6987] 1571900013.822900: NEGOEXTS<       INITIATOR_NEGO[00] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 96 MsgLength 112
[6987] 1571900013.822902: NEGOEXTS<           AP_REQUEST[01] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 126
Sending init_sec_context token (size=280)...continue needed...
[6987] 1571900013.822904: NEGOEXTS>        ACCEPTOR_NEGO[02] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 96 MsgLength 112
[6987] 1571900013.822906: NEGOEXTS>            CHALLENGE[03] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 92
[6987] 1571900013.822908: NEGOEXTS:            Available[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[6987] 1571900013.822910: NEGOEXTS<           AP_REQUEST[04] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 105
Sending init_sec_context token (size=118)...continue needed...
[6987] 1571900013.822912: NEGOEXTS>            CHALLENGE[05] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 109
[6987] 1571900013.822914: NEGOEXTS<           AP_REQUEST[06] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[6987] 1571900013.822916: NEGOEXTS>            CHALLENGE[07] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
[6987] 1571900013.822918: NEGOEXTS<           AP_REQUEST[08] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 402
Sending init_sec_context token (size=423)...continue needed...
[6987] 1571900013.822920: NEGOEXTS>            CHALLENGE[09] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 1093
[6987] 1571900013.822922: NEGOEXTS<           AP_REQUEST[10] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[6987] 1571900013.822924: NEGOEXTS>            CHALLENGE[11] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 1093
[6987] 1571900013.822926: NEGOEXTS<           AP_REQUEST[12] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[6987] 1571900013.822928: NEGOEXTS>            CHALLENGE[13] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 624
[6987] 1571900013.822930: NEGOEXTS<           AP_REQUEST[14] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 220
Sending init_sec_context token (size=237)...continue needed...
[6987] 1571900013.822932: NEGOEXTS>            CHALLENGE[15] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 148
[6987] 1571900013.822934: NEGOEXTS<           AP_REQUEST[16] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 154
Sending init_sec_context token (size=171)...continue needed...
[6987] 1571900013.822936: NEGOEXTS>            CHALLENGE[17] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 158
[6987] 1571900013.822938: NEGOEXTS<           AP_REQUEST[18] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 199
Sending init_sec_context token (size=216)...continue needed...
[6987] 1571900013.822940: NEGOEXTS>            CHALLENGE[19] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 170
[6987] 1571900013.822942: NEGOEXTS<           AP_REQUEST[20] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[6987] 1571900013.822944: NEGOEXTS>            CHALLENGE[21] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 91
[6987] 1571900013.822946: NEGOEXTS>               VERIFY[22] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 80 MsgLength 92
GSS-EAP: state transition AUTHENTICATE->INITIATOR_EXTS
GSS-EAP: state transition INITIATOR_EXTS->ACCEPTOR_EXTS
[6987] 1571900013.822949: NEGOEXTS<           AP_REQUEST[23] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 111
[6987] 1571900013.822952: NEGOEXTS<               VERIFY[24] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 80 MsgLength 92
Sending init_sec_context token (size=220)...continue needed...
[6987] 1571900013.822954: NEGOEXTS>            CHALLENGE[25] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 99
[6987] 1571900013.822956: NEGOEXTS>               VERIFY[26] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 80 MsgLength 92
GSS-EAP: state transition ACCEPTOR_EXTS->ESTABLISHED

context flag: GSS_C_MUTUAL_FLAG
context flag: GSS_C_REPLAY_FLAG
context flag: GSS_C_SEQUENCE_FLAG
context flag: GSS_C_CONF_FLAG 
context flag: GSS_C_INTEG_FLAG 
```

Acceptor

```
[6983] 1571900013.898890: NEGOEXTS>       INITIATOR_NEGO[00] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 96 MsgLength 112
[6983] 1571900013.898892: NEGOEXTS>           AP_REQUEST[01] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 126
[6983] 1571900013.898894: NEGOEXTS:             Proposed[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
GSS-EAP: vendor: JANET(UK)
GSS-EAP: state transition INITIAL->AUTHENTICATE
[6983] 1571900013.898897: NEGOEXTS:            Available[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[6983] 1571900013.898899: NEGOEXTS<        ACCEPTOR_NEGO[02] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 96 MsgLength 112
[6983] 1571900013.898901: NEGOEXTS<            CHALLENGE[03] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 92
[6983] 1571900013.898903: NEGOEXTS>           AP_REQUEST[04] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 105
libevent: [warn] kq_init: detected broken kqueue; not using.: Undefined error: 0
[6983] 1571900013.898906: NEGOEXTS<            CHALLENGE[05] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 109
[6983] 1571900013.898908: NEGOEXTS>           AP_REQUEST[06] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
[6983] 1571900013.898910: NEGOEXTS<            CHALLENGE[07] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
[6983] 1571900013.898912: NEGOEXTS>           AP_REQUEST[08] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 402
[6983] 1571900013.898914: NEGOEXTS<            CHALLENGE[09] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 1093
[6983] 1571900013.898916: NEGOEXTS>           AP_REQUEST[10] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
[6983] 1571900013.898918: NEGOEXTS<            CHALLENGE[11] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 1093
[6983] 1571900013.898920: NEGOEXTS>           AP_REQUEST[12] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
[6983] 1571900013.898922: NEGOEXTS<            CHALLENGE[13] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 624
[6983] 1571900013.898924: NEGOEXTS>           AP_REQUEST[14] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 220
[6983] 1571900013.898926: NEGOEXTS<            CHALLENGE[15] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 148
[6983] 1571900013.898928: NEGOEXTS>           AP_REQUEST[16] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 154
[6983] 1571900013.898930: NEGOEXTS<            CHALLENGE[17] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 158
[6983] 1571900013.898932: NEGOEXTS>           AP_REQUEST[18] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 199
[6983] 1571900013.898934: NEGOEXTS<            CHALLENGE[19] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 170
[6983] 1571900013.898936: NEGOEXTS>           AP_REQUEST[20] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 93
GSS-EAP: state transition AUTHENTICATE->INITIATOR_EXTS
[6983] 1571900013.898938: NEGOEXTS<            CHALLENGE[21] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 91
[6983] 1571900013.898941: NEGOEXTS<               VERIFY[22] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 80 MsgLength 92
[6983] 1571900013.898943: NEGOEXTS>           AP_REQUEST[23] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 111
[6983] 1571900013.898945: NEGOEXTS>               VERIFY[24] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 80 MsgLength 92
GSS-EAP: state transition INITIATOR_EXTS->ACCEPTOR_EXTS
GSS-EAP: state transition ACCEPTOR_EXTS->ESTABLISHED
[6983] 1571900013.898948: NEGOEXTS<            CHALLENGE[25] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 64 MsgLength 99
[6983] 1571900013.898951: NEGOEXTS<               VERIFY[26] -- ConvId e305e6ed-c68b-114e-5481-345bcddd937d HdrLength 80 MsgLength 92
context flag: GSS_C_MUTUAL_FLAG
context flag: GSS_C_REPLAY_FLAG
context flag: GSS_C_SEQUENCE_FLAG
context flag: GSS_C_CONF_FLAG 
context flag: GSS_C_INTEG_FLAG 
```

lhoward:
OK, well the proof will be whether it interoperates. I'll test soon.

greghudson:
In that old interop test, the server end was MS?  If so, I may have gotten the wrong answer about multiple VERIFY messages, since no VERIFY accompanies the CHALLENGE[19] from the acceptor.

lhoward:
Yes, the server was MS. Interesting.

lhoward:
Here's another interoperating trace from a different mechanism (BrowserID).

```
SASL/GSS-SPNEGO authentication started
NEGOEXTS:             Proposed[0] -- AuthScheme 53553811-836047f5-bc624bd8-076949f0
NEGOEXTS<       INITIATOR_NEGO[0] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 96 MsgLength 112
NEGOEXTS<           AP_REQUEST[1] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 64 MsgLength 343
NEGOEXTS>        ACCEPTOR_NEGO[2] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 96 MsgLength 112
NEGOEXTS>   ACCEPTOR_META_DATA[3] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 64 MsgLength 2316
NEGOEXTS>            CHALLENGE[4] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 64 MsgLength 267
NEGOEXTS>               VERIFY[5] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 80 MsgLength 92
NEGOEXTS:            Available[0] -- AuthScheme 53553811-836047f5-bc624bd8-076949f0
NEGOEXTS<               VERIFY[6] -- ConvId 9f1fa486-f8afabdb-c84e841f-1b3ec8ff HdrLength 80 MsgLength 92
SASL username: lukeh@padl.com
SASL SSF: 56
SASL data security layer installed.
dn: CN=Luke Howard,CN=Users,DC=de,DC=padl,DC=com
altSecurityIdentities: BrowserID:lukeh@padl.com
altSecurityIdentities: BrowserID:lukeh@lukktone.com
altSecurityIdentities: EAP:lukeh@MIT.DE.PADL.COM
```

lhoward:
Windows 10 client, MIT server (your branch) works. There are three verify messages.

lhoward:
MIT client (your branch), Windows 10 server works. Interestingly though, the verify behavior differs: it looks like Windows doesn't send the last verify message from the acceptor.

**Windows client, MIT server**

```
[6983] 1571968986.952495: NEGOEXTS>       INITIATOR_NEGO[00] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 96 MsgLength 128
[6983] 1571968986.952497: NEGOEXTS:             Proposed[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[6983] 1571968986.952499: NEGOEXTS:             Proposed[01] -- AuthScheme 127dd739-00e5-e011-ac64-cd534650acb9
[6983] 1571968986.952502: NEGOEXTS:            Available[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[6983] 1571968986.952504: NEGOEXTS<        ACCEPTOR_NEGO[01] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 96 MsgLength 112
[6983] 1571968987.210854: NEGOEXTS>           AP_REQUEST[02] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 129
GSS-EAP: vendor: JANET(UK)
GSS-EAP: state transition INITIAL->AUTHENTICATE
[6983] 1571968987.210856: NEGOEXTS<            CHALLENGE[03] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 92
[6983] 1571968987.210858: NEGOEXTS>           AP_REQUEST[04] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 105
libevent: [warn] kq_init: detected broken kqueue; not using.: Undefined error: 0
[6983] 1571968987.210860: NEGOEXTS<            CHALLENGE[05] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 109
[6983] 1571968987.210862: NEGOEXTS>           AP_REQUEST[06] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 93
[6983] 1571968987.210864: NEGOEXTS<            CHALLENGE[07] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 93
[6983] 1571968987.210866: NEGOEXTS>           AP_REQUEST[08] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 172
[6983] 1571968987.210868: NEGOEXTS<            CHALLENGE[09] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 1093
[6983] 1571968987.210870: NEGOEXTS>           AP_REQUEST[10] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 93
[6983] 1571968987.210872: NEGOEXTS<            CHALLENGE[11] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 1093
[6983] 1571968988.39817: NEGOEXTS>           AP_REQUEST[12] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 93
[6983] 1571968988.39819: NEGOEXTS<            CHALLENGE[13] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 272
[6983] 1571968988.39821: NEGOEXTS>           AP_REQUEST[14] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 421
[6983] 1571968988.39823: NEGOEXTS<            CHALLENGE[15] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 156
[6983] 1571968989.451967: NEGOEXTS>           AP_REQUEST[16] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 162
[6983] 1571968989.451969: NEGOEXTS<            CHALLENGE[17] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 166
[6983] 1571968990.151283: NEGOEXTS>           AP_REQUEST[18] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 211
[6983] 1571968990.151285: NEGOEXTS<            CHALLENGE[19] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 182
[6983] 1571968990.151287: NEGOEXTS>           AP_REQUEST[20] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 93
GSS-EAP: state transition AUTHENTICATE->INITIATOR_EXTS
[6983] 1571968990.151289: NEGOEXTS<            CHALLENGE[21] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 91
[6983] 1571968990.151292: NEGOEXTS<               VERIFY[22] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 80 MsgLength 92
[6983] 1571968990.151294: NEGOEXTS>           AP_REQUEST[23] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 111
[6983] 1571968990.151296: NEGOEXTS>               VERIFY[24] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 80 MsgLength 92
GSS-EAP: state transition INITIATOR_EXTS->ACCEPTOR_EXTS
GSS-EAP: state transition ACCEPTOR_EXTS->ESTABLISHED
[6983] 1571968990.151299: NEGOEXTS<            CHALLENGE[25] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 64 MsgLength 99
[6983] 1571968990.151302: NEGOEXTS<               VERIFY[26] -- ConvId 17459f92-7808-1c97-ac5a-fffa7c8cf367 HdrLength 80 MsgLength 92
```

**MIT client, Windows server**

```
GSS-EAP: state transition INITIAL->AUTHENTICATE
[26884] 1571969926.190176: NEGOEXTS:             Proposed[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[26884] 1571969926.190178: NEGOEXTS<       INITIATOR_NEGO[00] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 96 MsgLength 112
[26884] 1571969926.190180: NEGOEXTS<           AP_REQUEST[01] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 141
Sending init_sec_context token (size=296)...continue needed...
[26884] 1571969926.190182: NEGOEXTS>        ACCEPTOR_NEGO[02] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 96 MsgLength 112
[26884] 1571969926.190184: NEGOEXTS>            CHALLENGE[03] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 92
[26884] 1571969926.190186: NEGOEXTS:            Available[00] -- AuthScheme 117dd739-00e5-e011-ac64-cd534650acb9
[26884] 1571969926.190188: NEGOEXTS<           AP_REQUEST[04] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 105
Sending init_sec_context token (size=118)...continue needed...
[26884] 1571969927.248554: NEGOEXTS>            CHALLENGE[05] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 109
[26884] 1571969927.248556: NEGOEXTS<           AP_REQUEST[06] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[26884] 1571969927.248558: NEGOEXTS>            CHALLENGE[07] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 93
[26884] 1571969927.248560: NEGOEXTS<           AP_REQUEST[08] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 402
Sending init_sec_context token (size=423)...continue needed...
[26884] 1571969927.248562: NEGOEXTS>            CHALLENGE[09] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 1093
[26884] 1571969927.248564: NEGOEXTS<           AP_REQUEST[10] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[26884] 1571969928.11588: NEGOEXTS>            CHALLENGE[11] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 1093
[26884] 1571969928.11590: NEGOEXTS<           AP_REQUEST[12] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[26884] 1571969928.11592: NEGOEXTS>            CHALLENGE[13] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 624
[26884] 1571969928.11594: NEGOEXTS<           AP_REQUEST[14] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 220
Sending init_sec_context token (size=237)...continue needed...
[26884] 1571969928.11596: NEGOEXTS>            CHALLENGE[15] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 148
[26884] 1571969928.11598: NEGOEXTS<           AP_REQUEST[16] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 154
Sending init_sec_context token (size=171)...continue needed...
[26884] 1571969928.11600: NEGOEXTS>            CHALLENGE[17] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 158
[26884] 1571969928.11602: NEGOEXTS<           AP_REQUEST[18] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 215
Sending init_sec_context token (size=232)...continue needed...
[26884] 1571969929.154034: NEGOEXTS>            CHALLENGE[19] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 186
[26884] 1571969929.154036: NEGOEXTS<           AP_REQUEST[20] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 93
Sending init_sec_context token (size=106)...continue needed...
[26884] 1571969929.154038: NEGOEXTS>            CHALLENGE[21] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 91
[26884] 1571969929.154040: NEGOEXTS>               VERIFY[22] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 80 MsgLength 92
GSS-EAP: state transition AUTHENTICATE->INITIATOR_EXTS
GSS-EAP: state transition INITIATOR_EXTS->ACCEPTOR_EXTS
[26884] 1571969929.154043: NEGOEXTS<           AP_REQUEST[23] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 111
[26884] 1571969929.154046: NEGOEXTS<               VERIFY[24] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 80 MsgLength 92
Sending init_sec_context token (size=220)...continue needed...
[26884] 1571969929.154048: NEGOEXTS>            CHALLENGE[25] -- ConvId 71ad9147-b813-8a2b-05f8-88c7082ba45f HdrLength 64 MsgLength 99
GSS-EAP: state transition ACCEPTOR_EXTS->ESTABLISHED
```

lhoward:
Could this be a bug in Windows, where if the keys are available before the last leg, and a VERIFY token has been sent by a peer, it doesn't send an additional one?

As it happens, it's not a security issue for GSS EAP because the unprotected CHALLENGE is a MIC of the conversation at the mechanism level. But, clearly, it could be for another hypothetical mechanism.

lhoward:
I could modify the GSS EAP mechanism to withhold the keys until the last leg, of course, but... that seems like not something we should have to fix.

lhoward:
In any case, we do appear to have interoperability (no regressions), which is fantastic. Thanks for your work on this Greg!

lhoward:
Also, I agree, the < > symbols in the trace logs are very confusing!

greghudson:
I suspect dochelp just didn't understand my question, and the intent is to send only one VERIFY message (unless there is a VERIFY_NO_KEY alert, or unless the first VERIFY is for the optimistic mechanism and a different mechanism gets negotiated).  I will go back to only sending a checksum if we haven't already sent one unless we get an alert.

In SPNEGO, the MIC exchange only covers the mechlist (to prevent downgrades).  I don't know if NegoEx's transcript checksum is a promise (to integrity-protect the entire exchange) or just a different way to protect against downgrade attacks.  I guess the mechanism can control how much of the exchange is protected by when it supplies a key.

lhoward:
Hmm, I guess we want to ensure the metadata exchange is protected as well? I’m sure you’ve thought about this.

greghudson:
Metadata is always in the first NegoEx token from each side.  As long as the VERIFY messages appear last in their respective tokens, they should protect the metadata.

greghudson:
Did you get a chance to check if Windows debugging resembles your trace log format, and if so whether it has the angle brackets pointing the same way?

lhoward:
I forgot to do this unfortunately (mostly because I had trouble getting any debugging output on the Windows side this time). But I will.

greghudson:
Update 19:

* We're back to sending just one VERIFY messages under normal circumstances.

* Alert marshalling is implemented.  We send a VERIFY_NO_KEY alert if we don't have a key to verify the peer's VERIFY message, and we process such an alert by clearing the keys and sent_checksum flag.

* We re-query the mech for keys on each message in case the keys change.

* Tests updated to match the verify behavior changes, and test cases are added for VERIFY_NO_KEY alert generation and processing.

lhoward:
BTW if you want to change the license to the standard MIT license that's fine by me.

lhoward:
BTW could be nice if `NEGO_MESSAGE_HEADER_LENGTH` were similarly indented to the other constants, in `gssapiP_negoex.h`.

greghudson:
Update 20: create one krb5_context per negoex context, not one per libkrb5 call.  Also align the constant definitions.

greghudson:
There is a change to spnego_mech.c with the comment "add NegoEx mech if cred had it; NegoEx will filter further".  The new code is in get_negotiable_mechs(), and only applies if the caller supplied a claimant_cred_handle and also called gss_set_neg_mechs().  It tests if GSS_NEGOEX_MECHANISM is present in cred_mechs and adds it to the intersection.

Would a credential ever have the NegoEx mech ID in its mechanism list, though?  My understanding is that applications don't interact directly with the NegoEx OID.

lhoward:
I'm not at my desk right now, but I imagine it might be something to do with SPNEGO interacting with NegoEx itself (as NegoEx is invoked through the mechglue). And with wanting to support `gss_set_neg_mechs()` with NegoEx mechs.

greghudson:
Update 21: removed the GSS_S_PROMPTING_NEEDED changes.

greghudson:
I rebased, and then pushed a third commit (to be squashed with the first one at some point) moving the NegoEx code into the SPNEGO implementation.

In this implementation, if a mech supports NegoEx, SPNEGO won't negotiate it directly (there is no NEGOEX_ONLY mech attribute).  I believe that's consistent with Windows, but if that will be a problem for Moonshot let me know.

Other notes:

* Most of the mechglue changes aren't needed, so they are reverted in this commit.  Of course we still need the three new mech entry points.

* I defined the NegoEx minor codes without com_err for consistency with the existing SPNEGO minor codes.  I switched to a table implementation of spnego_gss_display_status() as the switch statement was getting unwieldy.

* The NegoEx mech list is now set up by the SPNEGO code (with a NegoEx helper call per mechanism) in get_negotiable_mechs(), so it's already in place when NegoEx processing begins.  The NegoEx code now just prunes or reorders that list.

* Probably the most awkward part of this integration is making it possible for the many SPNEGO pass-through functions to find the active context in the way the currently do, while still maintaining a context handle per mechanism as we must do for NegoEx (because the metadata APIs can create or modify contexts).  This is handled by moving the active context to ctx->ctx_handle at the end of a NegoEx step, and moving it back at the start of the next step.

* We now only offer or accept NegoEx if we find NegoEx-capable mechanisms in the mech list or credential, so the changes to t_spnego.c are reverted.

* The tighter integration could in principle allow the SPNEGO acceptor to discard an optimistic NegoEx token if the initiator and acceptor don't have overlapping NegoEx mechs, allowing a non-NegoEx SPNEGO mechanism to be negotiated.  Actually implementing this fallback would be tricky with the current SPNEGO code structure, and I don't know if it would help much in practice.

lhoward:
I haven't had a chance to look at the changes yet, but it will create an issue with Moonshot as existing versions export the NegoEx SPIs.

I suppose you could ignore NegoEx mechanisms that don't also export a NEGOEX attribute, but then that would require a flag day if Moonshot is ever to support NegoEx (which would offer it a better Windows interop story).

I am hoping to integrate your changes into Heimdal but it might be easier to take a snapshot of the pre-SPNEGO integration for this.

lhoward:
Actually, I take back my comment about Heimdal integration, looks like the SPNEGO mech changes were very non-intrusive.

lhoward:
The way you've handled `gss_set_neg_mechs()` is very nice.

greghudson:
> I haven't had a chance to look at the changes yet, but it will create an issue with Moonshot

I brought back GSS_C_MA_NEGOEX_ONLY and changed get_negotiable_mechs() to add NegoEx mechs to the SPNEGO list list as well, unless they have that attribute.

(This OID, as well as the two GSS_C_INQ OIDs, will need to be reassigned.  The ones in the PR conflict with OIDs used in the krb5 mech.)

lhoward:
Thanks @ghudson. I realize that allowing negotiation at multiple layers is not ideal, and that NegoEx-only should be the default if the applicable SPIs are exported. But doing this will make things easier for Moonshot (copying @alejandro-perez).

There is no deployed code that cares about the actual OID values.

How close to merging are you? Do you want me to do one final test with Windows before you do?

greghudson:
Update 22: squashed in the SPNEGO integration.

Update 23: modified the negoex_mechs list to use TAILQ macros from k5-queue.h.  (Heimdal has its own copy of those macros in libheimbase.)

I have a dozen or so smaller changes to make, which I should be able to burn through pretty quickly.  I'll let you know when to do a final test.  (And I'm still curious about the Windows debugging output.)

lhoward:
Great. Haven't forgotten about checking the Windows debug output, although I'm more and more certain that it's something I made up myself (and should be changed into something less confusing).

lhoward:
BTW, I noticed a naked call to `realloc()`. I introduced `negoex_alloc()` and `negoex_calloc()` for convenience and portability, although it appears I didn't add a `realloc()` version. I would fix this in the interests of consistency (either ditch the `negoex_XXX()` wrappers or add one for `realloc()`).

lhoward:
Also, only personal preference here, but I prefer checking `if (GSS_ERROR(major))` or `if (major != GSS_S_COMPLETE)` rather than `if (major)`.

greghudson:
I was planning to do a pass over the code to stop using gssalloc_malloc() for internal allocations (everything but the output token buffers, I think).  At that point it would probably make sense to remove the wrapper.

I should be using `if (major != GSS_S_COMPLETE)`; if I introduced any `if (major)`s that was a mistake.  (Nico has expressed objections to using GSS_ERROR() because overusing it can lead to security holes when using gss_unwrap() and similar functions.  That's not really an issue within the NegoEx code, but I've developed a general preference against using that macro.)

lhoward:
Makes sense.

greghudson:
Update 23:

* modified the negoex_mechs list to use TAILQ macros from k5-queue.h.  (Heimdal has its own copy of those macros in libheimbase.  I thought I pushed this yesterday but I hadn't.)

* Removed the allocation wrappers and used malloc/calloc for everything but the output tokens.  I'm not entirely sure how the PR code was working on Windows, as it was using gssalloc_malloc() in buffer_set_to_key() but krb5_c_free_keyblock_contents() to free the keyblocks.  I guess any recent Windows probably uses HeapAlloc(GetProcessHeap(), ...) for malloc/free.

* Converted most `if (GSS_ERROR(major))` and all `if (major)` to `if (major != GSS_S_COMPLETE)`.

greghudson:
Update 24:

* buffer_set_to_key() only worked for small positive enctype values.  I changed it to use generic_gss_oid_decompose().  A mech now must use a specific OID prefix; I used the one Moonshot's EAP mech is using (equal to GSS_KRB5_SESSION_KEY_ENCTYPE_OID).

* negoex_get_session_keys() was allowing minor codes to filter down from gss_inquire_sec_context_by_oid(), but always returned GSS_S_COMPLETE.  I changed it to return errors from buffer_set_to_key() so that it hard-fails if the mech returns garbage, and to use a temporary minor status for gss_inquire_sec_context_by_oid() calls.

* make_checksum() wasn't doing anything with the return code from krb5_c_make_checksum(); I fixed it to check the code.

* I renamed two static functions not to have a negoex_ prefix.

lhoward:
Could there be a more elegant solution than swapping the mechanism context handles back and forth between the SPNEGO and NegoEx contexts (along with using `kctx` as a sentinel to know if the context is initialized for NegoEx)? I can't think of one off-hand, it just feels a bit awkward.

lhoward:
Also: are `negoex_prep_context_for_negoex()` and `negoex_prep_context_for_spnego()` supposed to be paired? If so, then should the code before the former in `negoex_accept()` return immediately on failure, rather than jumping to the cleanup handler.

lhoward:
Instead of swapping context pointers around, how about having a helper function that SPNEGO could call that would return the active mechanism context depending on whether NegoEx was in use or not?

greghudson:
I can basically three solutions:

* Move the context pointers around (what I did).  Inelegant, but requires the least SPNEGO code churn.  Not too dangerous since there's only one pointer aimed at a given context at a time.

* Create an accessor, and change all of the SPNEGO passthrough functions to use it (your suggestion).

* Restructure the SPNEGO code to use also use the mechanism list, with a flag in each mech entry indicating whether it supports NegoEx.  Could lead to some simplification of the SPNEGO code, and could allow us to avoid trying a mechanism both inside and outside NegoEx.  But it would involve major SPNEGO surgery.

spnego_mech.c has a fair amount of technical debt, so I opted for the first option.  Changing all of the passthrough functions wouldn't be too difficult or dangerous, though, so I can see why the accessor approach might be better.

> are negoex_prep_context_for_negoex() and negoex_prep_context_for_spnego() supposed to be paired?

negoex_prep_context_for_spnego() is idempotent, and harmless if no NegoEx processing has been done, so we don't have to worry about jumping to cleanup too early.

lhoward:
I agree that avoiding SPNEGO code churn is a good idea. From memory, getting it correct and interoperable was difficult (although I imagine these days you have tests, right?). The abstraction-liking part of me preferred it being a separate mechanism but, obviously, one has to avoid it being invoked directly (I don't know how Windows does that, but given the API is completely different it's perhaps not too relevant). I suppose pushing and popping the context pointers is no worse than pushing and popping some thread-specific data to indicate we're inside SPNEGO (my original proposal when it was a separate mechanism).

Anyway, I'm rambling now. Do whatever works for you and, when I integrate to Heimdal, I'll figure out what works best for Heimdal. :)

greghudson:
Update 25: switched to the recently-added k5-buf.h integer helpers for marshalling.

greghudson:
Update 26: added a macro to compare auth schemes.  Adjusted some names and whitespace.

greghudson:
Thinking about documentation and OID registration leads me to consider the interface for retrieving verify keys in more detail.

First, draft-zhu-negoex only talks about one verify key, with two key usages.  This would be fine, but the PR code retrieves different keys for sending and receiving checksums, and Moonshot derives different keys for each.  What are the SSPI interfaces for this?

Second, the interface is patterned after gss_krb5int_inq_session_key(), with the krb5 tree taking the opposite role (consuming a keyblock from a mech rather than producing one for a caller).  This interface is pretty miserable, chiefly because it asks the mech to compose an OID from a prefix and the enctype, and stuff the DER encoding of that OID into a buffer.  There are internal functions to do this (generic_gss_oid_compose() and generic_gss_oid_decompose()), but the most convenient public interfaces would be gss_str_to_oid() and gss_oid_to_str().  The OID prefix also isn't declared anywhere.  It would be a lot simpler to ask the application to marshal the enctype into four bytes in a defined by order.

I realize making any changes here would present an issue for Moonshot--recognizing mech_eap as supporting NegoEx but not being able to compatibly retrieve keys could lead to auth failures.

greghudson:
Update 27: reassigned the OIDs.  In buffer_set_to_key(), decode the enctype as a four-byte little-endian integer if it is four bytes long, and otherwise decompose the OID for compatibility with existing mech_eap code.

lhoward:
SSPI allows separate keys.

lhoward:
Do we really need to have two ways to encode the key? I agree not using an OID is probably better but having backwards compatibility code in a new feature seems odd.

greghudson:
Unfortunately, by defining the NegoEx SPIs Moonshot is putting some unilateral compatibility pressure on the API.  I don't know how to get around that without breaking it.

lhoward:
Is the existing SPI so bad? If it is then let’s fix Moonshot.

greghudson:
I took out the compatibility code, under the assumption that the Moonshot dependency can be managed.

lhoward:
I will open a PR for Moonshot. Perhaps as a compromise and a sop to the deployed code, we could keep the compatibility code (perhaps behind an #ifdef?) one major release?

lhoward:
Actually: it's already broken anyway, because mech_eap hard-codes the references to the old OIDs for GSS_C_INQ_NEGOEX_KEY and GSS_C_INQ_NEGOEX_VERIFY_KEY (apologies, I mislead you earlier). So if I've stomped on those OIDs, then we can just fix both things in Moonshot at once.

lhoward:
I'm seeing some unused variable warnings in negoex_ctx.c in the latest version.

lhoward:
cf janetuk/mech_eap#2

lhoward:
Because old Moonshot will fail retrieving keys, we need to make sure SPNEGO will be preferred over NegoEx at least for the Moonshot mechanisms. Otherwise we will break it completely.

lhoward:
Hmm. I hope I haven’t backed myself into a corner here. Arguably if both parties support NegoEx, and a mechanism is advertised under both SPNEGO and NegoEx, we should negotiate it under NegoEx as the negotiation is unambiguously protected and there is the opportunity to exchange metadata which may aid the client in choosing a preferred mechanism.

But, right now, the deployed version of Moonshot is going to fail under NegoEx because it uses a different OID to advertise keys.

I think we may need a way of advertising that the mechanism supports the correct key advertisement OID, perhaps through an attribute that advertises that it supports NegoEx. I need to think about this some more.

greghudson:
Back in September we talked about using the gssspi_ prefix for the new NegoEx entry points, and we were both ambivalent about it (https://github.com/krb5/krb5/pull/883#issuecomment-532965487 is the last message).  If we made that change, deployed Moonshot wouldn't appear as supporting NegoEx and would continue to work via SPNEGO.

lhoward:
That sounds like a good solution, shall we do that?

greghudson:
Update 28: switched to a gssspi_ prefix for the three entry points, and moved the declarations to gssapi_ext.h so mech implementations can see them.

lhoward:
I think this is a good solution.

greghudson:
I'm considering reversing the sense of GSS_C_MA_NEGOEX_ONLY (and changing the name, of course) so that SPNEGO-and-NegoEx mechanisms have to go to extra work, rather than NegoEx-only mechanisms.  Does that seem reasonable?

lhoward:
Yes, I think this is a good idea.

lhoward:
BTW: is NOLOOP required in g_initialize.c, if the symbols aren't exported from libgssapi_krb5? Can't hurt I guess.

greghudson:
I hadn't thought about that.  However, we don't apply export lists to all platforms, so I think we still need to guard against that possibility.

greghudson:
Update 29: GSS_C_MA_NEGOEX_ONLY is now GSS_C_MA_NEGOEX_AND_SPNEGO.  Documentation added to plugindev/gssapi.rst.  secure_getenv() is now used for NEGOEX_NO_OPTIMISTIC_TOKEN.

greghudson:
This should be ready for a final interop test; the remaining changes on my list are all stylistic.

alejandro-perez:
Sorry I've come late for this conversation, but I was really caught offside. If there is any way I can help with this, let me know. I've seen @lhoward has already sent a PR to the mech_eap code.

lhoward:
@greghudson, I can confirm Windows interop has not regressed. I was unable to find any similar NegoEx debug output on Windows so feel free to make cosmetic changes.

greghudson:
> @greghudson, I can confirm Windows interop has not regressed. I was unable to find any similar NegoEx debug output on Windows so feel free to make cosmetic changes.

Thanks.  Right now, for each message we log:

* The direction (perhaps confusingly)
* The message type
* The sequence number
* The conversation ID
* The header length
* The message length

I think it might be nice to log per-message information (and we could eliminate negoex_trace_auth_schemes() if we log the schemes in a NEGO message).  On the other hand, I'm not sure the conversation ID, header length, and message length are interesting to trace.

lhoward:
@alejandro-perez: I made a few changes since the initial pull request but I think it should be fine now. Unfortunately whilst I'd initially hoped that existing deployments of Moonshot would ”just work” with NegoEx once MIT merged this code in, this hasn't been possible. I doubt it's a big issue in practice given how packages are built and deployed.

lhoward:
> Thanks. Right now, for each message we log:
> 
> * The direction (perhaps confusingly)
> * The message type
> * The sequence number
> * The conversation ID
> * The header length
> * The message length
> 
> I think it might be nice to log per-message information (and we could eliminate negoex_trace_auth_schemes() if we log the schemes in a NEGO message). On the other hand, I'm not sure the conversation ID, header length, and message length are interesting to trace.

Yes, probably the lengths aren't that interesting. Perhaps the conversation ID is? The current logging of the direction is very confusing, agreed.

alejandro-perez:
> @alejandro-perez: I made a few changes since the initial pull request but I think it should be fine now. Unfortunately whilst I'd initially hoped that existing deployments of Moonshot would ”just work” with NegoEx once MIT merged this code in, this hasn't been possible. I doubt it's a big issue in practice given how packages are built and deployed.

If the problem is at the protocol level then you might eventually find initiators supporting NegoEx and acceptors not doing so. If the issue is at the level of the API/Interface, I doubt it will be an issue at all, since our binary packages are per-distribution, so we have different packages built specifically for Debian X, CentOS Y, etc.

lhoward:
The protocol is fine as NegoEx itself is negotiated by SPNEGO, and (although this wouldn’t typically be the case for new mechanisms) we allow GSS EAP to be negotiated by both.

greghudson:
> Perhaps the conversation ID is?

The NegoEx conversation ID is weird.  GSS leaves it to the application protocol to correlate requests and responses.  For this reason, you don't find a conversation ID in RFC 2743 framing or SPNEGO or mechs.  But for whatever reason you do in NegoEx, and it's repeated for each NegoEx message within a token, leading to substantial message bloat.  I'm really not sure why it's there, and that leads me to think of it as a low-level marshalling chore and not want to trace it.

To be fair, in a server process handling concurrent client connections, it could be used to distinguish one NegoEx exchange from another in the trace log.  But none of the higher levels (currently untraced, but that could change) have a conversation ID to do that with.

greghudson:
Update 30: style changes to g_negoex.c (untabifying and bin-packing).  I've identified some additional changes I want to make, but don't want to conflate those with a bunch of whitespace changes in the force-push link.

greghudson:
Update 31: in g_negoex,

* Added the various runes and gestures required for interposing mechanisms.

* Removed argument validation since we control the only caller.

* Fixed a leak of internal_name when we allocate it (keep the allocated value in imported_name for safe freeing).

* Added code to error out if internal_cred isn't found in the provided credential.  Most likely this can't happen, but it's not obvious enough to leave open the possibility of continuing on with the default cred.

* Used gssint_create_union_context(), and de-indented that bit of code.  Moved deleting the new context to the cleanup handler.

greghudson:
Update 32: borrowed PR #1002 until that is merged.  Removed the header guards on gssapiP_negoex.h; it can only be included in a specific place from gssapiP_spnego.h so there's no point.  Removed the unsightly include of gssapiP_generic.h from the negoextest mech.

This clears my list of intended code changes.

greghudson:
Update 33: tracing rework.  I put the new tracing routines into negoex_util.c so we don't need a separate negoex_trace.c.  Also, since this work involved me looking more at the guid-to-string conversion function, I wound up renaming AUTH_SCHEME_LENGTH and CONVERSATION_ID_LENGTH to GUID_LENGTH and made a GUID_EQ macro which can be used for both.

greghudson:
I have no more todo list items for negoex; @rharwood, this is ready for a re-review from you.  (This is still rebased on PR #1002, which also needs review.)

src/lib/gssapi/spnego/gssapiP_spnego.h
```diff
@@ -12,7 +12,12 @@
 extern "C" {
 #endif
 
+typedef struct spnego_ctx_st *spnego_gss_ctx_id_t;
```
frozencemetery: Why did this definition move?

src/lib/gssapi/spnego/negoex_util.c
```diff
@@ -0,0 +1,810 @@
...
+    uint64_t protocol_version;
+    uint32_t extension_type;
+    size_t offset, count, i;
+
+    p = k5_input_get_bytes(in, sizeof(msg->random));
+    if (p != NULL)
+        memcpy(msg->random, p, sizeof(msg->random));
+    protocol_version = k5_input_get_uint64_le(in);
+    if (protocol_version != 0) {
+        *minor = ERR_NEGOEX_UNSUPPORTED_CRITICAL_EXTENSION; /* XXX */
```
frozencemetery: Comment should probably be more descriptive.

src/lib/gssapi/spnego/gssapiP_negoex.h
```diff
@@ -0,0 +1,208 @@
...
+    gss_OID oid;
+    auth_scheme scheme;
+    gss_ctx_id_t mech_context;
+    gss_buffer_desc metadata;
+    krb5_keyblock key;
+    krb5_keyblock verify_key;
+    int complete;
+    int sent_checksum;
+    int verified_checksum;
+} *negoex_auth_mech_t;
```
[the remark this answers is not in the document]
greghudson: We don't really have that rule, but I will probably change this to be just `struct negoex_auth_mech` with no typedef.

src/lib/gssapi/spnego/gssapiP_spnego.h
greghudson: The NegoEx function prototypes refer to spnego_gss_ctx_id_t, but the SPNEGO types refer to NegoEx types.  So right now gssapiP_negoex.h has to be included in a particular place, after spnego_gss_ctx_id_t is declared but before the SPNEGO types are declared.  It's possible that switching to `struct negoex_auth_mech` instead of a typedef will allow a less awkward coupling; I will check.

src/lib/gssapi/spnego/negoex_util.c
greghudson: I think Luke might have left that marker as a reminder to allocate a new error.  I can do that.

There are possibly too many different errors for "a protocol constraint was violated".  But a separate error for "the other side used a major version I don't recognize" probably makes sense.

lhoward: Yes, that would've been a note to self to allocate a new error.

src/lib/gssapi/spnego/gssapiP_spnego.h
greghudson: > It's possible that switching to struct negoex_auth_mech instead of a typedef will allow a less awkward coupling; I will check.

The remaining sticking point is "conversation_id negoex_conv_id;" in the SPNEGO context.  I'm not sure there's a good way to finesse that.

(There are of course ways around this.  The NegoEx stuff could be placed in a separate, dynamically allocated sub-context.  Or we could split up the NegoEx type declarations from the NegoEx function prototypes so they're in separate files.  Or we could combine everything into gssapiP_spnego.h.  I'm just not sure if any of those options is better than the current awkwardness where gssapiP_negoex.h has to be included from gssapiP_spnego.h in a specific place.)

frozencemetery: Thank you for clarifying.  Mostly it just wasn't immediately obvious why it was happening.

lhoward:
Whilst integrating with Heimdal, I noticed that we are not passing through channel bindings to NegoEx mechanisms. It's a simple fix.

greghudson:
The pre-existing SPNEGO code doesn't pass channel bindings down to any mechanism (going back to the imported Solaris code).  I'm not sure why, but I chose not to address it in this PR.

lhoward:
Another question from reading through the code with fresh eyes: should `mech_init()` ignore if `gss_init_sec_context()` errors on the first leg? In theory a metadata exchange could make non-optimistic context establishment succeed where optimistic failed.

lhoward:
This diff is for Heimdal, but you get the idea

```diff
[rand:heimdal/lib/gssapi] lukeh% git diff
diff --git a/lib/gssapi/spnego/negoex_ctx.c b/lib/gssapi/spnego/negoex_ctx.c
index 84a786551..c91bfff84 100644
--- a/lib/gssapi/spnego/negoex_ctx.c
+++ b/lib/gssapi/spnego/negoex_ctx.c
@@ -428,7 +428,14 @@ mech_init(OM_uint32 *minor,
         input_token = GSS_C_NO_BUFFER;
     }
 
-    if (HEIM_TAILQ_EMPTY(&ctx->negoex_mechs)) {
+    /*
+     * Ignore errors when generating an optimistic token as acceptor
+     * metadata may make it possible to establish a context later.
+     */
+    if (ctx->negoex_step == 1) {
+       major = GSS_S_COMPLETE;
+       *minor = 0;
+    } else if (HEIM_TAILQ_EMPTY(&ctx->negoex_mechs)) {
         major = first_major;
         *minor = first_minor;
     }
```

lhoward:
But... maybe it's better for the mechanism to return GSS_S_CONTINUE_NEEDED if thinks it may be able to succeed after metadata is exchanged.

greghudson:
That's an interesting question; it depends on the role metadata takes in NegoEx mechanisms.  If it's primarily a way to weed out mechs that won't work (no overlapping trust root, for instance) then the current code is fine.  If they are potentially a necessary first step in some mechanisms' interoperation, then we might need to make a change along those lines.

draft-zhu-negoex says "The optimistic mechanism token may be accompanied by the meta-data tokens and the optimistic mechanism token MUST be that of the first mechanism in the list of the mechanisms proposed by the initiator."  So if we make this change, we must stop at the first step-1 failure, rather than continue on and try to produce an optimistic token for a mech we aren't going to list first.

lhoward:
Leaving as-is seems like the right thing to do. Heimdal is working well with MIT NegoEx; I need to test it against Windows, too.

PR #982, opened 2019-09-27
Log unknown enctypes as unsupported in KDC

greghudson:
[This part is easy, but I may also want to change the behavior on PKINIT enctypes after discussion.]

Commit 8d8e68283b599e680f9fe45eff8af397e827bd6c logs both invalid and
deprecated enctypes as "DEPRECATED:".  An invalid enctype might be too
old or marginal to be supported (like single-DES) or too new to be
recognized.  For clarity, prefix invalid enctypes with "UNSUPPORTED:"
instead.

### DIFF
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1073,20 +1073,22 @@ void limit_string(char *name)
 static krb5_error_code
 enctype_name(krb5_enctype ktype, char *buf, size_t buflen)
 {
-    char *name;
+    const char *name, *prefix = "";
     size_t len;
 
     if (buflen == 0)
         return EINVAL;
     *buf = '\0'; /* ensure these are always valid C-strings */
 
-    if (krb5int_c_deprecated_enctype(ktype)) {
-        len = strlcpy(buf, "DEPRECATED:", buflen);
-        if (len >= buflen)
-            return ENOMEM;
-        buflen -= len;
-        buf += len;
-    }
+    if (!krb5_c_valid_enctype(ktype))
+        prefix = "UNSUPPORTED:";
+    else if (krb5int_c_deprecated_enctype(ktype))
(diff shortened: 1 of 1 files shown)


greghudson:
PKINIT enctypes:

* In practice, I think these are only sent by clients using pkinit-nss (the Red Hat preauth module which was never in the upstream krb5 tree).  Conceivably also by Microsoft clients, as it's not easy for me to check.  MIT krb5 (with its own PKINIT code in 1.6+) and Heimdal don't seem to have any code for sending or recognizing the enctypes.

* The pre-RFC PKINIT drafts had various different language about enctypes.  In draft 33 (the second-to-last draft), the language changed to include "The practice of assigning Kerberos encryption type numbers to indicate support for CMS algorithms is considered deprecated".  I take this remark as aimed at future standardization efforts, not as a deprecation statement for the enctype assignments or the corresponding algorithms.

My preference is to log these with a "PKINIT:" prefix.  That doesn't make any statement about their status but does hint that they aren't actually Kerberos enctypes.

greghudson:
After more reflection, I think logging PKINIT enctypes with an `UNSUPPORTED:` prefix is fine.  So the patch should be okay as-is.

src/kdc/kdc_util.c
```diff
@@ -1073,20 +1073,22 @@ void limit_string(char *name)
 static krb5_error_code
 enctype_name(krb5_enctype ktype, char *buf, size_t buflen)
 {
-    char *name;
+    const char *name, *deprecated = "";
```
frozencemetery: nit: since this can take values other than "DEPRECATED:", I'd prefer a different name - perhaps `status` or `support` or somesuch?

greghudson: Sure, I've changed it to `prefix`.

PR #983, opened 2019-09-30
Aliases support in S4U

iboukris:
Windows clients use aliases extensively, simple aliases like LHS of the UPN, no dollar suffix for machine accounts and a variety of enterprise names are used for TGT client name, as well as SPNs and enterprise names for servers. Specifically in S4U2Self the request server may be a different principal name than header client, and in S4U2Proxy the second ticket server may be a different principal name than header client.

This also helps with integrating in samba, and plugin in its current test-suits, see related wip:
https://gitlab.com/samba-team/samba/merge_requests/818

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2379,6 +2379,12 @@ krb5_get_credentials_for_user(krb5_context context, krb5_flags options,
                               krb5_creds **out_creds);
 
 krb5_error_code KRB5_CALLCONV
+krb5_get_creds_for_user_to_self(krb5_context context, krb5_flags options,
+                                krb5_ccache ccache, krb5_creds *in_creds,
+                                krb5_data *subject_cert, krb5_principal self,
+                                krb5_creds **out_creds);
+
+krb5_error_code KRB5_CALLCONV
 krb5_get_credentials_for_proxy(krb5_context context,
                                krb5_flags options,
                                krb5_ccache ccache,
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -736,6 +736,11 @@ krb5_error_code krb5_db_get_authdata_info(krb5_context context,
 
 void krb5_db_free_authdata_info(krb5_context context, void *ad_info);
 
+krb5_error_code krb5_db_check_alias(krb5_context kcontext,
+                                    const krb5_db_entry *self,
+                                    krb5_const_principal princ,
+                                    krb5_boolean *is_self);
+
 /**
  * Sort an array of @a krb5_key_data keys in descending order by their kvno.
  * Key data order within a kvno is preserved.
@@ -1523,6 +1528,14 @@ typedef struct _kdb_vftabl {
 
     void (*free_authdata_info)(krb5_context context,
                                void *ad_info);
+
+    /*
+     *
+     */
+    krb5_error_code (*check_alias)(krb5_context kcontext,
+                                   const krb5_db_entry *self,
+                                   krb5_const_principal princ,
+                                   krb5_boolean *is_self);
 
     /* End of minor version 0 for major version 8. */
 } kdb_vftabl;
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -599,9 +599,10 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
     /*
      * Note that according to the referrals draft we should
      * always canonicalize enterprise principal names.
+     * -- that didn't make it to RFC 6806 and Windows KDC does not implicitly
+     *    canonicalize enterprise principal names.
      */
-    if (isflagset(state->request->kdc_options, KDC_OPT_CANONICALIZE) ||
-        state->request->client->type == KRB5_NT_ENTERPRISE_PRINCIPAL) {
+    if (isflagset(state->request->kdc_options, KDC_OPT_CANONICALIZE)) {
         setflag(state->c_flags, KRB5_KDB_FLAG_CANONICALIZE);
         setflag(state->c_flags, KRB5_KDB_FLAG_ALIAS_OK);
     }
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1466,6 +1466,19 @@ cleanup:
     return code;
 }
 
+static krb5_error_code
+aliases(krb5_context context, const krb5_db_entry *server,
+        krb5_const_principal name, krb5_boolean *is_self)
+{
+    *is_self = FALSE;
+    if (krb5_principal_compare(context, server->princ, name)) {
+        *is_self = TRUE;
+        return 0;
+    }
+
+    return krb5_db_check_alias(context, server, name, is_self);
+}
+
 /*
  * Protocol transition (S4U2Self)
  */
@@ -1537,12 +1550,6 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
      *
      * (3) The requested service is some other name type: an exact
      *     match is required.
-     *
-     * An alternative would be to look up the server once again with
-     * FLAG_CANONICALIZE | FLAG_CLIENT_REFERRALS_ONLY set, do an exact
-     * match between the returned name and client_princ. However, this
-     * assumes that the client set FLAG_CANONICALIZE when requesting
-     * the TGT and that we have a global name service.
      */
     flags = 0;
     switch (krb5_princ_type(kdc_context, request->server)) {
@@ -1557,12 +1564,19 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
         break;
     }
 
-    if (!krb5_principal_compare_flags(kdc_context,
-                                      request->server,
-                                      client_princ,
-                                      flags)) {
-        *status = "INVALID_S4U2SELF_REQUEST";
-        return KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN; /* match Windows error code */
+    if (!isflagset(c_flags, KRB5_KDB_FLAG_ISSUING_REFERRAL) &&
+        !krb5_principal_compare_flags(kdc_context, request->server,
+                                      client_princ, flags)) {
+        krb5_boolean self;
+
+        code = aliases(kdc_context, server, client_princ, &self);
+        if (code != 0 || !self) {
+            if (code == 0)
+                code = KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN;
+            *status = "INVALID_S4U2SELF_REQUEST";
+            /* match Windows error code */
+            return code;
+        }
     }
 
     /*
@@ -1693,6 +1707,7 @@ check_rbcd_policy(kdc_realm_t *kdc_active_realm, unsigned int flags,
                   krb5_const_principal header_client_princ,
                   void *header_ad_info, const krb5_db_entry *proxy)
 {
+    krb5_error_code code;
     krb5_principal client_princ = stkt_client_princ;
 
     /* Ensure that either the evidence ticket server or the client matches the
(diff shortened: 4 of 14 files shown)


mikelpr:
can we please get this merged? samba as a DC is currently broken otherwise

iboukris:
@mikelpr this is not related, you're probably after this work (we hope to get it in fedora soon): https://gitlab.com/samba-team/samba/merge_requests/829

mikelpr:
Sorry I don't want to add noise, but is this related to not being able to access other computers via compmgmt and printers on them because of access denied 5? because I've had a really hard time trying to set them up unsuccessfully and after reading Dario's and your comment today on the red hat buzgilla I'm thinking it might not be misconfiguration on my part after all and should deploy better alternatives while this is fixed instead of trying more of everything harder

iboukris:
@mikelpr sorry, I don't know if it is related, but anyway, this is work in progress.

metze-samba:
Can we please add a krb5_tkt_creds_set_impersonate() instead?

I think all future work should be done via the  krb5_tkt_creds*() interface.
As it allows applications to use  krb5_tkt_creds_step() and do all network
processing itself. All other APIs should be wrappers on top of it.

I've started with that for samba/heimdal here:
https://git.samba.org/?p=metze/samba/wip.git;a=commitdiff;h=563c2d9f0922e409320ddb3103326a005097d5fc
part of:
https://git.samba.org/?p=metze/samba/wip.git;a=shortlog;h=refs/heads/master3-s4u2self

greghudson:
I initially implemented krb5_tkt_creds* in service of Luke's IAKERB implementation.  I have come to regret accepting IAKERB support, as it created numerous problems at the GSS layer and appeared to have no users.

By extension I have come to regret adding krb5_tkt_creds*.  The TGS request logic has been very complicated ever since referrals support was added, and expressing that logic as a resumable state machine in C made that logic even harder to understand.  S4U adds considerable additional complexity to that logic, and there are requests from Nico and company to add even more (such as trying multiple domain suffixes for short hostnames).

Because of that, recent features have been added at the synchronous layer, such as dns_canonicalize_hostname=fallback (commit 6c20cb1c89acaa03db897182a3b28d5f8f284907), and the recently added RBCD work moved the S4U complexity from the state machine to separate synchronous code.  (Isaac initially implemented RBCD support within the state machine, but I balked at seeing the double-referral-chasing implemented that way.  See https://github.com/krb5/krb5/pull/912#issuecomment-499152044 for that conversation.)

So, to turn that ship around, I would like compelling motivation.  Setting aside IAKERB, what client applications want to do their own network processing for TGS requests?

metze-samba:
Samba will need it, we really need a way to do our own side aware KDC lookups
and may fire parallel tcp connects to multiple KDCs and select the first one that works.

And all that needs to work completely async/non-blocking and the
krb5_{init,tkt}_creds_step() API is perfect for that.

That way we can get rid of our hacks with semi-async send_to_kdc hooks.

For gsskrb5 I'll try to add a ways to force KRB5_GC_CACHED and assume
the caller already filled the ccache with the required tickets.

greghudson:
> Samba will need it, we really need a way to do our own side aware KDC lookups
and may fire parallel tcp connects to multiple KDCs and select the first one that works.

I know Robbie is working on an improved KDC location interface; could that be sufficient for the lookups?  Parallel TCP connections could conceivably be implemented in libkrb5, if that's a must-have.

> For gsskrb5 I'll try to add a ways to force KRB5_GC_CACHED and assume
the caller already filled the ccache with the required tickets.

If the application is already tied to libkrb5, it can move the service ticket into a memory ccache with no accompanying TGT and import that as a cred.

iboukris:
I think it would be unrealistic to move back S4U code to get_creds.c, but maybe we can still support the async mode somehow (both self and proxy) but leave the code apart. I haven't look closely at the code however.

The need for async GSS goes beyond samba, is that something being worked on ? even if samba couldn't use it, it would assure the lower level krb5 api at least supports it. Otherwise even if we disable dns-canon (samba never wants it), and use all the tricks, we could still hit synchronous code at some point i'm afraid.

frozencemetery:
> I know Robbie is working on an improved KDC location interface; could that be sufficient for the lookups? Parallel TCP connections could conceivably be implemented in libkrb5, if that's a must-have.

Yes, not as much as I'd like though.  The main goal of the rework is to enable SSSD's locator plugin to feed back MS-KKDCP endpoints.  However, if you don't need that, the current locator interface should work fine I think.  Descheduling wasn't a goal in the new interface either (i.e., my current design doesn't expose a verto_ctx though it does use callbacks), but we could revisit that if it'd be useful.

However, it sounds like @metze-samba is looking for asnyc *after* DNS - if I understand correctly, the goal is to send requests to all possible KDCs at once and use the first one to respond.  Is that right?

greghudson:
> Descheduling wasn't a goal in the new interface either (i.e., my current design doesn't expose a verto_ctx though it does use callbacks), but we could revisit that if it'd be useful.

I'm looking for ways to meet Samba's needs with a synchronous design for TGS.  So far the stated Samba needs are KDC location and parallel requests, both of which seem possible within the confines of a synchronous krb5_get_credentials() API.

metze-samba:
No, the sync interfaces won't really fit. We also need non-blocking interfaces. Decoupling of the kerberos/crypto parts and the transport parts is ideal for us.

I think it's possible to cleanup some of the _step() related code and make it sane to maintain.
And even easier to understand then some of the current sync code. Once I test/verify
the protocol behavior of the current s4u2self/s4u2proxy code against a complex
Windows AD forest trust setup, I try to propose cleanup patches.

A split up like this for heimdal doesn't really add complexity:
https://git.samba.org/?p=metze/samba/wip.git;a=commitdiff;h=5fdc9a26ff52c5921886fc7c52391dab75b51788
I think having a few shorter functions and a state structure
is easier than one very long function.

greghudson:
> No, the sync interfaces won't really fit. We also need non-blocking interfaces. Decoupling of the kerberos/crypto parts and the transport parts is ideal for us.

This is a big ask, so it needs a strong justification.  Control of the KDC locations and parallel queries don't by themselves demand an async interface.

> A split up like this for heimdal doesn't really add complexity: [...]
I think having a few shorter functions and a state structure
is easier than one very long function.

That appears to be a split of the function to make a single TGS request to a KDC.  Async code in C is indeed very easy for a single request and response.  It becomes much harder to maintain for the full TGS request flow (especially including S4U, which not only adds a double referral-chase for RBCD but adds a series of AS requests to identify the client realm for S4U2Self).  I am always interested in learning about new patterns for async code in C, but the ones I have seen and tried become difficult to understand for complicated processes.

No one is advocating for one very long function.

greghudson:
> I think it would be unrealistic to move back S4U code to get_creds.c, but maybe we can still support the async mode somehow (both self and proxy) but leave the code apart. I haven't look closely at the code however.

I think I can imagine what Isaac is getting at here.  We can know early on whether we are doing S4U2Self, S4U2Proxy, or a regular TGS request.  Therefore, the first krb5_tkt_creds_step() could dispatch to three different state machines, each in its own code file and with its own state structure.  (referenced from krb5_tkt_creds_context, which would still have to be there as a master context to hold the inputs and to keep the API consistent).  That's still more complexity than synchronous S4U2Self implementations and I still have concerns about the maintainability of the resulting code, but it's not nearly as much complexity as a single giant state machine encompassing traditional TGS, S4U2Self, and S4U2Proxy.

We could even potentially decouple the referrals part of the existing state machine from the traditional cross-realm part, with handoffs where necessary.  Unfortunately there's a lot of bleed-through in the current behavior (going back to 1.6), so any decoupling would be likely to change some edge-case scenarios.

Some of the complexity we have to deal with comes from hostname canonicalization in the service principal, which can apply to both regular TGS and S4U2Proxy.  Right now that is limited to dns_canonicalize_hostname=fallback (which currently acts like dns_canonicalize_hostname=false for S4U2Proxy due to the separate implementations); in Heimdal there can be multiple options to try.  This could be managed as an outer state machine, where a regular or S4U2Proxy state machine is instantiated and stepped through for each potential canonicalization.

iboukris:
> Therefore, the first krb5_tkt_creds_step() could dispatch to three different state machines, each in its own code file and with its own state structure.

I'm starting make sense of it, I'll try to develop something like this.

> We could even potentially decouple the referrals part of the existing state machine from the traditional cross-realm part ...

Indeed, I think if it had been decoupled it would been a bit easier to reuse it for cross realm S4U which requires using only referrals (both, self and proxy).

> Some of the complexity we have to deal with comes from hostname canonicalization in the service principal, which can apply to both regular TGS and S4U2Proxy.

Changing the server name in a S4U2Proxy request might cause unexpected policy failure.

greghudson:
>> Some of the complexity we have to deal with comes from hostname canonicalization in the service principal, which can apply to both regular TGS and S4U2Proxy.

> Changing the server name in a S4U2Proxy request might cause unexpected policy failure.

The name candidates are an outer loop around the TGS or S4U2Proxy requests, and apply to the target service name, not the intermediate service in S4U2Proxy.  See MIT's get_creds.c:krb5_get_credentials() starting after the first try_get_creds() call, and Heimdal's krb5_name_canon_iterate().

iboukris:
> The name candidates are an outer loop around the TGS or S4U2Proxy requests, and apply to the target service name, not the intermediate service in S4U2Proxy. 

It's the target service name that must match exactly an entry in the msDS-AllowedToDelegateTo list.

greghudson:
> It's the target service name that must match exactly an entry in the msDS-AllowedToDelegateTo list (in old constrained-delegation).

I'm not quite following the argument.  If we tried an earlier candidate for the canonical name and got KDC_ERR_S_PRINCIPAL_UNKNOWN, surely that wasn't the name on the AllowedToDelegateTo list.

(Obviously in a pure Microsoft or Samba environment, there should be no name canonicalization and it's up to the KDC to canonicalize.  I'm not sure whether we should assume that all S4U2Proxy requests should behave like that.)

iboukris:
> If we tried an earlier candidate for the canonical name and got KDC_ERR_S_PRINCIPAL_UNKNOWN, surely that wasn't the name on the AllowedToDelegateTo list.

Technically the name could been on the list but the principal got removed or renamed. It could still make sense as fallback, just noting that legacy S4U2Proxy is more fragile in that.

iboukris:
Side note: as mentioned I noticed that i missed the handing of in_creds->authdata when moving the code to s4u_creds.c, the old code would unset it upon the first referral, and make a copy to rep_creds in the complete() call, both which i missed, need to look into it.
I also wonder who's in_creds->authdata belong to in S4U call, impersonator or impersonated ?

greghudson:
> I also wonder who's in_creds->authdata belong to in S4U call, impersonator or impersonated ?

When the client asks for authdata during a TGS request, it is in theory trying to place a restriction on the issued ticket, such as KRB5_AUTHDATA_FX_ARMOR restricting the ticket to be used only as FAST armor.  I wouldn't say that such a restriction "belongs to" either the impersonator or the impersonated, but the ticket it applies to is issued from the impersonated client.

iboukris:
Okay thanks for clearing that, so I guess we should only leave the authdata in the initial s4u2proxy request (like it was before), and also on the initial non-s4u2proxy request when chasing referrals as impersonator to get a tgt to proxy realm. (i'll incorporate it in the new changes, unless you prefer a small fix sooner).

greghudson:
> (i'll incorporate it in the new changes, unless you prefer a small fix sooner)

It's a bugfix, so it needs to be in a separate backportable commit.  It can be in the same PR if you want, though.

iboukris:
> It's a bugfix, so it needs to be in a separate backportable commit.

I'll submit a small PR asap, then I hope to test it in some meaningful way but that would have to wait.

iboukris:
I added this work on top of the stepwise PR #1014, closing here.

PR #986, opened 2019-10-11
Allow names to be canonicalized in non-krbtgt AS reply

iboukris:
if we verified returned PA-REQ-ENC-PA-REP, per RFC 6806.

I'm working on a test, see also related work in samba:
https://gitlab.com/samba-team/samba/merge_requests/840

### DIFF
--- a/src/lib/krb5/krb/get_in_tkt.c
+++ b/src/lib/krb5/krb/get_in_tkt.c
@@ -230,17 +230,12 @@ verify_as_reply(krb5_context            context,
     if (canon_req) {
         canon_ok = IS_TGS_PRINC(request->server) &&
             IS_TGS_PRINC(as_reply->enc_part2->server);
-        if (!canon_ok && (request->kdc_options & KDC_OPT_REQUEST_ANONYMOUS)) {
-            canon_ok = krb5_principal_compare_any_realm(context,
-                                                        as_reply->client,
-                                                        krb5_anonymous_principal());
-        }
     } else
         canon_ok = 0;
 
     if ((!canon_ok &&
-         (!krb5_principal_compare(context, as_reply->client, request->client) ||
-          !krb5_principal_compare(context, as_reply->enc_part2->server, request->server)))
+         !krb5_principal_compare(context, as_reply->enc_part2->server, request->server))
+        || (!canon_req && !krb5_principal_compare(context, as_reply->client, request->client))
         || !krb5_principal_compare(context, as_reply->enc_part2->server, as_reply->ticket->server)
         || (request->nonce != as_reply->enc_part2->nonce)
         /* XXX check for extraneous flags */
--- a/src/tests/t_kdb.py
+++ b/src/tests/t_kdb.py
@@ -389,6 +389,9 @@ realm.run([kadminl, 'modprinc', '+requires_preauth', 'canon'])
 realm.kinit('canon', password('canon'))
 realm.kinit('alias', password('canon'), ['-C'])
 
+# Test client name canonicalization in non-krbtgt AS reply
+realm.kinit('alias', password('canon'), ['-C', '-S', 'kadmin/changepw'])
+
 mark('LDAP password history')
 
 # Test password history.


greghudson:
I've looked over the Samba bug report and I have two questions:

1. When do we see the server name change in response to a kadmin/changepw AS request?  Is it just the realm that changes, or does kadmin/changepw change to another name?

2. What is the advantage of setting the canonicalize bit in an AS request for a password change?

The functional change is probably fine; I just want to understand the impact of the bug better.

"proctected" should be "protected".

iboukris:
> When do we see the server name change in response to a kadmin/changepw AS request? Is it just the realm that changes, or does kadmin/changepw change to another name?

It's the client name that we care about here to be canonicalized though windows KDC will canonicalize the server name too if the canon flag is set (that is HTTP/hostname server becomes machinename$).

> What is the advantage of setting the canonicalize bit in an AS request for a password change?

To my understanding this was done as part of a wider general change to always request canonicalization in AS exchange to make sure we always use the canonical name.

iboukris:
I fixed the typo and added a test.

iboukris:
Added a fixup commit, not sure it's needed if all we worry about is that the request was altered, which I'm not entirely sure of.

Also, I'm not sure now that PA-REQ-ENC-PA-REP is relevant here as it does not protect the client name in the reply, which is what we care about here. See quote from RFC 6806 below.
I think the correct fix here might be, to allow the client name to change even if the server is not krbtgt, if only the server name was not changed, would that make sense ?

```
Quote from RFC 6806:

   If only the negotiation mechanism is used, then the request from the
   client to the KDC is protected, but not all of the response is
   protected.  In particular, the client name is not protected; the
   ticket is also not protected.  An attacker can potentially modify
   these fields.  Modification of the client name will result in a
   denial of service.  When the client attempts to authenticate to a
   service (including the TGS), it constructs an AP-REQ message.  This
   message includes a client name that MUST match the client name in the
   ticket according to RFC 4120.  Thus, if the client name is changed,
   the resulting ticket will fail when used.  This is undesirable
   because the authentication is separated from the later failure, which
   may confuse problem determination. 
```

iboukris:
I guess the question is why do we only allow canonicalization for krbtgt service?  If all we worry about is that the request was altered, then I think it should be fine if protected_request is true.

iboukris:
This is the relevant section from the RFC I think, per this we should only reject if the server name has changed and the request was not protected.
```
   Clients SHOULD reject an AS response that changes the server name
   unless the response is protected by such a mechanism or the new
   server name is one explicitly expected by the client.  For example,
   many clients permit the realm name to be changed in an AS response,
   even if the response is not protected.
```

greghudson:
The comment ("We only allow the AS-REP server name to be changed if...") suggests that may have been the intent, so canon_ok applying to client changes may be unintentional.

iboukris:
Right, missed that. I also misread the last quote from the RFC, it allows servers names to change only if the response was protected, not the request. I've updated the branch, I think this should be fine by  itself.

greghudson:
I'm looking at this pre-existing code:
```C
        if (!canon_ok && (request->kdc_options & KDC_OPT_REQUEST_ANONYMOUS)) {
            canon_ok = krb5_principal_compare_any_realm(context,
                                                        as_reply->client,
                                                        krb5_anonymous_principal());
        }
```
to decide if it it is still necessary.  The conditional went in with the main commit to add anonymity support (ec49e6e673ab229462ef18aa2986167eaa643643), and I suspect was motivated by making anonymous kadmin work.

My conclusion is that (1) we don't need it any more, and (2) it was a little confused going in.  Neither type of anonymous request causes the server name to change.  Both types of anonymous request cause the client-principal-with-realm to change--fully anonymous requests cause the realm to change to WELLKNOWN:ANONYMOUS, while partially anonymous requests preserve the realm but change the cname from a regular principal name to WELLKNOWN/ANONYMOUS.

So that conditional never should have been specific to fully anonymous requests, but in practice that didn't matter because we only implement full anonymity on the client.  And the conditional is no longer needed after Isaac's change because canon_ok now only applies to server name changes.

greghudson:
(The force-push also rebases to master; only the change to get_in_tkt.c is new.)

iboukris:
That makes sense :)
Looks good to me.

PR #988, opened 2019-10-17
Improve argument validation in some GSS APIs

greghudson:
The prevailing discpline of public GSS APIs is to set output
parameters to default values, then validate input parameters.  Some
more recent APIs did not do this consistently, leading to the
possibility of minor_status retaining its previous value or similar
issues.

### DIFF
--- a/src/lib/gssapi/generic/gssapi_generic.c
+++ b/src/lib/gssapi/generic/gssapi_generic.c
@@ -413,6 +413,8 @@ generic_gss_display_mech_attr(
 {
     size_t i;
 
+    if (minor_status != NULL)
+        *minor_status = 0;
     if (name != GSS_C_NO_BUFFER) {
         name->length = 0;
         name->value = NULL;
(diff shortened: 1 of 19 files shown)


greghudson:
When there are multiple mandatory output parameters, we have to do this dance where we conditionally assign the output parameters, and then validate that they were passed.  Otherwise we might return an error but not set one of the output parameters.

Elsewhere in our APIs we would just crash out if a mandatory output parameter isn't passed, which simplifies things greatly (we can just assign default values to all of the outputs immediately).  Heimdal appears to do that for GSS as well, but we don't.  Since it's always a bit dangerous to make code less forgiving in version upgrades, I'm making the GSS APIs consistent rather than making them simpler but less forgiving.

greghudson:
That explanation probably wasn't totally clear, since one of the commits PR #973 also had to do with GSS.  The difference is that there we were talking about the krb5 mech, not the mechglue.  Because the mechglue does all of this argument validation and default-setting, the mech implementations can  assume that mandatory output parameters are always non-null.

frozencemetery:
Can you help me understand - is this a change in preference from https://github.com/krb5/krb5/pull/973#issuecomment-527565865 or is there a subtlety I'm missing?

PR #990, opened 2019-10-21
Fix gssalloc_realloc() on Windows

greghudson:
gss_inquire_sec_context_by_oid(GSS_C_INQ_SSPI_SESSION_KEY) fails on
Windows because generic_gss_add_buffer_set_member() relies on the
ability to realloc() a null pointer.  Unlike realloc(), HeapReAlloc()
requires an input pointer that (from the MSDN documentation) "is
returned by an earlier call to the HeapAlloc or HeapReAlloc function".
So gssalloc_realloc() must test for null inputs and call HeapAlloc()
instead.

Reported by Eric Pauly.


frozencemetery:
The change seems fine, but Travis seems upset - problem with the RADIUS daemon it looks like?

PR #993, opened 2019-10-23
kdc: Record the kdc_time and authtime earlier

cryptomilk:
If you want to know how long an authication takes for auditing, we
should note the time earlier as looking up the principal is part of it.

### DIFF
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -546,6 +546,11 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
     /* Seed the audit trail with the request ID and basic information. */
     kau_as_req(kdc_context, TRUE, au_state);
 
+    errcode = krb5_timeofday(kdc_context, &state->kdc_time);
+    if (errcode)
+        goto errout;
+    state->authtime = state->kdc_time;
+
     if (fetch_asn1_field((unsigned char *) req_pkt->data,
                          1, 4, &encoded_req_body) != 0) {
         errcode = ASN1_BAD_ID;
@@ -670,10 +675,6 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
     state->rock.local_tgt = state->local_tgt;
 
     au_state->stage = VALIDATE_POL;
-
-    if ((errcode = krb5_timeofday(kdc_context, &state->kdc_time)))
-        goto errout;
-    state->authtime = state->kdc_time; /* for audit_as_request() */
 
     if ((errcode = validate_as_request(kdc_active_realm,
                                        state->request, *state->client,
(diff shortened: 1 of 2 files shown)


src/kdc/do_as_req.c
```diff
@@ -546,6 +546,10 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
     /* Seed the audit trail with the request ID and basic information. */
     kau_as_req(kdc_context, TRUE, au_state);
 
+    if ((errcode = krb5_timeofday(kdc_context, &state->kdc_time)))
```
frozencemetery: Nit: put the assignment on its own line outside the conditional.  (I recognize you're just copying it from below.)

cryptomilk:
We have support for authentication logging for auditing purposes in Samba. We calculate how long it takes based on that timestamp which gets passed down to KDB in `audit_as_req` as `authtime`.

My current WIP branch is:

https://git.samba.org/?p=asn/samba.git;a=shortlog;h=refs/heads/master-mit-kdc-auditlog

greghudson:
Although the change is an improvement, I think I need to do a little digging.  It's not clear why we have separate kdc_time and authtime fields in struct as_req_state, and the comment mentions an audit_as_request() function which does not exist (there's a prototype for it in kdc_util.h but no calls or implementation).

greghudson:
The separation of kdc_time and authtime goes back to 1.0, where they are local variables in process_as_req().  I think there was a lot of copy-paste between the AS and TGS processing code when those functions were created, and in the TGS code the authtime is different (it comes from the header ticket).

The confusing comment about audit_as_request() comes from commit bb058d2c252ce8f2ba9f92c6beb283a701485bd9 on Luke's mskrb-integ branch back in 2008.  The assignment of authtime was moved up from a later piece of code (without actually removing the later assignment, so it is pointlessly assigned twice) so that the cleanup handler could call `audit_as_request(request, &client, &server, authtime, errcode);` .  However, audit_as_request() was removed in commit 671ceced3b5fec90cd2ad894a083d1b8b3d89997 (for reasons obscured by an accidental use of the wrong svn commit option), leaving behind the prototype and that cryptic comment.

greghudson:
Not really an objection, but: kdc_time has a resolution of 1s; can you really measure anything interesting about the processing time of an AS request with that?

cryptomilk:
I started to implement the feature in the KDB module and as the authtime is not documented (yes, the doc for audit_as_req needs some love) I've digged into the code and wondered if it is the right time stamp I'm looking for.

I didn't look at the time resolution yet. I will compare it with heimdal and we could do a follow up patch if we need a more accurate timestamp. Step by step, feature by feature :-)

PR #995, opened 2019-11-06
Implement enforce_ok_as_delegate libdefaults boolean

vdukhovni:
If this flag is set to true, then GSSAPI credential delegation will be
disabled when the "ok-as-delegate" flag is not set in the service ticket.


greghudson:
Looks good at first glance; I will likely make minor revisions before merging.

PR #999, opened 2019-11-12
Update test suite default message digest to sha256

frozencemetery:
Newer openssl will fail out the sha1 certificates with errors like
"ssl.SSLError: [SSL: CA_MD_TOO_WEAK] ca md too weak (_ssl.c:3833)" or
similar.  Also update the certs in question.


frozencemetery:
(Problem can be observed on Debian testing, for instance.)

kaduk:
FWIW I think this is a Debian-openssl thing, not an upstream-openssl one, at least for now.  Debian increases the default SECLEVEL with a global config file.

frozencemetery:
Thanks, tweaked commit message slightly to give that detail.

PR #1004, opened 2019-11-20
Match kpasswd_server definitions in kadmin's test config

frozencemetery:
kpaswd_server is defined in kdc.conf.proto, but not krb5.conf proto.
Based on a patch by Nalin Dahyabhai.

[This is part of some glue code we have downstream to make sure multiple concurrent test runs don't step on each other.  It's a trap waiting to happen, in any case.]

### DIFF
--- a/src/kadmin/testing/proto/krb5.conf.proto
+++ b/src/kadmin/testing/proto/krb5.conf.proto
@@ -9,6 +9,7 @@
 	__REALM__ = {
 		kdc = __KDCHOST__:1750
 		admin_server = __KDCHOST__:1751
+		kpasswd_server = __KDCHOST__:1752
 		database_module = foobar_db2_module_blah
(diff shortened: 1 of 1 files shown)


greghudson:
Does anything on the client side actually look up the kpasswd server in these tests?

frozencemetery:
Some testing suggests that no, nothing does.  Would you prefer a different approach?

greghudson:
What's the reason for the change?  If client code were added to contact kpasswd in those tests, it would fail (because it would try to contact the admin server on port 464), so I don't see a trap.  Concurrent test runs don't seem like a use case since nothing is looking up this relation.

frozencemetery:
Mismatch between client and server configurations.  It sounds like you'd prefer to just leave this, so I'll drop it.

PR #1005, opened 2019-11-21
WIP: TODO port GSS_KRB5_CRED_NO_TRANSIT_CHECK_X from Heimdal

metze-samba:
THIS NEEDS TO WAIT FOR: https://github.com/heimdal/heimdal/pull/656

Members of Active Directory domains should just rely on there
KDCs (domain controllers) to do SID-Filtering (and name checking)
on trust boundaries, I have verified this with a modified Samba KDC
and a Windows 2012R2 DC. The Windows DC rejects invalid cross-realm tickets
with KRB5KDC_ERR_POLICY, before generating a new (service or referral)
ticket. So any service ticket is already policy checked by the KDC
even if this does not result in setting the transited_policy_checked in the ticket.

This means an accepting service can tell gss_accept_sec_context()
to skip any transited checking, as the trust topoligy is only
fully known to the KDC anyway.

The detailed background for this can be found in the bug report
and the mailing list:
https://lists.samba.org/archive/samba-technical/2019-September/thread.html#134285
https://lists.samba.org/archive/samba-technical/2019-November/thread.html#134553
http://mailman.mit.edu/pipermail/krbdev/ should also have references.

BUG: https://bugzilla.samba.org/show_bug.cgi?id=12907


metze-samba:
@greghudson can you please approve the workflow?

greghudson:
Done.  I haven't had to do that before; I wonder if it's a new-ish change in github.

metze-samba:
> Done. I haven't had to do that before; I wonder if it's a new-ish change in github.

It told me it had to be done once for first time contributors...

kaduk:
> It told me it had to be done once for first time contributors...

IIRC it needs to be done for each push (including updates to a PR) until a PR is merged in that repo.
Spammers were submitting changes that caused the CI workflow to (e.g.) mine bitcoin, so this type of protection became needed.

greghudson:
Continuing an email conversation here as it's a bit in the weeds for some of the lists it was previously on (especially kitten):

> Ok, should I just use a different oid (I can allocate one from the Samba pool) and submit the changes to MIT without the "wait for heimdal first" tag?

That would be fine.

> I just found the "reject_bad_transit" option that's already implemented for the MIT kdc, but I guess we want an extra option, correct?

How is reject_bad_transit related?

To be clear, I'm not specifically asking for a krb5.conf option.  If the API is necessary and sufficient for your purposes, I am happy to stop there.

> Do we want the new "no_transit_check" option to be used via: krb5_appdefault_boolean()?

I would prefer just a realm option, if we have a krb5.conf option at all.

metze-samba:
@greghudson I used "[libdefaults] acceptor_skip_transit_check = true" instead of something in the [realms] section
because I didn't found a realm that decrypt_ticket() => conf_skip_transit_check() could check against.
And using the default realm from [libdefaults] seems to be a pointless redirection.

I hope that's ok.

It would be great to get this into 1.20!

metze-samba:
> @greghudson I used "[libdefaults] acceptor_skip_transit_check = true" instead of something in the [realms] section because I didn't found a realm that decrypt_ticket() => conf_skip_transit_check() could check against. And using the default realm from [libdefaults] seems to be a pointless redirection.
> 
> I hope that's ok.
> 
> It would be great to get this into 1.20!

On https://github.com/heimdal/heimdal/pull/656 @lhoward commented that he would prefer
a different name than "acceptor_skip_transit_check".

@greghudson any suggestions for a better name?

Something generic like:

[realms]
      SOME.REALM = {
            hosted_by_active_directory = true
      }

Seems a bit too generic and I have no idea how to get 'SOME.REALM' in a reliable
way within decrypt_ticket() => conf_skip_transit_check().

greghudson:
> I didn't found a realm that decrypt_ticket() => conf_skip_transit_check() could check against.

What's wrong with server->realm?

metze-samba:
> > I didn't found a realm that decrypt_ticket() => conf_skip_transit_check() could check against.
> 
> What's wrong with server->realm?

The one from the ticket is unprotected on the wire (for the GSSAPI AP-REQ), which makes me a bit nervous to use,
but I see try_one_entry() replaces req->ticket->server with the principal in the keytab, so I guess
we can use it.

And the if ((*auth_context)->key) { /* User to User authentication */ code path if not reachable by 
kg_accept_krb5(), correct?

> ETA: if you mean specifically for srv_dec_tkt.c, I think we want to pass in the keytab principal, and use that for both this purpose and the transited check (which we currently use ticket->server->realm for).

Should I also patch decrypt_ticket_keyblock()? There we only operate on values from our
ccache (or directly from a TGS-Rep), which means the server principal was checked against
the value from the TGS-Rep->enc_part, can be trusted as well, correct?

So what keyname should I use as $somekeyname in:
```
[realms]
   SOME.REALM = {
      $somekeyname = true
   }
```

Should it be generic like 'hosted_by_active_directory', which could also trigger other
active directory specific behavior changes, or should we better use something specific like:
implicit_transit_policy_check_by_kdcs

@greghudson , @lhoward which name would you prefer?

abartlet:
```
[realms]
SOME.REALM = {
hosted_by_active_directory = true
}
```
would have advantages for allowing some way for clients to assert that a PAC should have been present and so allow the administrator to address the name-aliasing issues and apply the workarounds provided by Samba and Windows AD in
CVE-2020-25721 https://bugzilla.samba.org/show_bug.cgi?id=14835 https://bugzilla.samba.org/show_bug.cgi?id=14557 (which  I've just made public). 

A broader discussion of this point should be elsewhere but I wanted to mention here that a generic name might not be so bad.

metze-samba:
I think I'd prefer 'transit_checked_by_active_directory'

greghudson:
The problem with using a generic name is that changing its behavior can break existing deployments if there isn't a perfect 1:1 correspondence.  For instance, if someone has a different reason to skip the transited check, and we then add requiring service ticket PACs to the behavior, our upgrade would break their setup.

User-to-user has to be within the same realm as far as I know, so there isn't a transited concern there.

> Should I also patch decrypt_ticket_keyblock()?

Ideally yes.  Although that API is only used by kvno, we have an easy opportunity to make it safer.

metze-samba:
> The problem with using a generic name is that changing its behavior can break existing deployments if there isn't a perfect 1:1 correspondence. For instance, if someone has a different reason to skip the transited check, and we then add requiring service ticket PACs to the behavior, our upgrade would break their setup.
> 

is 'transit_checked_by_active_directory' fine for you, as it makes it clear that it's only intended for pure active directory
realms. At the API level I'd keep GSS_KRB5_CRED_SKIP_TRANSIT_CHECK_X as name, but will be used
only by applications, which know what they are doing like samba.

@lhoward is that name also fine for you?

> User-to-user has to be within the same realm as far as I know, so there isn't a transited concern there.

Does it use GSSAPI based AP-Req messages? I guess not.

> > Should I also patch decrypt_ticket_keyblock()?
> 
> Ideally yes. Although that API is only used by kvno (probably), we have an easy opportunity to make it safer.

I can try to adjust it too.

jaltman:
I believe that it is a very bad idea to add a krb5.conf setting to permit disabling local transit policy checks when the KDC has refused to set the ` transited_policy_checked` flag.   Misuse of the setting will be a security disaster.   

If the KDC is not setting ` transited_policy_checked` when it should, that is bug that should be fixed in the KDC.  If the manufacturer of the KDC believes that it cannot set ` transited_policy_checked` because it believes the flag doesn't represent the checks that are being performed, then perhaps a new flag should be allocated for the purpose.

metze-samba:
As written in the Heimdal pull request, I think we could use a verified PAC as indication to skip the transit checking,
and do that by default without the need of any admin interaction at all.

What do you think?

greghudson:
A verified PAC doesn't prove that the KDC was AD, particularly after #1225.  It could be an MIT KDC which was asked not to check.  (By default the MIT KDC will error out if asked not to check transited policy, but that can be configured off.)

abartlet:
That will still be an Samba or FreeIPA KDC isn't it, so why wouldn't it be doing the required checks?

If this is a big worry, could the PAC not be permitted to be generated in situations that are not doing such checking?

greghudson:
> That will still be an Samba or FreeIPA KDC isn't it, so why wouldn't it be doing the required checks?

No, we will be issuing minimal PACs (to replace AD-SIGNEDPATH) even if a normal database is used.

I suppose we could make reject_bad_transit=false imply disable_pac.  Still, it doesn't seem very safe to assume that providing a PAC also promises a transited check.

jaltman:
@abartlet The existence of the PAC given today's servers cannot be relied upon to indicate that transited checks have been performed.   That is what the ticket flag is for.  If the flag is unset, the client must perform the checks.  That is the reality of the world we live in as opposed to the one some wish were present.

Changing behavior of MIT or Heimdal KDCs going forward will not alter the behavior of the KDCs that are deployed today.

lhoward:
I confess I'm not entirely across the whole transited realm checking thing. But what @jaltman says makes sense, so I'm inclined to agree with him. If we just leave this as a credential option with no configuration file knobs, is it useful to Samba? Or is it a question of working with arbitrary non-Windows acceptors?

lhoward:
Sorry, I forgot what repository we were on :) My comment above applies to Heimdal, not MIT, obviously.

abartlet:
My understanding is that this would be quite useful for cases where Samba is deployed, where Samba is not the accepting software.  For example, it would avoid neeting to patch BIND9 for Samba's supported mode with BIND9's DLZ database option, where the GSSAPI acceptor is in BIND. 

So far DNS updates by clients outside our realm is a niche case, but it is a good example of where a non-Samba acceptor would benefit from being able to make the reasonable assertion that PAC-present == AD behaviour and delegate these decisions back to the KDC.

abartlet:
> @abartlet The existence of the PAC given today's servers cannot be relied upon to indicate that transited checks have been performed. That is what the ticket flag is for. If the flag is unset, the client must perform the checks. That is the reality of the world we live in as opposed to the one some wish were present.
> 
> Changing behavior of MIT or Heimdal KDCs going forward will not alter the behavior of the KDCs that are deployed today.

@jaltman I'm confused, which MIT or Heimdal KDCs are issuing a PAC and not implementing AD?  (Samba AD with trusts is explicitly a full-trust situation, we don't implement SID filtering yet and so fully trust our fellow trusted realms, there is no boundary).

lhoward:
I think AD-SIGNTICKET got replaced with a PAC only containing server and KDC signatures irrespective of whether acting as an AD DC. @greghudson will remember better than I.

abartlet:
Yeah, that's what it looks like on the Heimdal side.  I was involved in that discussion, and will note my new reservations on #1225.  But to finish beating this dead horse, we could still use this as an indicator, to determine the (AD/not AD) origin of the ticket, it would just be signed_pac && has_LOGON_INFO.

greghudson:
Using logon_info as an indicator seems more likely to be safe, since there is actual SID information from another realm in there and it must have been vetted somehow (or there are bigger problems than transited).

Apple is also a potential player in this space, in addition to Samba and FreeIPA and Novell.  Does anyone know if they have a KDC implementation that issues PACs?  If we can verify that FreeIPA always does the equivalent of a transited check when issuing a PAC with logon_info and can make an educated guess that Apple does (or doesn't issue signed PACs with logon_info) then I'm tentatively okya with using that as an indicator.

lhoward:
I don't think Apple is doing anything in authorisation data beyond KRB5SignedPath; I believe they may have once but not anymore. I looked at `lib/hdb/hdb-od.c` in Heimdal-597.40.10.

lhoward:
As far as I can tell from looking at the Novell DSfW binaries in OES 2018.1 (4OCT2019), the transited checking is just performed by the MIT KDC front-end (apparently 1.16).

```bash
% strings kdb_xad.so |grep -i transit
Added transited service %s[%d] to delegation chain for proxy %s
Denied transitive trust %s between client realm %s and server realm %s
Failed to check transitivity of trust to realm %s: %s
% strings kdb_xad.so | grep -i realm
Failed to unparse principal name for cross-realm PAC: %s
Failed to decode cross-realm PAC for %s: %s
Rejecting cross-realm PAC for %s issued by non-uplevel KDC
Failed to add resource groups to cross-realm PAC for %s: %s
Failed to encode cross-realm PAC for %s: %s
cross-realm
(&(principal)->realm) != ((void *)0)
Principal realm %s differs from local realm: cannot rename
xad_kdc__make_cross_realm_keys
Denied transitive trust %s between client realm %s and server realm %s
Failed to check transitivity of trust to realm %s: %s
```

I don't have any contacts to reach out to anymore unfortunately.

metze-samba:
> I confess I'm not entirely across the whole transited realm checking thing. But what @jaltman says makes sense, so I'm inclined to agree with him. If we just leave this as a credential option with no configuration file knobs, is it useful to Samba? Or is it a question of working with arbitrary non-Windows acceptors?

Having GSS_KRB5_CRED_SKIP_TRANSIT_CHECK_X only would help Samba and would be much better than nothing,
but it would not help other services hosted with the same machine/service account, like ssh or nfs ganesha.

metze-samba:
> Yeah, that's what it looks like on the Heimdal side. I was involved in that discussion, and will note my new reservations on #1225. But to finish beating this dead horse, we could still use this as an indicator, to determine the (AD/not AD) origin of the ticket, it would just be signed_pac && has_LOGON_INFO.

Yes, when I meant a verified PAC, I was talking about a varified PAC coming from AD (or FreeIPA) containing a LOGON_INFO,
sorry if that wasn't clear.

jaltman:
> Using logon_info as an indicator seems more likely to be safe, since there is actual SID information from another realm in there and it must have been vetted somehow (or there are bigger problems than transited).

Using LOGON_INFO is a more narrowly scoped test to confirm that the source of the PAC is AD or an AD-clone.   However, such as test does not qualm my fears.   The TRANSITED-POLICY-CHECKED flag and the PAC are as distinct as Apples and Elephants.   

[RFC 4120 Section 2.7](https://datatracker.ietf.org/doc/html/rfc4120/#section-2.7)  is very clear about the required behavior.   If the TRANSITED-POLICY-CHECKED flag is unchecked, then the application server is required to verify the contents of the transit field against its own local policy.

Here is what MS-KILE has to say about TRANSITED-POLICY-CHECKED:

> The TRANSITED-POLICY-CHECKED flag ([RFC4120] section 2.7): KILE does not check for transited domains on servers or a KDC. Application servers MUST ignore the TRANSITED-POLICY-CHECKED flag. For details on decoding a cross-realm TGT and crealm filtering see [MS-PAC] section 4.1.2.3.

and [MS-PAC] section 4.1.2.3 says

> 4.1.2.3 crealm Filtering
>
> When decoding a cross-realm TGT, the crealm fields inside the TGT are compared to the expected name of the realm for the interrealm trust. If the names do not match the TGT, they are rejected, subject to other mitigating constraints. [32]
>
> These constraints can include allowing fully trusted domains to supply any crealm name on the basis that it would have validated it prior to passing it along, or any other settings that are established out of band. The full set of constraints is implementation-specific.
>
>  [32] Section 4.1.2.3: The TGT's crealm field is compared against the realm names listed on the TDO (trusted domain object), as specified in [MS-ADTS], corresponding to the cross-realm trust. If there is a mismatch, the TGT is rejected. TDOs marked as within the forest pass all crealm names through. TDOs marked as forest transitive indicate that the server will only accept crealm names if it is a name claimed by the forest on the TDO. If the TDO used for the cross-realm TGT has neither indicator set, the server checks if the fully qualified domain name (FQDN) (2) matches the FQDN (2) of any domain in the server's forest; if so, the TGT is accepted. Finally, if the crealm field matches the FQDN (2) of the TDO, then it is accepted.

MS-KILE makes it clear that AD KDCs do not apply any policy to the contents transited path field.   Instead AD KDCs assume that the KDC from the fully trusted domains have performed the validation checks.   However, checks performed by the fully trusted domain (if any) apply solely to application servers in that realm.   They do not apply at all to any application servers belonging to subsequent realms in the transit path.  

It is important to revisit the problems that led to the creation of the MIT [capaths] section in the first place.   There was a realm Alpha which was trusted by an application service in realm Epsilon.   However, there were multiple possible transit paths involving realms Beta, Gamma and Delta.   Any transit path involving Beta could not be trusted.  

  Alpha -> Beta -> Gamma -> Epsilon  (not trusted)
  Alpha -> Beta -> Delta -> Epsilon (not trusted)
  Alpha -> Gamma -> Delta -> Epsilon (trusted)
  Alpha -> Delta -> Gamma -> Epsilon (trusted)

It does not matter that the administrators of Delta and Gamma trust the administrators of Beta because the administrators of Epsilon do not trust the administrators of Beta.   As a result, blindly trusting the transit path because Delta and Gamma are fully trusted is unsafe.

If the crealm Alpha is trusted then KILE will pass it through.   The presence of a PAC with LOGON_INFO says nothing about the validation of the transited path because KILE KDCs and application servers explicitly do not validate the transit path contents.   This is a serious problem.

My opposition to the disabling of transited path checks remains.    If any KDC is enforcing transited path checks for its realm, then it must set the TRANSITED-POLICY-CHECKED flag.   Application servers that are compliant with RFC4120 must not ignore the TRANSITED-POLICY-CHECKED flag be unset and must perform their own checks.

abartlet:
Doesn't the fact that AD trusts are administrator-controlled to be
transitive (typically within a forest, or between trusted forests, so
within the trust boundary) or non-transitive (typically inter-forest
trusts, eg across the trust boundary) address your concerns?

It seems to me that the application server transit checks are most
important when no information is available that the KDC considered the
risks of transitive trusts.  While my research shows this is often done
poorly, it is on the KDC (Active Directory trust configuration) where
any fixes should be undertaken, not on individual application servers.

The security concern in an AD domain is essentially in regards to
EXTERNAL, inter-forest trusts.  These are the trusts outside ones own
organisation and are not transitive.

lhoward:
Forest trusts aren’t transitive to a third forest, but they are between the domains in the trusted/trusting domains.

If you have trusts between MIT realms A <-> B <-> C trusted by AD realm D, can I add ***@***.*** to altSecurityIdentities on a user in D? Are there any implications for path checking here?

abartlet:
On Mon, 2021-12-13 at 16:00 -0800, Luke Howard wrote:
> Forest trusts aren’t transitive to a third forest, but they are
> between the domains in the trusted/trusting domains.

Agreed, but the forest is the security boundary.

> If you have trusts between MIT realms A <-> B <-> C trusted by AD
> realm D, can I add ***@***.*** to altSecurityIdentities on a user in
> D? Are there any implications for path checking here?

I agree that D could then add a PAC, using altSecurityIdentities to
identify the user.  That trust would presumably be EXTERNAL and so non-
transitive, but regardless it would be for the administrator of D to
set that, based on how much they trust the trust, right?

Andrew

lhoward:
> I agree that D could then add a PAC, using altSecurityIdentities to
> identify the user. That trust would presumably be EXTERNAL and so non-
> transitive, but regardless it would be for the administrator of D to
> set that, based on how much they trust the trust, right?

Ah, OK, didn’t realise MIT trusts were non-transitive.

nicowilliams:
I've left commentary on the Heimdal version of this PR, 656.

nicowilliams:
@jaltman makes a good point that this is a bug that Microsoft should fix rather than the rest of us have to work around.  I'm in agreement with this.

Any workaround for Microsoft's bug should preferably be configuration-only.  For example, mark all AD realms as AD realms in `[realms]` in `krb5.conf` and implement some logic for checking if the KDC did the transit path checking.

On the whole I'm uncomfortable with this approach though.

nicowilliams:
Also, even with a way to mark a realm as an AD realm, it's not enough.  You'd have to be able to list all the realms that may transit it.  This gets ugly fast, but an API for this is ugly too.

We also need a GSS naming attribute for getting at the transit path.  Which brings me to exploring a possible compromise:

a) add a GSS flag / cred option / name attr / whatever that let's the application indicate that it knows how to do transit path checking,
b) add a GSS name attribute for extracting the transit path from the peer name (this should be symmetric).

Though, of course, no apps will _really_ know how to do transit path checking apart from Samba (if that), which, again, concerns me about cargo-culting.

I've seen too many things cargo-culted that create problems.  This one, if cargo-culted, would create a vulnerability.  Seems like a bad ide.

abartlet:
Why do you need to know the realms that transited AD?  Once you know that the ticket came from your local AD (which it has to for the signature to verify) then the rest is AD's problem, it doesn't matter what the transit path is, as those are configured and trusted/verified in AD.  There is no reason for the application server to be double-checking that.

PR #1007, opened 2019-11-22
Propogate default-sha1 behavior in s2k_pbkdf2.c

frozencemetery:
~~The NULL check confuses Coverity because we unconditionally dereference
shortly after.~~

CI tells me that the NULL check is important, so propogate it instead.

### DIFF
--- a/src/lib/crypto/krb/s2k_pbkdf2.c
+++ b/src/lib/crypto/krb/s2k_pbkdf2.c
@@ -168,9 +168,8 @@ pbkdf2_string_to_key(const struct krb5_keytypes *ktp, const krb5_data *string,
     if (err)
         goto cleanup;
 
-    err = krb5int_derive_keyblock(ktp->enc, ktp->hash, tempkey, key, &usage,
+    err = krb5int_derive_keyblock(ktp->enc, hash, tempkey, key, &usage,
                                   deriv_alg);
-
 cleanup:
     if (sandp.data)
         free(sandp.data);


greghudson:
krb5int_derive_keyblock() on ktp->hash isn't an unconditional dereference; the hash is only dereferenced when deriv_alg is DERIVE_SP800_108_HMAC (see krb5int_derive_random()).  If this defect were on Coverity Scan, I would click around to try to figure out why Coverity thinks the call is necessarily a dereference.  But it doesn't seem to be there, or I'm failing to find it.

frozencemetery:
I've asked the folks running it for the full trace, but all the tool gives me right now is:

```
15. krb5-1.17/src/lib/crypto/krb/s2k_pbkdf2.c:162: var_compare_op: Comparing "ktp->hash" to null implies that "ktp->hash" might be null.
18. krb5-1.17/src/lib/crypto/krb/s2k_pbkdf2.c:171: var_deref_model: Passing null pointer "ktp->hash" to "krb5int_derive_keyblock", which dereferences it.
#   169|           goto cleanup;
#   170|   
#   171|->     err = krb5int_derive_keyblock(ktp->enc, ktp->hash, tempkey, key, &usage,
#   172|                                     deriv_alg);
#   173|   
```

greghudson:
To check in: I am waiting to find out if the second commit of PR #1012 squashes that defect.  If it does, we can add an assert to k5_sp800_108_counter_hmac() to squash this one.  If it doesn't, we can add a runtime check (returning KRB5_CRYPTO_INTERNAL).  I don't want to pass a default value, as unintentionally using SHA-1 could create an interop issue.

Unfortunately, Coverity Scan seems to be wedged.  The regularly scheduled build from Tuesday morning is still in queue 48 hours later.  I have turned off the buildbot worker process so it won't try to submit any more builds until I can get one in manually.

greghudson:
Coverity Scan is still wedged.  Do you have the ability to submit a tree to RH's scan that isn't from the master branch?

greghudson:
I got the information I wanted about PR #1012, but now that I look at the code more closely, I am even more confused about what Coverity is thinking.

Coverity says "Passing null pointer "ktp->hash" to "krb5int_derive_keyblock", which dereferences it."  The only thing krb5int_derive_keyblock() does with its hash parameter is pass it down to krb5int_derive_random().  The only thing krb5int_derive_random() does with its hash parameter is possibly pass it down to k5_sp800_108_counter_hmac().  k5_sp800_108_counter_hmac() already has a check for hash == NULL (returning KRB5_CRYPTO_INTERNAL), so there can be no dereference of a null pointer there.  (The same conditional does use hash->hashsize, but short-circuit `||` will prevent that clause from being evaluated if hash is NULL.)

This seems worse than the usual Coverity false positive, and it doesn't show up in Coverity Scan.  I don't think there's anything we can reasonably do about it.

frozencemetery:
Thanks for your patience while I looked into this further.  Our Coverity instance doesn't expose the normal frontend, which makes investigating difficult, and I've been unable to get a hold of the admins.

Further investigation suggests this is my fault (so I'm closing this).  It seems to be due to how I've implemented `openssl_kbdkf_counter_hmac()` - which includes a dereference.  I'll push an update which fixes that, though I think it's impossible in practice.

PR #1009, opened 2019-11-25
Oldnames removed

ohjl:
Removed the dependency on oldnames.lib for Windows build.

### DIFF
--- a/src/clients/kinit/kinit.c
+++ b/src/clients/kinit/kinit.c
@@ -24,6 +24,7 @@
  * or implied warranty.
  */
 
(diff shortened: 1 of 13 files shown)


greghudson:
We commit all changes initially against the master branch, and there have been significant changes (such as the elimination of wshelper) since 1.16.  Can you recreate this against the master branch?

Thanks.

ohjl:
Sure, I will. It’s just that we use 1.16 so I’ve been working off 1.16 branch.


Regards,
Peter Lee

PR #1010, opened 2019-11-30
Qualify short hostnames when not using DNS

greghudson:
[I promised to do this many months ago and got sidetracked.  It should make the `dns_canonicalize_hostname = fallback` configuration significantly more functional, so I pushed to do this before branching for 1.18.  I haven't tested this yet on Windows or MacOS but will make an effort to do so.]

When DNS forward canonicalization is turned off or fails, qualify
single-component hostnames with the first DNS search domain.  Add the
qualify_shortname relation to override this suffix.

For one of the tests we need to disable qualification, which is
accomplished with an empty value.  Adjust k5test.py to correctly emit
empty values when writing profiles.

### DIFF
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -314,6 +314,15 @@ The libdefaults section may contain any of the following relations:
     If this flag is true, initial tickets will be proxiable by
     default, if allowed by the KDC.  The default value is false.
 
+**qualify_shortname**
+    If this string is set, it determines the domain suffix for
+    single-component hostnames when DNS canonicalization is not used
+    (either because **dns_canonicalize_hostname** is false or because
+    forward canonicalization failed).  The default value is the first
+    search domain of the system's DNS configuration.  To disable
+    qualification of shortnames, set this relation to the empty string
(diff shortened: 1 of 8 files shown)


src/lib/krb5/os/dnsglue.h
```diff
@@ -153,4 +153,5 @@ int krb5int_dns_expand(struct krb5int_dns_state *,
 void krb5int_dns_fini(struct krb5int_dns_state *);
 
 #endif /* KRB5_DNS_LOOKUP */
+
```
kaduk: Is this spurious?

greghudson:
I awkwardly tested k5_primary_domain() on MacOS using Travis, and it appears to work:

    ./s2p abcde SVC srv-hst  # with dns_canonicalize_host=false, no qualify_shortname set
    SVC/abcde.macstadium.travisci.net@

greghudson:
The Windows k5_primary_domain() does not work, and I've been unable to figure out how to make it work (I'm having the same issue as described in https://www.experts-exchange.com/questions/23177592/Getting-the-local-machine-domain-name.html ).  I will replace it with `return NULL;` pending useful advice from dochelp.

PR #1011, opened 2019-12-04
Improve type safety of serialization code

greghudson:
The serialization code has many variances from current coding practice, but the knottiest problem is the unnecessary use of a table mapping data types to serialization functions.  We always know statically what data type we want to serialize (except in one function in the test program), so there is no need for this table.  The table has several problems:

* It isn't type-safe.

* The use of void pointers in the generic deserialization signature is a strict aliasing violation (as currently used; calling code could be changed to deserialize to a temporary void * and assign that to the actual pointer).

* The casts involved in these strict aliasing violations generate warnings under clang on macOS, because of the `#pragma pack`s we have in krb5.h on that platform.

* It makes it harder to determine what code is used and what code isn't.

The serialization code could certainly use further cleanup, such as deindentation.  It could probably also be discarded entirely, with an analysis of what exactly we need to save and restore in the krb5 gss_export_sec_context() and gss_import_sec_context().  But this seemed like a reasonable unit of work and should facilitate some of the ongoing CI work.


src/lib/gssapi/krb5/ser_sctx.c
```diff
@@ -760,11 +661,10 @@ kg_ctx_internalize(kcontext, argp, buffer, lenremain)
                 if (initiator_name == NULL) {
                     kret = EINVAL;
                 } else {
-                    kret = krb5_internalize_opaque(kcontext,
-                                                   KV5M_AUTHDATA_CONTEXT,
-                                                   (krb5_pointer *)&initiator_name->ad_context,
-                                                   &bp,
-                                                   &remain);
+                    kret = k5_internalize_authdata_context(kcontext,
+                                                           &initiator_name->
```
frozencemetery: Fixing this seems more invasive than it's worth right now, but wow...

src/lib/krb5/krb/ser_actx.c
```diff
@@ -319,25 +234,22 @@ krb5_auth_context_externalize(krb5_context kcontext, krb5_pointer arg, krb5_octe
...
     krb5_error_code ret;
 
-    ret = krb5_internalize_opaque(ctx, KV5M_KEYBLOCK,
-                                  (krb5_pointer *) &keyblock, bp, sp);
+    ret = k5_internalize_keyblock(&keyblock, bp, sp);
     if (ret != 0)
         return ret;
-    ret = krb5_k_create_key(ctx, keyblock, key);
-    krb5_free_keyblock(ctx, keyblock);
+    ret = krb5_k_create_key(NULL, keyblock, key);
```
frozencemetery: It doesn't cause problems here, but is this something we're okay with doing in general in internal code?

greghudson: Yes.  I have plans to document that contexts are optional in many API functions.  That's not a commitment to never use the context, just to behave reasonably when there isn't one.  In the meantime I think it's fine to pass null internally (and we already do in many places).

PR #1012, opened 2019-12-10
Three code hygiene fixes

greghudson:
Two of these address Coverity defects (one real, one a false positive).  The third addresses warnings from clang under macOS.  (The macOS build is still not warning-clean because of deprecations warnings; macOS has decided that daemon() and the entire libldap API are deprecated.  We might be able to except deprecation warnings from -Werror.)

I still need to verify that the assert() squashes the false-positive defect by manually submitting a build to Coverity scan.  There's a regularly scheduled build in the queue right now so that will take a bit.


greghudson:
Coverity Scan finally unwedged (I would guess that it needed attention from someone who was away for the holidays) and I was able to verify that the assert squashes the false-positive forward-null defect.

PR #1013, opened 2019-12-11
s4u2proxy: only send client authdata in initial request

iboukris:
I think this should suffice for now, skipping the realm identification part as I hope it would be simpler later. I'll try to add a test though. (see #983).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -444,6 +444,7 @@ local.properties
 /src/tests/responder
 /src/tests/s2p
 /src/tests/s4u2proxy
+/src/tests/s4u2self
 /src/tests/unlockiter
 
 /src/tests/asn.1/expected_encode.out
--- a/src/lib/krb5/krb/s4u_creds.c
+++ b/src/lib/krb5/krb/s4u_creds.c
@@ -582,6 +582,10 @@ krb5_get_self_cred_from_kdc(krb5_context context,
         if (code != 0)
             goto cleanup;
 
+        /* The authdata in this referral TGT will be copied into the final
+         * credentials, so we don't need to request it again. */
+        s4u_creds.authdata = NULL;
+
         /* Only include a cert in the initial request to the client realm. */
         s4u_user.user_id.subject_cert = empty_data();
 
@@ -707,6 +711,12 @@ krb5_get_credentials_for_user(krb5_context context, krb5_flags options,
         }
         code = 0;
     }
+
+    /* Note the authdata we asked for in the output creds. */
+    code = krb5_copy_authdata(context, in_creds->authdata,
+                              &(*out_creds)->authdata);
+    if (code)
+        goto cleanup;
 
     if ((options & KRB5_GC_NO_STORE) == 0) {
         code = krb5_cc_store_cred(context, ccache, *out_creds);
@@ -1042,6 +1052,10 @@ k5_get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
             goto cleanup;
         }
 
+        /* The authdata in this referral TGT will be copied into the final
+         * credentials, so we don't need to request it again. */
+        mcreds.authdata = NULL;
+
         /*
          * Make sure the KDC supports S4U and resource-based constrained
          * delegation; otherwise we might have gotten a regular TGT referral
@@ -1111,6 +1125,11 @@ k5_get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
         if (code)
             goto cleanup;
     }
+
+    /* Note the authdata we asked for in the output creds. */
+    code = krb5_copy_authdata(context, in_creds->authdata, &tkt->authdata);
+    if (code)
+        goto cleanup;
 
     if (!(options & KRB5_GC_NO_STORE))
         (void)krb5_cc_store_cred(context, ccache, tkt);
--- a/src/tests/Makefile.in
+++ b/src/tests/Makefile.in
@@ -8,10 +8,10 @@ RUN_DB_TEST = $(RUN_SETUP) KRB5_KDC_PROFILE=kdc.conf KRB5_CONFIG=krb5.conf \
 
 OBJS= adata.o etinfo.o forward.o gcred.o hist.o hooks.o hrealm.o \
 	icinterleave.o icred.o kdbtest.o localauth.o plugorder.o rdreq.o \
-	replay.o responder.o s2p.o s4u2proxy.o unlockiter.o
+	replay.o responder.o s2p.o s4u2self.o s4u2proxy.o unlockiter.o
 EXTRADEPSRCS= adata.c etinfo.c forward.c gcred.c hist.c hooks.c hrealm.c \
 	icinterleave.c icred.c kdbtest.c localauth.c plugorder.c rdreq.c \
-	replay.c responder.c s2p.c s4u2proxy.c unlockiter.c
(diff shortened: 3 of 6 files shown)


iboukris:
Added a test for the regression, and also ported it locally to before rbcd changes and it passed.

src/tests/s4u2proxy.c
```diff
@@ -66,6 +87,12 @@ main(int argc, char **argv)
     krb5_keytab defkt;
     krb5_creds mcred, ev_cred, *new_cred;
     krb5_ticket *ev_ticket;
+    krb5_authdata **req_authdata = NULL;
+
+    if (argc == 4) {
+        req_authdata = make_request_authdata(-2, argv[3]);
```
iboukris: I'm trying to change this to [ad-type ad-contents].

greghudson:
> I think this should suffice for now, skipping the realm identification part as I hope it would be simpler later.

By realm identification part, you mean cross-realm S4U2Self?  It seems like that should just be an adjustment to the referral loop in krb5_get_self_cred_from_kdc().

iboukris:
> By realm identification part, you mean cross-realm S4U2Self? It seems like that should just be an adjustment to the referral loop in krb5_get_self_cred_from_kdc().

No, I only looked at S4U2Proxy but I'll take a look at the self code.
I meant in rbcd when chasing referrals as impersonator to identify target proxy's realm and get a tgt to it. Since we discard the ticket it didn't seem important to send the authdata, but I'll look again.

greghudson:
> I meant in rbcd when chasing referrals as impersonator to identify target proxy's realm and get a tgt to it. Since we discard the ticket it didn't seem important to send the authdata, but I'll look again.

Ah, I think of that as getting the target realm TGT, although it also serves to identify the realm. 
 You are correct, we don't need to (and shouldn't) include the authdata for that part of the process.

iboukris:
Added s4u2self fixes, still untested though.

iboukris:
I tried to add tests for the changes of not resending request authdata in cross-realm S4U referral chasing, but got into more issues related to kdb plugin handling I think. I pushed the partial test as wip commit, but I think we shouldn't wait on it as it may take long.

iboukris:
I got the cross-realm S4U2Self test to work as expected, leaving the RBCD test as a comment for now.

iboukris:
I figured it's our old friend 'only_pac_p()' striking again, added a wip commit that shows all is fine with the authdata handling, i can workaround it in a couple of ways but i'd like to address it, maybe we can make an assumption that if we have a PAC and krb5_db_get_authdata_info() is implemented and returned ok, then we can skip ?
Also added a commit about enterprise name canon handling that I borrowed from #983 as it would be good to get in sooner (we already document it i think, i'll look again).

iboukris:
> maybe we can make an assumption that if we have a PAC and krb5_db_get_authdata_info() is implemented and returned ok, then we can skip ?

Actually I'm starting to think the reason for "if it is the only authorization data element" was to not possibly break it for windows client who may not understand signedpath authdata, so basically we always fine if see a PAC, but if the PAC is not the only ad then we can add and expect signedpath, except we don't add signedpath in cross-realm, so we should just trust the PAC as the patch currently does.

greghudson:
@lhoward: we're looking at the only_pac_p() predicate added in release 1.8, where both verification and addition of ad-signedpath are skipped if the authdata contains a PAC and nothing else.  Do you recall the reasoning behind that check?

(I haven't followed exactly how this is getting in the way of these tests.  I assume that the addition of client-requested authdata causes only_pac_p() to return false, and therefore signedpath handling kicks in, but I'm not sure how that breaks the test.)

iboukris:
> I assume that the addition of client-requested authdata causes only_pac_p() to return false, and therefore signedpath handling kicks in, but I'm not sure how that breaks the test.

Yeah, my understanding is that signedpath can't work in cross ream as it only signs with local tgt.

lhoward:
I think it may have had something to do with Windows (or at least some release of it) getting confused with non-PAC AD types. Or simply have been because it’s redundant as the PAC has a KDC signature. It’s a long time ago, but I’ll check my email archives when I get to work. What does Heimdal do?

iboukris:
I don't recall something like that in heimdal (pac only ad) but i'll look again, the issue at hand however is irrelevant to heimdal as we don't have rbcd yet so s4u2proxy is only local (i'd guess this was the assumption here to).

iboukris:
Yeah, in heimdal we set ad_signedpath for s4u2proxy based either on PAC or on signedpath (the PAC verification is more explicit in heimdal though).

iboukris:
btw, at some point maybe we should design KRB5SignedPathV2, to allow rbcd interoperability with heimdal.

iboukris:
I think the concern about S4U2Proxy, is to make sure the kdc has issued the 2nd ticket and not made by the impersonating service (like kimpersonate).
In cross-realm RBCD we require the authorization-data of the 2nd ticket to be verified by krb5_db_get_authdata_info() in order to get the reply client principal, so it should be fine, but maybe we should make it clearer that if krb5_db_get_authdata_info() return ad_info (or success) it means the authdata were verified, then we can just check that there was a PAC in the authorization-data.

iboukris:
I added a fixup commit the way I suggested above (and removed the kdc enterprise name canon handling commit for now, will submit it separately).

iboukris:
I'm having second thoughts about not assuming KDC_OPT_CANONICALIZE in S4U, maybe we should leave the assumption it in the sync _for_user/_for_proxy calls but not in the stepwise layer (and krb5_get_credentials for proxy), I'll update the PR shortly.

iboukris:
I settled with removing only the S4U2Self commit for now, the cross-realm S4U2Proxy couldn't never work unless the flag was requested so I think it's ok.

iboukris:
> It’s a long time ago, but I’ll check my email archives when I get to work.

@lhoward ping, any chance you got something on that, thanks!

lhoward:
> > It’s a long time ago, but I’ll check my email archives when I get to work.
> 
> @lhoward ping, any chance you got something on that, thanks!

Will follow up via email.

lhoward:
Summary: circa Windows 2000, the Windows LSA would crash if you returned additional authorization data. I recall triggering this, returning a proposed NFSv4 ACE type from XAD. So conservatively we avoided generating `KRB5SignedPath` if the PAC is the *only* element in the authorization data. Things may have changed now so, if Windows 2000 interoperability is no longer a concern, likely you can return both.

iboukris:
Looking at some packet captures, moderns clients are certainly fine with more authorization data elements besides the PAC, I'll test adding KRB5SignedPath to make sure, and revisit the PR, thanks!

iboukris:
I've tested with samba and a windows 2019 as client, and AD-SIGNTICKET (512) has no impact.
BTW, i noticed this ad-type used to conflict with Windows, maybe related:
https://github.com/krb5/krb5-assignments/blob/master/ad-type

iboukris:
Also experimented with KDB_CAP_HANDLE_ENTERPRISE, some tests needed update so it does constitute a change, but not much (I left the ldap module though).

iboukris:
Revised as discussed, removed the cross-realm RBCD tests and the capabilities, but kept the changes in the tests as they harmless.

greghudson:
I believe I'm done revising this, and will merge it soon barring feedback.

iboukris:
Looks good, thanks.

PR #1014, opened 2019-12-16
wip: stepwise S4U interface

iboukris:
Initial draft following discussion in PR #983

### DIFF
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -448,6 +448,11 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
 #define TRACE_TKT_CREDS(c, creds, cache)                            \
     TRACE(c, "Getting credentials {creds} using ccache {ccache}",   \
           creds, cache)
+#define TRACE_TKT_CREDS_FOR_USER_TO_SELF(c, user, self)              \
+    TRACE(c, "Getting credentials {princ} -> {princ} (S4U2S)", user, self)
+#define TRACE_TkT_CRED_VIA_TGT(c, request, reply)              \
+    TRACE(c, "Get cred via TGT {princ} after requesting {princ}",    \
+          reply, request)
 #define TRACE_TKT_CREDS_ADVANCE(c, realm)                               \
     TRACE(c, "Received TGT for {data}; advancing current realm", realm)
 #define TRACE_TKT_CREDS_CACHED_INTERMEDIATE_TGT(c, tgt)                 \
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -734,6 +734,11 @@ krb5_error_code krb5_db_get_authdata_info(krb5_context context,
 
 void krb5_db_free_authdata_info(krb5_context context, void *ad_info);
 
+krb5_error_code krb5_db_check_alias(krb5_context kcontext,
+                                    const krb5_db_entry *self,
+                                    krb5_const_principal princ,
+                                    krb5_boolean *is_self);
+
 /**
  * Sort an array of @a krb5_key_data keys in descending order by their kvno.
  * Key data order within a kvno is preserved.
@@ -1518,6 +1523,14 @@ typedef struct _kdb_vftabl {
 
     void (*free_authdata_info)(krb5_context context,
                                void *ad_info);
+
+    /*
+     *
+     */
+    krb5_error_code (*check_alias)(krb5_context kcontext,
+                                   const krb5_db_entry *self,
+                                   krb5_const_principal princ,
+                                   krb5_boolean *is_self);
 
     /* End of minor version 0 for major version 8. */
 } kdb_vftabl;
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -7659,6 +7659,40 @@ krb5_tkt_creds_get_times(krb5_context context, krb5_tkt_creds_context ctx,
  * @return
  * Kerberos error codes
  */
+
+/**
+ * Set impersonate client.
+ *
+ * @param [in]  context         Library context
+ * @param [in]  impersonate     Impersonate client principal
+ *
+ * @retval
+ * 0 Success
+ * @return
+ * Kerberos error codes
+ */
+
+krb5_error_code KRB5_CALLCONV
+krb5_tkt_creds_set_impersonate(krb5_context context,
+                               krb5_tkt_creds_context ctx,
+                               krb5_principal impersonate);
+
+/**
+ * Set impersonate client certificate.
+ *
+ * @param [in]  context         Library context
+ * @param [in]  cert            Impersonate client certificate
+ *
+ * @retval
+ * 0 Success
+ * @return
+ * Kerberos error codes
+ */
+krb5_error_code KRB5_CALLCONV
+krb5_tkt_creds_set_impersonate_cert(krb5_context context,
+                                    krb5_tkt_creds_context ctx,
+                                    krb5_data *cert);
+
 krb5_error_code KRB5_CALLCONV
 krb5_get_init_creds_keytab(krb5_context context, krb5_creds *creds,
                            krb5_principal client, krb5_keytab arg_keytab,
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1415,6 +1415,19 @@ cleanup:
     return code;
 }
 
+static krb5_error_code
+aliases(krb5_context context, const krb5_db_entry *server,
+        krb5_const_principal name, krb5_boolean *is_self)
+{
+    *is_self = FALSE;
+    if (krb5_principal_compare(context, server->princ, name)) {
+        *is_self = TRUE;
+        return 0;
+    }
+
+    return krb5_db_check_alias(context, server, name, is_self);
+}
+
 /*
  * Protocol transition (S4U2Self)
  */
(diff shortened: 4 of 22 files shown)


iboukris:
I got it working fully stepwised, it is still a real mess but the tests are passing :)

iboukris:
I think the main issue to consider is cache lookup. Currently in S4U2Proxy the krb5_get_credentials() (and thus the stepwise api before rbcd), can't perform any cache lookup since the client name is in encrypted second ticket, so the sync APIs _for_proxy() and GSS caller do the check on their own.

I think it is rather a good thing, as I like the approach of using separate caches in S4U for the impersonator and impersonated client, so I wouldn't expect to find client's ticket in cache.
I'd like to do the same for S4U2Self and let the sync caller do the cache check, especially since in S4U2Self the client realm (and theoretically the name) can get canonicalized so the caller should take note of reply->client (init a new cache) and use that name in future requests to avoid IO.

Currently the sync _for_user() API performs an uncanonicalized cache lookup, then it identifies enterprise client realm and performs a second cache lookup, then once it gets the tickets it performs a third cache lookup on the fully canonicalized client to avoid saving a duplicate in cache.

So if we want to avoid cache lookup in the stepwise layer, we'd have to skip the middle cache lookup since the realm canonicalization will be done in the stepwise layer (recusing on get_init_creds) which will cause a slight overhead if the applications does not take note of reply->client as above (note however that the overhead would only be in case we already had to perform IO for realm canonicalization).

greghudson:
I'll try to break that down:

* Should S4U2Proxy have implicit caching of results, or should we expect that applications will want to cache them separately?  S4U2Proxy results used to not be cached before commit f388eb9673062d8e4eb7d359be2dcaff2acba768; I know I changed that by someone's request and not of my own volition.  So I think some people expect those tickets to be cached.  The S4U2Proxy model is similar to regular TGS requests in that the impersonator cache (TGT+evidence ticket) acts like a TGT that can be used for multiple target servers, so I think caching should ideally behave similarly.

* Should S4U2Self have implicit cache of results?  The only current public interface for S4U2Self is gss_acquire_cred_impersonate_name(), and it looks like caching is not done at that layer-- kg_impersonate_name() passes KRB5_GC_NO_STORE to krb5_get_credentials_for_user().  I could see value in caching S4U2Self results in the service's TGT cache, but each result is a credential for a different client, which doesn't fit the normal usage of ccaches.  Unlike S4U2Proxy, I don't think anyone has asked for implicit caching of S4U2Self results.

* Should the sync and async APIs behave differently for caching?  For regular TGS requests, they behave the same way; krb5_tkt_creds* implicitly caches unless told not to.  Ideally I'd say that the sync and stepwise APIs for S4U2Self and S4U2Proxy should also behave the same way, but I recognize that the issue is historically messy--we provide a nominally public libkrb5 S4U2PROXY API (in_creds->second_ticket and KRB5_GC_CONSTRAINED_DELEGATION, which used to work for either the sync or stepwise interfaces) that can't cache because the caller doesn't provide the client name.

iboukris:
While I had hoped to also not implicitly store the resulting ticket in S4U, that is not so important as the caller can always set NO_STORE.
My question is mainly about looking up the requested creds in the provided ccache, I'd like to avoid it in S4U2Self for the reasons above. While it would be inconsistent with the sync api, it would be similar to the inconsistency we have in S4U2Proxy. Another aspect in which the low level stepwise API differs is in the lack of the abstraction of in_creds->client we do in _for_user and _for_proxy.

greghudson:
The sync libkrb5 API for S4U2Self is not public; we can change its behavior as long as the GSS behavior remains consistent.

iboukris:
Indeed the GSS api with the 'proxy' credentials is fine, but maybe we can simplify the krb5 api a bit.

iboukris:
> The sync libkrb5 API for S4U2Self is not public; we can change its behavior as long as the GSS behavior remains consistent. Similarly, the sync libkrb5 API for S4U2Proxy is not public, so its behavior isn't really a consideration.

So the only public S4U krb5 api is S4U2Proxy via krb5_get_credentials and its stepwise counterpart, both of which do not perform cache lookup and do not use the in_creds->client abstraction, exactly the way I'd like to add S4U2Self stepwise (it can't work via krb5_get_credentials, it needs set_impersonate).

iboukris:
> The S4U2Proxy model is similar to regular TGS requests in that the impersonator cache (TGT+evidence ticket) acts like a TGT that can be used for multiple target servers, so I think caching should ideally behave similarly.

I think that's only true for the GSS layer, where we also handle the cache lookup manually (PR #381), in krb5 it's just a ccache with mixed tickets (matched by client name).

> I could see value in caching S4U2Self results in the service's TGT cache, but each result is a credential for a different client, which doesn't fit the normal usage of ccaches. Unlike S4U2Proxy, I don't think anyone has asked for implicit caching of S4U2Self results.

I think of S4U2Self as kinit, it may canonicalize the client name and we initialize the ccache with it.

iboukris:
I suspect we may need something like commit "kdc_alias_fixup" in master, in legacy constrained-delegation I think we shouldn't canonicalize the proxy server name we pass to krb5_db_check_allowed_to_delegate(), still unclear though.

iboukris:
Oh, that's wrong, second-ticket server is the impersonator, not the proxy target, need to revise the commit.

iboukris:
> These two stages can easily be applied to cross-realm S4U2Self, just s/srealm/icrealm, but not to RBCD as far as I can think of.

BTW, it turned out that S4U2Self wasn't that easy either, we have the following input:
Client: client@CREALM
Server: server@SREALM
Impersonate: impersonate@IREALM

(we assume that client and server are the same principal, not necessarily that CREALM and SREALM are the same).

First we canonicalize impersonate realm, let's leave the how for to later. Now we need to set SREALM = IREALM, which isn't right if server is not an SPN or enterprise server. So it seemed better to only explicitly override the requested realm if it is the referral realm, otherwise like we do in RBCD we'd just assume that SREALM == IREALM and produce a packet for SREALM and then send it to IREALM.

Then I took the same approach with client referrals when initially canonicalizing IREALM, in the krb5 api we'd only canonicalize IREALM if it is the referral realm (regardless of enterprise/canonicalize), otherwise we assume the realm to be canonicalized and we start with getting a TGT to that realm and so on. While this complicates it for the caller (it now sets the referral realm for enterprise clients), I think it makes the krb5 API more flexible and more strict at the time, which fits well.

metze-samba:
If I just want to test the client side, do I need the whole branch or would a just pick this?

commit 193cd868c2bfff4e5b290e96d62a771e6dfbe933
Author:     Isaac Boukris <iboukris@gmail.com>
AuthorDate: Mon Dec 16 01:32:13 2019 +0100
Commit:     Isaac Boukris <iboukris@gmail.com>
CommitDate: Sat Jan 4 18:51:54 2020 +0100

    wip: stepwise S4U interface

iboukris:
I squashed the first two minor commits, just the first commit now: "wip: stepwise S4U interface"

greghudson:
> BTW, it turned out that S4U2Self wasn't that easy either, we have the following input:
Client: client@CREALM
Server: server@SREALM
Impersonate: impersonate@IREALM
> 
> (we assume that client and server are the same principal, not necessarily that CREALM and SREALM are the same realm).

This feels like inventing a new kind of cross-realm S4U2Self, and perhaps inventing the idea that it's meaningful for a principal to exist in two different realms.  Does Microsoft do this ever?

iboukris:
> This feels like inventing a new kind of cross-realm S4U2Self, and perhaps inventing the idea that it's meaningful for a principal to exist in two different realms. Does Microsoft do this ever?

I'm  trying to allow setting different server / client aliases in the S4U2Self request (as windows clients do) and not to ignore the requested realm. For cross-realm, the API caller should set server to a name that is valid for referrals (spn or enterprise) and the referral realm, so we can safely overwrite it. Maybe we can think of server->realm it as start-realm, or just require it to be referral realm for cross-realm request.

iboukris:
I pushed a new revision, unified the S4U client code, and removed the wrong kdc fixup. Still needs lots of cleanup.

Another unexpected problem I had with moving realm-identification to stepwise, is that now the caller doesn't know the canonicalized client realm, so it can't compare it with the realm of the service to decide whether or not convert the server to enterprise (with referral realm).
For now, if the impersonate client is enterprise I convert the server to enterprise unconditionally since the client realm may change, but this doesn't work well with other KDB modules which can't handle enterprise server, but do support S4U2Self.

We have discussed it before and I think you suggested the KDC can unwrap the enterprise servers, and I suspected it will complicate it for samba (although I don't quite understand how samba currently distinguishes between client and server lookups in TGS, as client enterprise needs db lookup, unlike server).

iboukris:
I added a practical fix that simply ignores the server realm in both S4U2S/S4U2P requests.  I'm thinking maybe we can go about it with documentation, stating that server realm is ignored in S4U.

I think for S4U2Self there is nothing to worry about as it must be the same principal as the client, which is actually verified by the KDC in the last request.

The S4U2Proxy case is more complicated, the application may be using the API in parallel to regular TGS request, but when it will ask the same server, like PC$OTHER it may get a tickets to different principals, if PC$ principal also exist in the local client realm.
However, I'm not so convinced this is so bad, getting a ticket for PC$@LOCAL is the best we can do, since we must use referrals and this name is not valid for referrals, and if the application got a wrong ticket in the S4U2Proxy case, it will just fail to use it against the target server.

iboukris:
As a middle way, we can try to guess based on the name and type, and only ignore the realm if:
(name->length > 1 && has_dot(name->data[1])) || (name->type == 10 && has_at(name->data[0]))

This is similar to what heimdal does in get_cred_kdc_referral (later the server realm is ignored);
https://github.com/heimdal/heimdal/blob/cf940e15f4eac71d7b739bbb6672d7be9f9b98cb/lib/krb5/get_cred.c#L979

PR #1015, opened 2019-12-17
Simplify keytab creation in kadmin tests

greghudson:
[This is a preparatory change for working around macOS System Integrity Protection in the test suite. 
 With SIP turned on, the macOS shell discards DYLD_LIBRARY_PATH at startup, so using the shell to call kadmin.local from make-host-keytab.plin is a problem.  But switching to a direct exec of kadmin.local creates a different problem when the operation might fail, as Perl has no way to redirect the output of a `system` command without using the shell, and two of the callers are Tcl scripts which respond to stderr output by throwing an exception.  The Tcl behavior can be overridden with `-ignorestderr`, but we don't want error messages in the output anyway.  So the best easy approach is to get rid of the `kadmin.local ank` operations that might fail.]

In init_db and init.exp, do not create an ovsec_adm.keytab; kadmind
has authenticated directly against the KDB since commit
416d9a774090ee78c30a844025887bd2b9e79d16.  Since we no longer create
ovsec_adkm principals, perform the deletion and recreation tests with
kadmin/ principals.

With these changes, make-host-keytab.plin does not need to create
principals.


greghudson:
It turns out that macOS /usr/bin/perl also discards DYLD_LIBRARY_PATH, posing an additional problem for make-host-keytab.pl.  The changes in this PR eliminated all but one use of that script.  I eliminated that use by running kadmin ktadd from helpers.exp in the RPC tests.

So, there are now two commits.  The first removes all use of make-host-keytab.pl, and the second removes the script along with five other unnecessary scripts and an autoconf test.

greghudson:
Yeah, I think if we found an alternative for def-check.pl, we could purge Perl code from the tree.  (def-check.pl runs unconditionally on Windows, so simply rewriting it in Python would trade a Perl dependency for a Python dependency for the Windows build.)

PR #1016, opened 2019-12-17
Fix LDAP policy enforcement of pw_expiration

frozencemetery:
In only the LDAP backend, the change mask is used to determine what to
update.  As a result, password expiration was not set from policy when
running during addprinc, among other issues.  However, when the mask did
not contain KADM5_PRINCIPAL, pw_expiration would be applied regardless,
which meant that (for instance) changing the password would cause the
password application to be applied.

Remove this check, and fix the mask to contain KADM5_PW_EXPIRATION where
appropriate.  Add a regression test to the LDAP suite.

### DIFF
--- a/src/lib/kadm5/srv/svr_principal.c
+++ b/src/lib/kadm5/srv/svr_principal.c
@@ -356,6 +356,11 @@ kadm5_create_principal_3(void *server_handle,
     kdb = calloc(1, sizeof(*kdb));
     if (kdb == NULL)
         return ENOMEM;
+
+    /* In all cases the principal entry is new and key data is set; let the
+     * database provider know. */
+    kdb->mask = mask | KADM5_KEY_DATA | KADM5_PRINCIPAL;
+
     memset(&adb, 0, sizeof(osa_princ_ent_rec));
 
     /*
@@ -405,14 +410,12 @@ kadm5_create_principal_3(void *server_handle,
         kdb->expiration = handle->params.expiration;
 
     kdb->pw_expiration = 0;
-    if (have_polent) {
-        if(polent.pw_max_life)
-            kdb->pw_expiration = ts_incr(now, polent.pw_max_life);
-        else
-            kdb->pw_expiration = 0;
-    }
-    if ((mask & KADM5_PW_EXPIRATION))
+    if (mask & KADM5_PW_EXPIRATION) {
         kdb->pw_expiration = entry->pw_expiration;
+    } else if (have_polent && polent.pw_max_life) {
+        kdb->mask |= KADM5_PW_EXPIRATION;
+        kdb->pw_expiration = ts_incr(now, polent.pw_max_life);
+    }
 
     kdb->last_success = 0;
     kdb->last_failed = 0;
@@ -502,9 +505,6 @@ kadm5_create_principal_3(void *server_handle,
 
         adb.policy = entry->policy;
     }
-
-    /* In all cases key and the principal data is set, let the database provider know */
-    kdb->mask = mask | KADM5_KEY_DATA | KADM5_PRINCIPAL ;
 
     /* store the new db entry */
     ret = kdb_put_entry(handle, kdb, &adb);
@@ -601,6 +601,9 @@ kadm5_modify_principal(void *server_handle,
     if (ret)
         return(ret);
 
+    /* Let the mask propagate to the database provider. */
+    kdb->mask = mask;
+
     /*
      * This is pretty much the same as create ...
      */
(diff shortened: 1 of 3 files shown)


src/lib/kadm5/srv/svr_principal.c
```diff
@@ -405,14 +405,13 @@ kadm5_create_principal_3(void *server_handle,
...
-    if (have_polent) {
-        if(polent.pw_max_life)
+    if ((mask & KADM5_PW_EXPIRATION)) {
+        kdb->pw_expiration = entry->pw_expiration;
+    } else if (have_polent) {
+        mask |= KADM5_PW_EXPIRATION;
+        if (polent.pw_max_life)
             kdb->pw_expiration = ts_incr(now, polent.pw_max_life);
-        else
-            kdb->pw_expiration = 0;
```
greghudson: I don't think we want to set the mask bit if polent.pw_max_life is 0, so line 410 could be written `} else if (have_polent && polent.pw_max_life)`.

I am concerned about modifying the value of mask, since we currently don't do that (we set kdb->mask to mask plus the bits we want to set) and it muddies the semantics of the variable.  It might be better to set kdb->mask earlier in the function, and then add to it here.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -616,16 +615,19 @@ kadm5_modify_principal(void *server_handle,
             free(adb.policy);
         adb.policy = strdup(entry->policy);
     }
-    if (have_pol) {
+
+    kdb->pw_expiration = 0;
```
greghudson: I think this has the new effect of unsetting the password expiration time whenever the policy isn't changed, for KDB modules other than LDAP.

src/lib/kadm5/srv/svr_principal.c
frozencemetery: I think we do - actually this is one of my test cases.  If the policy is updated to remove expiration limits, we'd like that to propgate, I think.

src/lib/kadm5/srv/svr_principal.c
frozencemetery: I don't think so?  At least, all I've done is hoisted the assignment out of the `else`.

src/lib/kadm5/srv/svr_principal.c
frozencemetery: (Stopped modifying `mask`, though.)

src/lib/kadm5/srv/svr_principal.c
frozencemetery: Agreed, fixed.

src/lib/kadm5/srv/svr_principal.c
greghudson: For this particular block we're creating a new new principal, though; there is no need to set the mask bit if we don't want pw_expiration set.

frozencemetery: Updated.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1437,12 +1442,9 @@ kadm5_chpass_principal_3(void *server_handle,
             }
         }
 
+        kdb->mask |= KADM5_PW_EXPIRATION;
```
greghudson: I think this belongs before the if statement, since in chpass we always want to change pw_expiration.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1599,13 +1599,12 @@ kadm5_randkey_principal_3(void *server_handle,
         if (ret)
             goto done;
     }
+
+    kdb->pw_expiration = 0;
     if (have_pol) {
+        kdb->mask |= KADM5_PW_EXPIRATION;
```
greghudson: Same here.

src/lib/kadm5/srv/svr_principal.c
```diff
@@ -1864,13 +1862,12 @@ kadm5_setkey_principal_4(void *server_handle, krb5_principal principal,
...
```
greghudson: Likewise here.

src/tests/t_kdb.py
```diff
@@ -482,6 +482,23 @@ def get_princ(princ):
...
+out = realm.run([kadminl, 'getprinc', 'pwuser'],
+                expected_msg='Password expiration date: ')
+if 'Password expiration date: [never]' in out:
+    fail('pw_expiration not applied at principal creation')
+realm.run([kadminl, 'modpol', '-maxlife', '0', 'pw_e'])
+realm.run([kadminl, 'cpw', '-pw', 'password_', 'pwuser'])
+realm.run([kadminl, 'getprinc', 'pwuser'],
+          expected_msg='Password expiration date: [never]')
+realm.run([kadminl, 'delprinc', 'pwuser'])
+realm.run([kadminl, 'delpol', 'pw_e'])
```
greghudson: We probably don't need these since we're about to stop the realm anyway.

frozencemetery:
Rebased and addressed comments.

PR #1017, opened 2019-12-19
WIP: Fix cross-realm S4U2Self checks

iboukris:
I've been thinking again about PR #970 and came up with two changes, follow up to: https://github.com/krb5/krb5/pull/970/commits/e12e890f063f41bf8aef45e44a3ee329f64139d2


iboukris:
In a closer look, they are both wrong, closing sorry.

PR #1018, opened 2019-12-19
Do not assume KRB5_KDB_FLAG_CANONICALIZE for enterprise principals

iboukris:
As discussed, i think it would be good to have this with the api change.

### DIFF
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -1057,15 +1057,18 @@ typedef struct _kdb_vftabl {
      * in-realm alias, fill in a different value for entries->princ than the
      * one requested.
      *
-     * A module can return out-of-realm referrals if KRB5_KDB_FLAG_CANONICALIZE
-     * is set.  For AS request clients (KRB5_KDB_FLAG_CLIENT_REFERRALS_ONLY is
-     * also set), the module should do so by simply filling in an out-of-realm
-     * name in entries->princ and setting all other fields to NULL.  Otherwise,
-     * the module should return the entry for the cross-realm TGS of the
-     * referred-to realm.  For TGS referals, the module can also include
-     * tl-data of type KRB5_TL_SERVER_REFERRAL containing ASN.1-encoded Windows
-     * referral data as documented in draft-ietf-krb-wg-kerberos-referrals-11
-     * appendix A; this will be returned to the client as encrypted padata.
+     * A module can return a referral to another realm if
+     * KRB5_KDB_FLAG_CANONICALIZE is set, or if
+     * KRB5_KDB_FLAG_CLIENT_REFERRALS_ONLY is set and search_for->type is
+     * KRB5_NT_ENTERPRISE_PRINCIPAL.  If KRB5_KDB_FLAG_CLIENT_REFERRALS_ONLY is
+     * set, the module should return a referral by simply filling in an
+     * out-of-realm name in (*entry)->princ and setting all other fields to
+     * NULL.  Otherwise, the module should return the entry for the cross-realm
+     * TGS of the referred-to realm.  For TGS referals, the module can also
+     * include tl-data of type KRB5_TL_SERVER_REFERRAL containing ASN.1-encoded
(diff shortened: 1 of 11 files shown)


iboukris:
Revised the code change and added tests and doc.

greghudson:
I pushed some wording changes.  The KDC change seems fine, but I have some concerns about the documentation and test module changes.

The KDC code already clamps the returned names to the request names in many circumstances; see https://github.com/krb5/krb5/blob/master/src/kdc/do_as_req.c#L696, https://github.com/krb5/krb5/blob/master/src/kdc/do_as_req.c#L723, and https://github.com/krb5/krb5/blob/master/src/kdc/do_tgs_req.c#410.  (The "XXX careful for realm" comment on that last line actually dates back to a 1990 commit, and isn't talking about Microsoft.)
So the test module code to sometimes set alias returns to the requested name seems likely redundant.

The documentation change to get_principal seems to say that it is okay to return an out-of-realm AS client referral for an enterprise client, if the canonicalize bit is not set.  Should that actually happen?  Will Windows do that?

It seems likely that we don't need the ALIAS_OK flag (added by me in commit e220a6288b10e1f9280e8aa2cb18ad56391e0a45).  We could instead just permit the KDB module to return an in-realm alias under all circumstances.  That would be a behavior change, but I don't think it would cause any problems.

iboukris:
> The documentation change to get_principal seems to say that it is okay to return an out-of-realm AS client referral for an enterprise client, if the canonicalize bit is not set. Should that actually happen? Will Windows do that?

Yes, Windows does that. In fact, Windows would also return server referrals for TGS request without canonicalize flag.
```log
kinit -E administrator@acom@MECA.ORG
Password for administrator\@acom@ACME.COM: 
[admin@f ~]$ klist 
Ticket cache: FILE:/tmp/krb5cc_1000
Default principal: administrator\@acom@ACME.COM
```

iboukris:
> So the test module code to sometimes set alias returns to the requested name seems likely redundant.

I'll can get rid of the kdb_test tests, and leave only the ldap one (at first I didn't think of leaving alias-ok so I needed the kdb_test module for testing).

iboukris:
>> The documentation change to get_principal seems to say that it is okay to return an out-of-realm >> AS client referral for an enterprise client, if the canonicalize bit is not set. Should that actually
>> happen? Will Windows do that?
>
> Yes, Windows does that.

On the other hand, in Windows for client referrals the name must be enterprise as short names would never yield a referral to another realm.

> It seems likely that we don't need the ALIAS_OK flag

Indeed, it doesn't convey much information.

iboukris:
>> It seems likely that we don't need the ALIAS_OK flag
>
> Indeed, it doesn't convey much information.

In a closer look, it still conveys the implicit alias-ok for TGS server lookup, which can't otherwise be distinguished from TGS client lookup. Perhaps we can add a flag to denote client vs server, I could have used such flag in a recent issue with samba's module, although I still unclear if it is really needed there.

greghudson:
My thinking is that we can let the KDB module always return in-realm aliases (setting princ to the canonical name), and the KDC will decide whether to use the requested name or the canonical name in the reply.  That would be a behavior change from how the LDAP module currently works; "kinit alias" (without -C) currently fails, and would instead issue a ticket for the alias name.

iboukris:
Added wip commit, I'm still unclear about the t_referral change, need to debug.

iboukris:
Actually it looks ok, the client still doesn't follow referrals if no canon and no enterprise were requested:
https://github.com/krb5/krb5/blob/c6c19b1d35c6523cb7ed220c1f2e97e12e039293/src/lib/krb5/krb/get_in_tkt.c#L1510

iboukris:
Actually not, the module should have returned error, fixed now.

greghudson:
Revised.  The second commit still had "wip" in it before I pushed the revision; is there more to check?

iboukris:
> is there more to check?
Looks good now, thanks.

PR #1020, opened 2019-12-20
Work around macOS SIP in the test suite

greghudson:
[I have included a second commit with Travis configuration for a macOS build, but don't intend to merge it.  Travis is often slow to even start macOS jobs, and the actual build is over twice as slow as a Linux build.  Github Actions appears to have significantly more performance macOS runners, so we might use that instead.  Travis and Github Actions both use Macstadium for their macOS runners, but I guess Github might be paying them more.]

[On Github Actions, but not Travis, I occasionally see a failure from kdc/t_bigreply.py where it doesn't output the expected trace message about retrying with TCP.  I haven't managed to capture log output for this problem.  So there might need to be one additional workaround.]

[The SO_REUSEPORT workaround addresses a problem where the second krb5kdc process run by the test suite sometimes fails to start with an address reuse error.  Puzzlingly, I never saw this problem on Travis.  I saw it pretty often on Github Actions, and consistently on a desktop Mac when I had access to one.]

In macOS 10.11 and later with System Integrity Protection enabled,
system programs (including the shell) purge DYLD_LIBRARY_PATH from the
environment at startup.  As a result, any part of "make check" which
runs via a shell script must explicitly restore the runtime
environment.  Add a common rule for runenv.sh, and create and source
it where shell scripts are run.  Dejagnu's runtest is a shell script,
so create a tcl file for the kadmin and RPC unit tests and source it
from unix.exp.  Avoid using the shell to run commands in several
places.  Add a "tracefile" parameter to realm.run() in k5test.py to
allow a trace file to be generated without indirecting through
/usr/bin/env.

Do not include <malloc.h> in t_parse_host_string.c, as it does not
exist on macOS and is not needed.

Skip the iprop tests on macOS when SIP is enabled, as signal
restrictions appear to prevent the kpropd child process from informing
the parent process that a full resync has completed.

In net-server.c, set SO_REUSEPORT as well as SO_REUSEADDR on listener
sockets.  Otherwise the krb5kdc processes run by the test suite
sometimes fails to start with an "address in use" error.

In configure.ac, only generate po/Makefile if we will descend into it.


greghudson:
Added a commit to enable IPv6 pktinfo support on macOS.

src/util/k5test.py
```diff
@@ -241,7 +241,8 @@
   return code other than 0, expected_msg=MSG to expect a substring in
   the command output, and expected_trace=('a', 'b', ...) to expect an
   ordered series of line substrings in the command's KRB5_TRACE
-  output.
+  output, or tracefile=FILENAME to write trace output to the specified
```
frozencemetery: All uses of tracefile immediately open said file.  I'm wondering if it wouldn't be better to move that logic into `run` somehow - maybe a parameter to return the trace output instead of the command output, or perhaps interleave the two?

Not a huge issue though - if it's too complex to do that, this is fine as-is.

greghudson:
I managed to produce the t_bigreply.py failure twice on github actions with verbose logging turned on.  In the failure case, the client sees a timeout rather than a RESPONSE_TOO_BIG reply, like this:

    [38276] 1576877648.334486: Sending initial UDP request to dgram 10.79.3.177:61000
    [38276] 1576877649.476454: Initiating TCP connection to stream 10.79.3.177:61000

I don't have any good hypotheses here.  The KDC clearly hasn't died, since the request completes successfully.  I don't think packets should be getting lost over a local interface.  The KDC shouldn't be taking over one second to process a simple request, or successful runs of the test suite would be glacially slow.

src/util/k5test.py
greghudson: I added a commit with my best idea along these lines, which is a "return_trace=True" parameter.  Let me know if it seems better or just awkward.  (This is the case I've run into where C's approach to multiple results is arguably more elegant than the functions-can-return-tuples approach taken by most modern languages.)  If we like it, I will either squash it into the first commit or make it a separate first commit.

frozencemetery: I think I slightly prefer it, but if you don't I'm fine with leaving it alone.

greghudson:
I reorganized the commits, but did not change the final result, so no re-review is needed.  I will wait for a CI run, remove the last commit, wait for another CI run, and merge.

PR #1024, opened 2020-01-12
RBCD KDC fixups

iboukris:
As discussed, I tested each commit with the broken client and it gets an error.


greghudson:
local_tgs leaks in the second patch, in test_get_principal().

iboukris:
Pushed a fix.

greghudson:
Minor edits.  I changed KRB_AP_ERR_NOT_US to KRB5_KDB_NOENTRY.  The former is a protocol  code (KRB5KRB_AP_ERR_NOT_US would be the corresponding com_err code), and KRB5_KDB_NOENTRY more accurately reflects what a regular KDB would return for a non-local principal lookup.

iboukris:
Looks good thanks, btw Windows would return WRONG_REALM (with no hint).

PR #1025, opened 2020-01-14
Apply permitted_enctypes to client sesion enctypes

frozencemetery:
permitted_enctypes was initially intended only to apply to application
servers.  In ef9b79c0dd6bdc5d7b198dc1d681086d84b86b22 it was generalized
to apply to the KDC as well.  However, documentation did not mention
that permitted_enctypes did not apply to clients, and as a result, many
configurations assume that it does.

In light of the existing configurations, take the simple way out and use
permitted_enctypes as the default for default_tkt_enctypes and
default_tgs_enctypes.  Update documentation to match.

Add a test to explicitly check this behavior, and remove now-unnecessary
configuration from the test suite.


greghudson:
I think get_profile_etype_list() probably doesn't make sense any more as a helper after this change, with each accessor having different behavior.

It's out of scope for this PR, but I noticed that krb5_set_default_in_tkt_ktypes() and krb5_set_default_tgs_ktypes() aren't public (dating back to release 1.3) and aren't used outside of test programs.  They and context->in_tkt_etypes could be removed.

(Unfortunately krb5_set_default_tgs_enctypes() is public, is used by the krb5 mech's gss_init_sec_context(), and is the subject of the most technical debt: it affects krb5_get_permitted_enctypes() for unclear reasons, and context->use_conf_ktypes suppresses it for krb5_get_tgs_ktypes().)

greghudson:
I pushed a fixup commit with the helper function unrolled and with the zero-length-list case pushed down into krb5int_parse_enctype_list().  The line count is roughly the same, although some comments would probably need to be added.

There is an unsightly amount of repetition between the the in_tkt and tgs cases, with only `context->use_conf_ktypes` and the first profile key really separating them.  (If we get rid of `context->in_tkt_etypes`, that is another difference.)  I think that's better than having an awkwardly conditionalized helper, but I'll keep thinking about the code structure.

(Due diligence: the only other caller of krb5int_parse_enctype_list() is dbentry_supports_enctype() in the KDC.  It considers an empty list the same way as an error, at least for now, so pushing the empty list error down into the parsing function doesn't change any behavior.  The context lists we copy with k5_copy_etypes() can't be empty because of how set_default_etype_var() works, so we don't need to check for an empty list if we use them.)

greghudson:
I revised the commit message, documentation, and test case.

PR #1026, opened 2020-01-15
Fix KDC crash in handle_signticket

iboukris:
Regression since RBCD changes, commit d47f7dba3779 changed the
parameters passed to sign_authdata() in S4U2Proxy so that client
is the impersonted client (used to be impersonator), and added
header_server for the impersonator.  But it failed to fix the call
to handle_signticket() to use header_server instead of client which
now can be NULL when the client is not local.


greghudson:
Should we also change the name of the handle_signticket() parameter?

iboukris:
Yeah I'll revise, the test can use some cleanup too.

iboukris:
Updated, also extended the usage of subject_server to make the parameter clearer.

greghudson:
I don't have permission to push to the PR branch (this happens occasionally, no idea why in this case).  I only had whitespace and commit message edits, so I went ahead and pushed 96e5d384acf174e6079b0aeeec14bd8100d24840 to master.

PR #1027, opened 2020-01-18
Allow cross-realm RBCD with PAC not the only ad element

iboukris:
Follow-up to #1013


iboukris:
Pushed a new fixup, sorry for all the back and forth on this.

iboukris:
And another update, we shouldn't try to verify singed-path on cross-realm tickets, squashed now.

iboukris:
One more, it wouldn't harm to check for the PAC in cross-realm tickets (at least for the time being).

greghudson:
Pushed changes mostly just to the commit messages; the only code change I made was to check the cross-realm flag before calling has_pac() to more efficiently process within-realm requests.  This should be ready to merge.

iboukris:
Looks good, thanks.

PR #1028, opened 2020-01-23
Remove unexposed mutators for session enctype lists

frozencemetery:
(Follow-up to https://github.com/krb5/krb5/pull/1025#issuecomment-576038319 )
    
krb5_set_default_in_tkt_ktypes() and krb5_set_default_tgs_ktypes() are
not publicly exposed, so remove them.  Also remove the now-unused
in_tkt_etypes field from krb5_context.  Update test suite consumers, and
remove etype info tests that no longer test anything.

### DIFF
--- a/src/lib/krb5/krb/t_copy_context.c
+++ b/src/lib/krb5/krb/t_copy_context.c
@@ -70,7 +70,6 @@ check_context(krb5_context c, krb5_context r)
     int i;
 
     /* Check fields which should have been propagated from r. */
-    compare_etypes(c->in_tkt_etypes, r->in_tkt_etypes);
     compare_etypes(c->tgs_etypes, r->tgs_etypes);
     check(c->os_context.time_offset == r->os_context.time_offset);
     check(c->os_context.usec_offset == r->os_context.usec_offset);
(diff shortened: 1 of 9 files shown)


frozencemetery:
Based on top of #1025.

While etinfo.c could theoretically be invoked without an enctype list, in practice it never is because the only caller is t_etype_info.py.  This tests session enctype negotiation, which has value but doesn't strike me as particularly fraught.  While we could rewrite the test to use permitted_enctypes throughout, I'm not sure it's worth the effort to eliminate a mutator and a variable.

frozencemetery:
Rebased onto master now that #1025 merged.

greghudson:
See https://github.com/greghudson/krb5/commit/a6a3759cb9aac291b32846827646a98e7f6757fc for changing the etype-info tests to use krb5.conf.

frozencemetery:
Updated, pulling that in.

src/lib/krb5/krb/t_copy_context.c
```diff
@@ -113,7 +112,6 @@ main(int argc, char **argv)
 {
     krb5_context ctx, ctx2;
     krb5_plugin_initvt_fn *mods;
-    const krb5_enctype etypes1[] = { ENCTYPE_DES3_CBC_SHA1, 0 };
```
greghudson: etypes2 should be renamed to etypes or the resulting code will be confusing.

frozencemetery: Updated.

greghudson:
I restored the t_etypes.c tests by making them test krb5_set_default_tgs_enctypes() instead.

PR #1029, opened 2020-01-23
Further simplify test KDB module authdata code

greghudson:
Commit 94f7c9705879500b1dc8dda8592490efce05688f simplified the
generation of authdata elements, but left behind some unnecessary
conditionalization when assembling the elements into a list, causing a
Coverity defect.  Further simplify the code.


iboukris:
Ah right, looks good, thanks.

frozencemetery:
Would be convenient if this were pulled into the release when 94f7c9705879500b1dc8dda8592490efce05688f is.

PR #1032, opened 2020-01-29
Change KDC constrained-delegation order of precedence

iboukris:
Per MS-SFU errata changes from 2019/12/09:
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-winerrata/68c4fd08-207c-4353-b59d-4d281edfb6bf

### DIFF
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1622,65 +1622,6 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
     return 0;
 }
 
-static krb5_error_code
-check_allowed_to_delegate_to(krb5_context context, krb5_const_principal client,
-                             const krb5_db_entry *server,
-                             krb5_const_principal proxy)
-{
-    /* Must be in same realm */
-    if (!krb5_realm_compare(context, server->princ, proxy))
-        return KRB5KDC_ERR_POLICY;
-
-    return krb5_db_check_allowed_to_delegate(context, client, server, proxy);
-}
-
-static krb5_error_code
-check_rbcd_policy(kdc_realm_t *kdc_active_realm, unsigned int flags,
-                  const krb5_principal stkt_client_princ,
-                  krb5_principal stkt_authdata_client,
-                  const krb5_db_entry *stkt_server,
-                  krb5_const_principal header_client_princ,
-                  void *header_ad_info, const krb5_db_entry *proxy)
-{
-    krb5_principal client_princ = stkt_client_princ;
-
-    /* Ensure that either the evidence ticket server or the client matches the
-     * TGT client. */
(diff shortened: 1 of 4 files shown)


iboukris:
The logical change looks ok now, it could be simplified but I tried not to change the returned error codes, mainly to avoid impact in the tests.  I hope to run some experiments and take notes of the exact error codes, but that will take some time.

greghudson:
What's the motivation for the first commit?

iboukris:
I had something similar in development, in legacy-cd, if the kdb for some reason will return ad_info but not stkt_ad_info, we end up passing ad_info  to sign_authdata(), while it expects stkt_ad_info.

greghudson:
Is there a reason to still have an stkt_ad_info variable, instead of nulling ad_info and reading the kdc_process_s4u2proxy_req() result into ad_info?

iboukris:
We need to pass ad_info to kdc_process_s4u2proxy_req() for RBCD policy check, maybe we can pass pointer to it, if that's what you mean.

greghudson:
Aha, I understand now.

src/kdc/kdc_util.c
```diff
@@ -1698,6 +1686,8 @@ kdc_process_s4u2proxy_req(kdc_realm_t *kdc_active_realm, unsigned int flags,
     krb5_error_code errcode;
     krb5_boolean support_rbcd;
 
+    *status = "NOT_ALLOWED_TO_DELEGATE";
```
greghudson: This assignment happens on success as well as failure.  So if a later call in process_tgs_req() fails without setting the status string, we will use "NOT_ALLOWED_TO_DELEGATE" instead of "UKNOWN_REASON".  I think it's better to keep setting a status string on each error exit path.

iboukris: Fixed now.

greghudson:
If we return KRB5_TKT_NOT_FORWARDABLE from kdc_process_s4u2proxy_req(), the KDC returns KRB_ERR_GENERIC (60).  [MS-SFU] says it should return KDC_ERR_BADOPTION (13).

The current PR logic returns this error code in two cases (both requiring intermediate and target in the same realm, and a non-forwardable evidence ticket):

* allowed_to_delegate says yes
* allowed_to_delegate says no, and the client doesn't support RBCD (or the PAC isn't verified)

I wonder if Windows returns KDC_ERR_POLICY(12) in the second case; it seems like the path of least resistance.

iboukris:
Windows KDC returns KDC_ERR_BADOPTION in all these cases, i'll work out a fixup.

iboukris:
I added a change to make it easier to follow why stkt_authdata_client is valid in cross-realm.

greghudson:
I merged the first commit.

iboukris:
I pushed a new revision.

PR #1033, opened 2020-02-01
Allow KDB authdata to be the first ad-elements

iboukris:
See: http://mailman.mit.edu/pipermail/krbdev/2020-January/013200.html

This still needs some research, testing and tests, but that's basically the idea.
It is cleat that the PAC needs to proceed other ad-elements (at least unknown to Windows), while for other ad-elements, as far as i can tell the order doesn't matter.

With this patch:
```code
PYTHONPATH=../util VALGRIND="" python3 ./t_authdata.py -v
...
*** [130] Executing: ./adata -p user@B service/rb.b -2 cross_s4u_proxy_ad
?128: 0500000000000000010000000100000058000000000000000B0000001B00000060000000000000000A000000120000008000000000000000060000001000000098000000000000000700000010000000A8000000000000000000000000000000736572766963652F72622E623A696D706572736F6E61746F724041000000000000BCF73213D9D501080075007300650072000000000000000F000000F16C33374BF00452556DA63C0F000000E01827F9524CC8CD9D4B5413
 -456: db-authdata-test
 -456: db-authdata-test
 -2: cross_s4u_proxy_ad
 -456: db-authdata-test
 -456: db-authdata-test
 -2: cross_s4u_self_ad
 -456: db-authdata-test
 -456: db-authdata-test
?512: 3041A003020111A1173015A00302010FA10E040CE318D6D524DAAD5AC1B29AC3A221301F301DA0163014A003020102A10D300B1B066B72627467741B0142A1031B0141
*** [130] Completed with return code 0
```

### DIFF
--- a/src/kdc/kdc_authdata.c
+++ b/src/kdc/kdc_authdata.c
@@ -372,11 +372,14 @@ fetch_kdb_authdata(krb5_context context, unsigned int flags,
     if (ret)
         return (ret == KRB5_PLUGIN_OP_NOTSUPP) ? 0 : ret;
 
-    /* Add the KDB authdata to the ticket, without copying or filtering. */
-    ret = merge_authdata(context, db_authdata,
-                         &enc_tkt_reply->authorization_data, FALSE, FALSE);
+    /* Put the KDB authdata first in the ticket.  A successful merge places the
+     * combined list in db_authdata and releases the old ticket authdata. */
+    ret = merge_authdata(context, enc_tkt_reply->authorization_data,
+                         &db_authdata, FALSE, FALSE);
     if (ret)
         krb5_free_authdata(context, db_authdata);
+    else
+        enc_tkt_reply->authorization_data = db_authdata;
     return ret;
 }
 
--- a/src/tests/adata.c
+++ b/src/tests/adata.c
@@ -56,7 +56,8 @@
 static krb5_context ctx;
 
 static void display_authdata_list(krb5_authdata **list, krb5_keyblock *skey,
-                                  krb5_keyblock *tktkey, char prefix_byte);
+                                  krb5_keyblock *tktkey, char prefix_byte,
+                                  krb5_boolean pac_expected);
 
 static void
 check(krb5_error_code code)
@@ -206,7 +207,7 @@ display_binary_or_ascii(krb5_authdata *ad)
  * must be the ticket session key. */
 static void
 display_authdata(krb5_authdata *ad, krb5_keyblock *skey, krb5_keyblock *tktkey,
-                 int prefix_byte)
+                 int prefix_byte, krb5_boolean pac_expected)
 {
     krb5_authdata **inner_ad;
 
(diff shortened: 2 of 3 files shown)


iboukris:
I think the PR is ready for review. I also found an old capture "rbcd_win_two_transits_with_keys.tgz", which i've uploaded back then to https://wiki.wireshark.org/SampleCaptures page, in packets 30/32, i can see that the client sends encrypted request autdata of types 141 and 142 inside an ad-if-relevant container, and the windows kdc places this container after the first container which contains the PAC.

src/kdc/kdc_authdata.c
```diff
@@ -373,10 +373,12 @@ fetch_kdb_authdata(krb5_context context, unsigned int flags,
         return (ret == KRB5_PLUGIN_OP_NOTSUPP) ? 0 : ret;
 
     /* Add the KDB authdata to the ticket, without copying or filtering. */
-    ret = merge_authdata(context, db_authdata,
-                         &enc_tkt_reply->authorization_data, FALSE, FALSE);
+    ret = merge_authdata(context, enc_tkt_reply->authorization_data,
```
frozencemetery: Changes here should probably have a comment.

src/tests/adata.c
```diff
@@ -216,11 +217,16 @@ display_authdata(krb5_authdata *ad, krb5_keyblock *skey, krb5_keyblock *tktkey,
         ad->ad_type == KRB5_AUTHDATA_CAMMAC) {
         /* Decode and display the contents. */
         inner_ad = get_container_contents(ad, skey, tktkey);
-        display_authdata_list(inner_ad, skey, tktkey, get_prefix_byte(ad));
+        display_authdata_list(inner_ad, skey, tktkey, get_prefix_byte(ad),
+                              (ad->ad_type == KRB5_AUTHDATA_IF_RELEVANT &&
```
frozencemetery: I think it'd be clearer to give this condition its own variable.

src/kdc/kdc_authdata.c
greghudson: I will probably follow up this PR with a cleanup for merge_authdata().  Having allocation behavior dependent on a flag is difficult for static analysis tools (and humans) to reason about, so I will probably split the helper into a "merge without copying" variant (which nulls the list it invalidates on success, like we decided k5_add_pa_data_element() should do) and a "merge with copying" wrapper.

(Side note: there is a relatively hefty function named concat_authorization_data() which

iboukris: Right, this code can use some cleanup, I'll leave the comment to Greg.

src/tests/adata.c
iboukris: I've updated the tests slightly.

iboukris:
BTW @frozencemetery for the 1.17, i think the kdc commit from this PR should suffice to address both the request authdata issue and the cammac issue.

greghudson:
I added a comment to the first commit, edited the tests a little bit for simplicity and style conformance, and edited the commit messages.

iboukris:
Thanks, looks good.

abbra:
For the record, the commit 331fa4b backported to 1.17 actually fixes the issue with CAMMAC and PAC for FreeIPA 4.8.2 or later. I tested and reported the status in this FreeIPA ticket: https://pagure.io/freeipa/issue/8185
@greghudson would it make sense to have this commit backported to 1.17?

frozencemetery:
@abbra The change is marked for backport already, though the test suite change is not.

greghudson:
> The change is marked for backport already, though the test suite change is not.

It is, actually.  It has the same ticket number as the code change, and the RT metadata apply to the ticket, not the commit.

PR #1034, opened 2020-02-03
Handle NULL result from OpenSSL evp functions

frozencemetery:
Starting in OpenSSL 3.0, the evp_*() functions can fail if the
appropriate provider is not loaded.  This is most likely observed with
the legacy evp_md4().  While the OpenSSL functions cope with NULL value,
we incorrectly return ENOMEM from hash_evp(), which complicates
debugging.

Analyzed by the OpenSSL folks (@kaduk, @levitte, @mspncp, @mattcaswell) at https://github.com/openssl/openssl/pull/10992

### DIFF
--- a/src/lib/crypto/openssl/hash_provider/hash_evp.c
+++ b/src/lib/crypto/openssl/hash_provider/hash_evp.c
@@ -42,6 +42,11 @@ hash_evp(const EVP_MD *type, const krb5_crypto_iov *data, size_t num_data,
     size_t i;
     int ok;
 
+    /* Starting in OpenSSL 3.0, the evp_*() calls can fail if the appropriate
+     * provider is not loaded. */
+    if (type == NULL)
+        return KRB5_CRYPTO_INTERNAL;
+
     if (output->length != (unsigned int)EVP_MD_size(type))
         return KRB5_CRYPTO_INTERNAL;
 


src/lib/crypto/openssl/hash_provider/hash_evp.c
```diff
@@ -42,6 +42,11 @@ hash_evp(const EVP_MD *type, const krb5_crypto_iov *data, size_t num_data,
     size_t i;
     int ok;
 
+    /* Starting in OpenSSL 3.0, the evp_*() calls can fail if the appropriate
+     * provider is not loaded. */
+    if (type == NULL)
```
mattcaswell: I don't think this is quite right. Calls like `EVP_md4()` still will not fail. But if you attempt to use them then the initialisation will fail. So I suspect it must be the `EVP_DigestInit_ex()` call that was failing in the discussion in https://github.com/openssl/openssl/pull/10992

mspncp:
@frozencemetery as @matcaswell already pointed out, `EVP_md4 ()` returns a nonzero pointer, but it "doesn't work".

```
#0  EVP_md4 () at crypto/evp/legacy_md4.c:32
#1  0x00007ffff7eba52e in hash_md4 (data=0x7fffffffdd10, num_data=4, output=0x7fffffffdc70) at hash_evp.c:67
#2  0x00007ffff7ec1edf in krb5int_old_encrypt (ktp=0x7ffff7ed1660 <krb5int_enctypes_list+128>, key=0x55555555c750, usage=7, ivec=0x0, data=0x7fffffffdd10, num_data=4) at enc_old.c:100
#3  0x00007ffff7ebf72c in krb5_k_encrypt (context=0x0, key=0x55555555c750, usage=7, cipher_state=0x0, input=0x7fffffffdf40, output=0x7fffffffde20) at encrypt.c:72
#4  0x00007ffff7ebf7c6 in krb5_c_encrypt (context=0x0, keyblock=0x55555555c790, usage=7, cipher_state=0x0, input=0x7fffffffdf40, output=0x7fffffffde20) at encrypt.c:91
#5  0x00005555555556cd in main () at t_encrypt.c:150
```

As a consequence, `EVP_DigestInit_ex()` fails in `hash_evp()` and `ok` is set to zero. At the end, `hash_evp()` erroneously returns ENOMEM, because it guesses the wrong cause for the error:

https://github.com/krb5/krb5/blob/a3a5c9b31cd9b421941eff9cd11dbc3876f4ff90/src/lib/crypto/openssl/hash_provider/hash_evp.c#L61

```
#0  hash_evp (type=0x7ffff7e61760 <md4_md>, data=0x7fffffffdd10, num_data=4, output=0x7fffffffdc70) at hash_evp.c:61
#1  0x00007ffff7eba545 in hash_md4 (data=0x7fffffffdd10, num_data=4, output=0x7fffffffdc70) at hash_evp.c:67
#2  0x00007ffff7ec1edf in krb5int_old_encrypt (ktp=0x7ffff7ed1660 <krb5int_enctypes_list+128>, key=0x55555555c750, usage=7, ivec=0x0, data=0x7fffffffdd10, num_data=4) at enc_old.c:100
#3  0x00007ffff7ebf72c in krb5_k_encrypt (context=0x0, key=0x55555555c750, usage=7, cipher_state=0x0, input=0x7fffffffdf40, output=0x7fffffffde20) at encrypt.c:72
#4  0x00007ffff7ebf7c6 in krb5_c_encrypt (context=0x0, keyblock=0x55555555c790, usage=7, cipher_state=0x0, input=0x7fffffffdf40, output=0x7fffffffde20) at encrypt.c:91
#5  0x00005555555556cd in main () at t_encrypt.c:150
```

mspncp:
This is the reason why the program aborts with the misleading error message
```
Continuing.
Encrypting (c): . . . Failed: Cannot allocate memory

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
```

mspncp:
Debugged with 890ca2f401924cdcb88f493950b04700bbe52db3.

mspncp:
(Didn't see that you mentioned the ENOMEM thing already at the top)

frozencemetery:
Debugging advanced after making this.  Will investigate and see if we can make the aborting return a better message; if not, it's not clear that KRB5_CRYPTO_INTERNAL is more useful than ENOMEM.

PR #1036, opened 2020-02-06
Refactor KDC authdata list management helpers

greghudson:
Remove the unused concat_authorization_data().  Split merge_authdata()
into two helpers, one to destructively merge without filtering and one
to merge a copy with filtering.  Remove context parameters where they
aren't needed (taking advantage of knowledge that some libkrb5
functions don't use their context parameters).

[It turns out that all of the existing calls to merge_authdata() either want to copy and filter, or destructively merge without filtering.  I'm taking advantage of that to simplify the helper function contracts.]

### DIFF
--- a/src/kdc/kdc_authdata.c
+++ b/src/kdc/kdc_authdata.c
@@ -108,7 +108,7 @@ unload_authdata_plugins(krb5_context context)
 /* Return true if authdata should be filtered when copying from untrusted
  * authdata.  If desired_type is non-zero, look only for that type. */
 static krb5_boolean
-is_kdc_issued_authdatum(krb5_context context, krb5_authdata *authdata,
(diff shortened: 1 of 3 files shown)


greghudson:
> A part of me would prefer an early return if the new list is essentially empty, but the code is nice.

While it's common for new_elements to be null, I think it should almost never be non-null but empty.  RFC 4120 says "AuthorizationData is always used as an OPTIONAL field and should not be empty", although this is not enforced on decoding.  And of course when we locally generate authdata to merge in, we don't generate empty lists.

An early return for non-null but empty new_elements is complicated code-wise because we would still need to free the list.

The end result is the same, as at the end we check if the composed list is empty and replace it with null.

PR #1037, opened 2020-02-13
Rename KDB authdata param server_princ to request_server

iboukris:
to distinguish it from PAC server.

### DIFF
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -724,7 +724,7 @@ krb5_error_code krb5_db_get_authdata_info(krb5_context context,
                                           unsigned int flags,
                                           krb5_authdata **in_authdata,
                                           krb5_const_principal client_princ,
-                                          krb5_const_principal server_princ,
+                                          krb5_const_principal request_server,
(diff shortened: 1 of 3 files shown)


greghudson:
I'm not sure I like this change.  The existing names distinguish between principal names and KDB entries, and server_princ is documented properly.  I don't see any misleading suggestion in the existing names that server_princ corresponds to header_server.

iboukris:
Ok, it confused me that server_princ and server can be different entities.

greghudson:
> Ok, it confused me that server_princ and server can be different entities.

In the referral case, or in other cases?  I'm not sure how this relates to the PAC server, which as far as I know is header_server.

iboukris:
My confusion was about get_authdata_info() that actually takes a server_key and that server_princ (used for delegation-info blob), and then there is the referral case. But the doc is ok.

greghudson:
The change at kdb.h:1294 threw me off, because it changes the comment for sign_authdata.  Is that an error?

iboukris:
It's the same argument to both functions, so I changed both

PR #1038, opened 2020-02-13
Use GitHub Actions for CI

greghudson:
[Github Actions has marginally better average build times than the alternatives in my experience, and it can support all three of the platforms we want to build on.  For now I don't have a macOS build here, due to the unsolved t_bigreply.py issue]

Use Github Actions instead of Travis and AppVeyor.

In the Windows installer config, add support for Visual Studio 2019
(aka 16.0).

[ghudson@mit.edu: switched to Ubuntu 18.04 for Linux builds; removed
macOS build job for now; added more packages to avoid skipping tests;
made it easier to see skipped tests and to see files not cleaned;
added make install command; adjusted Windows build path]


greghudson:
For whatever reason, Github Actions uses the master branch contents to decide what to do, rather than the contents of each commit like Travis or Appveyor do (as currently configured, at least).  You can look at https://github.com/greghudson/krb5 to see what it looks like; there is an "Actions" tab near the top, and also next to the "Latest commit" text there is a checkmark you can click to see the build results.  (The checkmark is a red X if the CI failed, and a yellow dot if it's still in progress.)  I don't know what the UI is for pull requests.

PR #1041, opened 2020-02-24
Allow certauth modules to set hw-authent flag

greghudson:
[Pursuant to http://mailman.mit.edu/pipermail/krbdev/2020-February/013206.html ]

In PKINIT, if a certauth module returns KRB5_CERTAUTH_HWAUTH from its
authorize method, set the hw-authent flag in the ticket.


src/plugins/preauth/pkinit/pkinit_srv.c
```diff
@@ -347,9 +347,10 @@ authorize_cert(krb5_context context, certauth_handle *certauth_modules,
 
     /*
      * Check the certificate against each certauth module.  For the certificate
-     * to be authorized at least one module must return 0, and no module can an
-     * error code other than KRB5_PLUGIN_NO_HANDLE (pass).  Add indicators from
-     * modules that return 0 or pass.
+     * to be authorized at least one module must return 0 or
+     * KRB5_CERTAUTH_HWAUTH, and no module can an error code other than
```
frozencemetery: I recognize this was a problem in the original, but can we fix "can an error code" here?  I think it should be something like "can return an error code".

PR #1042, opened 2020-02-26
Fix to musl libc issue introduced by bf5953c

TBK:
#992 - Commit bf5953c549a6d279977df69ffe89b2ba51460eaf introduced  __GLIBC_PREREQ which causes compilation against musl libc to fail.

### DIFF
--- a/src/util/support/plugins.c
+++ b/src/util/support/plugins.c
@@ -62,8 +62,7 @@
  * dlopen() with RTLD_NODELETE, we weren't going to unload the plugin objects
  * anyway.
  */
-#ifdef __linux__
-#include <features.h>
+#ifdef __GLIBC__PREREQ
 #if ! __GLIBC_PREREQ(2, 25)
 #define dlclose(x)
 #endif


src/util/support/plugins.c
```diff
@@ -64,10 +64,12 @@
  */
 #ifdef __linux__
 #include <features.h>
+#if defined(__GLIBC__)
```
frozencemetery: Is there a reason not to use `#ifdef` here?

TBK: No, I will change it.

TBK: Thanks for the git commit link.

greghudson:
I pushed a proposed simplification (as a second commit for now, but it would of course be squashed), based on the following research:

* features.h is implicitly included by just about any standard header on Linux systems, because it is included by <bits/types.h> which is included by other headers like <stdio.h>.

* features.h defines both __GLIBC__ and __GLIBC_PREREQ on glibc.  The compiler does not define __GLIBC__ (as it does __linux__).

* We have vendored code that checks for __GLIBC__ without explicitly including <features.h> in two places: the bison-generated lib/krb5/krb/deltat.t, and util/verto/ev.c.  So it's pretty common to rely on features.h being implicit.

Therefore we can just check for __GLIBC_PREREQ directly--no need to check for __linux__ or include <features.h>.

frozencemetery:
Also fine with this approach.

puckipedia:
This did not fix musl builds, as the commit contains a typo: It checks if `__GLIBC__PREREQ` is defined, with two underscores; not `__GLIBC_PREREQ` with one underscore.

greghudson:
Whoops.  But, shouldn't the change fix musl builds anyway, with the side effect of also reverting the workaround introduced in commit bf5953c549a6d279977df69ffe89b2ba51460eaf?

greghudson:
I did a bit of poking around, the typo does (as I would expect) cause the tests to fail on Ubuntu 16.04 due to the libc bug.  I don't see why musl builds would still be failing, as `__GLIBC__PREREQ` wouldn't be defined in that environment (or any environment).  I guess I will fix the typo and we can revisit if there is still an issue.

PR #1044, opened 2020-02-26
Attempt to refresh credentials in the ccache when expired

frozencemetery:
Fixes an issue where client keytabs would be ignored if expired
credentials were present in the ccache.

ticket: 7976

### DIFF
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -557,15 +557,23 @@ set_refresh_time(krb5_context context, krb5_ccache ccache,
 krb5_boolean
 kg_cred_time_to_refresh(krb5_context context, krb5_gss_cred_id_rec *cred)
 {
-    krb5_timestamp now;
+    krb5_timestamp now, soon;
 
     if (krb5_timeofday(context, &now))
         return FALSE;
+    soon = ts_incr(now, 30);
     if (cred->refresh_time != 0 && !ts_after(cred->refresh_time, now)) {
-        set_refresh_time(context, cred->ccache,
-                         ts_incr(cred->refresh_time, 30));
+        set_refresh_time(context, cred->ccache, soon);
         return TRUE;
     }
+
+    /* If the creds will expire soon, try to refresh even if they weren't
+     * acquired with a client keytab. */
+    if (ts_after(soon, cred->expire)) {
+        set_refresh_time(context, cred->ccache, soon);
+        return TRUE;
+    }
+
     return FALSE;
 }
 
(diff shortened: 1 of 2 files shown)


frozencemetery:
Force pushed to rerun CI.

The failure was:

```
PYTHONPATH=../util VALGRIND="" python3 ./t_otp.py 
Traceback (most recent call last):
  File "./t_otp.py", line 268, in <module>
    verify(daemon1, queue, False, realm.user_princ.split('@')[0], 'accept2')
  File "./t_otp.py", line 149, in verify
    assert data['reply'] is reply
AssertionError
For details, see: /home/runner/work/krb5/krb5/src/tests/testlog
Or re-run this test script with the -v flag:
    cd /home/runner/work/krb5/krb5/src/tests
    PYTHONPATH=/home/runner/work/krb5/krb5/src/util /usr/bin/python3 ./t_otp.py -v

Use --debug=NUM to run a command under a debugger.  Use
--stop-after=NUM to stop after a daemon is started in order to
attach to it with a debugger.  Use --help to see other
options.
make[1]: *** [check-pytests] Error 1
Makefile:765: recipe for target 'check-pytests' failed
make[1]: Leaving directory '/home/runner/work/krb5/krb5/src/tests'
make: *** [check-recurse] Error 1
Makefile:1546: recipe for target 'check-recurse' failed
##[error]Process completed with exit code 2.
```

Anecdotally, I think I've seen this sporadically from CI in the past as well.

greghudson:
Some surrounding context:

* The ideal way to use the client keytab is to give the application it's own ccache, and let GSS populate the ccache initially.  In that case, the refresh_time ccconfig var causes the cache to be refreshed when it is halfway to expired, with retries spaced out by 30s.

* In some cases, the admin thinks they should prime the ccache using kinit.  If they do that there is no refresh_time config var and the cache doesn't get auto-refreshed.  If the credentials present in the ccache match the keytab name, then it's relatively safe for GSS to refresh the ccache on or near expiration.  (Isaac points out that if the keytab entry name isn't the canonical client name, canonicalization might make the ccache client principal not match, making it impossible for GSS to verify that a refresh is safe.)

* More serious misconfigurations have been observed.  When this came up in a 2014 email thread ( http://mailman.mit.edu/pipermail/kerberos/2014-July/020041.html ) the client keytab was being used with the default ccache for (I think) root, and sysadmins would sometimes kinit as various different principals.  It's harder to behave well in this case.  Refreshing the ccache changes the client identity, which could be surprising, but there's a good chance the application was already behaving surprisingly by authenticating as the wrong identity.

* It probably doesn't help that we don't seem to have user or admin documentation on using client keytabs.

frozencemetery:
Updated.  Added another commit on top which would address the duplication of `krb5_timeofday()` as an alternate approach.

greghudson:
Rebased and adjusted.  While adding test cases, I noted that the refresh from the client keytab is attempted using the name of the existing cred, which should prevent some surprises (except with KEYRING where the expired ccache vanishes).  I took that into account in the test cases.

PR #1047, opened 2020-03-09
Match Heimdal behavior for channel-bindings

iboukris:
This is a subset of PR #685, following recent discussions on krbdev.

### DIFF
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -430,6 +430,32 @@ kg_process_extension(krb5_context context,
 /* A zero-value channel binding, for comparison */
 static const uint8_t null_cb[CB_MD5_LEN];
 
+/* Look for AP_OPTIONS in authdata.  If present and the options include
+ * KERB_AP_OPTIONS_CBT, set *cbt_out to true. */
+static krb5_error_code
+check_cbt(krb5_context context, krb5_authdata **authdata,
+          krb5_boolean *cbt_out)
+{
+    krb5_error_code code;
+    uint32_t ad_ap_options;
+    const uint32_t KERB_AP_OPTIONS_CBT = 0x4000;
+
+    *cbt_out = FALSE;
+
+    code = krb5_find_authdata(context, NULL, authdata,
+                              KRB5_AUTHDATA_AP_OPTIONS, &authdata);
+    if (code || authdata == NULL)
+        return code;
+    if (authdata[1] != NULL || authdata[0]->length != 4)
+        return KRB5KRB_AP_ERR_MSG_TYPE;
+
+    ad_ap_options = load_32_le(authdata[0]->contents);
+    if (ad_ap_options & KERB_AP_OPTIONS_CBT)
+        *cbt_out = TRUE;
+
+    return 0;
+}
+
 /*
  * The krb5 GSS mech appropriates the authenticator checksum field from RFC
  * 4120 to store structured data instead of a checksum, indicated with checksum
@@ -454,7 +480,7 @@ process_checksum(OM_uint32 *minor_status, krb5_context context,
     krb5_error_code code = 0;
     OM_uint32 status, option_id, token_flags;
     size_t cb_len, option_len;
-    krb5_boolean valid, token_cb_present = FALSE, cb_match = FALSE;
+    krb5_boolean valid, client_cbt, token_cb_present = FALSE, cb_match = FALSE;
     krb5_key subkey;
     krb5_data option, empty = empty_data();
     krb5_checksum cb_cksum;
@@ -580,6 +606,23 @@ process_checksum(OM_uint32 *minor_status, krb5_context context,
                 goto fail;
             }
         }
+    }
+
+    /*
+     * If the client asserts the KERB_AP_OPTIONS_CBT flag (from MS-KILE) in the
+     * authenticator authdata, and the acceptor passed channel bindings,
+     * require matching channel bindings from the client.  The intent is to
+     * prevent an authenticator generated for use outside of a TLS channel from
+     * being used inside of one.
+     */
+    code = check_cbt(context, authenticator->authorization_data, &client_cbt);
+    if (code) {
+        status = GSS_S_FAILURE;
+        goto fail;
+    }
+    if (client_cbt && acceptor_cb != GSS_C_NO_CHANNEL_BINDINGS && !cb_match) {
+        status = GSS_S_BAD_BINDINGS;
+        goto fail;
     }
 
     status = GSS_S_COMPLETE;
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -427,6 +427,9 @@ kg_process_extension(krb5_context context,
                          GSS_C_SEQUENCE_FLAG | GSS_C_DCE_STYLE |        \
                          GSS_C_IDENTIFY_FLAG | GSS_C_EXTENDED_ERROR_FLAG)
 
+/* A zero-value channel binding, for comparison */
+static const uint8_t null_cb[CB_MD5_LEN];
+
 /*
  * The krb5 GSS mech appropriates the authenticator checksum field from RFC
  * 4120 to store structured data instead of a checksum, indicated with checksum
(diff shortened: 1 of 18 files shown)


greghudson:
I took a brief glance and didn't see any red flags.

Do we know what the Windows GSS server code does if the client sends no authenticator checksum (the `authdat->checksum == NULL` case) but also sends the authdata flag?

iboukris:
I haven't tested that case yet, it's in my todo list, I'll update.

src/lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -424,6 +424,25 @@ kg_process_extension(krb5_context context,
 /* A zero-value channel binding, for comparison */
 static const uint8_t null_cb[CB_MD5_LEN];
 
+static krb5_error_code
+check_ap_options_cbt(krb5_context context, krb5_authdata **authdata,
+                     krb5_boolean *client_asserted_cd)
```
simo5: shouldn't this be client_asserted_cb ?
(for channel bindings I assume ?)

src/lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -424,6 +424,25 @@ kg_process_extension(krb5_context context,
...
+{
+    uint32_t ad_ap_options;
+
+    *client_asserted_cd = FALSE;
+
+    if (authdata[1] != NULL || authdata[0]->length != sizeof(uint32_t))
+        return KRB5KRB_AP_ERR_MSG_TYPE;
+
+    memcpy(&ad_ap_options, authdata[0]->contents, sizeof(uint32_t));
```
simo5: shouldn't this be converted via endian aware macros ?

src/lib/krb5/krb/mk_req_ext.c
```diff
@@ -253,6 +253,34 @@ krb5_mk_req_extended(krb5_context context, krb5_auth_context *auth_context,
...
+static krb5_error_code
+add_ad_ap_options(krb5_context context, krb5_authdata **out)
+{
+    krb5_error_code code;
+    krb5_authdata ad, *list[2], **ifrel;
+
+    uint32_t ad_ap_options = KRB5_AP_OPTIONS_CBT;
+
+    ad.magic = KV5M_AUTHDATA;
+    ad.ad_type = KRB5_AUTHDATA_AP_OPTIONS;
```
simo5: same endianess question here.

src/lib/gssapi/krb5/accept_sec_context.c
iboukris: I guess it should, i'll look for examples.

src/lib/gssapi/krb5/accept_sec_context.c
iboukris: Yeah typo, nice catch :)

src/lib/gssapi/krb5/accept_sec_context.c
greghudson: You want load_32_le() from k5-platform.h, and store_32_le() for the reverse operation.

iboukris:
I noticed in the email i've sent to krbdev about Windows HTTPS client, that it sends KRB5_AUTHDATA_ETYPE_NEGOTIATION in the same ad-if-relevant container with KRB5_AP_OPTIONS_CBT, while in my code it would be separate.  I think this should not be a problem (the PAC is likely an exception), but I want to test that windows understand it this way.

iboukris:
Nope, with two ad-if-relevant elements it seems windows just ignores the second one.  It only applies the flag if it is in the first container.  I'll combine them both somehow.

src/lib/krb5/krb/mk_req_ext.c
```diff
@@ -253,6 +253,35 @@ krb5_mk_req_extended(krb5_context context, krb5_auth_context *auth_context,
     return retval;
 }
 
+static krb5_error_code
+add_ad_ap_options(krb5_context context, krb5_authdata **out)
+{
+    krb5_error_code code;
+    krb5_authdata ad, *list[2], **ifrel;
+    uint32_t ad_ap_options;
```
greghudson: This should be `uint8_t ad_ap_options[4];'.

simo5: Given store_32_le is used on this variable, wouldn't it potentially cause alignment problems on some platform if it were declared as an array of bytes ?

greghudson: store_32_le() writes an integer value (the first argument) into four bytes (the second argument).  The pointer to the bytes is not required to be aligned.

simo5: ah, good to know, I thought it might be using optimized instructions that do the swap, those tend to require alignment, so nevermind

iboukris:
I added some spnego changes, not needed for samba I think, but tested against Windows server.

In our tests, it looks like when I pass "mech_spengo" it goes via negoex path, I wonder how can I test the other path.

iboukris:
I checked in Heimdal as well, and the bindings are being passed in spnego to the sub mechanism.

greghudson:
> In our tests, it looks like when I pass "mech_spengo" it goes via negoex path, I wonder how can I test the other path. Any idea?

Which tests?  negoex should only apply to mechanisms which provide gssspi_query_mechanism_info().  In our tree that's only the negoextest mech.

iboukris:
In the tests I'm adding, in this commit: https://github.com/krb5/krb5/pull/1047/commits/9642e16ac173858d530cbc4df6f812f8202e2838

When I debug the server, it always enters this if (negoex): https://github.com/krb5/krb5/blob/02b3171380d5683212ae1deae67a8c6cd19e290a/src/lib/gssapi/spnego/spnego_mech.c#L1571

greghudson:
> When I debug the server, it always enters this if (negoex):

I checked out the PR branch and ran the first few t_channel_bindings invocations from t_gssapi.py under gdb.  For me it always took the false branch of that if statement, and didn't make any calls to negoex_accept().

iboukris:
Right, I can't reproduce it, I must have been wrong, sorry about that.

I'll try to add channel-bindings tests to negoextest.

iboukris:
Added a test for the negoex path, looks ready for review.

greghudson:
Does anything stop the initiator from including GSS_C_CHANNEL_BOUND_FLAG in the on-the-wire flags and getting it included in the flags seen by the acceptor, without providing channel bindings?

iboukris:
I added a fix to clear GSS_C_CHANNEL_BOUND_FLAG if the channel was not bound.

src/man/krb5.conf.man
```diff
@@ -487,6 +487,11 @@ attempt fails.
 If this flag is true, then an attempt to verify initial
 credentials will fail if the client machine does not have a
 keytab.  The default value is false.
+.TP
+\fBclient_aware_channel_bindings\fP
```
frozencemetery, line 491: Man pages are generated from rst files in doc (this one is doc/admin/conf_files/krb5_conf.rst).  It's sufficient to modify the rst file.

src/tests/gssapi/t_channel_bindings.c
```diff
@@ -0,0 +1,113 @@
+/* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+/*
+ * Copyright 2020 by the Massachusetts Institute of Technology.
```
frozencemetery, line 3: Is this the right copyright?  It's not what we (Red Hat) use for this project.

greghudson: That's the old MIT copyright, probably copied from another source file.  We now use the one in src/prototype/prototype.c.

greghudson:
I pushed a revision of the first two commits, pushing more of the flag handling responsibility into process_checksum() so that the contract is more coherent.  (The other direction would have been to make process_checksum() return the token flags unmodified, and communicate the deleg_cred and channel_bound flags separately.  But that would have added to the already large list of parameters.)

greghudson:
I refactored the third commit a bit, primarily to minimize the code size increase to process_checksum().

greghudson:
I refactored the fourth commit to do more stack allocation, and to move some groups of related decisions within the same function (like whether to send an etype list and how long it is, for instance).

iboukris:
Thanks for the refactoring, looks good to me and works as expected against Windows.

greghudson:
I made a pass over the tests.  The biggest change is that the test program now outputs the channel-bound state instead of optionally erroring out if it isn't 0, allowing us to reduce the number of test cases.  I made all of the tests check the output of the test program, and split them out into a separate Python script.

src/tests/gssapi/t_bindings.c
```diff
@@ -0,0 +1,110 @@
...
+ * Establish contexts (without and with GSS_C_DCE_STYLE) with the default
+ * initiator name, a specified principal name as target name, initiator
+ * bindings, and acceptor bindings.  If any call is unsuccessful, display an
+ * error message.  Output a pair of integers "0" or "1" indicating whether the
+ * contexts were reported as channel-bound on the acceptor.  Exit with status 0
+ * if all operations are successful, or 1 if not.
+ *
+ * Usage: ./t_bindings [-s] targetname icb acb
+ *
+ * An icb or abc value of "none" will not specify channel bindings.
```
iboukris, line 47: Should say: value of "-"

iboukris:
Perhaps the test could output "bound" or "unbound" instead of "00/11" (and require the dce case to be the same), looks good otherwise.

greghudson:
I went with "yes" or "no" since it helps keep test invocations on one line, but checking for consistency within the harness is a good idea.

PR #1048, opened 2020-03-10
Employ safe coding practices also in src/util/et/error_message.c

yurashek:
Samba dumps core on SIGSEGV in atexit() procedure. Problem was identified the et_list is not NULLed after free the list so its value points to freed memory and SIGSEGV is raised in remove_error_table() . also the "et" removal is unsafe and should fail if "et" is pointed by the first member in the list.

### DIFF
--- a/src/util/et/error_message.c
+++ b/src/util/et/error_message.c
@@ -26,7 +26,7 @@
 
 static struct et_list *et_list;
 static k5_mutex_t et_list_lock = K5_MUTEX_PARTIAL_INITIALIZER;
-static int terminated = 0;      /* for debugging shlib fini sequence errors */
+static int terminated = 0;      /* for safety and finalization debugging */
 
 MAKE_INIT_FUNCTION(com_err_initialize);
 MAKE_FINI_FUNCTION(com_err_terminate);
@@ -69,6 +69,7 @@ void com_err_terminate(void)
         enext = e->next;
         free(e);
     }
+    et_list = NULL;
     k5_mutex_unlock(&et_list_lock);
     k5_mutex_destroy(&et_list_lock);
     terminated = 1;
@@ -279,6 +280,10 @@ errcode_t KRB5_CALLCONV
 remove_error_table(const struct error_table *et)
 {
     struct et_list **ep, *e;
+
+    /* Safety check in case libraries are finalized in the wrong order. */
+    if (terminated)
+        return ENOENT;
 
     if (CALL_INIT_FUNCTION(com_err_initialize))
         return 0;


src/util/et/error_message.c
```diff
@@ -278,24 +278,32 @@ add_error_table(const struct error_table *et)
...
     /* Remove the entry that matches the error table instance. */
-    for (ep = &et_list; *ep; ep = &(*ep)->next) {
-        if ((*ep)->table == et) {
-            e = *ep;
-            *ep = e->next;
-            free(e);
-            k5_mutex_unlock(&et_list_lock);
-            return 0;
-        }
+    for (epp = NULL, ep = et_list; ep && ep->table != et; epp = ep, ep = ep->next) ;
```
frozencemetery: Style here is:
- check values against NULL explicitly
- limit lines to 79 or fewer characters
- no space before semicolon.

src/util/et/error_message.c
```diff
@@ -278,24 +278,32 @@ add_error_table(const struct error_table *et)
...
+    if (!ep) {
```
frozencemetery: Explicit NULL here too.

greghudson:
> also the "et" removal is unsafe and should fail if "et" is pointed by the first member in the list.

I don't think I agree; can you elaborate on the reasoning?

src/util/et/error_message.c
```diff
@@ -69,9 +69,9 @@ void com_err_terminate(void)
         enext = e->next;
         free(e);
     }
+    et_list = NULL; terminated = 1;
```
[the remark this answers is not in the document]
yurashek: Library calls should be coded with respect to safe coding practices so static pointers should be NULLed to be able to test it for NULL to avoid it to point on already freed object.
Also the critical section should be mutexed all not only the part of it.

src/util/et/error_message.c
yurashek: similar as above

yurashek:
remove_error_table() is called from atexit() handler
... 
fffffbefb82cfa54 sigacthandler (b, 0, fffffc771015e170, 0, fffffbefb49044b8, fffffbefb8424000) + 54
 --- called from signal handler with signal 11 (SIGSEGV) ---
 fffffbefb4802334 remove_error_table (fffffbefb4d410b0, fffffbefb4904500, 0, 0, 10, c01365dda0) + a4
 fffffbefb4b89704 krb5int_lib_fini (1, 6d28, 6c00, fffffbefb4d38000, fffffbefb842c900, 0) + 74
 fffffbefb85253e8 call_array (fffffbefb4d3ed18, 0, fffffbefb54e2e48, f, 0, fffffbefb8744448) + c0
 fffffbefb85257a4 call_fini (fffffbefb8744160, fffffbefa21ee6e0, 0, fffffbefb54e2e48, 0, fffffbefb8744e58) + dc
 fffffbefb8525990 atexit_fini (fffffbefb87442b0, fffffffffffff458, fffffbefb8744178, fffffbefb8744e58, fffffbefb87442fc, fffffbefb8744160) + 98
 fffffbefb81f9440 _exithandle (2880, fffffbefb842c900, 2800, ea15fa71, c013712d70, b) + 60
 fffffbefb81e7bcc exit (0, fffffbefab5c04d0, c0136de800, c013713020, fffffbefb4100b50, fffffbefb4100b58) + 4
 fffffbefb76e8688 exit_server_common (0, fffffbefab5c04d0, c0136e8f20, fffffbefb7900150, 0, fffffbefb7900158) + 364
 fffffbefb76e8a80 smbd_exit_server_cleanly (fffffbefab5c04d0, 0, 0, 0, 0, 0) + 8
 fffffbefb1d01390 exit_server_cleanly (fffffbefab5c04d0, c01367c0e8, 1, c00000b5, 1, 0) + 28
 fffffbefb76be0e8 smbd_server_connection_terminate_ex (c01367c020, fffffbefab5c04d0, fffffbefb7526848, 0, 0, c0136e8f20) + 114
 fffffbefb76c1b58 smbd_smb2_connection_handler (1, c01367c0d8, ffffffff, fffffc771015f4f7, c01367c0e8, c01367c020) + 1c8
...

greghudson:
What platform was this error observed on?

A correctly functioning platform should finalize libkrb5 before libcom_err, so that no calls into libcom_err are made after the libcom_err finalizer runs.  The finalizer should also not be contending with calls into libcom_err in other threads, so the existing locks around the for loop are probably unnecessary (though to be conservative, I have no intention of removing them).

If we can determine that there is a finalization order issue on a non-niche platform, it's cheap to add an `et_list = NULL;` after the for loop, but I can't currently see a need for any of the other code changes.

yurashek:
remove_error_table() is called from atexit() when "terminated" is already set to 1. It happens in Samba occasionally. I understand Samba is complicated FSM so it can happen there the atexit() callback is not disarmed when com_err_terminate() call is already finished. I also understand this is misuse of the code but even though I expect the good (safe) code will return error but poor code will trap what is the case of remove_error_table() here:

287      /* Remove the entry that matches the error table instance. */
288      for (ep = &et_list; *ep; ep = &(*ep)->next) {
289          if ((*ep)->table == et) {
290              e = *ep;
291              *ep = e->next;
                                ^ just here
292              free(e);

reason is the strong OS like Solaris rewrites freed memory by 0xdeadbeef's so the "next" member is not pointing to (already also freed ;-) ) next member of the list as it was before the memory was freed. Weak OS's [Linux, Windows,...] is not doing it so the code will not trap here but...

core is from SPARC (I prefer SPARC debugging because it is much more easier) but i86pc is doing the same.

Set et_list = NULL only is not enough because the code in remove_error_table() is unsafe and:
ep = &et_list; then (*ep)->table  or (*ep)->next will cause NULL pointer redirection in this case.

My proposal is to change poor code to safe code. Nothing more.

greghudson:
> remove_error_table() is called from atexit() when "terminated" is already set to 1. It happens in Samba occasionally. I understand Samba is complicated FSM so it can happen there the atexit() callback is not disarmed when com_err_terminate() call is already finished.

Samba does not directly call remove_error_table(), and I don't think Samba's own atexit handler is at issue.  The errant call to remove_error_table() is coming from the libkrb5 finalizer, which in turn comes from a Solaris rtld atexit handler (atexit_fini()).  If the platform is behaving properly, the libkrb5 finalizer should run before the libcom_err finalizer.

I would ordinarily expect the Solaris dynamic linker to finalizer libraries in the proper order, so I still can't fully account for the crash.

> Set et_list = NULL only is not enough because the code in remove_error_table() is unsafe and:
> ep = &et_list; then (*ep)->table or (*ep)->next will cause NULL pointer redirection in this case.

The for loop header is:

    for (ep = &et_list; *ep; ep = &(*ep)->next) {

If et_list is NULL, then *ep is false before the first iteration, and the loop will not run.

> reason is the strong OS like Solaris rewrites freed memory by 0xdeadbeef's so the "next" member is not pointing to (already also freed ;-) ) next member of the list as it was before the memory was freed. Weak OS's [Linux, Windows,...] is not doing it so the code will not trap here but...

Linux platforms would absolutely crash in remove_error_table() if it ran after the library finalizer.

src/util/et/error_message.c
frozencemetery: I'm telling you a style practice - in krb5 we check values against NULL explicitly.  See https://k5wiki.kerberos.org/wiki/Coding_style/Practices

src/util/et/error_message.c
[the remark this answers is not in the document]
frozencemetery: My point here and in the next comment is that `terminated` is not considered part of the critical section.  (It's a debugging aid, as the comment at its definition indicates.)  Your change on line 284 is therefore unsafe, as it happens unsynchronized.

Also, statements still each get their own line.

src/util/et/error_message.c
```diff
@@ -278,24 +278,32 @@ add_error_table(const struct error_table *et)
 errcode_t KRB5_CALLCONV
 remove_error_table(const struct error_table *et)
 {
-    struct et_list **ep, *e;
+    struct et_list *epp, *ep;
 
+    /* be safe */
+    if (et_list == NULL || terminated)
```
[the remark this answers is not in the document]
yurashek: Setting of "terminated" is now moved into the mutexed section so it is "fully synchronized" with destroying of the et_list_lock mutex.
Re-initialization is not a good idea.
"et_list == NULL" should be removed from here to enable of call of init function if et_list is NULL but "terminated" is not set (initial state).

> > also the "et" removal is unsafe and should fail if "et" is pointed by the first member in the list.
> 
> I don't think I agree; can you elaborate on the reasoning?

It was my mistake.

yurashek:
I adopted greghudson's comments and re-uploaded the changeset.

src/util/et/error_message.c
yurashek: code test for null is unchanged against the original code

src/util/et/error_message.c
yurashek: See the original code.
note: I returned it back by greghudson's comment

greghudson:
It may actually be necessary to test terminated if we want to prevent a crash when remove_error_table() is improperly called after finalization, as locking a destroyed mutex can trigger an assertion failure.

(And of course we can't synchronize the check, as the whole goal is to detect when the mutex is destroyed.  So this would only help when the final accesses to libcom_err are within a single thread--which I would hope would be normal during process termination or library unloading.)

I don't think it's meaningful to move the setting of terminated into the critical section.  That critical section was added to shut up helgrind (see https://krbdev.mit.edu/rt/Ticket/Display.html?id=2878 ).  It can't possibly be solving an actual runtime issue because the mutex is destroyed immediately afterwards, so any concurrent thread locking et_list_lock at the same time would run the risk of having its lock operation happen at the same time as the mutex is destroyed.

yurashek:
> 
> I would ordinarily expect the Solaris dynamic linker to finalize libraries in the proper order, so I still can't fully account for the crash.
> 
Ali our Solaris RT linker "guru" sent me the link to old Rod's blog:
http://www.linker-aliens.org/blogs/rie/entry/init_and_fini_processing_who/

PR #1056, opened 2020-03-24
Call warn_pw_expiry() from krb5_init_creds_get_creds()

sumit-bose:
Currently warn_pw_expiry() is called from krb5_get_init_creds_password()
and as a result only if this call is used the expiration callback is
called.

By moving the warn_pw_expiry() to krb5_init_creds_get_creds() other
way of requesting the initial credentials, e.g. the step-by-step
approach with krb5_init_creds_context, can benefit from the expiration
callback, too.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -7172,11 +7172,10 @@ typedef void
  *
  * Set a callback to receive password and account expiration times.
  *
- * This option only applies to krb5_get_init_creds_password().  @a cb will be
- * invoked if and only if credentials are successfully acquired.  The callback
- * will receive the @a context from the krb5_get_init_creds_password() call and
- * the @a data argument supplied with this API.  The remaining arguments should
- * be interpreted as follows:
+ * @a cb will be invoked if and only if credentials are successfully acquired.
+ * The callback will receive the @a context from the calling function and the
+ * @a data argument supplied with this API.  The remaining arguments should be
+ * interpreted as follows:
  *
  * If @a is_last_req is true, then the KDC reply contained last-req entries
  * which unambiguously indicated the password expiration, account expiration,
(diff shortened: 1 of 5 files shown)


greghudson:
krb5_init_creds_get_creds() is just an accessor; calling the prompter or callback from it is really weird behavior.  I think it should happen from krb5_init_creds_step().

sumit-bose:
> krb5_init_creds_get_creds() is just an accessor; calling the prompter or callback from it is really weird behavior. I think it should happen from krb5_init_creds_step().
Hi Greg,

thanks for the comment. I will the call to `krb5_init_creds_step()` and send an updated version.

> 
> It's also reasonable, even encouraged, for a caller to use the stepwise init_creds interface without calling krb5_init_creds_get_creds(), instead using krb5_get_init_creds_opt_set_out_ccache(). sssd should be doing that, actually; right now when it acquires creds it's not storing any associated ccache config settings (including FAST negotiation and the pa_type and pa_config_data entries Nalin added back in 2012).

Thanks for the hint, I will create a ticket for SSSD to change this.

bye,
Sumit

greghudson:
I added tests.  (The first force-push is just a rebase after merging other PRs.)

greghudson:
I made some changes:

* Call warn_pw_expire() from init_creds_step_reply(), unconditionally where we set ctx->complete. 
 (This necessitated moving the code earlier in the file, which is the bulk of the diff.)  This is simpler than putting it in the main logic of init_creds_step().  We do set ctx->complete in two other places, but those are both for realm identification, which is a private interface internal to S4U2Self.

* There is no need to enumerate which init_creds functions krb5_get_init_creds_opt_set_expire_callback() applies to if it applies to all of them, so I took that sentence out.

* I rewrote the commit message.

sumit-bose:
Thanks, I'm fine with the changes, SSSD still gets the expected password expiration data.

PR #1057, opened 2020-03-30
Correctly import "service@" GSS host-based name

greghudson:
[As discussed in the krbdev thread, we may also want to make changes to krb5_sname_to_principal().  But this part is pretty clearly an improvement.]

The intended way to specify only a service in a GSS host-based name is
to omit the "@" separator.  Some applications include the separator
but no hostname, and this happened to yield wildcard hostname behavior
prior to commit 996353767fe8afa7f67a3b5b465e4d70e18bad7c when
shortname qualification was added.  To restore this behavior, check in
parse_hostbased() that at least one character is present after the "@"
separator before copying the hostname.  Add a test case to t_gssapi.py.

### DIFF
--- a/src/lib/gssapi/krb5/import_name.c
+++ b/src/lib/gssapi/krb5/import_name.c
@@ -102,8 +102,8 @@ parse_hostbased(const char *str, size_t len,
     memcpy(service, str, servicelen);
     service[servicelen] = '\0';
 
-    /* If present, copy the hostname. */
-    if (at != NULL) {
+    /* Copy the hostname if present (at least one character after '@'). */
+    if (len - servicelen > 1) {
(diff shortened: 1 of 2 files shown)


greghudson:
I added a second commit changing how krb5_sname_to_principal() behaves on an empty hostname, but then decided to remove it for now.  Any such change would be tied up with ticket 8809 ( https://krbdev.mit.edu/rt/Ticket/Display.html?id=8809 ) which requests that we validate hostnames before applying DNS canonicalization.  I may need to revisit whether that's a valid use case (why is KRB5_NT_SRV_HST or GSS_C_NT_HOSTBASED_SERVICE being used with a non-hostname?) but I don't want to make a summary judgment on that right now.

PR #1060, opened 2020-04-03
gsstest from SAP reports too large lifetime skew between acceptor and initiator ctx 

Sashan:
this fixes the [gsstest](https://archive.sap.com/documents/docs/DOC-14051) from SAP  to complain on 5 minute difference between initiator and acceptor security contexts lifetime (acceptor lifetime is 5 minutes greater).

change also fixes tests, which complain on mismatch between GSS context lifetime and credential lifetime.

I'm still not sure if it is a hack or right approach. I did poke around and it looks MIT kerberos is the only gssapi implementation, which gives acceptor context lifetime exta `clockskew` seconds. I understand the [change](https://github.com/krb5/krb5/commit/b496ce4095133536e0ace36b74130e4b9ecb5e11) has been introduced to fix issues with kerberized NFS. I don't understand why `gsstest` requires acceptor and initiator lifetime close to each other (according to source code to skew must not exceed 20 secs).

We were notified about the issue by one of our customers, who happens to be fighting with upgrade from S11.3 to S11.4. The kerberos upgrade from 1.8.4 to 1.16.1 is part of that game. The addition of `clockskew` arrived to kerberos 1.14.

Also the fix in this pull request does not get kerberized SAP to run. There seems to be yet another issue around integrity/message sequence check. As I'm still learning the kerberos stuff, so I would rather ask questions:
   - is `gsstest` requirement to keep acceptor and initiator context lifetime close to each other legitimate? I've briefly poked to RFCs but could not find anything on this?

   - does change in this pull request make sense? or are there some undesired implications?

thanks a lot
sashan

### DIFF
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -352,8 +352,8 @@ kg_accept_dce(minor_status, context_handle, verifier_cred_handle,
         *mech_type = ctx->mech_used;
 
     if (time_rec) {
-        *time_rec = ts_delta(ctx->krb_times.endtime, now) +
-            ctx->k5_context->clockskew;
+	    *time_rec = ts_delta(ctx->krb_times.endtime, now) +
+		ctx->k5_context->clockskew;
     }
 
     /* Never return GSS_C_DELEG_FLAG since we don't support DCE credential
@@ -1133,7 +1133,7 @@ kg_accept_krb5(minor_status, context_handle,
     /* Add the maximum allowable clock skew as a grace period for context
      * expiration, just as we do for the ticket. */
     if (time_rec)
-        *time_rec = ts_delta(ctx->krb_times.endtime, now) + context->clockskew;
+	    *time_rec = ts_delta(ctx->krb_times.endtime, now) + context->clockskew;
 
     if (ret_flags)
         *ret_flags = ctx->gss_flags;
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -591,7 +591,7 @@ kg_cred_set_initial_refresh(krb5_context context, krb5_gss_cred_id_rec *cred,
 
     /* Make a note to refresh these when they are halfway to expired. */
     refresh = ts_incr(times->starttime,
-                      ts_delta(times->endtime, times->starttime) / 2);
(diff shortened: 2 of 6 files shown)


frozencemetery:
I had also gotten this complaint from gsstest.  However, my opinion is that gsstest is just being fussy - there's no reason to change the krb5 behavior here, and it doesn't break anything in SAP to my knowledge.

frozencemetery:
Sorry, hit send a bit too early.  Wanted to add that this comes from commit https://github.com/krb5/krb5/commit/b496ce4095133536e0ace36b74130e4b9ecb5e11 and we've been shipping it this way since 2016.

gsstest is really fussy about some other things - it's best to take what it spits out with a grain of salt.

greghudson:
While looking at the changes I noticed that commit b496ce4095133536e0ace36b74130e4b9ecb5e11 didn't hit krb5_gss_inquire_cred(), and we'll want to address that.

This scenario comes up periodically: someone has an interop issue between SAP and the MIT krb5 GSS libraries; SAP says they don't support that but here's a gsstest program you can run; the gsstest program finds something fussy to complain about in our GSS libraries which has nothing to do with the interoperability issue, but the user (understandably) focuses on the gsstest output.

I don't see anything in RFC 2743 to require that the initiator and acceptor return the same lifetime.  We want the acceptor to be more lax about the lifetime (and therefore to report a larger lifetime value) because of potential clock skew between the initiator and acceptor hosts.

I reviewed the most recent SAP integration issue we had at MIT, but it doesn't seem applicable.  (It was an issue upgrading from 1.6 to 1.12, and the fix was commit 8cbfec302659812efee9b67031f5ae0c5400953f, which went into release 1.15; since you are upgrading to version 1.16.x you would have that fix.  It also had nothing to do with sequence checks.)

Sashan:
Thank you for a quick reply. It indeed looks like those failures are red herrings. The logs from SAPware also contain messages, which match `GSS_S_GAP_TOKEN` error. This seems to be a problem with message ordering/sequence numbers. Given customer is essentially upgrading from kerberos 1.8.4 to 1.16.1 it can be caused by change sequence number [validation](https://github.com/krb5/krb5/commit/cb3db58b1942998a5e2c4d46c21ca0554e769a8cit), which went to 1.13 (I think). I'm preparing some debug bits in a hope to gather some more info about this. 

I'm going to close this pull request. Will eventually open another one, if something useful will be found.

Thank you for your help.

PR #1067, opened 2020-05-06
Omit KDC indicator check for S4U2Self requests

greghudson:
[Prompted by http://mailman.mit.edu/pipermail/krbdev/2020-May/013261.html ]

As there was no initial ticket exchange from the client for an
S4U2Self request, the auth indicator check is inapplicable (and would
always fail if any auth indicators are required).


abbra:
LGTM. Thank you for the prompt fix!

PR #1074, opened 2020-05-22
Return GSS_S_NO_CRED from krb5 gss_acquire_cred

greghudson:
Earlier versions of the GSS-API spec (RFCs 1508 and 2078) do not list
GSS_S_NO_CRED as a valid error code for gss_acquire_cred.  As a
result, the OpenVision developers of the GSSAPI krb5 mech created
GSS_S_CRED_UNAVAIL as an alias for GSS_S_FAILURE and returned it when
no valid credentials could be obtained.  RFC 2743 lists GSS_S_NO_CRED
as the proper return code when matching credentials cannot be
accessed.  Change the krb5 gss_acquire_cred() implementation to return
GSS_S_NO_CRED where it currently returns GSS_S_CRED_UNAVAIL.

Also stop using GSS_S_CRED_UNAVAIL in the krb5 gss_store_cred(), but
change it to explicitly use GSS_S_FAILURE instead.  RFC 5588 specifies
GSS_S_NO_CRED as indicating a problem with input_cred_handle, not the
receiving store, so GSS_S_NO_CRED would be inappropriate.

### DIFF
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -211,7 +211,7 @@ acquire_accept_cred(krb5_context context, OM_uint32 *minor_status,
         }
     }
     if (code) {
-        major = GSS_S_CRED_UNAVAIL;
+        major = GSS_S_NO_CRED;
         goto cleanup;
     }
 
@@ -223,7 +223,7 @@ acquire_accept_cred(krb5_context context, OM_uint32 *minor_status,
                 k5_change_error_message_code(context, code, KG_KEYTAB_NOMATCH);
                 code = KG_KEYTAB_NOMATCH;
             }
-            major = GSS_S_CRED_UNAVAIL;
+            major = GSS_S_NO_CRED;
             goto cleanup;
         }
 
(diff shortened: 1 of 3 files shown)


frozencemetery:
What's the reason behind the `gss_store_cred()` change?  5588 says that "GSS_S_UNAVAILBLE indicates that the credential store is not available", which seems appropriate, while GSS_S_FAILURE is "could not be stored for some other reason".

greghudson:
> What's the reason behind the gss_store_cred() change?

I wanted to stop using GSS_S_CRED_UNAVAIL, since it's not a standard constant and the value is just the same as GSS_S_FAILURE.

I can see the argument for GSS_S_UNAVAILABLE (which is distinct from GSS_S_CRED_UNAVAIL and is actually a standard constant) instead, but I'm not sure what "the credential store is not available" actually means.  krb5_cc_resolve() or krb5_cc_initialize() could fail for any number of reasons, including ENOMEM or ENOSPC.  Heimdal never returns GSS_S_FAILURE from gss_store_cred(), so we wouldn't be aligning with them.

frozencemetery:
That makes sense, thanks.

PR #1076, opened 2020-05-27
Default dns_canonicalize_hostname to "fallback"

greghudson:
This change should mitigate some of the pain caused by the rdns=true
default (generally associated with unwanted PTR records that cannot
easily be changed), with a minimum of fallout.

Update the documentation and tests accordingly.


greghudson:
I fixed the test suite issues after several tries; this should be ready.

PR #1078, opened 2020-06-04
WIP: Support tgt-less creds via delegation policy

iboukris:
This could be useful for local auth (such as sudo), for S4U2Self and for S4U2Proxy use cases, the latter when used alongside a privileged daemon proxy with access to impersonator TGT.

Following the mail thread on krbdev:
http://mailman.mit.edu/pipermail/krbdev/2020-June/013277.html
Discussion summary:
http://mailman.mit.edu/pipermail/krbdev/2020-June/013278.html

### DIFF
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -114,19 +114,24 @@
 #ifndef LEAN_CLIENT
 
 static OM_uint32
-create_constrained_deleg_creds(OM_uint32 *minor_status,
+create_constrained_deleg_creds(OM_uint32 *minor_status, krb5_context context,
                                krb5_gss_cred_id_t verifier_cred_handle,
-                               krb5_ticket *ticket,
-                               krb5_gss_cred_id_t *out_cred,
-                               krb5_context context)
+                               krb5_ticket *ticket, delegation_types type,
+                               krb5_gss_cred_id_t *out_cred)
 {
     OM_uint32 major_status;
     krb5_creds krb_creds;
     krb5_data *data;
     krb5_error_code code;
+    krb5_gss_cred_id_t impersonator_creds = verifier_cred_handle;
 
     assert(out_cred != NULL);
-    assert(verifier_cred_handle->usage == GSS_C_BOTH);
+    if (type == DELEG_PROXY_CREDS)
+        assert(verifier_cred_handle->usage == GSS_C_BOTH);
+    else {
+        assert(type == DELEG_CLIENT_TICKET);
+        impersonator_creds = NULL;
+    }
 
     memset(&krb_creds, 0, sizeof(krb_creds));
     krb_creds.client = ticket->enc_part2->client;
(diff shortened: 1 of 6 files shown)


greghudson:
On parse_delegation_policy():

* Why use strncmp() instead of just strcmp()?  It means we'll match "client-tgtwithtrailinggarbage" but I don't see how that's desirable.

* Don't we have to remember the order, instead of creating a bitmask?

simo5:
If I want maximum flexibility I need to set: client-ticket,client-tgt,proxy-creds as delegation_policy ?
Can you please provide a truth table with each combination of configuration and each valid combination of received credential and tell what you think should happen ?
Here is what my understanding is (question marks where I am unsure):

TKT= client-ticket, TGT = client-tgt, PRX = proxy-creds

Configuration in columns, Received data in rows, result in delegated_creds handle in the cell

```
    | TKT | TGT | PRX | TKT/TGT | TKT/PRX | TGT/PRX | TKT/TGT/PRX |
----|--------------------------------------------------------------
TKT | TKT | NaN | PRX |   TKT   |    ?    |   PRX   |     ?       |
TGT | NaN | TGT |  ?  |   TGT   |    ?    |   TGT   |    TGT      |
----|--------------------------------------------------------------
```

iboukris:
> If I want maximum flexibility I need to set: client-ticket,client-tgt,proxy-creds as delegation_policy ?

Using "client-ticket,client-tgt,proxy-creds" will just always give you "client-ticket" because it is always available, to get the 'best' creds you want "client-tgt,proxy-creds,client-ticket".

The table is incomplete as we also depend on GSS_C_BOTH, and the order matters.

simo5:
Alright, it was unclear to me that order mattered, and the GSS_C_BOTH also made a difference, can you please provide a table or a full explanation of each case? I think this is fundamental to make sure what is implement is what people agreed/expect out of this.

simo5:
I think we basically need all this documented in the man-page too, it is critical that admins understand what they are setting if we delegate the decision to them.

iboukris:
> Don't we have to remember the order, instead of creating a bitmask?

Simplified now.

src/lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -633,6 +638,81 @@ process_checksum(OM_uint32 *minor_status, krb5_context context,
...
+    *out_delegation_type = DELEG_NONE;
+
+    token = strtok_r(policy, ",", &save);
+    while (token != NULL) {
+        if (have_tgt && strcmp(token, "client-tgt") == 0) {
+            *out_delegation_type = DELEG_CLIENT_TGT;
+            break;;
+        }
+        if (strcmp(token, "proxy-creds") == 0 &&
+            (cred_usage == GSS_C_BOTH || cred_usage == GSS_C_INITIATE)) {
```
greghudson, line 660: Can cred_usage be GSS_C_INITIATE?  This is from the gss_accept_sec_context() acceptor_cred_handle, right?

iboukris: That's just a preparation to move parse_delegation_policy() to s4u_gss_glue.c  and to call it also from gss_acquire_cred_impersonate_name() in order to handle the cred_store option.

PR #1080, opened 2020-06-06
Omit PA_FOR_USER if we can't generate it due to hardcoded HMAC_MD5

iboukris:


greghudson:
I'm guessing this happens when using the OpenSSL crypto back end in FIPS mode?

GSS channel bindings are probably a bigger problem when we can't compute MD5 checksums.  We're long overdue for implementing RFC 6542, and even with that, sending the MD5 channel bindings is mandated by the spec.

greghudson:
See also:

https://bugzilla.redhat.com/show_bug.cgi?id=1838628
https://mailman.mit.edu/pipermail/krbdev/2020-June/013309.html

iboukris:
I thought channel-bindings were less of a problem because it is in encrypted envelop, so we can treat it as simple plain-text hash, while pa-for-user is sent in clear.

iboukris:
I wonder if Windows currently supports RFC 6542, I see several references to it as GSS-EXTS.

iboukris:
Didn't find an easy way to test it, leaving without.

iboukris:
> I'm guessing this happens when using the OpenSSL crypto back end in FIPS mode?

Yes, it fixes this failure:
```
$ kvno -I admin@ACME.COM apache@ACME.COM
kvno: Cryptosystem internal error while getting credentials for apache@ACME.COM

simo5:
just fyi: I think we can easily claim that in CB MD5 is not used as a protection mechanism, but just as way to compress the input into a quantity of a known size, so it is just a hash, not a secure hash that would fall under FIPS consideation.

PR #1082, opened 2020-06-09
Fix missing return code initialization in negoex_parse_token()

frozencemetery:
Discovered by clang.


frozencemetery:
(This is a problem if `in.status == 0`; major won't be initialized by line 487.)

greghudson:
in.status is always zero for the first iteration, but in.len could be zero if token->length is zero.  That's ruled out by the calling code (negoex_init() and negoex_accept() both reject zero-length tokens) but the compiler is right to flag this.

What version of clang flags the warning?  Presumably it's recent since CI with clang and -Werror has been passing.

frozencemetery:
I wish I knew; that team doesn't respond when I ask questions like that.  Probably the analyzer from 9?

PR #1083, opened 2020-06-09
Allow no rep_s4u_padata with newer enctypes

iboukris:


iboukris:
With the heimdal patch landed in samba, this can actually be useful:
https://gitlab.com/samba-team/samba/-/merge_requests/1390

greghudson:
From research:

* RFC 4120 and 4121 defines "newer" enctypes, which are basically anything from aes-sha1 onward.  These enctypes are sometimes defined by what they are not: DES, 3DES, and RC4 enctypes are not newer.

* The use of newer enctypes are used as ratchets for some protocol behavior.  RFC 4120 mandates client support for etype-info2 if the client implements a newer enctype.  RFC 4121 mandates the use of its new token types for newer enctypes.

* [MS-SFU] also defines behavior based on the use of newer enctypes:

  - Section 2.2.2 requires the KDC to send a copy of the PA_S4U_X509_USER checksum in encrypted padata if the enctype is not-newer (defined by reference to RFC 4121).

  - Section 3.1.5.1.1.1 says the KDC SHOULD send PA_S4U_X509_USER if the TGT session key is newer (defined as not DES or RC4 based).

* Luke's client code treats both of these as KDC requirements, enforced in verify_s4u2self_reply().

* Heimdal does not implement PA_S4U_X509_USER, but does of course implement newer enctypes.  Therefore our S4U2Self client code does not interop with Heimdal or Samba when it uses Heimdal, unless an older enctype is used.

I think it would be much better if Heimdal implemented PA_S4U_X509_USER.  But as this is only stated as a SHOULD in the relevant protocol spec, we should consider interoperating if it's safe.  It should be no less safe than interoperating with an older enctype, but I need to do some further research.

iboukris:
> Section 3.1.5.1.1.1 says the KDC SHOULD send PA_S4U_X509_USER if the TGT session key is newer (defined as not DES or RC4 based).

I think this section came to fix bug with checksum in the returned PA_S4U_X509_USER, but PA_FOR_USER never required any returned padata, and shouldn't been impacted by this. See: 
https://github.com/krb5/krb5/blob/42e21c57261f78cf0c68cbef16d58ca4a106f0e0/src/kdc/kdc_util.c#L1399

iboukris:
I quoted the wrong section, fixed now.

iboukris:
> I think this section came to fix bug with checksum in the returned PA_S4U_X509_USER

Actually no, the checksum in the request may be problematic as well, that's why they are both returned in the encrypted padata, but only for older enctypes per the logic in MS-SFU 2.2.2, see also: https://github.com/krb5/krb5/blob/03f122bdb22cfa53c7d855ed929c9541e56365e0/src/lib/krb5/krb/s4u_creds.c#L217

PR #1084, opened 2020-06-12
add refresh_time to ccache created on behalf of kinit

Sashan:
This PR rather hopes to raise awareness of issue we've seen on Solaris.
Things currently works as described in [ccache_file_format](https://web.mit.edu/kerberos/krb5-devel/doc/formats/ccache_file_format.html).
The issue comes from combination of two factors:

* system services/daemons use root user's credential cache (ccache is shared between user and system)
* kinit never adds a `refresh_time` configuration key to credential cache, although a renewal option is specified

We see the issue on system, where user accounts are provisioned from AD/LDAP
and SASL/GSSAPI is being used as authentication. Things work fine until root user
performs a `kdestroy` followed by `kinit...` to obtain krbtgt for its principal.
At this point the ccache, which is shared with system daemons loses `refresh_time`
configuration key, causing tickets to stop renewing. Once root's krb5tgt expires
the cache stays around, which makes system daemons (nscd/ldap client) to fail to
authenticate with AD/LDAP server. If there will be no cache, the daemons would
create a brand new. But as long as *poisoned* cache sits around, the SASL/GSSAPI
authentication for daemons is doomed to fail.

The patch in this pull request solves the issue for us. As I've said
I don't expect it will get accepted I rather hope to document the issue.
It took a great deal of time to figure out what was going on. Hence it's
perhaps worth to share our findings with others.

### DIFF
--- a/src/clients/kinit/kinit.c
+++ b/src/clients/kinit/kinit.c
@@ -660,6 +660,22 @@ kinit_prompter(krb5_context ctx, void *data, const char *name,
     return krb5_prompter_posix(ctx, data, name, banner, num_prompts, prompts);
 }
 
+/* Make a note in ccache that we should attempt to refresh it from the client
+ * keytab at refresh_time. (comes from GSSAPI acquire cred, modified to
+ * make it better fit for kinit) */
+static void
+set_refresh_time(struct k5_data *k5, krb5_timestamp refresh_time)
+{
+    char buf[128];
+    krb5_data d;
+
+    snprintf(buf, sizeof(buf), "%u", (unsigned int)ts2tt(refresh_time));
+    d = string2data(buf);
+    (void)krb5_cc_set_config(k5->ctx, k5->out_cc, NULL, KRB5_CC_CONF_REFRESH_TIME,
+                             &d);
+    krb5_clear_error_message(k5->ctx);
+}
+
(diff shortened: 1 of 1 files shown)


greghudson:
I think commit 729896467e3c77904666019d6cbbda583ae49b95 should solve the refresh_time issue.  However, sharing a ccache between the root user and services using a client keytab may still be inadvisable, as the root user may kinit as a different principal and cause services to authenticate as a different client than they usually do.

Sashan:
> I think commit [7298964](https://github.com/krb5/krb5/commit/729896467e3c77904666019d6cbbda583ae49b95) should solve the refresh_time issue. However, sharing a ccache between the root user and services using a client keytab may still be inadvisable, as the root user may kinit as a different principal and cause services to authenticate as a different client than they usually do.

thank you for a pointer to changeset. I gave it a try, but it does not work. However I have to look at it more closely. There is a Solaris specific change in `acquire_cred` area, which might be coming to play. I need to look at it more closely. Is always devil is in detail.

Sashan:
I'm going to close this request. It has turned out my test system was dirty. There was other change deployed, which prevented MIT changeset from doing a right thing. Everything works. Thank you very much for your help.

PR #1087, opened 2020-06-18
Add three kvno options from Heimdal kgetcred

greghudson:
Add the flags --cached-only and --no-store, which pass the
corresponding options to krb5_get_credentials().  Add the option
--out-cache to write the retrieved credentials to a specified output
cache.

Add a Python test script for kvno command-line options, including
tests for the new options.

### DIFF
--- a/doc/user/user_commands/kvno.rst
+++ b/doc/user/user_commands/kvno.rst
@@ -74,6 +74,19 @@ OPTIONS
     client principal with the X.509 certificate in *cert_file*.  The
     certificate file must be in PEM format.
 
+**--cached-only**
+    Only retrieve credentials already present in the cache, not from
+    the KDC.
+
+**--no-store**
+    Do not store retrieved credentials in the cache.  If
+    **--out-cache** is also specified, credentials will still be
+    stored into the output credential cache.
+
+**--out-cache** *ccache*
+    Initialize *ccache* and store all retrieved credentials into it.
+    Do not store acquired credentials in the input cache.
+
 **--u2u** *ccache*
     Requests a user-to-user ticket.  *ccache* must contain a local
     krbtgt ticket for the server principal.  The reported version
--- a/src/clients/kvno/t_kvno.py
+++ b/src/clients/kvno/t_kvno.py
@@ -1,0 +1,75 @@
+from k5test import *
+
+realm = K5Realm()
+
+def check_cache(ccache, expected_services):
+    # Fetch the klist output and skip past the header.
+    lines = realm.run([klist, '-c', ccache]).splitlines()
+    lines = lines[4:]
+
+    # For each line not beginning with an indent, match against the
+    # expected service principals.
+    svcs = {x: True for x in expected_services}
+    for l in lines:
+        if not l.startswith('\t'):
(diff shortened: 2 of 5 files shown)


doc/user/user_commands/kvno.rst
```diff
@@ -74,6 +74,18 @@ OPTIONS
     client principal with the X.509 certificate in *cert_file*.  The
     certificate file must be in PEM format.
 
+**--cached-only**
+    Only retrieve credentials already present in the cache, not from
+    the KDC.
+
+**--no-store**
+    Do not store acquire credentials in the cache.  If **--out-cache**
+    is also specified, credentials will still be stored into the
```
iboukris, line 83: I think it would be simple to error out in that case.

iboukris:
Testing you patch (both regularly and with -U), i still get the ticket the input cache in addition to the output cache, i think it shouldn't.

iboukris:
BTW, we could probably get rid of the required server argument to 'kvno -IUF' and figure it from the default cache principal (like we do when -P is passed).

greghudson:
I had thought kgetcred treats "copy into specified output cache" orthogonally from "don't store in input cache", but now that I check, Isaac is correct.  I will change the PR.

kgetcred does not appear to error if --out-cache and --no-store are both given (the --no-store has no additional effect), so I won't add an error case.

iboukris:
Looks good to me.

src/clients/kvno/t_kvno.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+mark('no options')
+realm.run([kvno, realm.user_princ], expected_msg='user@KRBTEST.COM: kvno = 1')
+check_cache(realm.ccache, [realm.krbtgt_princ, realm.user_princ])
+
+mark('-e')
+msgs = ('etypes requested in TGS request: des3-cbc-sha1',
+        '/KDC has no support for encryption type')
+realm.run([kvno, '-e', 'des3-cbc-sha1', realm.host_princ],
```
frozencemetery, line 32: How would you feel about using a more futureproof enctype here?

src/clients/kvno/t_kvno.py
```diff
@@ -0,0 +1,76 @@
...
+    for l in lines:
+        if not l.startswith('\t'):
+            svcprinc = l.split()[4]
+            if svcprinc in svcs:
+                del svcs[svcprinc]
+            else:
+                fail('unexpected service princ ' + svcprinc)
+
+    if svcs:
+        fail('services not found in klist output: ' + ' '.join(keys((svcs))))
```
frozencemetery, line 22: I believe it's `dict.keys()`, not `keys(dict)`.

src/clients/kvno/t_kvno.py
greghudson: I'll switch to a Camellia enctype; we're less likely to remove those in the forseeable future than des3 (although I have considered it; I don't think there's much use).

(aes-sha2 isn't future-proof for a different reason: we might add it to the default supported_enctypes, causing kvno to succeed where we currently expect it to fail.)

PR #1089, opened 2020-06-23
Fix KDC choice to send encrypted S4U_X509_USER

greghudson:
The KDC's decision to send a PA_S4U_X509_USER entry in encrypted has
no connection to the client USE_REPLY_KEY_USAGE flag.  Only
conditionalize on the enctype.


iboukris:
Looks good to me!

PR #1090, opened 2020-06-25
Use the term "primary KDC" in source and docs

greghudson:
Where it does not affect program behavior, use the term "primary KDC".
This commit does not change any profile variables, DNS labels,
pathnames, or externally visible identifiers, nor does it change the
term "master key".


doc/admin/conf_files/krb5_conf.rst
```diff
@@ -515,10 +515,10 @@ following tags may be specified in the realm's subsection:
     host will be tried.
 
 **master_kdc**
-    Identifies the master KDC(s).  Currently, this tag is used in only
+    Identifies the primary KDC(s).  Currently, this tag is used in only
```
frozencemetery, line 518: This seems like it's setting us up for admin confusion.  Our docs change to talk only about primary, but there's no `primary_kdc` parameter.  It seems like it'd be more clear to either rename the parameter (and keep backward compatibility) if this change is desired.

There's some precedent for this - for instance, `iprop_replica_poll` in 07d9f95e433cf59674faedf1121fe810ab3683c7 as part of the rename in #851.

greghudson:
I pushed a followup commit to rename master_kdc to primary_kdc and iprop_master_ulogsize to iprop_ulogsize.

Both fallbacks were awkward to code for different reasons.

doc/admin/conf_files/krb5_conf.rst
frozencemetery: (This has been resolved, though I'm unable to mark it as such for some reason.)

PR #1091, opened 2020-06-27
Trace plugin module loading errors

greghudson:
Add trace messages when dlopen() or dlsym() fails when loading a
plugin module.


greghudson:
I originally included more context in the new trace messages, such as the module pathname and symbol name.  But at least in glibc, dlerror() includes all that information anyway, so you wind up with a highly redundant trace message.

PR #1094, opened 2020-07-15
Ignore bad enctypes in krb5_string_to_keysalts()

frozencemetery:
Fixes a problem where the presence of legacy/unrecognized keysalts in
supported_enctypes would prevent the kadmin programs from starting.

### DIFF
--- a/src/lib/kadm5/str_conv.c
+++ b/src/lib/kadm5/str_conv.c
@@ -340,9 +340,10 @@ krb5_string_to_keysalts(const char *string, const char *tupleseps,
     while ((ksp = strtok_r(p, tseps, &tlasts)) != NULL) {
         /* Pass a null pointer to subsequent calls to strtok_r(). */
         p = NULL;
-        ret = string_to_keysalt(ksp, ksaltseps, &etype, &stype);
-        if (ret)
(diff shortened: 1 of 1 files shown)


frozencemetery:
By way of justification: krb5kdc starts just fine when it finds unrecognized enctypes.  Additionally, the error for unrecognized keysalts in supported_enctypes from kadmin is deeply mysterious ("kadmind: Required parameters in kdc.conf missing while initializing, aborting")

greghudson:
We should have a trace log message, similar to TRACE_ENCTYPE_LIST_UNKNOWN.

(Ideally this would show up in the kadmind log, but I don't know how to make that happen without a much more complicated library API.)

frozencemetery:
krb5_string_to_keysalts() is used in several places in the code, which means that we won't have as much information from a TRACE_KEYSALT_LIST_UNKNOWN as we would from ENCTYPE.  It lacks a krb5_context parameter; this is fixable (since it's not a function in a header or exported in a stable interface), but a bit messy.  It also probably has the wrong prefix, as do several adjacent functions.

I added a commit that does all this on top, in case it's desirable or I've missed something.

PR #1095, opened 2020-07-18
Expand dns_canonicalize_hostname=fallback support

greghudson:
This isn't complete, but I may have to switch focus next week so I'm putting up what I have so far.  The goal is to iron out a lot of edge cases, and make the server side of things less likely to break regardless of what the qualify_shortname default is.  The design is:

* A lightweight iterator over the one or two service principal candidates, so we can do fallback in more places without duplicating code.

* When dns_canonicalize_hostname=fallback, delay all processing of the sn2princ inputs until use, so that both canonicalizations can be dependent on the original sn2princ inputs and realm lookup can be done properly on each candidate.  (Heimdal does the same thing if the canonicalization rules are non-trivial.)

* Integrate fallback support into the stepwise TGS state machine, so that we don't break IAKERB with delayed canonicalization.

* Add fallback support to k5_get_proxy_cred_from_kdc(), krb5_kt_get_entry(), and the krb5_init_creds_set_keytab() enctype lookup.

* Cache credentials under the unprocessed sn2princ name.

TBD:

* Address the S4U2Proxy must-use-referrals requirement in a way that works with delayed canonicalization.  Currently the GSS code unsets the realm before making an S4U2Proxy request, but the delayed realm lookup defeats that.  [Edit: this is done]

* Address kprop's sn2princ_realm(), which works like the old sn2princ.  [Edit: this is done.  It would be nice to avoid the realm lookup like the old version did, but the logic has gotten too complicated for that, and simply overwriting the realm should serve.  In the default dns_canonicalize_hostname=fallback configuration no realm lookup will be performed because it is deferred.]

* Address KDB creation.

* The k5_get_cached_cred() refactor probably isn't needed for this change and could be separated out.  [Edit: this is done]

* The handling of KRB5_NT_UNKNOWN is incorrect and causes an assertion failure.  The correct behavior is to look up the realm but not canonicalize the hostname.  (Though, if we don't know it's a host-based service name, why look up the realm in host-to-realm?  But that is the historical behavior.)  [Edit: this is fixed, without changing the behavior.  We might still consider changing the behavior, which would simplify the code a bit.]

* Add client principal canonicalization to the get_init_creds state machine.  This is necessary for kprop as well as kinit -k and kadmin -k.  [Edit: every in-tree operation using sn2princ results as AS client principals uses a keytab, so it's probably fine to just use the keytab as an authority.  This is now done.]

* The iterator should check for the second result being identical to the first and not yield it if so.  [Edit: this is done]

* Tests.  Also check for memory errors (including leaks) with an asan build.  [Edit: I think the important cases are tested.  The asan test run is clean with #1097 fixed.]

See also: https://krbdev.mit.edu/rt/Ticket/Display.html?id=8925

### DIFF
--- a/src/lib/krb5/krb/s4u_creds.c
+++ b/src/lib/krb5/krb/s4u_creds.c
@@ -26,6 +26,7 @@
 
 #include "k5-int.h"
 #include "int-proto.h"
+#include "os-proto.h"
 
 /* Convert ticket flags to necessary KDC options */
 #define FLAGS2OPTS(flags) (flags & KDC_TKT_COMMON_MASK)
@@ -981,10 +982,10 @@ get_target_realm_proxy_tgt(krb5_context context, const krb5_data *realm,
     return 0;
 }
 
-krb5_error_code
-k5_get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
-                           krb5_ccache ccache, krb5_creds *in_creds,
-                           krb5_creds **out_creds)
+static krb5_error_code
+get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
+                        krb5_ccache ccache, krb5_creds *in_creds,
+                        krb5_creds **out_creds)
 {
     krb5_error_code code;
     krb5_flags flags, req_kdcopt = 0;
@@ -1120,21 +1121,10 @@ k5_get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
         }
     }
 
-    if (!krb5_principal_compare(context, in_creds->server, tkt->server)) {
-        krb5_free_principal(context, tkt->server);
-        tkt->server = NULL;
-        code = krb5_copy_principal(context, in_creds->server, &tkt->server);
-        if (code)
-            goto cleanup;
-    }
-
     /* Note the authdata we asked for in the output creds. */
     code = krb5_copy_authdata(context, in_creds->authdata, &tkt->authdata);
     if (code)
         goto cleanup;
-
-    if (!(options & KRB5_GC_NO_STORE))
-        (void)krb5_cc_store_cred(context, ccache, tkt);
 
     *out_creds = tkt;
     tkt = NULL;
@@ -1146,6 +1136,48 @@ cleanup:
     krb5_free_pa_data(context, in_padata);
     krb5_free_pa_data(context, enc_padata);
     return code;
+}
+
+krb5_error_code
+k5_get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
+                           krb5_ccache ccache, krb5_creds *in_creds,
+                           krb5_creds **out_creds)
+{
+    krb5_error_code code;
+    krb5_const_principal canonprinc;
+    krb5_creds copy, *creds;
+    struct canonprinc iter = { in_creds->server, .no_hostrealm = TRUE };
+
+    *out_creds = NULL;
+
+    copy = *in_creds;
+    while ((code = k5_canonprinc(context, &iter, &canonprinc)) == 0 &&
+           canonprinc != NULL) {
+        copy.server = (krb5_principal)canonprinc;
+        code = get_proxy_cred_from_kdc(context, options, ccache, &copy,
+                                       &creds);
+        if (code != KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN)
+            break;
+    }
+    if (!code && canonprinc == NULL)
+        code = KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN;
+    free_canonprinc(&iter);
+    if (code)
+        return code;
+
+    krb5_free_principal(context, creds->server);
+    creds->server = NULL;
+    code = krb5_copy_principal(context, in_creds->server, &creds->server);
+    if (code) {
+        krb5_free_creds(context, creds);
+        return code;
+    }
+
+    if (!(options & KRB5_GC_NO_STORE))
+        (void)krb5_cc_store_cred(context, ccache, creds);
+
+    *out_creds = creds;
+    return 0;
 }
 
 /*
--- a/src/lib/krb5/krb/s4u_creds.c
+++ b/src/lib/krb5/krb/s4u_creds.c
@@ -1119,6 +1119,13 @@ get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
             code = KRB5KRB_AP_WRONG_PRINC;
             goto cleanup;
         }
+
+        /* Put the original evidence ticket in the output creds. */
+        krb5_free_data_contents(context, &tkt->second_ticket);
+        code = krb5int_copy_data_contents(context, &in_creds->second_ticket,
+                                          &tkt->second_ticket);
+        if (code)
+            goto cleanup;
     }
 
     /* Note the authdata we asked for in the output creds. */
@@ -1145,10 +1152,32 @@ k5_get_proxy_cred_from_kdc(krb5_context context, krb5_flags options,
 {
     krb5_error_code code;
     krb5_const_principal canonprinc;
-    krb5_creds copy, *creds;
+    krb5_creds mcreds, copy, *creds, *ncreds;
+    krb5_flags fields;
     struct canonprinc iter = { in_creds->server, .no_hostrealm = TRUE };
 
     *out_creds = NULL;
+
+    code = krb5int_construct_matching_creds(context, options, in_creds,
+                                            &mcreds, &fields);
+    if (code != 0)
+        return code;
+
+    ncreds = calloc(1, sizeof(*ncreds));
+    if (ncreds == NULL)
+        return ENOMEM;
+    ncreds->magic = KV5M_CRED;
+
+    code = krb5_cc_retrieve_cred(context, ccache, fields, &mcreds, ncreds);
+    if (code) {
+        free(ncreds);
+    } else {
+        *out_creds = ncreds;
+    }
+
+    if ((code != KRB5_CC_NOTFOUND && code != KRB5_CC_NOT_KTYPE) ||
+        options & KRB5_GC_CACHED)
+        return code;
 
     copy = *in_creds;
     while ((code = k5_canonprinc(context, &iter, &canonprinc)) == 0 &&
@@ -1195,9 +1224,6 @@ krb5_get_credentials_for_proxy(krb5_context context,
                                krb5_creds **out_creds)
 {
     krb5_error_code code;
-    krb5_creds mcreds;
-    krb5_creds *ncreds = NULL;
-    krb5_flags fields;
     krb5_data *evidence_tkt_data = NULL;
     krb5_creds s4u_creds;
 
@@ -1218,30 +1244,6 @@ krb5_get_credentials_for_proxy(krb5_context context,
         code = EINVAL;
         goto cleanup;
     }
-
-    code = krb5int_construct_matching_creds(context, options, in_creds,
-                                            &mcreds, &fields);
-    if (code != 0)
-        goto cleanup;
-
-    ncreds = calloc(1, sizeof(*ncreds));
-    if (ncreds == NULL) {
-        code = ENOMEM;
-        goto cleanup;
-    }
-    ncreds->magic = KV5M_CRED;
-
-    code = krb5_cc_retrieve_cred(context, ccache, fields, &mcreds, ncreds);
-    if (code != 0) {
-        free(ncreds);
-        ncreds = in_creds;
-    } else {
-        *out_creds = ncreds;
-    }
-
-    if ((code != KRB5_CC_NOTFOUND && code != KRB5_CC_NOT_KTYPE)
-        || options & KRB5_GC_CACHED)
-        goto cleanup;
 
     code = encode_krb5_ticket(evidence_tkt, &evidence_tkt_data);
     if (code != 0)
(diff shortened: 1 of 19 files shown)


iboukris:
Quote from the ticket:
> * krb5_get_credentials() ordinarily handles both checking the cache and storing into the cache.  For S4U2Self requests, it calls k5_get_proxy_cred_from_kdc(), which stores into the cache but does not check the cache, so repeated krb5_get_credentials() S4U2Self calls will result in duplicate cache entries.  (GSSAPI does its own cache check before making the S4U2Proxy request, and kvno -P uses the krb5_get_credentials_for_proxy() wrapper which does a cache check.  So this is purely an issue with the krb5_get_credentials() API.)

I guess you mean S4U2Proxy throughout, as S4U2Self doesn't use krb5_get_credentials() afair.

I think we can't do a cache check in krb5_get_credentials() for S4U2Proxy because we do not know the client name to match in the cache as it is in a the (possibly) encrypted 2nd ticket, that's why we needed a fix in the gssapi caller (PR #381)  and why we have a cache check in krb5_get_credentials_for_proxy(), see also discussion at PR #1014.

See this comment removed in commit 2b29619aa27c2e63fea80cac60b5607a3fce972f (since the S4U2Proxy code moved).

    /* For constrained delegation, the expected result is in second_ticket, so
     * we can't really do a cache check here. */
    if (ctx->req_options & KRB5_GC_CONSTRAINED_DELEGATION)
        return (ctx->req_options & KRB5_GC_CACHED) ? KRB5_CC_NOTFOUND : 0;

iboukris:
> Address the S4U2Proxy must-use-referrals requirement in a way that works with delayed canonicalization. Currently the GSS code unsets the realm before making an S4U2Proxy request, but the delayed realm lookup defeats that.

With this hack the s4u tests pass: https://github.com/iboukris/krb5/commits/ghudson-getent-search

So the reason we do the unsetting in GSS code is because there we can assume that the realm was added from domain_realm and not explicitly requested by the caller, so we can ignore it.

As discussed in #1014, maybe we can simply ignore the requested realm in S4U2Proxy and set to referral realm for hostbased (like Heimdal does in general, in the referral call) and enterprise principals (similar to the above hack, just with referral realm).
This will keep everything working with the only caveat of ignoring the requested realm, so if the caller requested `host/pc7@B` we may end up requesting `host/pc7@A` and if this host exists in A then we may get a ticket for it instead (which will likely case an error when presented to the target server), while a regular call to krb5_get_credentials() without s4u for `host/pc7@B` will result the correct princ using regular cross-realm.

greghudson:
I did mean S4U2Proxy in the ticket.

I did notice that we can't check the cache in krb5_get_credentials() for an S4U2Proxy request.  With perfect hindsight, I wish we had never (going back to 1.7/1.8) tried to make krb5_get_credentials() work for S4U2Proxy requests, and just required a separate API.  But for now we will just leave that wrinkle.

> As discussed in #1014, maybe we can simply ignore the requested realm in S4U2Proxy and set to referral realm for hostbased (like Heimdal does in general, in the referral call) and enterprise principals (similar to the above hack, just with referral realm).

I think we want to do this, but caching under a realm we ignored seems like it could create problems.  So I guess the GSS code should continue removing the realm, and krb5_get_credentials_for_proxy() may also need to.

iboukris:
> I think we want to do this, but caching under a realm we ignored seems like it could create problems.

Not sure what problems, I think this may happen with current code, if you request `service/rb.c@B` you may end up with a for `service/rb.c@C` saved under the requested name, e.g. below.  So I think we can say that we treat the requested realm as a mere start-realm hint.

```log
diff --git a/src/tests/gssapi/t_s4u.py b/src/tests/gssapi/t_s4u.py
index 67eba44da..4550fcbd7 100755
--- a/src/tests/gssapi/t_s4u.py
+++ b/src/tests/gssapi/t_s4u.py
@@ -388,6 +388,9 @@ ra.run(['./t_s4u', 'p:' + ra.user_princ, 'h:service@rb.c'], env=domain_conf)
 ra.run(['./t_s4u', 'p:' + 'sensitive@A', 'h:service@rb.c'], expected_code=1)
 ra.run(['./t_s4u', 'p:' + rb.user_princ, 'h:service@rb.c'])
 
+ra.run([kvno, 'service/rb.c@B'])
+ra.run([klist, ])

Result:

*** [150] Executing: /home/admin/git/krb5/src/clients/kvno/kvno service/rb.c@B
service/rb.c@B: kvno = 1
*** [150] Completed with return code 0
*** [151] Executing: /home/admin/git/krb5/src/clients/klist/klist
Ticket cache: FILE:/home/admin/git/krb5/src/tests/gssapi/testdir/1/ccache
Default principal: impersonator@A

07/19/20 21:28:37  07/20/20 21:28:37  krbtgt/A@A
07/19/20 21:28:37  07/20/20 21:28:37  krbtgt/B@A
07/19/20 21:28:37  07/20/20 21:28:37  service/rb.c@B
        Ticket server: service/rb.c@C
*** [151] Completed with return code 0
```

greghudson:
In that case realm B's KDC thinks service/rb.c belongs in realm C, though.  In the S4U2Proxy case realm B is never consulted.

It probably wouldn't cause serious problems, but it is safer if, when possible, we check and store the name without the realm.  That's doable in the usual cases (GSS and krb5_get_credentials_for_proxy()).  We have less control if it's a third party calling krb5_get_credentials() because we have to assume that it used the supplied service name to check the cache, if it checked the cache at all.

iboukris:
> I did notice that we can't check the cache in krb5_get_credentials() for an S4U2Proxy request.

Actually, we do have a flag to match on second-ticket KRB5_TC_MATCH_2ND_TKT, maybe that's sufficient and we should ignore the client princ by passing a new flag (need to check that we save second-ticket on result cred).  Then we can get rid of the cache check in the callers, although we'd require the same 2nd ticket to be used.

greghudson:
> Actually, we do have a flag to match on second-ticket

I guess that would work, since we expect the same evidence ticket to be used for all S4U2Proxy requests within a cache.

iboukris:
Yeah, the important part is that a repeated requests won't add to the cache, custom S4U2Proxy requests with the krb5 API could end up with seemingly duplicate entries but limited to the number of evidence tickets used, so it should be fine.

iboukris:
> It probably wouldn't cause serious problems, but it is safer if, when possible, we check and store the name without the realm. That's doable in the usual cases (GSS and krb5_get_credentials_for_proxy()). 

I now recall that I had another motivation for not overriding with referral realm in the krb5 layer, so we can exercise requests with netbios realm on wire, so if we only override the realm in the callers we may be able to keep this ability.

greghudson:
I pushed a cleaned-up version of Isaac's workaround, by adding a no_hostrealm flag to the iterator structure.

I'm using the C99 .field = value initializer feature here, because it reads very well in this case and I think our compiler base can now support it (now that we require VS2013 in the Windows build).

iboukris:
> I guess that would work, since we expect the same evidence ticket to be used for all S4U2Proxy requests within a cache.

I gave it a little try but no luck so far, s4u tests fail with: "Expected 3 tickets but got 4"
https://github.com/iboukris/krb5/commits/ghudson-getent-search

greghudson:
Check the cache before canonicalizing, not after, since we cache it under the requested server name.

greghudson:
I checked Heimdal, and its krb5_compare_creds() allows mcreds->client and mcreds->server to be null.  We might consider doing that instead of adding a flag.

src/lib/krb5/krb/s4u_creds.c
```diff
@@ -1187,26 +1228,7 @@ krb5_get_credentials_for_proxy(krb5_context context,
...
-
-    code = krb5_cc_retrieve_cred(context, ccache, fields, &mcreds, ncreds);
-    if (code != 0) {
-        free(ncreds);
-        ncreds = in_creds;
-    } else {
-        *out_creds = ncreds;
-    }
-
+    code = k5_get_cached_cred(context, options, ccache, in_creds, out_creds);
```
iboukris, line 1231: This can go now.

iboukris:
> I checked Heimdal, and its krb5_compare_creds() allows mcreds->client and mcreds->server to be null. We might consider doing that instead of adding a flag.

This sounds good by itself, but won't be very helpful for this use case so the flag may still make sense.

iboukris:
Or instead of the new flag we can ignore the client comparison implicitly when KRB5_TC_MATCH_2ND_TKT is specified (unless KRB5_TC_MATCH_IS_SKEY?) but that's also cryptic.

iboukris:
This fixup feels a bit better: https://github.com/iboukris/krb5/commit/aebcdd5aeff47e84587988b4adaaaa4126ff8b47

iboukris:
> This fixup feels a bit better: iboukris@aebcdd5

Actually, i think we can get rid of the krb5_principal_compare() check altogether, we'll get the right ticket from cache based on 2nd ticket, and there is nothing wrong with requesting a S4U2Proxy request to self anyway.

greghudson:
> This sounds good by itself, but won't be very helpful for this use case so the flag may still make sense.

I was thinking construct_matching_creds() could null out the client field where it currently sets the flag.

> Actually, i think we can get rid of the krb5_principal_compare() check altogether, we'll get the right ticket from cache based on 2nd ticket, and there is nothing wrong with requesting a S4U2Proxy request to self anyway.

Sounds good.

iboukris:
> I was thinking construct_matching_creds() could null out the client field where it currently sets the flag.

Oh, now that construct_matching_creds is static and mcreds is confined to k5_get_cached_cred, that's fine indeed.

iboukris:
>> Actually, i think we can get rid of the krb5_principal_compare() check altogether, we'll get the right ticket from cache based on 2nd ticket, and there is nothing wrong with requesting a S4U2Proxy request to self anyway.
>
>Sounds good.

Actually2, we won't find the evidence ticket in cache by matching on 2nd ticket (as it was not acquired with S4U2Proxy), but still there is nothing wrong with such S4U2Proxy request to self.

iboukris:
> but still there is nothing wrong with such S4U2Proxy request to self.

That said, it is less likely to succeed due to policy, so it's better as is, sorry for the noise.

greghudson:
While adding tests I observed that basically nothing does krb5_kt_get_entry() with an uncanonicalized sn2princ result, since we have scan-and-match in rd_req, vfy_increds, and gic_keytab.  Accordingly, I reverted the krb5_kt_get_entry() iteration and added rd_req iteration.  vfy_increds is only matching against the service name (for "host") so it doesn't need attention.

For a while I thought I was going to have to add iteration to the get_in_tkt state machine, but since there's always a keytab involved when authenticating as an sn2princ result (at least for in-tree operations), we can use the keytab as an authority rather than the KDC.  I had to defer the restart_init_creds_loop() to the first krb5_init_cred_step() so that we use the principal found in the keytab to set up the server realm.

The CI is reporting a seg fault in one of the tests that was passing before, which doesn't happen on my dev machine, so I have to debug that by carrier pigeon.

greghudson:
Everything on the TBD list is done except for KDB creation, and I have written proper commit messages.  As mentioned in the ticket, newly created KDBs are populated with kadmin/hostname and kiprop/hostname principal entries.  Getting these wrong isn't really high-impact, since kadmin will fall back to kadmin/admin and kiprop/hostname is just one of several principals you need for iprop to work.  (We only started creating kiprop/hostname in 1.13; the change was contributed by a Solaris developer and justified as an admin convenience.)  Still, if we decide to change the qualify_shortname default to "", it will be at best galling if KDB creation usually creates the wrong hostbased principals.

We have a number of imperfect options:

1. Do nothing, or at least defer the problem to a later PR.

2. Stop creating the hostbased principals.  Perhaps change the libkadm5clnt init code to try kadmin/admin first and kadmin/hostname second, for efficiency.

3. Explicitly use DNS (well, NSS) to canonicalize the local hostname when creating a KDB, overriding the dns_canonicalize_hostname setting.

4. Create hostbased principals for both possible canonicalizations of the local hostname.

5. Document that KDC administrators should set qualify_shortname in the KDC's krb5.conf, unless the kernel hostname is an FQDN.

I went back and researched why our kadmin uses the hybrid of kadmin/hostname and kadmin/admin.  Originally it just used kadmin/admin, but for the 1.4 release the preferred form was changed for compatibility with the Solaris kadmin system (called SEAM).  SEAM used hostbased kadmin principals because it wanted to use GSS hostbased names, and kadmin/admin isn't hostbased.  Interestingly, Sam argued against the change in https://mailman.mit.edu/pipermail/krbdev/2003-August/001937.html and https://mailman.mit.edu/pipermail/krbdev/2003-September/001948.html for exactly the reason that the KDC can't easily tell which hostname clients will contact it with, but I guess he lost that argument.

greghudson:
I'm removing the WIP designation and declaring this ready for review.  Since KDB creation isn't a matter of fallback, it should probably be addressed in a separate PR.  (I am leaning towards option 2, stop creating the hostbased principals, and change libkadm5clnt to try kadmin/admin before kadmin/hostname.)

iboukris:
Ack on the S4U2Proxy commit, looks good, thanks!

src/lib/krb5/os/os-proto.h
```diff
@@ -83,6 +83,36 @@ struct sendto_callback_info {
     void *data;
 };
 
+/*
+ * Initialize with all zeros except for princ.  Set no_hostrealm to disable
+ * host-o-realm lookup, which ordinarily happens after canonicalizing the host
```
frozencemetery, line 88: "host-o-realm"?

greghudson: Typo for "host-to-realm", not actually a breakfast cereal.

PR #1097, opened 2020-07-24
Fix leak in KERB_AP_OPTIONS_CBT server support

greghudson:
[Found this with asan while checking a different PR.  CBT support hasn't gone into a release yet, so I'm combining this into the implementation ticket.]

In check_cbt(), use a local variable to hold the retrieved authdata
list, and free it before returning.

### DIFF
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -433,27 +433,30 @@ static const uint8_t null_cb[CB_MD5_LEN];
 /* Look for AP_OPTIONS in authdata.  If present and the options include
  * KERB_AP_OPTIONS_CBT, set *cbt_out to true. */
 static krb5_error_code
-check_cbt(krb5_context context, krb5_authdata **authdata,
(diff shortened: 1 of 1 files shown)


iboukris:
Ouch, the fix looks good, thanks!

iboukris:
Maybe make authdata const:
```diff
diff --git a/src/lib/gssapi/krb5/accept_sec_context.c b/src/lib/gssapi/krb5/accept_sec_context.c
index b0bcd7b41..85be82e9e 100644
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -433,7 +433,7 @@ static const uint8_t null_cb[CB_MD5_LEN];
 /* Look for AP_OPTIONS in authdata.  If present and the options include
  * KERB_AP_OPTIONS_CBT, set *cbt_out to true. */
 static krb5_error_code
-check_cbt(krb5_context context, krb5_authdata **authdata,
+check_cbt(krb5_context context, krb5_authdata *const *authdata,
           krb5_boolean *cbt_out)

greghudson:
Sure, done.

PR #1098, opened 2020-07-28
Two SPNEGO DER decoding changes

greghudson:
The first commit is the change requested by Richard Sharpe, to interoperate with a buggy NetApp DER encoder.  The supplied patch had some vulnerabilities, so I rewrote it.

The second commit fixes read length checking issues that I noticed while investigating.  (I don't think they have a material security impact.)


frozencemetery (approved):
Apologies, thought I'd reviewed this.  +1 to the second commit; still ambivalent about the first.

PR #1105, opened 2020-08-20
Kvno docs internal consistency

frozencemetery:
It was pointed out [downstream](https://bugzilla.redhat.com/show_bug.cgi?id=1869055) that the three places `kvno` options could be documented differ.  The full state of affairs is:

```
Unimplemented but documented somewhere: []
Undocumented but implemented: [] /* not that these would be a problem necessarily */
Missing from man page synopsis: ['--cached-only', '--no-store', '--out-cache', '-C', '-k', '-u']
Missing from man page long descriptions: ['-C', '-k', '-u']
Missing from xusage: ['-h', '-q']
Complete set: ['--cached-only', '--no-store', '--out-cache', '--u2u',
               '-C', '-F', '-I', '-P', '-S', '-U',
               '-c', '-e', '-h', '-k', '-q', '-u']
```

The first commit addresses the disparity, while including some missing dependency information and removing -h.

The second commit is a different approach which reworks kvno's docs to a (possibly) more clear presentation, treating kvno as a command which has three forms of invocation.

(I feel like this has been reported before, but was unable to find a ticket number, so it's possible I'm thinking of something else.)

PR #1106, opened 2020-08-25
wip: alias realm support

iboukris:
Following discussion on #1105 about netbios server realm, I did some testing but I think there are more places to look at.

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -143,6 +143,7 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
     krb5_audit_state *au_state = NULL;
     krb5_data **auth_indicators = NULL;
     void *ad_info = NULL, *stkt_ad_info = NULL;
+    krb5_principal_data reply_server;
 
     memset(&reply, 0, sizeof(reply));
     memset(&reply_encpart, 0, sizeof(reply_encpart));
@@ -208,6 +209,22 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
         goto cleanup;
     }
 
+    /* XXX make sure server here has the proper realm...taken from AP_REQ
+       header? */
+
+    if (isflagset(request->kdc_options, KDC_OPT_CANONICALIZE)) {
+        setflag(c_flags, KRB5_KDB_FLAG_CANONICALIZE);
+        setflag(s_flags, KRB5_KDB_FLAG_CANONICALIZE);
+    }
+
+    au_state->stage = SRVC_PRINC;
+
+    errcode = search_sprinc(kdc_active_realm, request, s_flags, &server,
+                            &status);
+    if (errcode != 0)
+        goto cleanup;
+    sprinc = server->princ;
+
     errcode = get_local_tgt(kdc_context, &sprinc->realm, header_server,
                             &local_tgt, &local_tgt_storage, &local_tgt_key);
     if (errcode) {
@@ -231,22 +248,6 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
      * use header_ticket freely.  The encrypted part (if any) has been
      * decrypted with the session key.
      */
-
-    au_state->stage = SRVC_PRINC;
-
-    /* XXX make sure server here has the proper realm...taken from AP_REQ
-       header? */
-
-    if (isflagset(request->kdc_options, KDC_OPT_CANONICALIZE)) {
-        setflag(c_flags, KRB5_KDB_FLAG_CANONICALIZE);
-        setflag(s_flags, KRB5_KDB_FLAG_CANONICALIZE);
-    }
-
-    errcode = search_sprinc(kdc_active_realm, request, s_flags, &server,
-                            &status);
-    if (errcode != 0)
-        goto cleanup;
-    sprinc = server->princ;
 
     /* If we got a cross-realm TGS which is not the requested server, we are
      * issuing a referral (or alternate TGT, which we treat similarly). */
@@ -296,8 +297,10 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
 
     /* Aside from cross-realm S4U2Self requests, do not accept header tickets
      * for local users issued by foreign realms. */
-    if (s4u_x509_user == NULL && data_eq(cprinc->realm, sprinc->realm) &&
-        isflagset(c_flags, KRB5_KDB_FLAG_CROSS_REALM)) {
+    if (s4u_x509_user == NULL &&
+        isflagset(c_flags, KRB5_KDB_FLAG_CROSS_REALM) &&
+        is_local_principal(kdc_context, server, cprinc)) {
+
         krb5_klog_syslog(LOG_INFO, _("PROCESS_TGS: failed lineage check"));
         retval = KRB5KDC_ERR_POLICY;
         goto cleanup;
@@ -417,10 +420,15 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
         }
     }
 
-    if (is_referral)
-        ticket_reply.server = server->princ;
+    if (is_referral || (isflagset(request->kdc_options, KDC_OPT_CANONICALIZE) &&
+                       is_local_tgs_principal(server->princ)))
+        reply_server = *server->princ;
     else
-        ticket_reply.server = request->server; /* XXX careful for realm... */
+        reply_server = *request->server; /* XXX careful for realm... */
+
+    /* Allow kdb module to canonicalize server realm ^^ */
+    reply_server.realm = server->princ->realm;
+    ticket_reply.server = &reply_server;
 
     enc_tkt_reply.flags = get_ticket_flags(request->kdc_options, client,
                                            server, header_enc_tkt);
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -1407,6 +1407,18 @@ is_client_db_alias(krb5_context context, const krb5_db_entry *entry,
     return is_self;
 }
 
+/* Return true if princ is in the same realm as entry->princ. */
+krb5_boolean
+is_local_principal(krb5_context context, const krb5_db_entry *entry,
+                   krb5_const_principal princ)
+{
+    if (krb5_realm_compare(context, entry->princ, princ))
+        return TRUE;
+
+    return krb5_db_is_realm_alias(context, &entry->princ->realm,
+                                  &princ->realm);
+}
+
 /*
  * Protocol transition (S4U2Self)
  */
@@ -1502,7 +1514,7 @@ kdc_process_s4u2self_req(kdc_realm_t *kdc_active_realm,
     /*
      * Do not attempt to lookup principals in foreign realms.
      */
-    if (data_eq(server->princ->realm, id->user->realm)) {
+    if (is_local_principal(kdc_context, server, id->user)) {
         krb5_db_entry no_server;
         krb5_pa_data **e_data = NULL;
 
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -734,6 +734,10 @@ krb5_error_code krb5_db_get_authdata_info(krb5_context context,
 
 void krb5_db_free_authdata_info(krb5_context context, void *ad_info);
 
+krb5_boolean
+krb5_db_is_realm_alias(krb5_context context, const krb5_data *realm,
+                       const krb5_data *alias);
+
 /**
  * Sort an array of @a krb5_key_data keys in descending order by their kvno.
  * Key data order within a kvno is preserved.
@@ -1523,6 +1527,9 @@ typedef struct _kdb_vftabl {
 
     void (*free_authdata_info)(krb5_context context,
                                void *ad_info);
+
+    krb5_boolean (*is_realm_alias)(krb5_context context, const krb5_data *realm,
+                                   const krb5_data *alias);
 
     /* End of minor version 0 for major version 8. */
 } kdb_vftabl;
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -182,6 +182,7 @@ struct as_req_state {
     const krb5_fulladdr *local_addr;
     const krb5_fulladdr *remote_addr;
     krb5_data **auth_indicators;
+    krb5_principal_data reply_server;
 
     krb5_error_code preauth_err;
 
@@ -674,10 +675,13 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
     if (isflagset(s_flags, KRB5_KDB_FLAG_CANONICALIZE) &&
         krb5_is_tgs_principal(state->request->server) &&
         krb5_is_tgs_principal(state->server->princ)) {
-        state->ticket_reply.server = state->server->princ;
+        state->reply_server = *state->server->princ;
     } else {
-        state->ticket_reply.server = state->request->server;
+        state->reply_server = *state->request->server;
     }
+    /* Allow kdb module to canonicalize ticket server realm */
+    state->reply_server.realm = state->server->princ->realm;
+    state->ticket_reply.server = &state->reply_server;
 
     /* Copy options that request the corresponding ticket flags. */
     state->enc_tkt_reply.flags = get_ticket_flags(state->request->kdc_options,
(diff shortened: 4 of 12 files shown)


iboukris:
The lineage check should be updated: https://github.com/krb5/krb5/blob/d4272f3f29ddbd6268c21be73887cd5b20326dd5/src/kdc/kdc_util.c#L270

As well as other usage of is_local_principal() such as:
https://github.com/krb5/krb5/blob/d4272f3f29ddbd6268c21be73887cd5b20326dd5/src/kdc/kdc_util.c#L1564

greghudson:
I would ideally like to get rid of all of the is_local_principal() calls and all uses of tgs_server, because I have dreams of making it possible to serve multiple realms out of the same KDB.  For is_local_principal(), we have:

* do_as_req.c wants to return a WRONG_REALM referral if the KDB returned a non-local client principal.  We could compare against the requested server realm, or wait until after the server lookup and compare against the canonicalized server realm, or detect the referral in another way (all other fields in the principal entry are supposed to be null, per the kdb.h comment).  When a Windows KDC sees an AS-REQ and the client realm is an alias or a case-mismatch for the canonical realm name, does it issue a WRONG_REALM error or just change the realm in the reply?

* do_tgs_req.c wants to set the CROSS_REALM flag if the header ticket server was issued by a foreign realm.  (The PR changes this to check against the canonicalized header server, although I don't know if we expect header tickets to contain non-canonicalized realms.)  We can compare the realms of sprinc and header_server->princ instead.

* kdc_process_tgs_req() has the lineage check, which errors out if the header ticket server is foreign but the header ticket client is local, except for S4U2Self requests.  This one may require some reordering of operations, since in the current order we haven't looked up the request server yet (so don't know the canonicalized request realm).

* kdc_process_s4u2self_req() bypasses client lookup if the S4U-requested realm is non-local.  Here we might need to do the krbtgt/realm@realm lookup to canonicalize the request realm as is done in this PR, and then compare the realm of the result against the realm of server->princ.

For tgs_server we have:

* do_tgs_req.c doesn't add a transited entry if the header ticket server is from our realm.  We can check the CROSS_REALM flag instead.

* fast_util.c:armor_ap_request() rejects the armor if ticket if isn't a TGT for the local realm (though it can be a TGT issued by a foreign realm).  This is the most difficult check; we want to compare with the canonicalized inner body's server realm, which we don't have until significantly later in processing.

* check_anon() enforces restrict_anonymous_to_tgt by rejecting anonymous requests to principals other than tgs_server.  We can pass in the server DB entry and check that its principal is a local TGT for any realm (krbtgt/X@X).

* kdc_process_s4u2proxy_req() requires that the second ticket server be a TGT to the realm being served.  I think we can check that it is a TGT to proxy->princ->realm.  (The parameter naming in this function is confusing; I would naively expect "server" to be the DB entry of the request server, not the second ticket server.)

* tgs_policy.c:check_tgs_u2u() requires that the second ticket server be a local TGT to the realm being served.  If we pass in the canonicalized server principal we can check that the second ticket server is a local TGT for the server realm instead.

iboukris:
> do_as_req.c wants to return a WRONG_REALM referral if the KDB returned a non-local client principal. We could compare against the requested server realm, or wait until after the server lookup and compare against the canonicalized server realm, or detect the referral in another way (all other fields in the principal entry are supposed to be null, per the kdb.h comment). When a Windows KDC sees an AS-REQ and the client realm is an alias or a case-mismatch for the canonical realm name, does it issue a WRONG_REALM error or just change the realm in the reply?

For alias realm Windows would just change the reply regardless of canonicalize flag (like i drated in this PR).
btw, i thought AS-REQ has only one realm for both client and server.

> do_tgs_req.c wants to set the CROSS_REALM flag if the header ticket server was issued by a foreign realm. (The PR changes this to check against the canonicalized header server, although I don't know if we expect header tickets to contain non-canonicalized realms.) We can compare the realms of sprinc and header_server->princ instead.

Ack, we could error if header_server->princ.realm canonicalizes. 

> kdc_process_s4u2self_req() bypasses client lookup if the S4U-requested realm is non-local. Here we might need to do the krbtgt/realm@realm lookup to canonicalize the request realm as is done in this PR, and then compare the realm of the result against the realm of server->princ.

Or we can lookup the client unconditionally, then check that entry->princ.realm is the same as heade_server->princ.realm to determine non-local.

Ack for the rest.

iboukris:
I also recall yet another place where we may want to use the new is_client_alias(), in u2u handling code where we currently may fail with '2ND_TKT_MISMATCH', like in this bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1836630

greghudson:
> For alias realm Windows would just change the reply regardless of canonicalize flag (like i drated in this PR).
> btw, i thought AS-REQ has only one realm for both client and server.

It does only have one realm, but if we're trying to determine whether the client lookup is a referral, we can't assume that the realm of client->princ is the canonicalized request realm.

I said we could perhaps detect client referrals another way besides a realm comparison, but looking at krb5_db_entry, there doesn't seem to be any field we can rely on being set for a non-referral DB lookup--that is, zero/null is a possible value in every field for a real principal entry.  So I think to be correct here, we need to canonicalize the server realm.  The PR does this by looking up krbtgt/realm@realm of the realm being served, but we could instead wait until after the server lookup, then check and issue a WRONG_REALM error if client->princ has a different realm from sever->princ.

> > kdc_process_s4u2self_req() bypasses client lookup if the S4U-requested realm is non-local. Here we might need to do the krbtgt/realm@realm lookup to canonicalize the request realm as is done in this PR, and then compare the realm of the result against the realm of server->princ.
> 
> Or we can lookup the client unconditionally, then check that entry->princ.realm is the same as heade_server->princ.realm to determine non-local.

I'm not sure if that would work.  If the client princ isn't found, we need to know whether that's because it's in a foreign realm or because it's just not an existing client principal.  So we need to know whether the named realm canonicalizes to the realm of server->princ.

iboukris:
I rebased poorly on top of #1120, and dropped the client vs server aliases commit (although I'm still unclear how to tell in samba if we are in AS server lookup or TGS server lookup).

Was thinking again about the logic of cross-realm S4U2Self lineage check, previously we added checks so only valid requests can go thru, but I still feel unclear why are they valid from lineage perspectives, and whether adding the realm to the client name in the PAC has anything to do with it (in which case PAC may be required in cross-realm).
Perhaps we can discuss in the call tomorrow?

greghudson:
> (although I'm still unclear how to tell in samba if we are in AS server lookup or TGS server lookup)

I'm not sure there's currently a way to distinguish.

greghudson:
Rebased and made trivial whitespace fixes.

src/kdc/kdc_util.c
```diff
@@ -1444,6 +1407,41 @@ is_client_db_alias(krb5_context context, const krb5_db_entry *entry,
...
+    code = krb5_db_get_principal(context, alias_krbtgt, 0, &alias_entry);
+    krb5_free_principal(context, alias_krbtgt);
+    if (code) {
+        if (code != KRB5_KDB_NOENTRY)
+            return code;
+        *alias = FALSE;
+        return 0;
+    }
+
+    *alias = krb5_realm_compare(context, entry->princ, alias_entry->princ);
```
iboukris, line 1440: Note: I think this would be better as "data_eq(alias_entry->princ->data[1], entry->princ->realm);"

iboukris:
I added some client changes to allow realm-canon in AS-REP even if the flag wasn't requested, but it looks like a cache with a ticket to krbtgt/alias@realm isn't very useful anyway, so I guess we can drop or deffer this.

Also I'm a bit hesitating about adding a lookup for any cross-realm request, and wonder if a dedicated kdb method could be more efficient, although I still favor the alias lookup.

greghudson:
> Also I'm a bit hesitating about adding a lookup for any cross-realm request, and wonder if a dedicated kdb method could be more efficient, although I still favor the alias lookup.

That is a concern.  One argument for a dedicated KDB method is that it could give dispatch() a way to do the right thing (without unreasonable efficiency) when the KDC is started with multiple `-r` flags.

iboukris:
> One argument for a dedicated KDB method is that it could give dispatch() a way to do the right thing (without unreasonable efficiency) when the KDC is started with multiple -r flags, and one or more of the realms has aliases.

Note that my idea of the kdb method was 'kdb_is_realm_alias(realm, alias)' rather than give me the whole list, this allows to easily implement any lower/upper case combination, like Acme.com etc, which windows supports fine.

greghudson:
> Note that my idea of the kdb method was 'kdb_is_realm_alias(realm, alias)' rather than give me the whole list, this allows to easily implement any lower/upper case combination, like Acme.com etc, which windows supports fine.

That's fine for dispatch().  Iterating through a list of realms asking "is Y an alias for you (X)?" is reasonably efficient; iterating through doing krbtgt/Y@Y lookups not as much.

iboukris:
Switched to a dedicated kdb method.

src/kdc/do_tgs_req.c
```diff
@@ -417,10 +420,15 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
         }
     }
 
-    if (is_referral)
-        ticket_reply.server = server->princ;
+    if (is_referral || (isflagset(request->kdc_options, KDC_OPT_CANONICALIZE) &&
+                        krb5_is_tgs_principal(server->princ)))
```
greghudson, line 424: I'm curious about this change.  It only has an effect if is_referral is false but server->princ is a TGS principal and is different from request->server.  is_referral is set by:

    is_referral = is_cross_tgs_principal(server->princ) &&
        !krb5_principal_compare(kdc_context, request->server, server->princ);

So we can infer that server->princ is a local TGS principal; that is, the client requested krbtgt/ALIAS@ALIAS (or krbtgt/ALIAS@REALM) and the KDB returned krbtgt/REALM@REALM.  I have two questions:

1. Is this a likely request, or just a theoretical point of compatibility with Windows?  This is the TGS path, so the client has already had a chance to canonicalize the realm.

2. Can we handle this more simply by just setting is_referral to true?  (That is, when we set is_referral, use krb5_is_tgs_principal() instead of is_cross_tgs_principal().)  It would be the KDB's responsibility not to canonicalize the TGS name if KRB5_KDB_FLAG_CANONICALIZE isn't set.

greghudson: Oh, but we now say that modules can always return in-realm aliases, so we'd be going back on that.  So I guess I just have the first question.

iboukris: I think it is likely, it happens for instance in the capture I attached to #1110, packets 80-84:
https://github.com/iboukris/S4U/blob/master/logs/xrealm_s4u2self.pcapng

And yeah, it is simpler when the kdb always return the canonical name, i'll look at the code again.

iboukris: I changed the condition a bit but I can't think of anything better, let me know if you prefer option 2. and we'll have to make an exception in samba for krbtgt alias lookup when the canon flag isn't set.

greghudson: > and we'll have to make an exception in samba for krbtgt alias lookup when the canon flag isn't set.

I'm not following; why does changing the condition to use is_local_tgs_principal() require a KDB change?

iboukris: That was unrelated to the is_local_tgs_principal() change, if we go with option 2. from the ones you gave above, that is always canonicalize krbtgt reply regardless of the the canon flag.

greghudson: Understood.  We'll stick with this method rather than making the KDB requirements more complicated.

iboukris:
Dropped the additional client changes, and started to clean up the tests, and found that validate_tgs_request() also needed fixing.

iboukris:
It was too strict in two cases, one when header server is fully canon (either as requested or -C) but request is for alias realm, and the other when the header server is krbtgt/alias@realm and request is for canon realm (or other alias).

iboukris:
Hmm, while windows allows a request for alias realm with header server fully canon, I think it shouldn't usually happen as the client would ask for krbtgt/alias and try to use that as header. I think this only happens in S4U2Proxy with MIT client where we do it in purpose in order to allow alias realm to work, but maybe we should instead simply override the realm to client realm as discussed previously.

greghudson:
I agree that we should probably change the client behavior; it was always a strange workaround and we don't want to be the only ones making a screwy kind of KDC request.

iboukris:
Ok, another way of fixing it without ignoring the requested realm could be to revert as before, and if the server realm isn't referral or client realm then we fetch a krbtgt for it first, in the alias case we'd end up with either krbtgt/alias@realm or krbtgt/realm@realm depending on whether the canon flag was set. At this point if the realm of the returned krbtgt princ is the same as client realm then we can make a final S4U2Proxy request, if it is not (ie it was another realm or an alias to another realm), then we can fail immediately - this case would fail now as well.

greghudson:
That seems reasonable.

iboukris:
Interestingly, client-referrals don't work well with alias-realm, because the returned wrong-realm isn't canon (although with -C it doesn't matter since the final reply is gets canon):
```
[root@fdc ~]# echo -n pwd  | kinit -E Administrator@ACME.COM@ACOM
Password for Administrator\@ACME.COM@ACOM: 
[root@fdc ~]# klist 
Ticket cache: FILE:/tmp/krb5cc_0
Default principal: Administrator\@ACME.COM@ACME.COM

10/15/2020 00:14:20  10/15/2020 10:14:20  krbtgt/ACOM@ACME.COM
	renew until 10/16/2020 00:14:20

**Comment by iboukris** [comment]
Dropped the changes in validate_tgs_request() but when adding tests I found another case where the kdc returns `The ticket isn't for us`, not S4U2Proxy related. See the XXX note in the last commit, it happens if you first ask for `krbtgt/NBB@NBB` with -C so we end up with:
```
10/15/20 01:22:27  10/16/20 01:22:27  krbtgt/NBB@NBB
	Ticket server: krbtgt/B@B
```
Then we ask for `kvno rbb@NBB`, so the above ticket gets fetch from cache and used to realm NBB, resulting failure (also reproduces against Windows kdc).

iboukris:
Perhaps we can canonicalize the request realm only in check_tgs_tgt() and keep assuming the header server's data[1] is canon. This should allow both these use cases to work like with Winodws KDC. While we can still fix the S4U2Proxy to behave more properly, I don't have a better solution for the cache problem.

greghudson:
My initial reaction is that the cache case should be fixed by coercing the request to be in the TGT's realm.  I'd have to do a lot of analysis to figure out where that change should be made, though, given all the complicated realm-traversal logic in get_creds.c.

iboukris:
Not sure how the cache issue can be fixed, given we cache by the requested name and given we end up accepting a canonicalized krbtgt/canon while we asked for krbtgt/alias (when the canon flag is used, we'll never get a krbtgt/alias).
On the other hand I don't get why is it so wrong, if we embrace the concept of alias realms then the two forms are equivalent and should work like in Windows KDC.

greghudson:
> Not sure how the cache issue can be fixed, given we cache by the requested name and given we end up accepting a canonicalized krbtgt/canon while we asked for krbtgt/alias (when the canon flag is used, we'll never get a krbtgt/alias).

The client's reasoning would be thus:

* We want to make a request for realm R.
* We asked the cache for krbtgt/R and the cache gave us krbtgt/R' instead.
* R must be an alias and R' must be its canonical name.
* Therefore, we should make the request to realm R' instead.

But again, I'm not sure exactly how that reasoning should translate into a code change.

Whether the KDC should be strict about this is a separate question.  The argument for being strict is relatively weak, given that we have the canonical realm names of both halves on hand.

iboukris:
> Therefore, we should make the request to realm R' instead.

This make sense and I think that's what Windows does as well, to test I created trust between samba-mit[1] and Windows, and I ran the lsa command from both sides using the netbios-realm name (as -d). I can see that if we canonicalize the tgs request for krbtgt/ALIAS to krbtgt/CANON then the next request would be a S4U2Self request to realm CANON, and if we don't then the next request is a S4U2Self request to ALIAS as realm.

I'd like to continue the research until I get all the lsa commands "working" from both sides of the trust, including alias realms, like in this comment: https://github.com/krb5/krb5/pull/1110#issuecomment-695202057

[1] using wip branch: https://gitlab.com/samba-team/samba/-/merge_requests/818

iboukris:
On the other hand, I won't have time for this in the near future, closing for now.

PR #1107, opened 2020-08-26
Remove S4U2Self post-canonicalization cache check

greghudson:
Isaac reminded me of https://colabti.org/irclogger/irclogger_log/krbdev?date=2020-01-03#l21 wherein I concluded that the second cache check in krb5_get_credentials_for_user() could probably be omitted.

The change is easy enough.  A few tests have to be adjusted because they look for a trace log message corresponding to the second cache lookup.

However, I did notice that kvno winds up caching under the final result, rather than the requested client principal:

    ====== cross-realm S4U2Self with enterprise name ======
    *** [110] Executing: ./t_s4u e:enterprise@abc@NOREALM - /me/krb5/build/tests/gssapi/testdir/1/keytab
    *** [110] Spawning shell
    $ /me/krb5/build/clients/kvno/kvno -U enterprise@abc@NOREALM user@SREALM
    user@SREALM: kvno = 1
    $ /me/krb5/build/clients/klist/klist
    Ticket cache: FILE:/me/krb5/build/tests/gssapi/testdir/1/ccache
    Default principal: user@SREALM
    
    08/26/20 00:07:05  08/27/20 00:07:05  krbtgt/SREALM@SREALM
    08/26/20 00:10:32  08/27/20 00:07:05  krbtgt/UREALM@SREALM
    08/26/20 00:10:32  08/27/20 00:07:05  user@SREALM
    	for client enterprise\@abc@UREALM

Repeating the operation doesn't cause the cache to grow because of the final (previously third) cache check.  But it does mean we perform some unnecessary TGS operations for successive kvno operations (on top of the AS requests to identify the realm).

If this were a server principal, I would say we should cache under the requested name.  But I don't know if that would fly for a client principal; having the client name in a ccache entry not match the tickets isn't familiar territory like it is for server names.

gss_acquire_cred_impersonate_name() is unaffected because it passes KRB5_GC_NO_STORE and instead stores the ticket in the memory cache for the proxy cred.

### DIFF
--- a/src/lib/krb5/krb/s4u_creds.c
+++ b/src/lib/krb5/krb/s4u_creds.c
@@ -665,11 +665,13 @@ krb5_get_credentials_for_user(krb5_context context, krb5_flags options,
         /* Uncanonicalised check */
         code = krb5_get_credentials(context, options | KRB5_GC_CACHED,
                                     ccache, in_creds, out_creds);
-        if (code != KRB5_CC_NOTFOUND && code != KRB5_CC_NOT_KTYPE)
+        if ((code != KRB5_CC_NOTFOUND && code != KRB5_CC_NOT_KTYPE) ||
+            (options & KRB5_GC_CACHED))
             goto cleanup;
-
-        if ((options & KRB5_GC_CACHED) && !(options & KRB5_GC_CANONICALIZE))
-            goto cleanup;
+    } else if (options & KRB5_GC_CACHED) {
+        /* Fail immediately, since we can't check the cache by certificate. */
+        code = KRB5_CC_NOTFOUND;
+        goto cleanup;
     }
 
     code = s4u_identify_user(context, in_creds, subject_cert, &realm);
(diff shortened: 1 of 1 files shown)


iboukris:
Yeah we mentioned that, maybe leave the cache check, just don't depend on KRB5_GC_CANONICALIZE (and obey KRB5_GC_CACHED).

greghudson:
Okay, I've changed the PR to do that and removed the WIP tag.

iboukris:
The patch looks good to me.

> If this were a server principal, I would say we should cache under the requested name. But I don't know if that would fly for a client principal;

I think we must use the returned cname or we may end up with ap-req->authenticator->client not equals to ticket->client (although i have not tested):
https://github.com/krb5/krb5/blob/8f2f0a2e8f65c4b39883129967301e3a8986218b/src/lib/krb5/krb/rd_req_dec.c#L532

greghudson:
It just occurred to me that KRB5_GC_CACHED will be ignored for cert-based requests (in_creds->client == NULL).  We can't successfully check the cache in that case, but I wonder if we should return an error (KRB5_CC_NOTFOUND? EINVAL?) rather than ignore the flag.

Now that we have kvno --cached-only, it is possible to run into this case with kvno --cached-only -F certfile.  We could of course check for that in kvno.

iboukris:
As we think of KRB5_GC_CACHED in terms of no IO, then I think returning KRB5_CC_NOTFOUND would be fine.

greghudson:
Amended to do that.

iboukris:
Thanks, looks good.

PR #1108, opened 2020-08-27
Allow KDC to canonicalize realm in TGS client

greghudson:
Active Directory canonicalizes the srealm field of TGS replies,
whether or not the client requests canonicalization.  Allow this for
regular TGS and S4U2Self referrals queries by comparing only the name
part of the service principal.  The S4U2Proxy code is already correct.

### DIFF
--- a/src/lib/krb5/krb/get_creds.c
+++ b/src/lib/krb5/krb/get_creds.c
@@ -571,10 +571,12 @@ step_referrals(krb5_context context, krb5_tkt_creds_context ctx)
     if (ctx->reply_code != 0)
         return try_fallback(context, ctx);
 
-    if (krb5_principal_compare(context, ctx->reply_creds->server,
-                               ctx->server)) {
-        /* We got the ticket we asked for... but we didn't necessarily ask for
-         * it with the right enctypes.  Try a non-referral request if so. */
+    /* Check if we got the ticket we asked for.  Allow the KDC to canonicalize
+     * the realm. */
+    if (krb5_principal_compare_any_realm(context, ctx->reply_creds->server,
+                                         ctx->server)) {
+        /* We didn't necessarily ask for it with the right enctypes.  Try a
+         * non-referral request if so. */
         if (wrong_enctype(context, ctx->reply_creds->keyblock.enctype)) {
             TRACE_TKT_CREDS_WRONG_ENCTYPE(context);
             return begin_non_referral(context, ctx);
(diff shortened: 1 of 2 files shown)


greghudson:
PR #1106 will allow test cases for this, so that will probably get merged first.

iboukris:
> Allowing krbtgt/ALIAS to be rewritten to krbtgt/REALM might be a reasonable behavior for KRB5_GC_CANONICALIZE to control, although it seems very niche. This is a tricky case from a protocol perspective because krbtgt/REALM looks just like an alternate TGT. When we ask REALM for krbtgt/ALIAS using the supplied krbtgt/REALM, we get krbtgt/REALM back again. At this point, with KRB5_GC_CANONICALIZE we could accept the result (instead of "falling back" to the same query without server principal checking), and without it we could retry without the canonicalize flag in order to get a krbtgt/ALIAS ticket.

Yes, this sounds good. If we only allow name canonicalization for krbtgt then we don't need special handling and can restrict to:
https://github.com/iboukris/krb5/commits/canon_fixup

greghudson:
The first of those two commits might break changing the realm for regular requests (service@ALIAS -> service@REALM) because check_reply_server() isn't using krb5_principal_compare_any_realm().

For the second commit, the history around the deleted fallback has to do with krb5_set_password_using_ccache() against AD 2000/2003.  That function makes a TGS request for kadmin/changepw, and apparently those versions of AD would issue a krbtgt ticket instead (probably with PA-SVR-REFERRAL-DATA as described in RFC 6806 appendix A) if the canonicalize bit was set.  The interop problem was first fixed in https://krbdev.mit.edu/rt/Ticket/Display.html?id=4955 and then re-fixed in https://krbdev.mit.edu/rt/Ticket/Display.html?id=7016 .  There's a description of the scenario in https://mailman.mit.edu/pipermail/kerberos/2011-November/017714.html after the regression.

iboukris:
> The first of those two commits might break changing the realm for regular requests (service@ALIAS -> service@REALM) because check_reply_server() isn't using krb5_principal_compare_any_realm().

Right missed that, maybe it should use any_realm().

> For the second commit, the history around the deleted fallback has to do with krb5_set_password_using_ccache() against AD 2000/2003. That function makes a TGS request for kadmin/changepw, and apparently those versions of AD would issue a krbtgt ticket instead

Then we can leave the fallback but still take the shortcut for alias krbtgt requests (btw this doesn't happen with newer version of Windows).

frozencemetery:
(Wrong window please ignore)

PR #1109, opened 2020-09-01
Harmonize macOS pack declarations with Heimdal

greghudson:
[The use of pragma pack in the ABI is an ancient MIT krb5 mistake that isn't easily fixed by either us or Apple.  We don't want to be incompatible with macOS on ARM or to give Apple reason to migrate it to ARM.]

Replace the TARGET_OS_MAC conditionals with the conditionals used in
Heimdal, so that we do not pack structures inconsistently with macOS
on ARM.  Suggested by Luke Howard.

### DIFF
--- a/src/include/CredentialsCache.h
+++ b/src/include/CredentialsCache.h
@@ -52,7 +52,7 @@
 extern "C" {
 #endif /* __cplusplus */
 
-#if TARGET_OS_MAC
+#if defined(__APPLE__) && (defined(__ppc__) || defined(__ppc64__) || defined(__i386__) || defined(__x86_64__))
 #pragma pack(push,2)
 #endif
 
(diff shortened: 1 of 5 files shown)


kaduk:
> We don't want to be incompatible with macOS on ARM or to give Apple reason to migrate it to ARM.

It looks like this sentence didn't make it into the commit message, but I'm confused how to parse it.  Why would anything we do give Apple reason to migrate macOS to ARM?

greghudson:
Unclear antecedent, sorry.  "It" refers to the mistake, which currently doesn't apply to macOS ARM (in Heimdal and therefore macOS native Kerberos), rather than macOS itself.

kaduk:
> Unclear antecedent, sorry. "It" refers to the mistake, which currently doesn't apply to macOS ARM (in Heimdal and therefore macOS native Kerberos), rather than macOS itself.

Ah, that makes sense; thanks for clarifying.

PR #1110, opened 2020-09-02
wip: S4U2Self cross-realm enterprise server

iboukris:
I'm clueless about the top commit, looks like if the server is overwritten in the cache, then encode_krb5_ticket() would yield different ticket data depending on whether krb5_server_decrypt_ticket_keytab() was called.

I was planning to make an alternative fix for this commit from #1014,  before i got caught with the above: https://github.com/krb5/krb5/pull/1014/commits/5887e98b768ab57d8a5d6655404a0ae38afe3da9


iboukris:
Matching by second ticket may not be as robust as i thought but probably still fine.

iboukris:
Commented out the krb5_server_decrypt_ticket_keytab() call, and added the intended new fix which becomes needed.

iboukris:
More specifically, here is what I find weird:

```
116	    check(krb5_cc_retrieve_cred(context, evcc, 0, &mcred, &ev_cred));
(gdb) n
118	    check(krb5_decode_ticket(&ev_cred.ticket, &ev_ticket));
(gdb) p *ev_ticket.server.data
$8 = {magic = 0, length = 14, data = 0x406100 "impersonator@A"}
(gdb) n
121	    check(krb5_server_decrypt_ticket_keytab(context, defkt, ev_ticket));
(gdb) p *ev_ticket.server.data
$9 = {magic = -1760647422, length = 12, data = 0x40ad70 "impersonator\377\177"}

**Comment by iboukris** [comment]
I now see that krb5_server_decrypt_ticket_keytab() rewrites ticket->server to the principal from keytab, i wonder why.

**Comment by iboukris** [comment]
Or alternatively to keep out of krb5_server_decrypt_ticket_keytab(), we can rewrite ticket->server too (and re-encode creds->ticket), so no one can tell we requested an enterprise server, i.e it is not in the encrypted part (pushed a commit this way).

**Comment by greghudson** [comment]
> I now see that krb5_server_decrypt_ticket_keytab() rewrites ticket->server to the principal from keytab, i wonder why (omitting the rewrite passes make-check).

Unfortunately that change does not have good revision control notes, but it was presumably to mirror the behavior of krb5_rd_req().

The history as I understand it is:

* krb5_rd_req() and krb5_server_decrypt_ticket_keytab() used to look up the unprotected server principal in the keytab and decrypt the ticket with the entry it found.  If someone tampered with the server principal, the lookup would fail, or it would find the wrong key and decryption would fail.

* Then, we wanted to add support for aliases.  Making admins store redundant keytab entries for every alias seemed unfriendly, and in some cases impossible.  So the two functions were changed to iterate over the keytab and decrypt with any entry (constrained by the application-provided server name if one was provided).

* This change raised the possibility that someone could tamper with the unprotected server principal and the ticket would still decrypt successfully.  This could pose a security issue if the application made any decisions based on the service principal (e.g. kadmind requires it to be a kadmin/* principal, although in kadmind we don't iterate because the keytab is the whole KDB).

* Therefore we rewrite the ticket server principal to match the keytab entry with the matching key.  This isn't perfect, because if there are multiple keytab entries with the same key then the apparent server principal depends on keytab entry order.

**Comment by iboukris** [comment]
> This change raised the possibility that someone could tamper with the unprotected server principal and the ticket would still decrypt successfully. This could pose a security issue if the application made any decisions based on the service principal (e.g. kadmind requires it to be a kadmin/* principal, although in kadmind we don't iterate because the keytab is the whole KDB).

It would still only be possible to rewrite to a valid alias (in tgs-rep we also don't allow canonicalization).

> Therefore we rewrite the ticket server principal to match the keytab entry with the matching key. This isn't perfect, because if there are multiple keytab entries with the same key then the apparent server principal depends on keytab entry order.

Maybe we can try to look for an entry with the requested name first, then fallback to any key to avoid the rewrite.

That said, i'm happy with my latest approach - full rewrite for the enterprise server, what do you think?

**Comment by greghudson** [comment]
> It would still only be possible to rewrite to a valid alias (in tgs-rep we also don't allow canonicalization).

The rewrite could happen at AP-REQ time though.

**Comment by iboukris** [comment]
Ok, but still the main point is that it has the same key so its a valid alias. If we worry about the case of principals with same keys, then the rewrite might actually cause access elevation.

**Comment by iboukris** [comment]
Ok got it.

**Comment by iboukris** [comment]
I forgot another issue with SPN like principals, for those we can't really convert to enterprise, that is windows server will error if the server is enterprise like 'host/apache@A@A', the only reason it currently works is because we only use these names when talking with foreign realm, which only look at the inner realm for purpose of routing and ignore the name. However in this case we should simply not convert as these names are valid across realms. Pushed a fixup.

**Comment by iboukris** [comment]
Added a commit which I think is needed, now that we allow uncanonicalized names in AS-REP like Windows.

**Comment by iboukris** [comment]
> Added a commit which I think is needed, now that we allow uncanonicalized names in AS-REP like Windows.

That is, in AD it is quite possible for one principal's client alias to be the same as another principal's server alias.
e.g. if users a and b have their UPN inversed (b@acme, and a@acme, respectively), then kinit -E a@acme@acme will result a tgt for b principal, while doing kvno to enterprise server a@acme@acme will get a ticket to principal a.

**Comment by greghudson** [comment]
I pushed a revision to the first commit.  Only trivial code edits, but I rewrote the commit message.

File: src/lib/krb5/krb/s4u_creds.c [hunk 1]
```diff
@@ -405,7 +405,8 @@ convert_to_enterprise(krb5_context context, krb5_principal princ,
     char *str;
 
     *eprinc_out = NULL;
-    code = krb5_unparse_name(context, princ, &str);
+    code = krb5_unparse_name_flags(context, princ,
+                                   KRB5_PRINCIPAL_UNPARSE_DISPLAY, &str);
```
greghudson, line 409: What's this change about?  If we're going to immediately re-parse the name, won't the quoted special characters be de-quoted at that point?

iboukris: Not quite, it causes the reparse to escape the first at, resulting 3 at's at the end, causing kinit -E followed by s4u2self to fail.

greghudson: So in this case, in_creds->server is already an enterprise principal?

iboukris: Yes, but it may be an enterprise without inner at, so by unparsing and reparsing we get the right result.

greghudson: You're saying if we have an enterprise principal with data[0] = 'abc' and a realm set, we want to translate that to data[0] = abc@REALM?

iboukris: Yes, that our best shot to get a routable server name (the name-type matters less, what matters is the inner at).

iboukris:
So the goal is for the stepwise API to use `in_creds->client` and `in_creds->server` as is with the exception of overriding the server realm, while the impersonate client will come from `krb5_tkt_creds_set_impersonate()` (not sure how about `krb5_get_credentials()` though).
This is similar to how S4U2Proxy is done now with the `krb5_get_credentials()` API (and used to, for stepwise), only with the impersonate client coming from `in_creds->additional_ticket`.
If you feel reluctant about changing the (very convenient) hack of using the enterprise server only in requests to foreign realms, then maybe we can keep it but with a special flag, for the gss caller?

src/lib/krb5/krb/s4u_creds.c
greghudson: I think we're introducing a bug here if princ is a regular principal with an intentional `@` (not intended as either a realm or enterprise domain separator) in the first data component.  But nobody does that as it would just be inviting trouble, so it's probably not a bug with real consequences.

I thought about making krb5_parse_name_flags() treat `\@` as the first at for an enterprise name, but Heimdal doesn't do that, so it doesn't seem like a good idea.

iboukris: I think if it's only one at sign, then the name-type will just change to enterprise but the name will stay as is.

greghudson:
As long as we're being more consistent rather than less consistent with Windows clients, I think the change is okay.

I am still understanding the change, though.  I am surprised that not converting SPNs when talking to foreign realms is correct; I thought SPNs were the usual case for S4U2Self (but perhaps people typically use the computer account name?) and I thought we had to introduce that conversion for SPNs to get cross-realm S4U2Self to work.

iboukris:
Windows uses the logged-on user as the server, so it's always the canonical name (as they asked for canonicalize). It then would use enterprise server when the client isn't in the same realm, but they keep using it for the final request to server realm.

Our SPN to enterprise conversion isn't right, as we get an error if we try it in the final request (foreign realms would accept anything@realm@localrealm and will give a referral to realm, and all works fine).

In Windows you have the canonical name (can't be SPN like), which works both as client and as server locally, then you have enterprise servers (canonical converted) which works as servers both locally and cross realm. Then you have the single value UPN which is client alias for kinit, and SPN multi-value for server aliases which do work across realms.
For linux clients, it is often important to have the UPN set as their SPN, as otherwise they aren't able to kinit with the keytab (which often has the SPN) and this is also what ktpass does (it sets the UPN by default, not to mention the salt related issues).
In #1106 i started changing our aliases logic in kdb_test to match that of Windows with client and server aliases.

However, i just noticed i'm missing something, we only know the real realm after the realm-identification part (i hit it before and forgot), so either we always use enterprise server if the client is enterprise, or just leave it as a special flag and handle it in the krb5 api, need to revise.

iboukris:
> Our SPN to enterprise conversion isn't right, as we get an error if we try it in the final request (foreign realms would accept anything@realm@localrealm and will give a referral to realm, and all works fine).

I think the reason for that is simply because enterprise servers only really work with the canonical name.

> However, i just noticed i'm missing something, we only know the real realm after the realm-identification part

To avoid this problem, maybe we can split the future S4U2Self stepwise API into two parts, for realm-identification the caller would have to loop over krb5_init_creds_(), and then call krb5_tkt_creds_() using the canonical realm?

iboukris:
> However, i just noticed i'm missing something, we only know the real realm after the realm-identification part

After thinking some more, forcing enterprise server if the client was enterprise (or certificate) isn't that bad, we only added these two options recently in gss (with the enterprise and certificate names), so it won't cause much change and technically it is fine too.
I pushed fixup in this direction (which is exactly what I had in #1014).

greghudson:
> However, i just noticed i'm missing something, we only know the real realm after the realm-identification part (i hit it before and forgot), so either we always use enterprise server if the client is enterprise, or just leave it as a special flag and handle it in the krb5 api, need to revise.

I'm not following.  Realm identification happens in the caller, before krb5_get_self_cred_from_kdc(), so we know at this point whether we're doing a cross-realm request.

iboukris:
> I'm not following. Realm identification happens in the caller, before krb5_get_self_cred_from_kdc(), so we know at this point whether we're doing a cross-realm request.

Now yes, but not once we change it to work via krb5_tkt_creds_(), which is my objective to sort it out as preparation to #1014, although we can hold on this change or the whole PR until i revise #1014.

greghudson:
I think I see where you're going, although I don't currently understand why the stepwise code can't convert the server principal on the fly when needed.

Do Windows clients use the enterprise server form even when doing a same-realm query, or only when there is cross-realm involved?

I assume I should at least merge the first commit.

iboukris:
> I think I see where you're going, although I don't currently understand why the stepwise code can't convert the server principal on the fly when needed.

a. because it shouldn't alter in_creds->server between requests, so we can properly implement S4U2Self like Windows clients.
b. because it actually doesn't work with netbios realm, as the client doesn't know it's the same realm, and is not happy with a ticket to enterprise server.

> Do Windows clients use the enterprise server form even when doing a same-realm query, or only when there is cross-realm involved?

Only when cross-realm is involved, but then keep using it in the final request to server realm. I'll run some more tests though.

> I assume I should at least merge the first commit.

I now wonder if we should pass KRB5_KDB_FLAG_CLIENT_REFERRALS_ONLY (and change the logic a bit), in order to make the call a bit less surprising to current implementations (in samba and ipa), I'll move it to a separate PR.

iboukris:
> Then you have the single value UPN which is client alias for kinit

Actually, I stand corrected by samba code which i've just tested, the UPN can be used as server principal (strangely only in TGS request, not AS):
https://github.com/samba-team/samba/blob/232054c09b1932b3940f08aa818703b51d29d968/source4/kdc/db-glue.c#L1973

iboukris:
> I think the reason for that is simply because enterprise servers only really work with the canonical name.

Which means i was wrong about the above as well, i'm now able to kvno HTTP/apache.acme.com@ACME.COM parsed as enterprise, i'll now try the S4U2Self case that i recall to have fail.

iboukris:
> Our SPN to enterprise conversion isn't right, as we get an error if we try it in the final request (foreign realms would accept anything@realm@localrealm and will give a referral to realm, and all works fine).

So this was all wrong, i just tested this PR branch only removing the condition on "s4u_creds.server->length == 1" and the SPN like name (the UPN) converted to enterprise work just fine even for the final request..

```
[admin@fdc src]$ kinit HTTP/apache.acme.com@ACME.COM
Password for HTTP/apache.acme.com@ACME.COM: 
[admin@fdc src]$ kvno -I Administrator@CDOM.COM HTTP/apache.acme.com@ACME.COM
HTTP/apache.acme.com@ACME.COM: kvno = 2
[admin@fdc src]$ kvno -I Administrator@ACOM HTTP/apache.acme.com@ACME.COM
HTTP/apache.acme.com@ACME.COM: kvno = 2

[admin@fdc src]$ klist 
Ticket cache: FILE:/tmp/krb5cc_1000
Default principal: HTTP/apache.acme.com@ACME.COM

09/19/2020 01:50:24  09/19/2020 11:50:24  krbtgt/ACME.COM@ACME.COM
	renew until 09/20/2020 01:50:21
09/19/2020 01:50:37  09/19/2020 11:50:24  krbtgt/CDOM.COM@ACME.COM
	renew until 09/20/2020 01:50:21
09/19/2020 01:50:37  09/19/2020 11:50:24  HTTP/apache.acme.com@ACME.COM
	for client Administrator@CDOM.COM, renew until 09/20/2020 01:50:21
09/19/2020 01:50:51  09/19/2020 11:50:24  krbtgt/ACOM@ACME.COM
	renew until 09/20/2020 01:50:21
09/19/2020 01:50:51  09/19/2020 11:50:24  HTTP/apache.acme.com@ACME.COM
	for client Administrator@ACOM, renew until 09/20/2020 01:50:21

**Comment by iboukris** [comment]
Note that this type of SPN as enterprise still only works for the UPN and the canonical names, not for the SPNs, which is what lead to my error.

> only removing the condition on "s4u_creds.server->length == 1" 

But this causes new failures in t_kvno.py ..

**Comment by iboukris** [comment]
I ran some tests on Windows with:
https://github.com/iboukris/S4U/blob/master/win/lsa.cpp

See result pcap (all succeeded):
https://github.com/iboukris/S4U/blob/master/logs/xrealm_s4u2self.pcapng

As you can see the win api let's you specify the realm or not to, in case it is
not specified it would be looked up with client referrals (and server realm).

This is similar to what we offer with kvno -U because we identify the realm
the same way, but in Windows what matters is whether the domain was specified,
it always uses enterprise type so it can fit in names with at sign.

Note that in cross-realm the final request is also using enterprise server, more
ever notice the case where netbios realm is specified (-d ACOM, packets 69-73), the
Windows client is not aware it is the same realm, asks for krbtgt/ACOM and then for
enterprise server because it thinks it is cross-realm, but then happy with the results.

```
@ECHO OFF

klist purge

REM these triggers realm identification

lsa -u Administrator
lsa -u Administrator@ACME.COM
lsa -u Administrator@ACOM

lsa -u Administrator@CDOM.COM
lsa -u Administrator@CDOM

klist purge

REM these do not

lsa -u Administrator -d ACME.COM
lsa -u Administrator@ACME.COM -d ACME.COM
lsa -u Administrator@ACOM -d ACME.COM

lsa -u Administrator@CDOM.COM -d CDOM.COM
lsa -u Administrator@CDOM -d CDOM.COM

klist purge

lsa -u Administrator -d ACOM
lsa -u Administrator@ACME.COM -d ACOM
lsa -u Administrator@ACOM -d ACOM

lsa -u Administrator@CDOM.COM -d CDOM
lsa -u Administrator@CDOM -d CDOM


ECHO Done
PAUSE

iboukris:
In a hindsight we could have implemented the same and only do realm-identification if the realm is referral realm, regardless of the name-type, similar to the use we make of referral (empty) realm in server requests.

iboukris:
Pushed a fixup to only determine cross-realm after realm-identification, as it is clearly what Windows is doing, and allowed for SPN to enterprise conversion. That triggered an issue that I've encountered before where profile_get_relation_names() finds non-existing entries causing referral logic error, as a workaround I added a check for "names[0] == NULL" but I don't really get it.

greghudson:
I can investigate, but the test suite passes for me if I remove the names[0] == NULL check.

iboukris:
True.. doesn't happen here anymore either, i'll try to chase it next time...

PR #1114, opened 2020-09-15
Begin deprecating des3-cbc-sha1

greghudson:
During initial authentication, if des3-cbc-sha1 is used for either the
reply key or session key, display a warning that it will be
deprecated.

### DIFF
--- a/doc/admin/conf_files/kdc_conf.rst
+++ b/doc/admin/conf_files/kdc_conf.rst
@@ -844,17 +844,17 @@ Encryption types
 
 Any tag in the configuration files which requires a list of encryption
 types can be set to some combination of the following strings.
-Encryption types marked as "weak" are available for compatibility but
-not recommended for use.
+Encryption types marked as "weak" and "deprecated" are available for
+compatibility but not recommended for use.
 
 ==================================================== =========================================================
 des3-cbc-raw                                         Triple DES cbc mode raw (weak)
-des3-cbc-sha1 des3-hmac-sha1 des3-cbc-sha1-kd        Triple DES cbc mode with HMAC/sha1
+des3-cbc-sha1 des3-hmac-sha1 des3-cbc-sha1-kd        Triple DES cbc mode with HMAC/sha1 (deprecated)
 aes256-cts-hmac-sha1-96 aes256-cts aes256-sha1       AES-256 CTS mode with 96-bit SHA-1 HMAC
 aes128-cts-hmac-sha1-96 aes128-cts aes128-sha1       AES-128 CTS mode with 96-bit SHA-1 HMAC
 aes256-cts-hmac-sha384-192 aes256-sha2               AES-256 CTS mode with 192-bit SHA-384 HMAC
 aes128-cts-hmac-sha256-128 aes128-sha2               AES-128 CTS mode with 128-bit SHA-256 HMAC
-arcfour-hmac rc4-hmac arcfour-hmac-md5               RC4 with HMAC/MD5
+arcfour-hmac rc4-hmac arcfour-hmac-md5               RC4 with HMAC/MD5 (deprecated)
 arcfour-hmac-exp rc4-hmac-exp arcfour-hmac-md5-exp   Exportable RC4 with HMAC/MD5 (weak)
 camellia256-cts-cmac camellia256-cts                 Camellia-256 CTS mode with CMAC
 camellia128-cts-cmac camellia128-cts                 Camellia-128 CTS mode with CMAC
--- a/doc/admin/enctypes.rst
+++ b/doc/admin/enctypes.rst
@@ -123,24 +123,83 @@ Enctype compatibility
 
 See :ref:`Encryption_types` for additional information about enctypes.
 
-========================== ===== ======== =======
-enctype                    weak? krb5     Windows
-========================== ===== ======== =======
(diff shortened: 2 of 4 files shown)


greghudson:
This approach mirrors the one used in the Apple fork of Heimdal.  That code base uses a more permanent code fixture, by adding a krb5_enctype_warning() function and a warning flag on enctypes.  I think we can just modify the code when we need to add or remove a warning.

This approach only finds users or initial ticket services (usually krbtgt/REALM) using triple-DES.  Services won't generally be detected.  We could add a check for the ticket encryption key, but I don't think it would be of much value.  We don't have a good way of displaying a warning for TGS requests (unless the request happens to be made with kvno, but that's not the common case).
a
My plan is to escalate to disabling with a setting to reenable (like allow_weak_crypto, but probably not the same variable) in either 1.20 or 1.21, and then remove support in some release after that.

frozencemetery (approved):
First, thanks for starting on this and laying out a timeline.  I was going to ask you about scope and what would be useful today, but that's all answered now.

I've checked that kinit and kadmin both trip the warning and a message is printed.  It might be nice to add a test for this, but I don't think I want to hold this up for that.

greghudson:
It appears that des3 is already deprecated according to the enctype table (and therefore in logs and kadmin output).  I changed the warning to say "will be desupported" (although I still plan to have some releases where it can be reenabled via a flag), and changed the commit summary.

src/lib/krb5/krb/get_in_tkt.c
```diff
@@ -1604,7 +1604,7 @@ warn_des3(krb5_context context, krb5_init_creds_context ctx,
         return;
 
     banner = _("Warning: encryption type des3-cbc-sha1 used for "
-               "authentication is weak and will be deprecated");
+               "authentication is weak and will be desupported");
```
frozencemetery, line 1607: I think this change is good in general but "desupported" is a bit unclear.  "will be nsupported in the future" is clearer, but long; maybe "will be removed"?

greghudson: I'll change it to "will be disabled" since that is the next step.

greghudson:
I added a documentation commit.  I expect to remove admin/advanced/retiring-des.rst at some point, but that's out of scope.

doc/admin/conf_files/kdc_conf.rst
```diff
@@ -844,22 +844,21 @@ Encryption types
...
 aes256-cts-hmac-sha384-192 aes256-sha2               AES-256 CTS mode with 192-bit SHA-384 HMAC
 aes128-cts-hmac-sha256-128 aes128-sha2               AES-128 CTS mode with 128-bit SHA-256 HMAC
-arcfour-hmac rc4-hmac arcfour-hmac-md5               RC4 with HMAC/MD5
+arcfour-hmac rc4-hmac arcfour-hmac-md5               RC4 with HMAC/MD5 (deprecated)
 arcfour-hmac-exp rc4-hmac-exp arcfour-hmac-md5-exp   Exportable RC4 with HMAC/MD5 (weak)
 camellia256-cts-cmac camellia256-cts                 Camellia-256 CTS mode with CMAC
 camellia128-cts-cmac camellia128-cts                 Camellia-128 CTS mode with CMAC
-des3                                                 The triple DES family: des3-cbc-sha1
-aes                                                  The AES family: aes256-cts-hmac-sha1-96, aes128-cts-hmac-sha1-96, aes256-cts-hmac-sha384-192, and aes128-cts-hmac-sha256-128
+des3                                                 The triple DES family: des3-cbc-sha1aes                                                  The AES family: aes256-cts-hmac-sha1-96, aes128-cts-hmac-sha1-96, aes256-cts-hmac-sha384-192, and aes128-cts-hmac-sha256-128
```
frozencemetery, line 861: This change seems accidental.

doc/admin/enctypes.rst
```diff
@@ -123,24 +123,83 @@ Enctype compatibility
...
+
+krb5 releases 1.17 and later flag deprecated encryption types
+(including ``des3-cbc-sha1`` and ``arcfour-hmac``) in KDC logs and
+kadmin output.  krb5 release 1.19 issues a warning during initial
+authentication if ``des3-cbc-sha1`` is used.  Future releases will
+disable ``des3-cbc-sha1`` by default and eventually remove support for
+it.
+
+
+Migrating away from older encryption types
```
frozencemetery, line 156: I like the step-by-step checklist feel of this.  Do you think it would make sense to link it from retiring-des.rst?

frozencemetery (changes requested):

doc/admin/enctypes.rst
greghudson: Since I plan to remove retiring-des.rst I don't think there's much value in adding the cross-reference.

frozencemetery: Ah okay, wasn't aware that was a candidate for deletion.  If this is to be the replacement I am in favor :)

kaduk: I am, of course, biased (being the main author of retiring-des.rst) but in some sense a lot of it is a very-expanded version of "changing the krbtgt key", and that part might still have value even once DES is well and truly gone.  That said, we have version control so nothing is permanently gone, and I can't ask anyone to sign up for doing the work of extracting out the generic bits.

PR #1115, opened 2020-09-16
Add password option to cred store

simo5:
This patch make the gss_acquire_cred_from_store() function also cover
the same functionality of gss_acquire_cred_with_password().

This makes the interface "complete" in the sense that it allows to cover
all possible acquire scenarios in a single API.

More importantly it makes it safer to acquire cred with password in
processes that need to do so in a thread safe manner. By specifying
unique ccache anmes in the cred store we avoid other non-thread safe
methods (or heavy synchronization) to use different ccaches which
are needed with the current gss_acquire_cred_with_password() interface.

### DIFF
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -1168,16 +1168,13 @@ gss_krb5int_import_cred(OM_uint32 *minor_status,
     return code;
 }
 
-OM_uint32 KRB5_CALLCONV
-krb5_gss_acquire_cred_from(OM_uint32 *minor_status,
-                           const gss_name_t desired_name,
-                           OM_uint32 time_req,
-                           const gss_OID_set desired_mechs,
-                           gss_cred_usage_t cred_usage,
-                           gss_const_key_value_set_t cred_store,
-                           gss_cred_id_t *output_cred_handle,
-                           gss_OID_set *actual_mechs,
-                           OM_uint32 *time_rec)
+static OM_uint32
+acquire_cred_from(OM_uint32 *minor_status, const gss_name_t desired_name,
+                  OM_uint32 time_req, const gss_OID_set desired_mechs,
+                  gss_cred_usage_t cred_usage,
+                  gss_const_key_value_set_t cred_store, krb5_boolean iakerb,
+                  gss_cred_id_t *output_cred_handle,
+                  gss_OID_set *actual_mechs, OM_uint32 *time_rec)
 {
     krb5_context context = NULL;
     krb5_error_code code = 0;
@@ -1246,7 +1243,7 @@ krb5_gss_acquire_cred_from(OM_uint32 *minor_status,
 
     ret = acquire_cred_context(context, minor_status, desired_name, NULL,
                                time_req, cred_usage, ccache, client_keytab,
-                               keytab, rcname, 0, output_cred_handle,
+                               keytab, rcname, iakerb, output_cred_handle,
                                time_rec);
 
 out:
@@ -1258,4 +1255,38 @@ out:
         krb5_kt_close(context, keytab);
     krb5_free_context(context);
     return ret;
+}
+
+OM_uint32 KRB5_CALLCONV
+krb5_gss_acquire_cred_from(OM_uint32 *minor_status,
+                           const gss_name_t desired_name,
+                           OM_uint32 time_req,
+                           const gss_OID_set desired_mechs,
+                           gss_cred_usage_t cred_usage,
+                           gss_const_key_value_set_t cred_store,
+                           gss_cred_id_t *output_cred_handle,
+                           gss_OID_set *actual_mechs,
+                           OM_uint32 *time_rec)
+{
+    return acquire_cred_from(minor_status, desired_name, time_req,
+                             desired_mechs, cred_usage, cred_store,
+                             FALSE, output_cred_handle, actual_mechs,
+                             time_rec);
+}
+
+OM_uint32 KRB5_CALLCONV
+iakerb_gss_acquire_cred_from(OM_uint32 *minor_status,
+                             const gss_name_t desired_name,
+                             OM_uint32 time_req,
+                             const gss_OID_set desired_mechs,
+                             gss_cred_usage_t cred_usage,
+                             gss_const_key_value_set_t cred_store,
+                             gss_cred_id_t *output_cred_handle,
+                             gss_OID_set *actual_mechs,
+                             OM_uint32 *time_rec)
+{
+    return acquire_cred_from(minor_status, desired_name, time_req,
+                             desired_mechs, cred_usage, cred_store,
+                             TRUE, output_cred_handle, actual_mechs,
+                             time_rec);
 }
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -595,9 +595,27 @@ kg_cred_set_initial_refresh(krb5_context context, krb5_gss_cred_id_rec *cred,
     set_refresh_time(context, cred->ccache, refresh);
 }
 
+struct verify_params {
+    krb5_principal princ;
+    krb5_keytab keytab;
+};
+
+static krb5_error_code
+verify_initial_cred(krb5_context context, krb5_creds *creds,
+                    const struct verify_params *verify)
+{
+    krb5_verify_init_creds_opt vopts;
+
+    krb5_verify_init_creds_opt_init(&vopts);
+    krb5_verify_init_creds_opt_set_ap_req_nofail(&vopts, TRUE);
+    return krb5_verify_init_creds(context, creds, verify->princ,
+                                  verify->keytab, NULL, &vopts);
+}
+
 /* Get initial credentials using the supplied password or client keytab. */
 static krb5_error_code
-get_initial_cred(krb5_context context, krb5_gss_cred_id_rec *cred)
+get_initial_cred(krb5_context context, const struct verify_params *verify,
+                 krb5_gss_cred_id_rec *cred)
 {
     krb5_error_code code;
     krb5_get_init_creds_opt *opt = NULL;
@@ -621,6 +639,11 @@ get_initial_cred(krb5_context context, krb5_gss_cred_id_rec *cred)
     }
     if (code)
         goto cleanup;
+    if (cred->password != NULL && verify != NULL) {
+        code = verify_initial_cred(context, &creds, verify);
+        if (code)
+            goto cleanup;
+    }
     kg_cred_set_initial_refresh(context, cred, &creds.times);
     cred->have_tgt = TRUE;
     cred->expire = creds.times.endtime;
@@ -632,7 +655,9 @@ cleanup:
 
 /* Get initial credentials if we ought to and are able to. */
 static krb5_error_code
-maybe_get_initial_cred(krb5_context context, krb5_gss_cred_id_rec *cred)
+maybe_get_initial_cred(krb5_context context,
+                       const struct verify_params *verify,
+                       krb5_gss_cred_id_rec *cred)
 {
     krb5_error_code code;
 
@@ -642,7 +667,7 @@ maybe_get_initial_cred(krb5_context context, krb5_gss_cred_id_rec *cred)
 
     /* Get creds if we have none or if it's time to refresh. */
     if (cred->expire == 0 || kg_cred_time_to_refresh(context, cred)) {
-        code = get_initial_cred(context, cred);
+        code = get_initial_cred(context, verify, cred);
         /* If we were trying to refresh and failed, we can keep going. */
         if (code && cred->expire == 0)
             return code;
@@ -652,11 +677,10 @@ maybe_get_initial_cred(krb5_context context, krb5_gss_cred_id_rec *cred)
 }
 
 static OM_uint32
-acquire_init_cred(krb5_context context,
-                  OM_uint32 *minor_status,
-                  krb5_ccache req_ccache,
-                  gss_buffer_t password,
+acquire_init_cred(krb5_context context, OM_uint32 *minor_status,
+                  krb5_ccache req_ccache, gss_buffer_t password,
                   krb5_keytab client_keytab,
+                  const struct verify_params *verify,
                   krb5_gss_cred_id_rec *cred)
 {
     krb5_error_code code;
@@ -741,7 +765,7 @@ acquire_init_cred(krb5_context context,
     }
 #endif
 
-    code = maybe_get_initial_cred(context, cred);
+    code = maybe_get_initial_cred(context, verify, cred);
     if (code)
         goto error;
 
@@ -759,6 +783,7 @@ acquire_cred_context(krb5_context context, OM_uint32 *minor_status,
                      OM_uint32 time_req, gss_cred_usage_t cred_usage,
                      krb5_ccache ccache, krb5_keytab client_keytab,
                      krb5_keytab keytab, const char *rcname,
+                     const struct verify_params *verify,
                      krb5_boolean iakerb, gss_cred_id_t *output_cred_handle,
                      OM_uint32 *time_rec)
 {
@@ -828,7 +853,7 @@ acquire_cred_context(krb5_context context, OM_uint32 *minor_status,
      */
     if (cred_usage == GSS_C_INITIATE || cred_usage == GSS_C_BOTH) {
         ret = acquire_init_cred(context, minor_status, ccache, password,
-                                client_keytab, cred);
+                                client_keytab, verify, cred);
         if (ret != GSS_S_COMPLETE)
             goto error_out;
     }
@@ -922,7 +947,8 @@ acquire_cred(OM_uint32 *minor_status, gss_name_t desired_name,
 
     ret = acquire_cred_context(context, minor_status, desired_name, password,
                                time_req, cred_usage, ccache, NULL, keytab,
-                               NULL, iakerb, output_cred_handle, time_rec);
+                               NULL, NULL, iakerb, output_cred_handle,
+                               time_rec);
 
 out:
     krb5_free_context(context);
@@ -1012,7 +1038,7 @@ kg_cred_resolve(OM_uint32 *minor_status, krb5_context context,
     }
 
     /* Resolve name to ccache and possibly get initial credentials. */
-    code = maybe_get_initial_cred(context, cred);
+    code = maybe_get_initial_cred(context, NULL, cred);
     if (code)
         goto kerr;
 
@@ -1181,7 +1207,10 @@ acquire_cred_from(OM_uint32 *minor_status, const gss_name_t desired_name,
     krb5_keytab client_keytab = NULL;
     krb5_keytab keytab = NULL;
     krb5_ccache ccache = NULL;
+    krb5_principal verify_princ = NULL;
     const char *rcname, *value;
+    struct verify_params vparams = { NULL };
+    const struct verify_params *verify = NULL;
     gss_buffer_desc pwbuf;
     gss_buffer_t password = NULL;
     OM_uint32 ret;
@@ -1265,10 +1294,33 @@ acquire_cred_from(OM_uint32 *minor_status, const gss_name_t desired_name,
         password = &pwbuf;
     }
 
+    ret = kg_value_from_cred_store(cred_store, KRB5_CS_VERIFY_URN, &value);
+    if (GSS_ERROR(ret))
+        goto out;
+    if (value != NULL) {
+        if (iakerb || password == NULL) {
+            /* Only valid if acquiring cred with password, and not supported
+             * with IAKERB. */
+            *minor_status = G_BAD_USAGE;
+            ret = GSS_S_FAILURE;
+            goto out;
+        }
+        if (*value != '\0') {
+            code = krb5_parse_name(context, value, &verify_princ);
+            if (code != 0) {
+                *minor_status = code;
+                ret = GSS_S_FAILURE;
+                goto out;
+            }
+        }
+        vparams.princ = verify_princ;
+        vparams.keytab = keytab;
+        verify = &vparams;
+    }
     ret = acquire_cred_context(context, minor_status, desired_name, password,
                                time_req, cred_usage, ccache, client_keytab,
-                               keytab, rcname, iakerb, output_cred_handle,
-                               time_rec);
+                               keytab, rcname, verify, iakerb,
+                               output_cred_handle, time_rec);
 
 out:
     if (ccache != NULL)
@@ -1277,6 +1329,7 @@ out:
         krb5_kt_close(context, client_keytab);
     if (keytab != NULL)
         krb5_kt_close(context, keytab);
+    krb5_free_principal(context, verify_princ);
     krb5_free_context(context);
     return ret;
 }
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -1182,6 +1182,8 @@ acquire_cred_from(OM_uint32 *minor_status, const gss_name_t desired_name,
     krb5_keytab keytab = NULL;
     krb5_ccache ccache = NULL;
     const char *rcname, *value;
+    gss_buffer_desc pwbuf;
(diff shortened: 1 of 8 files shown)


greghudson:
For historical context, see the thread starting at https://mailman.mit.edu/pipermail/krbdev/2012-July/011025.html and the wiki project page it references.

There was originally no password option for three reasons.  First, store options were intended to be read from config files, which perhaps shouldn't have passwords directly embedded in them (this rule isn't necessarily observed in the wider world; for instance, DB passwords show up directly in web app config files quite often).  Second, it conflates credential store location with the secrets required to obtain credentials--perhaps a problem already shared with the "client_keytab" input.  Third, a password input isn't general enough to encompass how krb5 initial credentials are acquired; this issue is of course shared with gss_acquire_cred_with_password().  Adding additional inputs can encompass additional scenarios like PKINIT, but in the general case a challenge-response framework is needed.

The first principle was espoused by Simo back in 2012, but obviously isn't a concern for him now.  Some people (Sam and Nico) did not find any of the concerns as important during the discussion.

gss_krb5_ccache_name() is thread-safe, and I wouldn't describe it as involving "heavy synchronization"; it uses a thread-local variable.  So this change isn't strictly necessary to allow an application to use gss_acquire_cred_with_password() with a specified ccache--although the change does of course allow the application to avoid using a krb5-specific function to do so.

frozencemetery:
Some additional context: Heimdal implements both `gss_krb5_ccache_name()` and "password" as a key to `gss_acquire_cred_from()`.

There's some amount of desire to have "GSSAPI kinit" due to our ability to interfere at the mechglue layer.  So while it's true that this doesn't capture the breadth of how credentials are acquired in Kerberos, perhaps that means we should look at how we would capture that better.

simo5:
@greghudson you bring up all valid points.

On my change of heart, after using the _from() interface, I appreciate a lot more the uniformity of using a 
single API for all the acquire functions, rather than having to hunt for specific ones.

The "hunting" point is especially painful when it comes to gss_krb5_ccache_name(), not a single interaction with users had them discover this function, they all tried to set KRB5CCNAME in the process environment.

On completeness, it is true that password authentication is only one of the methods to do 'kinit'. but it is still the most used by a large margin, so I decided that it is better to have a good enough interface than nothing. The good thing is that this patch does not preclude adding other methods (like passing a pkcs#11 URI for pkinit), so I think it is ok to introduce this now and more methods later in a "let's get the ball rolling" spirit.

HTH

simo5:
Also forgot to add that in my gss-ntlmssp mech I already support the "password" method, so this makes the experience consistent as it allows, in theory, to use SPNEGO and fallback from krb5 to ntlm should that be needed. I haven't tested if that will work yet, but we can fix that later if it comes up.

greghudson:
Has this been tested?  I don't think it will work, because acquire_init_cred() ignores req_ccache if a password is specified.

simo5:
I haven't tested specifying a cache indeed, hmmm

simo5:
I see that acquire_init_cred() wants to forcibly use a MEMORY cache when password is provided.
Is it ok to change this behavior ?
Or should this be handled at a higher level by copying the MEMORY ccache into a specified cache only if the cache was explicitly provided via cred_store ?

simo5:
As far as I can tell the only case where acquir_cred() will be passed both a ccache and a password would be from gss_acquire_cred_from() with this patch, all other code paths I see have ether NULL password argument or NULL ccache argument provided...

greghudson:
I think Simo is correct.  However:

* Although Heimdal implements "password" for gss_acquire_cred_from(), it errors out if "ccache" or "client_keytab" is set in the store alongside a password (see is_valid_password_cred_store()).  So it doesn't admit the desired functionality.

* If the store contains "password" and not "ccache", it's not clear to me whether the creds should wind up in a memory ccache or in the default ccache.

I also realized that the desired functionality can be implemented with gss_acquire_cred_with_password() and gss_store_cred_into().  (And gss_krb5_ccache_name() is irrelevant, because gss_acquire_cred_with_password() doesn't use it.)

simo5:
@nicowilliams what is the reasoning in Heimdal from erroring out if a ccache is provided?

@greghudson currently the code is basically just an other way to invoce gss_acquire_cred_with_password() through the gss_acquire_cred_from() interface, so it does what that function does when no ccache is provided. That said I think it would be valuable to use the default ccache with this interface, therefore emulating a kinit-like behavior.

simo5:
This are the specific semantics I have in mind:
- If a password is provided then passing in client_keytab should be considered an error in MIT too
- if no ccache is provided then the default ccache is used
- if a credentials exist in the selected ccache they are *removed* and replaced by new credentials obtained with the password provided.

Desiderata but open to discussion:
- if a "keytab" is provided then the credentials obtained will be validated by also acquiring a ticket to the first key in the keyab and then checking that the ticket can be decrypted with the same.

I realize this is a change in scope from what I proposed initially, but I think this would provide useful semantics and allow users to perform a very common operation in the safest possible way without descending into having to use dozen of krb5 calls.

greghudson:
> if no ccache is provided then the default ccache is used

While I think this is the most consistent answer given the other semantics, it conflicts with the current Heimdal behavior when only a password is provided.  (This behavior was implemented by @lhoward in January 2019, and I think is therefore shipped in a release).

> if a "keytab" is provided then the credentials obtained will be validated by also acquiring a ticket to the first key in the keyab and then checking that the ticket can be decrypted with the same.

This is an interesting idea, although it would be a departure from "keytab" only applying to acceptor creds.  The most natural implementation would not use the first key in the keytab, but would try all host/* keys in the keytab; see the current implementation of krb5_verify_init_creds().

> I think this would provide useful semantics and allow users to perform a very common operation in the safest possible way without descending into having to use dozen of krb5 calls.

For the basic functionality the current method is just gss_acquire_cred_with_password() and gss_store_cred_into(), neither of which are krb5 calls.  So I assume this statement is intended to include verification.

simo5:
> > if no ccache is provided then the default ccache is used
> 
> While I think this is the most consistent answer given the other semantics, it conflicts with the current Heimdal behavior when only a password is provided. (This behavior was implemented by @lhoward in January 2019, and I think is therefore shipped in a release).

I would like to learn from Heimdal maintainers if this was done on purpose or just happened.
I am not sure we should be tied to this implementation behavior though, heimdal can be changed to do what MIT will do (unless it was done on purpose for some reason that will be important to preserve and applies to MIT as well).

> > if a "keytab" is provided then the credentials obtained will be validated by also acquiring a ticket to the first key in the keyab and then checking that the ticket can be decrypted with the same.
> 
> This is an interesting idea, although it would be a departure from "keytab" only applying to acceptor creds. The most natural implementation would not use the first key in the keytab, but would try all host/* keys in the keytab; see the current implementation of krb5_verify_init_creds().

I think it is perfectly valid to try non-host keys. What we want here is just to validate that the credentials obtained can be verified, any service with access to their own keys should be allowed to do that. this is why I said first key.
Of course people can arrange keytabs so that the first key is the strongest/most appropriate host/ key available.
We could try to prefer host/ if present though I guess.

> > I think this would provide useful semantics and allow users to perform a very common operation in the safest possible way without descending into having to use dozen of krb5 calls.
> 
> For the basic functionality the current method is just gss_acquire_cred_with_password() and gss_store_cred_into(), neither of which are krb5 calls. So I assume this statement is intended to include verification.

Yes it is about all the 4 points taken together. A single API call to do full validated "kinit" in gssapi, fundamentally.

greghudson:
A couple of questions arise:

* If we preserve consistency with current Heimdal behavior for just "password" (contrary to Simo's preference), then is there a way to explicitly store into the default ccache?  Perhaps "ccache" with a zero-length value?

* Similarly, according to Simo's proposal, the presence of "keytab" causes validation that doesn't happen if you don't specify a keytab.  Is there a way to ask for the default keytab?  Again, perhaps with a zero-length value?

> I think it is perfectly valid to try non-host keys. What we want here is just to validate that the credentials obtained can be verified, any service with access to their own keys should be allowed to do that.

Prior to client keytabs, nothing depends on the order of entries in a keytab.  Because keytab entry deletion results in a hole that can be filled by an equal-or-smaller-sized entry, order can be unpredictable in a keytab that is managed over time, and we don't really have tools for manipulating the order of keytab entries.  I knew this when implementing client keytab support, but figured it wouldn't be much of a burden for people to create client keytabs with just the one desired principal in them.  Service keytabs are a less predictable animal; although it's probably best to have separate keytabs (with appropriate filesystem ownership) for each service, I don't think that is close to universal in practice.

krb5_verify_init_creds() does take a principal argument; if we're going to do verified credential acquisition in gss_acquire_cred_from(), it might be reasonable to have a key to indicate which principal to use for verification, with the default of using any host/* key.  (That default is traditionally safe because ownership of the host key is traditionally synonymous with root access, which is not true of other keys.)

simo5:
> A couple of questions arise:
> 
>     * If we preserve consistency with current Heimdal behavior for just "password" (contrary to Simo's preference), then is there a way to explicitly store into the default ccache?  Perhaps "ccache" with a zero-length value?

We could special case a value of 'default:'
I would like to avoid a zero lenght value because in higher level language/bindings it is sometimes easy to mistakenly pass an empty string for no string.

>     * Similarly, according to Simo's proposal, the presence of "keytab" causes validation that doesn't happen if you don't specify a keytab.  Is there a way to ask for verification with the default system keytab?  Again, perhaps with a zero-length value?

Again I would use 'default:' as special value to indicate that, if we have to.

> > I think it is perfectly valid to try non-host keys. What we want here is just to validate that the credentials obtained can be verified, any service with access to their own keys should be allowed to do that.
> 
> Prior to client keytabs, nothing depends on the order of entries in a keytab. Because keytab entry deletion results in a hole that can be filled by an equal-or-smaller-sized entry, order can be unpredictable in a keytab that is managed over time, and we don't really have tools for manipulating the order of keytab entries. I knew this when implementing client keytab support, but figured it wouldn't be much of a burden for people to create client keytabs with just the one desired principal in them. Service keytabs are a less predictable animal; although it's probably best to have separate keytabs (with appropriate filesystem ownership) for each service, I don't think that is close to universal in practice.
> 
> krb5_verify_init_creds() does take a principal argument; if we're going to do verified credential acquisition in gss_acquire_cred_from(), it might be reasonable to have a key to indicate which principal to use for verification, with the default of using any host/* key. (That default is traditionally safe because ownership of the host key is traditionally synonymous with root access, which is not true of other keys.)

Another option:
Rather then using the presence of 'keytab' as a special indicator we'd introduce a 'verify' key, that take a principal name OR the value 'default'/''/'some better term'.
If a principal name is specified then that is searched for in the keytab an used, if 'defalt' (etc) then the first key is used?
Passing the 'verify':'default' key/value w/o a 'keytab' value will attempt to verify against the default keytab. (therefore we do not need a special 'default' value for the 'keytab' case here)

greghudson:
"default" is a valid ccache name and a valid keytab name, which makes it a poor special case.  But I do take the point that null vs. empty is a common programming mistake.

simo5:
In one example I gave 'default:' as the name (note colon at then), I have not been consistent, but I believe the colon will prevent issues as it is a reserved character to define types.

frozencemetery:
Perhaps I'm misunderstanding, but: ':' is a valid filepath character.

```
(0) rharwood@eesha:/tmp$ touch default:
(0) rharwood@eesha:/tmp$ ls -l default:
-rw-r--r-- 1 rharwood rharwood 0 sep 21 17:41 default:
(0) rharwood@eesha:/tmp$
```

simo5:
While that is true the ambiguity can always be solved by using "FILE:default:", right ?

frozencemetery:
Technically yes.  I think the zero-length is slightly cleaner but that would work too.

simo5:
I am looking at the code to implement the additional proposed behavior and there is a case that stands out. that is when cred_usage is set to GSS_C_BOTH (I will return an error if set to GSS_C_ACCEPT only).
In the GSS_C_BOTH case the code will try to acquire creds from the default keytab (or the provided keytab) searching for a principal as specified in desired_name (which is normally a user principal when trying to acquire a cred with password ...
I wonder if I should try to do something smart or just error out also if GSS_C_BOTH is specified and basically allow only GSS_C_INITIATE.

simo5:
this is still WIP (compiles but untested), however I'd like a comment from @greghudson or @frozencemetery about the dirction before I do more testing and start working on additions for make check.

greghudson:
I was hoping we'd get input from @nicowilliams on the caching semantics.

I apparently misread the meaning of the proposed 'verify' key when I gave feedback before.  If 'verify' names a principal (not a keytab name), then there isn't really a good way to indicate a default, other than the empty string.  There are some obviously degenerate principal names like `@` but they are all awkward.  I would prefer if the semantics for default were to use any host key, just as krb5_verify_init_creds() does with a null server argument, and not the first key in the keytab.

The emergent semantics for GSS_C_BOTH (i.e. what will naturally happen if we don't make a special effort to detect it) seem reasonable to me.

simo5:
@greghudson so it looks like all I need to do is to skip krb5_parse_name() if verify == "" ?

simo5:
Ok updated the code to allow an empty value for verify, this will pas NULL to krb5_verify_init_creds() instead of a service name.

frozencemetery:
The use of an empty string for verify raises a consistency issue - now verify and ccache use different techniques for specifying default.

simo5:
> The use of an empty string for verify raises a consistency issue - now verify and ccache use different techniques for specifying default.

True but they represent different things.
An empty verify represents "not choosing", while the ccache 'default:' value is about explicitly choosing a specific ccache (whatever the system considers the default).

I guess we could argue that an empty principal is a bit like saying "choose the default principal", but there is no default principal, it fully depends on what's in the keytab (which does not need to be specified as 'default:' only because that's already the behavior krb5_verify_init_creds() takes if no keytab is specified...

But I am willing to drop 'default:' and allow empty ccache value if this is perceived as a major consistency issue.

greghudson:
It would be nice if we could verify the creds at the point they were retrieved by krb5_get_init_creds_password() instead of fetching them back out of the ccache.  I think that's possible--add a verify argument to maybe_get_initial_cred() and get_initial_cred(), and do the verification inside get_initial_cred() after the gic_pwd call.

If that doesn't work out, verify_init_cred() with its current contract doesn't need to copy the client princ to fetch the TGT, since the lifetime of the mcred structure is just this one function.  Use a separate local variable to hold the TGT name so that mcred contains only aliases and doesn't get freed.

This doesn't appear to work with IAKERB, which is fine, but perhaps we should check and return an error.

Are you still planning to add test cases?

simo5:
@greghudson I think I found a way to do verification as you ask.
I am not sure what should I do in the IAKERB case.
Just detect the mech in acquire_cred_context and immediately return an error if verification is requested ?

I am still planning on writing test cases, yes

greghudson:
I think we have a more fundamental problem with gss_acquire_cred_from() and IAKERB.  Unlike the regular gss_acquire_cred() and gss_acquire_cred_with_password(), we don't have a separate entry point for IAKERB for gss_acquire_cred_from().  That means we could go off and talk to the KDC instead of using IAKERB (currently for the client keytab, and in the future for the password).

So, we need a separate iakerb_gss_acquire_cred_from(), and to turn the current krb5_gss_acquire_cred_from() into a helper that takes an iakerb parameter.  It can error out if iakerb is true and verify is set.

simo5:
> So, we need a separate iakerb_gss_acquire_cred_from(), and to turn the current krb5_gss_acquire_cred_from() into a helper that takes an iakerb parameter. It can error out if iakerb is true and verify is set.

Does this need to be part of this PR?
Or can we make a separate change for that?

greghudson:
We can do it within this PR or prior to it.  I don't want to do it afterwards, since we'd be adding to brokenness.

simo5:
I pushed as revision with the change API to do verification within get_initial_cred()
I also added a set of tests

the last test fails, it indicate I may be doing something wrong with the cache, however I wanted to push this to get feedback on this revision.

I will take care of IAKERB as well as fixing the last test in the next few days.

simo5:
I am adding iakerb passthrough and looking at the code to see where this info is used I found this
https://github.com/krb5/krb5/blob/6dbf0b755bb1d8254c881cf21717441cc2e880b8/src/lib/gssapi/krb5/acquire_cred.c#L640

Shouldn't this check return an error ?
I am not sure what will make this function fail if the mech is iakerb otherwise.
Do we depend on the kg_cred_resolve() call in acquire_cred_context() to fail?

greghudson:
> Shouldn't this check return an error ?

No.  IAKERB can get initial credentials with a password via the mech.  What we want to make fail is if the caller asks for IAKERB and validation.

simo5:
Ok I changed the last patch to explicitly exclude IAKERB if verification is requested.
@greghudson let me know if any other change is necessary.

src/lib/gssapi/krb5/acquire_cred.c
```diff
@@ -595,9 +595,16 @@ kg_cred_set_initial_refresh(krb5_context context, krb5_gss_cred_id_rec *cred,
     set_refresh_time(context, cred->ccache, refresh);
 }
 
+struct _init_verify_creds {
+    krb5_principal srvprinc;
+    krb5_keytab keytab;
+} kg_verify_creds;
```
greghudson, line 601: My inclination is that it would be better to just pass two arguments, but maybe I'm missing a complication there.

simo5: Hi @greghudson ,
I did not want to bloat the APIs by passing multiple separate arguments.

Also when I started working on this I was not 100% sure the two members of the structure were sufficient, so the structure allowed me flexibility to add, remove, change things w/o changing all the intermediate functions that simply pass along this argument without touching it.

Finally the struct allows me the simple option to pass a NULL pointer as indication that verification is not needed. Without it I will have to pass a third boolean option through all these APIs, because it is legal to pass NULL for both the srvprinc and keytab arguments and still want verification, ie when no "keytab" key is present in the store and "verify" is the empty string.

Because of this I would prefer to keep this internal struct, but I am also ok to convert it to three additional argument to pass along if you request it.

greghudson: Understood.

I made some changes (keeping the structure); do they seem okay?

simo5: Yeah all changes make a lot of sense.
Thanks!

greghudson:
Sorry for the delay.  I've done some rebasing; I put the IAKERB fix first in the sequence, and I squashed the password and password+ccache commits.

I think I agree with Robbie that an empty string is probably better than "default:" given that we're using the empty string already.  I can make that change.

What happens if you specify a password value and no desired key?  I ran into the comment saying that the ccache has to match the desired_name if one is given, and that made me wonder about the semantics when one isn't given.  gss_acquire_cred_with_password() errors out at the mechglue layer if you pass in a null name.  Heimdal appears to (in this case and for gss_acquire_cred_with_password()) get the client principal of the default ccache or fall back to the login username.

I'll need to write documentation before merging this.  Unfortunately we never documented gss_acquire_cred_from() so I'll be starting from scratch on that.

greghudson:
> I think I agree with Robbie that an empty string is probably better than "default:" given that we're using the empty string already. I can make that change.

Another option here is not to have a way to name the default ccache, for now.

There are a lot of not-yet-implemented cache collection semantics in the way of making a GSS kinit-alike (on top of the preauth issues).  Since the default ccache might be a collection, we might not want to make it easy to acquire creds into it until we implement some of those semantics.

frozencemetery:
> I'll need to write documentation before merging this. Unfortunately we never documented gss_acquire_cred_from() so I'll be starting from scratch on that.

python-gssapi has some [documentation](https://github.com/pythongssapi/python-gssapi/blob/main/docs/source/credstore.rst) of options, though I'm sure it's a bit dated now, if that's helpful.

simo5:
@greghudson I am, ok "default:" -> "" wise if you find it preferable.
Also ok with the other code changes.
As for password with missing name, I think it would be prudent for now to fail if desired_name is not provided. Although in theory I can see how gssapi could simply use the name of a provided ccache, it is likely that any app that would want to do that would end up inquiring creds anyway to present a user with a reasonable prompt. And if they do that, they have a name to pass into desired_name anyway.

greghudson:
I removed the `default:` handling for ccache as I discussed above, and added a check for desired_name being set.  I folded the tests into the appropriate commits, moved them to a new script t_credstore.py (along with the few existing tests), and fleshed them out.

I ran into a big problem with the password+ccache+verify combination, which is the combination of a few issues:

1. With desired_mechs=GSS_C_NO_OID_SET, we attempt to get creds for both krb5 and SPNEGO->krb5.
2. If we get creds and fail to verify them, the creds are still stored in the ccache.
3. If we already have creds in the ccache and they aren't close to expiring, we don't try to get new ones and we don't verify them.

So if verification fails under krb5, cred acquisition still succeeds under SPNEGO->krb5 and the overall operation appears successful to the caller.  (If ccache is not set, we acquire two sets of creds into separate memory ccaches, one for krb5 and one for SPNEGO->krb5, and verify both them.  Not great but not broken.)

Fixing (1) is out of scope for the PR, but (2) and (3) seem like bugs.  (2) is unfortunately annoying to fix with current libkrb5 APIs; we'd need to fetch the initial creds into a memory cache, verify them, and then copy out the cache.

I was hoping to close this out and get it into 1.19 (which I wanted to branch for at the beginning of November), but this is a big wrinkle.  The easiest resolution is to drop password+ccache, leaving it to the application to store the creds with gss_store_cred_into().

simo5:
> The easiest resolution is to drop password+ccache, leaving it to the application to store the creds with gss_store_cred_into().

This would defeat the purpose IMO.

simo5:
I can take over the annoying task of fixing (2) by doing the ccache copy dance if you want.

simo5:
It would fix (3) as well, because we always use an empty ccache, so verification is always done regardless.

greghudson:
> It would fix (3) as well, because we always use an empty ccache, so verification is always done regardless.

If we use a memory cache as an intermediary within get_initial_cred(), then we don't solve (3), because we currently don't arrive at get_initial_cred() if we scanned the ccache and found creds.

If we use a memory cache for the cred object as a whole, then the ccache value in the cred store stops meaning "use this ccache for the GSSAPI handle" as it means for other acquire_cred operations, and starts meaning "incidentally, store the acquired creds into this ccache and then forget about it".

We could create a memory cache for the handle and then, during get_initial_cred(), switch the handle to point to the named ccache.  That doesn't feel very elegant to me, especially with a maybe_ call in the way.

> This would defeat the purpose IMO.

I may have lost track of the purpose.  The original stated purpose was to be able to acquire creds with a password into a named ccache without using mech-specific APIs.  That turned out to be already possible (acquire_cred_with_password + store_cred_into), at which point the purpose became verifying initial creds without using mech-specific APIs and storing them into a named ccache.  Narrowing the purpose to verifying initial credentials still greatly simplifies the process of fetching verified initial credentials into a named ccache (two generic API calls down from a dozen or so).

simo5:
Ok, I guess I can buy it this way.
And seem the complications of doing it all in one call are really truly ugly.
So the code now must bail out if a "ccache" key is passed in when "password" is also provided ?

greghudson:
> So the code now must bail out if a "ccache" key is passed in when "password" is also provided ?

Yes, as Heimdal currently does.

simo5:
Or should we just ignore the "ccache" ?
That would allow to gss_acquire_cred_from + gss_stored_cred_into using the same cred_store structure for both calls, but I do not know if the convenience of preparing a single cred_store is worth the potential confusion of allowing a ccache key to be passed in and then ignoring it.

simo5:
> > So the code now must bail out if a "ccache" key is passed in when "password" is also provided ?
> 
> Yes, as Heimdal currently does.

Consistency with Heimdal would be a good point in favor of a hard bail.

greghudson:
I made the changes discussed, and added a verify test case for verification failure.

simo5:
We need to change the second commit message, it still says you can pass a ccache option to store acquired credentials in it.

simo5 (commented):
All code changes look good to me!

greghudson:
I fixed the second commit message, and added a final commit documenting gss_acquire_cred_from() and gss_store_cred_into().

simo5 (commented):
LGTM

PR #1116, opened 2020-09-17
Allow aliases in U2U 2nd ticket matching

iboukris:
As mentioned in #1106, see bug below. Note that I don't have a way to test it, i'll ask Dario:
https://bugzilla.redhat.com/show_bug.cgi?id=1836630

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -648,7 +648,7 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
          */
         krb5_enc_tkt_part *t2enc = request->second_ticket[st_idx]->enc_part2;
         krb5_principal client2 = t2enc->client;
-        if (!krb5_principal_compare(kdc_context, request->server, client2)) {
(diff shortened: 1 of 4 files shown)


iboukris:
Got feedback from Dario, it looks good, updated the PR.

greghudson:
Just checking if I follow correctly on is_client_alias():

* Originally it took a DB entry (containing the canonical principal name) and a principal name to be looked up.
* It was changed to take two principal names (one canonical, the other to be looked up) because the proposed is_local_principal() change in PR #1106 needed to use tgs_server (a principal name) as the canonical principal to compare against.
* Since we expect to stop using tgs_server, we can go back to taking a DB entry.

iboukris:
Yep exactly, i like the db entry better because the canon is implied.

greghudson:
In the tests, does the "Try u2u against the client user" test anything specific to this commit, or is it leftover copy-paste?  (The klist invocation seems like leftover, and in fact I'm not sure why it's there in the existing tests.)

iboukris:
Yeah, mostly copy-paste.

PR #1118, opened 2020-09-21
Add doc build to Github Actions config

greghudson:
Add a second workflow to build documentation, with the HTML output as
a generated artifact.  Skip the doc workflow if no documentation files
were changed.  Skip the existing build workflow if no source files
were changed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,8 @@
 name: Build
 
-on: [push, pull_request]
+on:
+    push: {paths: [src/**, .github/workflows/build.yml]}
+    pull_request: {paths: [src/**, .github/workflows/build.yml]}
 
 jobs:
 
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,0 +1,27 @@
+name: Doc
+
+on:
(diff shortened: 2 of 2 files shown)


greghudson:
If you expand "Show all checks", click on details for the doc build, and then find the "Artifacts" drop-down, you can download a zip file of the generated HTML (within seven days of the last PR push).  I would prefer if it could be easy to just browse the resulting HTML documentation for a doc change, but that is still easier than my old process.

This should also catch semi-frequent bugs where we make a change to krb5.hin and don't make a corresponding change to doc/appdev/refs/*/index.rst.

frozencemetery (approved):
Nice, that's handy

frozencemetery:
It might be possible to adapt the github pages workflow for this.  Our automation for updating docs is https://github.com/pythongssapi/python-gssapi/blob/master/.github/workflows/deploy-docs.yml#L17-L26 (and similar logic for a released version).  I think it would be possible to make (e.g.) a separate folder for each PR.  This does have the distinct disadvantage of crudding up the repo with a gh-pages branch, though.

greghudson:
We could do it in a different repo, but that's starting to feel heavyweight.  I'll push what I have here for now.

PR #1119, opened 2020-09-22
Fix doc issues with newer Doxygen and Sphinx

greghudson:
I found several issues with the doc build on Ubuntu 20.04.  I think all of them stem from the doxygen 1.8.17 package.  I am not sure what, if anything, I will merge to address these issues, but I want a place to document the investigation.

* Seemingly harmless whitespace issues in comments cause doxygen to emit `<para><linebreak/>\n </para>` within the `<detaileddescription>` element.  A bug in our custom doxygen-to-rst script emits `\n**\n` for these, causing a Sphinx warning about unterminated inline markup.  (I don't fully understand the bug in our scripts; there are lots of `<linebreak/>`s emitted by some of the OTP responder comments, with either new or old Doxygen, and they don't cause the same output.)  The PR currently contains whitespace changes to avoid the problem.

* Python issues a warning about using `is` to compare against the empty string.  The PR has a fix.

* This version of Doxygen misrecognizes `typedef const krb5_principal_data *krb5_const_principal;` as a variable declaration rather than a typedef, so the corresponding file doesn't get generated (our script skips variable declarations).  This bug is fixed in Doxygen 1.8.18.  We could work around it with `typedef krb5_principal_data const *krb5_const_principal;` but that doesn't read well.  This problem is not currently addressed in the PR.

* doc/_templates/layout.html uses css_files, causing the warning `builder.css_files is deprecated. Please use app.add_stylesheet() instead`.  This problem is not currently yet addressed in the PR.

There is a doxygen-to-RST bridge called Breathe; if it can work for us then it might be better than our custom script (and would let us drop the dependency on Cheetah and lxml).

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -2,7 +2,6 @@
 {% set rellinks = [('search', 'Enter search criteria', 'S', 'Search')] +
                   rellinks +
                   [('index', 'Full Table of Contents', 'C', 'Contents')] %}
-{% set css_files = css_files + ["_static/kerb.css"] %}
 
 {# Add a "feedback" button to the rellinks #}
 {%- macro feedback_rellinks() %}
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -98,6 +98,9 @@ pygments_style = 'sphinx'
 
 # -- Options for HTML output ---------------------------------------------------
 
+def setup(app):
+    app.add_stylesheet('kerb.css')
+
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 # html_theme = 'default'
@@ -138,10 +141,6 @@ html_static_path = ['_static']
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
 # using the given strftime format.
 #html_last_updated_fmt = '%b %d, %Y'
-
-# If true, SmartyPants will be used to convert quotes and dashes to
-# typographically correct entities.
-html_use_smartypants = True
 
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}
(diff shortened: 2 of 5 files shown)


greghudson:
Initial, and possibly final, notes on Breathe, after trying it for one arbitrarily selected function (krb5_build_principal):

* We would probably still need some scripting to generate small .rst files for each function/type/macro.

* For functions, it wants to parse the function declaration, and it doesn't like the KRB5_CALLCONV prefix (as it has no access to the macro definition).

* It didn't seem to generate cross references for the argument types, as our current scripting does.

* It seemed to stumble over the first paragraph after the parameters, putting it above the parameter list, improperly formatting it, and not generating a reference to krb5_free_principal().  The rest of the text looks okay.

* Our current scripting seems to have punted a `@code` block containing an example, but Breathe outputs it.

greghudson:
I pushed some additional changes:

* Address the css_files deprecation, replacing it with a conf.py setup() function to call add_stylesheet().  This dates back to Sphinx 1.0 so shouldn't present any compatibility issue.  (Sphinx 1.8 and later have a slightly simpler way, but Ubuntu 18.04 ships Sphinx 1.6.x.)

* Removed the html_use_smartypants setting as it is deprecated and corresponds to the default behavior.

* Get rid of the explicit linebreaks in the JSON schemas, instead relying on markdown support (added in Doxygen 1.8, which is now old enough) to introduce a verbatim block.  Add support for verbatim tags in doxybuilder_types.rst.  Remove support for linebreak tags as there are no more linebreak tags in the Doxygen output and the code for it didn't really make sense.

This just leaves the Doxygen bug affecting krb5_const_principal typedef.  We can either work around that at the cost of readability or just skip Ubuntu 20.04 as a doc building platform.

greghudson:
I wrote a proper commit message and removed the WIP designation.

frozencemetery (approved):

PR #1120, opened 2020-09-25
Minimize usage of tgs_server in KDC

greghudson:
Where possible, use the realm of the request server principal
(canonicalized via KDB lookup, if available) in preference to
tgs_server.  This change facilitates alias realm support and potential
future support for serving multiple realms from the same KDB.

S4U2Self local user testing currently uses the uncanonicalized request
realm after this change, which will require attention for alias realm
support.

FAST armor ticket checking is unaffected by this change (it still
compares against tgs_server).  This check poses no issue for realm
aliases, as both tgs_server and the armor ticket server should have
canonical realms, but it will require attention for multi-realm KDB
support.

Remove is_local_principal() as it is no longer used.  Add an
is_local_tgs_principal() helper and shorten is_cross_tgs_principal().

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -267,7 +267,7 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
         goto cleanup;
     }
 
-    if (!is_local_principal(kdc_active_realm, header_ticket->server))
+    if (!data_eq(header_server->princ->realm, sprinc->realm))
         setflag(c_flags, KRB5_KDB_FLAG_CROSS_REALM);
     if (is_referral)
         setflag(c_flags, KRB5_KDB_FLAG_ISSUING_REFERRAL);
@@ -292,6 +292,15 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
         au_state->status = status;
         kau_s4u2self(kdc_context, errcode ? FALSE : TRUE, au_state);
         au_state->s4u2self_user = NULL;
+    }
+
+    /* Aside from cross-realm S4U2Self requests, do not accept header tickets
+     * for local users issued by foreign realms. */
+    if (s4u_x509_user == NULL && data_eq(cprinc->realm, sprinc->realm) &&
+        isflagset(c_flags, KRB5_KDB_FLAG_CROSS_REALM)) {
+        krb5_klog_syslog(LOG_INFO, _("PROCESS_TGS: failed lineage check"));
+        retval = KRB5KDC_ERR_POLICY;
+        goto cleanup;
     }
 
     if (errcode)
@@ -565,13 +574,12 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
 
     /*
      * Only add the realm of the presented tgt to the transited list if
-     * it is different than the local realm (cross-realm) and it is different
+     * it is different than the server realm (cross-realm) and it is different
      * than the realm of the client (since the realm of the client is already
      * implicitly part of the transited list and should not be explicitly
      * listed).
      */
-    /* realm compare is like strcmp, but knows how to deal with these args */
-    if (krb5_realm_compare(kdc_context, header_ticket->server, tgs_server) ||
+    if (!isflagset(c_flags, KRB5_KDB_FLAG_CROSS_REALM) ||
         krb5_realm_compare(kdc_context, header_ticket->server,
                            enc_tkt_reply.client)) {
         /* tgt issued by local realm or issued by realm of client */
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -78,12 +78,6 @@ static krb5_error_code find_server_key(krb5_context,
                                        krb5_kvno, krb5_keyblock **,
                                        krb5_kvno *);
 
-krb5_boolean
-is_local_principal(kdc_realm_t *kdc_active_realm, krb5_const_principal princ1)
-{
-    return krb5_realm_compare(kdc_context, princ1, tgs_server);
-}
-
 /*
  * Returns TRUE if the kerberos principal is the name of a Kerberos ticket
  * service.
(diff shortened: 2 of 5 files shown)


greghudson:
This should make PR #1106 easier, but does not actually implement alias realm support.  At least two changes will still be needed for alias realm support: moving get_local_tgt() to after the server lookup in process_tgs_req(), and checking for realm aliases in S4U2Self local principal testing.  My current belief is that we'll have to do a krbtgt/realm@realm lookup for the latter, if the realm doesn't textually match the server realm.

For multi-realm KDB support we will need to address the FAST armor ticket check.  I considered just checking for any TGS principal, but RFC 6113 wants the armor ticket to be a TGT to the target realm.  If we have a realm alias check (which we'll need for S4U2Self), we can check if the request realm is an alias for the target realm; this is less efficient than waiting for the server lookup, but in most cases I think the request realm will textually match the ticket realm and we won't have to do a DB lookup.

There is one more use of tgs_server that I hadn't analyzed: in make_too_big_error() we stuff it into the errpkt.server field.  I don't think this is really important since the client shouldn't generally look at that field for a too-big error.

iboukris:
> At least two changes will still be needed for alias realm support: moving get_local_tgt() to after the server lookup in process_tgs_req(), and checking for realm aliases in S4U2Self local principal testing. My current belief is that we'll have to do a krbtgt/realm@realm lookup for the latter, if the realm doesn't textually match the server realm.

I'm convinced as well, although as I think of it as S4U2Self lineage check, i wonder if we'd need the same in regular requests, I'll try to write it down on to get a better idea.

greghudson:
> I'm convinced as well, although as I think of it as S4U2Self lineage check, i wonder if we'd need the same in regular requests, I'll try to write it down on to get a better idea.

I think you're right, so that would be a third use for a realm-alias check.

src/kdc/do_tgs_req.c
```diff
@@ -294,6 +294,16 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
         au_state->s4u2self_user = NULL;
     }
 
+    /* Aside from cross-realm S4U2Self requests, do not accept header tickets
+     * for local users issued by foreign realms. */
+    if (s4u_x509_user == NULL &&
+        data_eq(cprinc->realm, server->princ->realm) &&
+        !data_eq(header_server->princ->realm, server->princ->realm)) {
```
iboukris, line 301: We could use KRB5_KDB_FLAG_CROSS_REALM here too.

iboukris:
btw, it looks like kdc_active_realm gets setup by req->server, so i'm not so clear why we need to get rid of.

greghudson:
> btw, it looks like kdc_active_realm gets setup by req->server, so i'm not so clear why we need to get rid of.

Every so often, a person wants to create an environment with a large and changing set of small realms.

Currently we have a semblance of multi-realm support in krb5kdc (but not kadmind), with multiple instances of the -r option.  It is limited to 32 realms,  you have to restart krb5kdc each time the set of realms changes, and the realm of the request has to exactly match one of the -r options.  We could relax those restrictions, but some of them might be hard to address, and there is still kadmind.

If the KDC is only serving one realm, it doesn't look at the request realm in dispatch, but just dispatches it to the one realm unconditionally.  This raises the possibility of serving multiple realms out of the same KDB.  That scenario is simultaneously limiting (because you can't configure any of the realms to have different realm options, or to be stored and backed up separately from the other realms) and simpler to administer.  There is of course still kadmind to consider, but it might be easy enough to do the same thing in kadmind.  We also lack administrative tools to add and remove realms from a KDB, particularly the special principals like krbtgt/REALM.

src/kdc/tgs_policy.c
```diff
@@ -252,19 +252,21 @@ check_tgs_s4u2proxy(kdc_realm_t *kdc_active_realm,
...
             *status = "NO_2ND_TKT";
             return KDC_ERR_BADOPTION;
         }
-        /* Check that second ticket is a TGT. */
-        if (!krb5_principal_compare(kdc_context,
-                                    req->second_ticket[0]->server,
-                                    tgs_server)) {
+        /* Check that second ticket is a TGT to the server realm. */
+        second_server_princ = req->second_ticket[0]->server;
+        if (!is_local_tgs_principal(second_server_princ) ||
```
iboukris, line 268: I'm curios what so special about local (root) tgts, say in this case what's wrong with a non-local tgt? The client can still be non-local, that is could ask for a "local" tgt, using its non-local one and proceed, so why local?

iboukris (approved):
Looks good otherwise, thanks!

src/kdc/do_tgs_req.c
```diff
@@ -565,13 +574,13 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
 
     /*
      * Only add the realm of the presented tgt to the transited list if
-     * it is different than the local realm (cross-realm) and it is different
+     * it is different than the server realm (cross-realm) and it is different
      * than the realm of the client (since the realm of the client is already
      * implicitly part of the transited list and should not be explicitly
      * listed).
      */
     /* realm compare is like strcmp, but knows how to deal with these args */
```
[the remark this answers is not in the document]
greghudson: About realm compare?  krb5_realm_compare() is still used once in the conditional.

src/kdc/tgs_policy.c
greghudson: I don't know if there's a security reason.  In user-to-user the second ticket is typically received from the other party (without its session key) before being presented to the KDC.  If the request is made using other party's realm, one would expect the other party to have simply acquired a local TGT.

src/kdc/do_tgs_req.c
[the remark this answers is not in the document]
iboukris: Ah right.

src/kdc/tgs_policy.c
iboukris: Ack, thanks for clarifying it.

src/kdc/kdc_util.c
```diff
@@ -265,15 +237,6 @@ kdc_process_tgs_req(kdc_realm_t *kdc_active_realm,
...
     }
 
-    /* make sure the client is of proper lineage (see above) */
-    if (foreign_server && !has_s4u2self_padata(request->padata) &&
-        is_local_principal(kdc_active_realm, ticket->enc_part2->client)) {
-        /* someone in a foreign realm claiming to be local */
-        krb5_klog_syslog(LOG_INFO, _("PROCESS_TGS: failed lineage check"));
-        retval = KRB5KDC_ERR_POLICY;
-        goto cleanup_authenticator;
-    }
```
frozencemetery, line 275: I'm afraid I don't see why it's okay to remove this check; do you mind clarifying?

greghudson: This check is moved to do_tgs_req.c line 297, after the kdc_process_s4u2self_req() call, where we know in a more natural way whether the request is an S4U2Self request and where we have the canonicalized server principal.  I will call that out in the commit message.

Where this check should live is an open question.  I don't think it really ever belonged in kdc_process_tgs_req() since it isn't about decoding the header authenticator.  (That function also contains an FX_ARMOR check which I think belongs elsewhere.)  Putting it in do_tgs_req.c makes a long function longer, which I don't like, but validate_tgs_request() doesn't have quite the right inputs (whether this is a cross-realm request, whether this is an S4U2Self request) and I'm still considering whether there is a natural way to change that.  I would prefer not to hold up this PR for that analysis, instead perhaps doing some refactoring work after the fact.

frozencemetery: Thanks, I think I understand now.  I agree it's not worth holding up the PR.

frozencemetery (approved):

PR #1121, opened 2020-09-30
Avoid passing DB entry structures in KDC

greghudson:
When validating AS or TGS requests, pass pointers to DB entry
structures, not the structures themselves.

### DIFF
--- a/src/kdc/do_as_req.c
+++ b/src/kdc/do_as_req.c
@@ -644,8 +644,8 @@ process_as_req(krb5_kdc_req *request, krb5_data *req_pkt,
     au_state->stage = VALIDATE_POL;
 
     if ((errcode = validate_as_request(kdc_active_realm,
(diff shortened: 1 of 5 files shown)


greghudson:
This just fixes a long-standing annoyance.  It will need merging with some recent PRs but the merge should be easy.  I initially started out using `const krb5_db_entry *` pointers, but krb5_db_check_policy_as() and krb5_db_check_policy_tgs() do not accept const pointers, and constifying those turns out to be a lot of churn.

As a separate cleanup, I would like to make some KDC functions accept `krb5_context context` parameters instead of kdc_active_realm parameters, such as kdc_process_for_user().  But that should probably wait until PR #1106 is done, since there will be more candidates after there is less usage of `tgs_server`.

frozencemetery (approved):

PR #1122, opened 2020-10-05
Move more KDC checks to validate_tgs_request()

greghudson:
[This will require a merge with PR #1106 because of the moved lineage and S4U2Self checks, but it shouldn't be too difficult.]

[I would have liked to move the FX_ARMOR check from kdc_process_tgs_req(), but unfortunately RFC 6113 requires us to check the authenticator authdata as well as the ticket authdata, and it doesn't seem worth it to preserve the authenticator beyond kdc_process_tgs_req().]

Move the following validity checks into validate_tgs_request():

* the INVALID ticket flag check from kdc_process_tgs_req()
* the lineage check from process_tgs_req()
* all S4U2Self validity checks from kdc_process_s4u2self_req()

In process_tgs_req(), add an is_crossrealm boolean to for convenience.

There are some minor behavior changes:

* For invalid S4U2Self request options, the status string is changed
  from "INVALID AS OPTIONS" to "INVALID S4U2SELF OPTIONS".

* For a header ticket with the INVALID flag, the reply code is changed
  to KRB_AP_ERR_TKT_NYV (as specified in RFC 4120) and the status
  string to "TICKET NOT VALID".

* For a lineage check failure, the explicit KDC log is removed, and
  the status string is changed to "INVALID LINEAGE".

* e_data returned from the KDB check_policy_as() method will be
  included in the error for S4U2Self requests.

### DIFF
--- a/src/kdc/do_tgs_req.c
+++ b/src/kdc/do_tgs_req.c
@@ -78,8 +78,8 @@ prepare_error_tgs(struct kdc_request_state *, krb5_kdc_req *,krb5_ticket *,int,
                   krb5_principal,krb5_data **,const char *, krb5_pa_data **);
 
 static krb5_error_code
-decrypt_2ndtkt(kdc_realm_t *, krb5_kdc_req *, krb5_flags, krb5_db_entry **,
-               krb5_keyblock **, const char **);
+decrypt_2ndtkt(kdc_realm_t *, krb5_kdc_req *, krb5_flags, const krb5_ticket **,
+               krb5_db_entry **, krb5_keyblock **, const char **);
 
 static krb5_error_code
 gen_session_key(kdc_realm_t *, krb5_kdc_req *, krb5_db_entry *,
@@ -112,7 +112,7 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
     krb5_kdc_rep reply;
     krb5_enc_kdc_rep_part reply_encpart;
     krb5_ticket ticket_reply, *header_ticket = 0;
-    int st_idx = 0;
+    const krb5_ticket *stkt = NULL;
     krb5_enc_tkt_part enc_tkt_reply;
     int newtransited = 0;
     krb5_error_code retval = 0;
@@ -133,7 +133,7 @@ process_tgs_req(krb5_kdc_req *request, krb5_data *pkt,
     krb5_pa_s4u_x509_user *s4u_x509_user = NULL; /* protocol transition request */
     krb5_authdata **kdc_issued_auth_data = NULL; /* auth data issued by KDC */
     unsigned int c_flags = 0, s_flags = 0;       /* client/server KDB flags */
-    krb5_boolean is_referral;
+    krb5_boolean is_referral, is_crossrealm;
     const char *emsg = NULL;
     krb5_kvno ticket_kvno = 0;
     struct kdc_request_state *state = NULL;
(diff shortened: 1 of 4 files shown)


frozencemetery (approved):

greghudson:
I originally left kdc_process_s4u2proxy_req() out of the mix, but on a second look there are some validity checks there that I think belong better in tgs_policy.c, and can be moved there if we decrypt the second ticket before validating.  With those checks moved, kdc_process_s4u2proxy_req() is reduced to the simpler problem of determining whether the KDB authorizes the request and who the subject client is for a cross-realm request.

When comparing the old checks to the new ones, be aware that kdc_process_s4u2req() uses Microsoft terminology for its parameters ("proxy" for the target server, "server" for the intermediate server), which is counter-intuitive from the perspective of do_tgs_req.c (to whom the target server is "server" and the intermediate server is the header ticket client).  Specifically:

* The parameter `proxy` is named `server` in the caller.
* The parameter `proxy_princ` is `request->sever` in the caller.
* The parameter `server` is `stkt_server` in the caller (the evidence ticket server).
* The parameter `server_princ` is `header_ticket->enc_part2->client` in the caller.

src/kdc/tgs_policy.c
```diff
@@ -237,40 +244,168 @@ check_tgs_times(krb5_kdc_req *req, krb5_ticket_times *times,
...
+         !data_eq(stkt_server->princ->data[1], server->princ->realm) ||
+         !krb5_principal_compare(kdc_context, stkt->enc_part2->client,
+                                 tkt->enc_part2->client))) {
+        *status = "XREALM_EVIDENCE_TICKET_MISMATCH";
+        return KDC_ERR_BADOPTION;
+    }
+
+    /* For an initial or same-realm request, the second ticket server and
+     * header ticket client must be the same principal. */
+    if (!is_crossrealm && !is_client_db_alias(kdc_context, stkt_server,
```
iboukris, line 382: Here I have a slight preference for something like this (or just the comment):
```
/* There are only two possible S4U2Proxy requests, initial request to impersonator realm,
   and final request to resource realm in case of cross realm. */
if (is_crossrealm) {
    if (is_referral || ..) {
        return error;
    }
} else if (!is_client_db_alias()) {
  return error;
}

**Review by iboukris — APPROVED** [review]
Looks good, great reorg!

File: src/kdc/do_tgs_req.c [hunk 2]
```diff
@@ -917,38 +896,52 @@ prepare_error_tgs (struct kdc_request_state *state,
...
+    krb5_keyblock *key = NULL;
     krb5_kvno kvno;
     krb5_ticket *stkt;
 
-    if (!(req->kdc_options & STKT_OPTIONS))
+    *stkt_out = NULL;
+    *server_out = NULL;
+    *key_out = NULL;
+
+    if (!(req->kdc_options & STKT_OPTIONS) || req->second_ticket == NULL ||
```
iboukris, line 920: btw, should we have a check that if `!(req->kdc_options & STKT_OPTIONS)` then `req->second_ticket == NULL` ?

greghudson: An interesting question.  Since someone could invent a new kdc-options flag requiring a second ticket, I think this check would only make sense if we also wanted to fail on unknown KDC options, and that would in turn only be desirable if the other major KDC implementations did so.  (RFC 4120 doesn't say what to do with unrecognized options.)

src/kdc/tgs_policy.c
greghudson: I added a comment along those lines and flipped the checks to match the order of mention in the comments, but maintained the basic structure.  The suggested structure puts the cross-realm check inside braces and the initial-request check inside the `else if`, which is asymmetric.  Putting both sides within arms of the `is_crossrealm` check would read oddly.

iboukris: Ack, thanks!

frozencemetery (approved):

greghudson:
Sorry, I thought I was done but I noticed one more good candidate: a user-to-user second-ticket client check.

frozencemetery (approved):
Still looks good.

PR #1124, opened 2020-10-13
Update Gladman AES code

greghudson:
Update lib/crypto/builtin/aes to commit
8798ad829374cd5ff312f55ba3ccccfcf586fa11 of
https://github.com/BrianGladman/aes .

The following changes are made to the upstream code:

* In aes.h, #defines are added to give the linker-visible symbols a
  prefix.

* In aes.h, AES_192 is undefined, since we only need AES-128 and
  AES-256.

* In aesopt.h, USE_INTEL_AES_IF_PRESENT and USE_VIA_ACE_IF_PRESENT are
  suppressed by changing the corresponding "#if 1"s to "#if 0"s.

* In aesopt.h, the conditionals for ENC_UNROLL, DEC_UNROLL, ENC_ROUND,
  LAST_ENC_ROUND, DEC_ROUND, LAST_DEC_ROUND, and KEY_SCHED are changed
  from "#if 1" to "#if !defined(CONFIG_SMALL) ||
  defined(CONFIG_SMALL_NO_CRYPTO)".


greghudson:
Additional notes:

* The large body of changes to aes.h, aescrypt.c, etc. are from upstream and I don't expect them to be reviewed here.

* As far as I can tell we wrote aes-gen.c ourselves, so I modified it to work with the new function names and kept it.  I added a comparison of the output with expected output (generated using the old tree) since it doesn't look like aes-gen.c internally verifies much of its results.

* Upstream contains AESNI code, using compiler features rather than yasm.  This initially appeared attractive (as it would let us ditch the yasm dependency and the assembly source files), but it appears that the way the compiler features work in gcc would require us to build with -msse3 -maes, at which point my understanding is that the compiler might generate code that doesn't work on some processors.

* There's a bunch of trailing whitespace in the upstream code, which I didn't remove since it could generate merge issues.  The "bigredbutton: whitespace" in the commit message should disable the update hook's whitespace check, which would otherwise deny the push.

* Since upstream renamed the block encryption function to aes_encrypt(), adding a krb5int_ prefix would conflict with the enc provider's encryption function.  Since I used the k5_ prefix for namespace protection this time around, it's not an immediate issue, but we'll have to do something about that if we ever make a big pass changing krb5int_ to k5_ in libk5crypto.

src/lib/crypto/libk5crypto.exports
```diff
@@ -83,9 +83,9 @@ krb5_k_make_checksum_iov
 krb5_k_prf
 krb5_k_reference_key
 krb5_k_verify_checksum
-krb5_k_verify_checksum_iov
 krb5int_aes_encrypt
 krb5int_aes_decrypt
+krb5_k_verify_checksum_iov
```
frozencemetery, line 88: This movement seems unrelated?

frozencemetery (changes requested):
In the commit message, I think the repo url got double-pasted.

One nit inline; otherwise looks fine.

src/lib/crypto/libk5crypto.exports
greghudson: I tried removing krb5int_aes_{enc,dec}rypt (which are used by a test program, it turns out) and accidentally put them back at a one-line offset.  Fixed along with the commit message.

frozencemetery (approved):

PR #1125, opened 2020-10-15
Port to BoringSSL

alexey-milovidov:


alexey-milovidov:
Sorry, wrong repo.

frozencemetery:
Hi @alexey-milovidov, I know this wasn't the repo you intended.  I did take a quick look at the changes out of morbid curiosity.  Please be aware of the license implications between krb5's licensing and openssl's - you can't just import openssl code into krb5 with no other changes.  (I am not a lawyer etc.)

alexey-milovidov:
Thank you, that looks true. The whole derivative krb5 library cannot longer be used in terms of MIT license if we include OpenSSL code. But there should not be a big problem for open-source software that is already using both OpenSSL/BoringSSL and krb5. Maybe we have to clearly separate files, or mention license restrictions in readme of modified krb5...

PS. I am not a lawyer either.

alexey-milovidov:
Just figured out that the license of MIT krb5 library is not MIT.

PR #1126, opened 2020-10-15
Static analyzer fixes

frozencemetery:
I believe these were found by clang.

### DIFF
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -671,7 +671,7 @@ kg_accept_krb5(minor_status, context_handle,
     krb5_auth_context auth_context = NULL;
     krb5_ticket * ticket = NULL;
     const gss_OID_desc *mech_used = NULL;
-    OM_uint32 major_status = GSS_S_FAILURE;
+    OM_uint32 major_status;
     OM_uint32 tmp_minor_status;
     krb5_error krb_error_data;
     krb5_data scratch;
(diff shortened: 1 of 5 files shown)


greghudson:
For the third commit, I see the unhandled return value from k5_expand_path_tokens(), but what is the potential leak of rcname?

greghudson:
For the last commit: I think kg_accept_krb5() was written with the pattern where the result code is assigned failure until the end.  The dead-assignment defect means that this is unnecessary; every path to the end of the function has an explicit success or failure assignment.  Since we don't encourage that pattern, we can stop using it, but we should also remove the initialization of major_status.

(There are also some pre-existing violations of the pattern; the calls to krb5_gss_acquire_cred(), kg_cred_resolve(), and create_constrained_deleg_creds() set major_status to 0 on success and there's no subsequent reset t to GSS_S_FAILURE.  This is a really easy mistake to make when editing a function using the failure-until-success pattern, which is main the reason we don't encourage it.  Sam even filed a bug about the use of that pattern in PKINIT, ticket 6959.)

frozencemetery:
You're right about the third commit - I must have been looking at an intermediate rework or something.

In any case, pushed a new version with requested changes.

greghudson:
I pushed a simpler variant of the third commit; I don't think a flow control is needed.  I probably also want to squash these.

frozencemetery:
Changes seem fine; squashing is also fine but I'll defer that to you since I'm not sure what you look for in a commit message like that.

frozencemetery (commented):
Looks good to me, thanks.

PR #1127, opened 2020-10-16
Unregister thread key in SPNEGO finalization

greghudson:
Commit d160bc733a3dbeb6d84f4e175234ff18738d9f66 (ticket 7045) added a
new thread key K5_KEY_GSS_SPNEGO_STATUS and registered it in SPNEGO
library initialization, but neglected to unregister it in
finalization.  As a result, loading, unloading, and reloading
libgssapi_krb5 could throw an assertion failure if libkrb5support
remained loaded.  Unregister the key in SPNEGO finalization and add a
test case.

Reported and investigated by Adam Dabrowski.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -468,6 +468,7 @@ local.properties
 
 /src/tests/gssapi/ccinit
 /src/tests/gssapi/ccrefresh
+/src/tests/gssapi/reload
 /src/tests/gssapi/t_accname
 /src/tests/gssapi/t_add_cred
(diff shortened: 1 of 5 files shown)


greghudson:
An unresolved concern here is that the test program relies on dlopen() (unconditionally) and the presence of shared libraries with a `.so` suffix.  I guess I can conditionalize running the program on the presence of `$(TOPLIBD)/libgssapi_krb5.so`.

frozencemetery (approved):

PR #1128, opened 2020-10-19
Stop adding clock skew to GSS acceptor lifetimes

greghudson:
Adding to the acceptor lifetime can cause authentication bridges to
extend the lifetime of bridged credentials, and also produces spurious
errors in SAP's GSS test program.  Instead, report a lifetime of 0 if
the context is within the clock skew window.


greghudson:
Nico and Jeff asked for this change.  We started adding the clockskew to the acceptor lifetime after this thread and ticket:

http://mailman.mit.edu/pipermail/krbdev/2015-October/012457.html
https://krbdev.mit.edu/rt/Ticket/Display.html?id=8268

Aside from the annoyance of having SAP's test program complain, there is a concern that a loop of authentication bridges could be convinced to extend a credential's lifetime indefinitely.  I don't know how realistic that scenario is.

The change deserves some thought, as we don't want to bring back the bad experience discussed in the mail thread.  (Summarizing briefly: service has a short max lifetime, so clients commonly send it tickets close to expiration.  With a bit of client-server clock skew, the client thinks its ticket is valid, and therefore doesn't fetch a fresh one, but the server thinks the ticket is expired.)  Most applications should be okay as long as gss_accept_sec_context() succeeds for a ticket within the clockskew window, regardless of the lifetime_rec and inquire_context behavior.  An application that is explicitly lifetime-aware could decide to discard the context immediately and get into a fight with a client that thinks there is still some time remaining on its ticket, however.

In particular, we might want to revisit the return status of gss_inquire_context() and gss_context_time() for expired contexts.  Looking at RFC 2744 and Heimdal, I don't think gss_inquire_context() should ever return GSS_S_CONTEXT_EXPIRED, so I might fold that change into this PR as a related fix.

gss_context_time() is specified as returning GSS_S_CONTEXT_EXPIRED for an expired context.  We might consider making it return success (but with a lifetime of 0) if an acceptor context is expired within the clockskew window.  That seems like confusing behavior, so maybe it's not a good change.

frozencemetery:
What was the reason for this request?  Was it just the bridges scenario?  

The change does what it says, and the refactor is nice.  I just don't see why we would want it (yet?).

greghudson:
I believe Jeff noticed that his AFS tokens were valid for five minutes longer than expected, and that led to discussion leading to this request.  The AFS token case is an example of an authentication bridge, I think; an AFS server receives a krb5 credential and issues its own credential with the lifetime reported by GSS.

This change also makes us more consistent with Heimdal.  (Heimdal also does an expiration check with no clockskew tolerance in the DCE third leg, but I don't see any reason why the DCE variant should be less forgiving.)

frozencemetery (approved):

PR #1130, opened 2020-10-28
Add interposition for saslname to mech operation

simo5:
We had a comment that interposition was not possible in this case, but
it turns out all we need to do is to cycle through interposed mechanism
if the saslname was not resolved after cycling through the regular ones.

I am not 100% convinced we really want/need this. In theory an interposed
mechanism could recognize a sasl name type that the original mechanism
does not. But in practice I wonder if there is any value in returning a positive
response in that case, given it may indicate that the mechanism loaded in the
application is newer/different than it actually is...

### DIFF
--- a/src/lib/gssapi/mechglue/g_saslname.c
+++ b/src/lib/gssapi/mechglue/g_saslname.c
@@ -167,7 +167,6 @@ OM_uint32 KRB5_CALLCONV gss_inquire_saslname_for_mech(
     return status;
 }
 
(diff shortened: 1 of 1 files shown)


frozencemetery:
This is conceptually interesting, but I'm not sure it's useful.  SASL names are supposed to be ("must") registered with IANA, which means that the problem space is somewhat constrained.  So unless we want this for an existing mechanism, I'm disinclined to add it.

simo5:
Agreed, I'm closing this PR, but it will be here should we need it in the future.

PR #1131, opened 2020-10-30
Don't mark FAST as available if our ccache disagrees

frozencemetery:
[In the scenario that this caused problems in, we first performed anonymous PKINIT to realm A.  Then requesting armor with the resulting ccache, we kinit -E someuser.  Realm A generates a "not local" response with a referral to another realm.  At this point, we chase referrals until we reach B, whereupon we try to upgrade to FAST due to presence of PA_FX_FAST in KDC reply because our FAST state says it's available.

The whole redacted trace is https://gist.github.com/frozencemetery/88e16e06ec706959778158d3286aa5be (from a patched 1.17) ]


greghudson:
I see the problem, but the solution isn't straightforward.

The PR proposes to ignore a FAST armor ccache unless it contains FAST negotiation config for the target realm.  This is at best an out-of-the-box solution.  FAST negotiation was specified in RFC 6806, eighteen months after FAST itself.  It was not intended to be a requirement to make FAST work, but rather an optimization and a security measure (to detect an attacker who pretends to be the real KDC and pretends not to support FAST).  That in itself may not be a serious consideration, as I think every reasonably modern KDC implementation supports FAST negotiation.  But in our client implementation, the FAST negotiation config state is pretty fragile; an initial-creds application might not write it out (because it stores the creds itself instead of using krb5_init_creds_opt_set_out_ccache()), and we also currently lose that state during renewal operations.

It's not clear to me how FAST is supposed to work with AS referrals.  I wasn't able to find text about it in RFC 6113 or 6806.  It is very unlikely that an armor ccache would happen to contain a TGT to the target realm of a WRONG_REALM referral.

The simplest option is not to use FAST after following a client realm referral.  Care should probably be taken not to follow an unprotected WRONG_REALM reply if FAST negotiation state was detected.

Another option is to use the armor TGT to get a cross-realm ticket to the referred-to-realm, perhaps writing the resulting TGT to the armor ccache.  There are three challenges here:

* The client AS state machine would have to call out to the TGS state machine, which involves more complexity than we would need in a synchronous implementation.

* If the TGS request fails, should we fail open or fail closed?  If the WRONG_REALM reply was FAST-protected, I would think we should fail closed.

* If the armor ticket contains AD-fx-fast-armor authdata, the KDC is prohibited from allowing its use in a TGS request, so the client couldn't use it to get a cross-realm TGT.  However, as far as I know, AD-fx-fast-armor is a dead feature; I've never heard of anyone using it, and Heimdal fouled the ecosystem by not implementing it in the KDC when it implemented FAST.  So this may not be a serious concern.

I am curious how Windows clients handle this scenario, but do not have the resources to investigate.

frozencemetery:
> I am curious how Windows clients handle this scenario, but do not have the resources to investigate.

@abbra @iboukris - do either of you know what Windows does here?

abbra:
I am only aware of the behavior described here:
- Client AS_REQ choices: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/1db3d89f-4eb6-4ca5-88c2-e4cc5097db86 (for FAST armor) and https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/a6adb34c-b6bb-42f4-afb8-8cd989cbadc1 (for CBAC and armor)
- Client TGS_REQ choices: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/2c42487d-2572-4090-999d-0a2d73d8c946 (for CBAC and armor)
- KDC consideration of ticket flags: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/f5993d64-a280-4555-9fcc-922fb934fc27
- KDC AS-REQ considerations: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/a076994a-47a9-4d16-af77-d4abb523afe4
- KDC TGS processing: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/2e5dcf34-4b51-44a0-b45a-277ed616ca39 (for CBAC and armor)
- KDC compound identity choices: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/ee7826f5-d985-4a6b-a167-0df0525daea0

I think their client choices described in the first link ([MS-KILE] 3.2.5.4) but they are still unclear if the client is not from the user's realm -- how can it get TGT for a computer from the user's realm?

abbra:
Any update on this?

frozencemetery:
Presumably it's waiting for an update from me, and I've unfortunately been a bit busy :)

It seems like the best course of action was Greg's simple proposal above, since it involved the fewest unknowns.  (That is, not to use FAST after following a client realm referral.)

greghudson:
I wrote above:

> Care should probably be taken not to follow an unprotected WRONG_REALM reply if FAST negotiation state was detected.

My concern here is that if we have an armor cache with FAST negotiation detected, we expect the KDC to protect its error replies.  However, krb5int_fast_process_error() accepts unprotected errors, with the comment:

            /*                                                                  
             * This can happen if the KDC does not understand FAST. We don't    
             * expect that, but treating it as the fatal error indicated by the
             * KDC seems reasonable.                                            
             */

The problem is, the error as interpreted by the caller isn't necessarily fatal.  Since we're adding a path in the AS client code where we can start with FAST required and continue without FAST, we allow an attacker to downgrade out of FAST by sending an unprotected error redirecting the client to another realm.  And we don't do any loop checking for AS referrals, so the attacker could then redirect the client back to the original realm, only without FAST.

To complicate matters, I believe that comment was written before FAST TGS was implemented.  For TGS requests we blindly try to use FAST with implicit armor, which means we do expect to hit that case in the TGS path under normal circumstances (even if FAST negotiation happened when obtaining the TGT, since Heimdal implements FAST negotiation but not FAST TGS).  So simply changing that error case to yield KRB5_KDCREP_MODIFIED probably wouldn't fly.

frozencemetery:
Ah, I thought you were referring to the checks `is_referral()` makes - thanks for clarifying.  Will think about this more.

nicowilliams:
It'd be nice if AS referrals could deliver a trust anchor for the target realm's AS so the client could just acquire an anonymous TGT for it using PKINIT without needing any local configuration.  This would allow referrals that don't trust the referrent.

greghudson:
I talked with Sam on the phone about this, and he thought it was disappointing that the client doesn't go out and fetch a TGT for the target realm.  The reason is that it would be nice if clients could set the FAST_REQUIRED flag (via krb5_get_init_creds_opt_set_fast_flags()) now that all modern KDCs support FAST for AS requests, without breaking realm referrals.

So I rechecked the code after the call and... it actually does what Sam thought it should!  Sort of.  fast_armor_ap_request() calls krb5_get_credentials() for service krbtgt/TARGETREALM@TARGETREALM.  (The fact that it's requesting a local TGT for the target realm isn't an obstacle; try "make testrealm CROSSNUM=2" and "kvno krbtgt/KRBTEST2.COM@KRBTEST2.COM".)  This is obvious a problem from an async perspective, but the FAST code was written for 1.7 and the stepwise interface was written for 1.8, evidently missing this synchronous call inside the FAST subroutines.  And in most circumstances it should still work, just not in a stepwise fashion like it ought to.

I think the real reason that the original scenario is visibly not working is that the armor ticket was obtained with anonymous PKINIT, and anonymous caches are broken for cross-realm or the referral realm (ticket 8332).

greghudson:
If I am right, PR #1143 should make this particular scenario work.

However, we should still make the following changes:

* For AS requests (but not TGS requests), error out more forcefully if we are using FAST and receive a non-FAST-protected error.

* Clear the DO_FAST flag across realm referrals.  The AS request will have to upgrade to FAST in the target realm.  (There is no specified FAST negotiation mechanism that survives across realms currently.)

I filed a ticket (8972) for the use of a synchronous TGS API inside fast_armor_ap_request().  I think this problem might be low-priority, since TGS requests are non-interactive and usually fast enough that they wouldn't visibly interrupt a GUI application or something.  But it's obviously not great.

PR #1133, opened 2020-11-06
src/aclocal.m4: add compatibility with upcoming autoconf-2.70

trofi:
`autoconf-2.70` removed empty statements assignment
and exposed no-op statement problem:

```
$ autoreconf-2.70_beta3
./configure: line 2898: syntax error near unexpected token `else'
./configure: line 2898: `else'
```

The change uses `AS_IF` to avoid no-op. On autoconf-2.69 the
`./configure` diff is the following:

```diff
--- configure-2.69.pre  2020-11-06 08:20:50.000000000 +0000
+++ configure-2.69.post 2020-11-06 08:20:26.000000000 +0000
@@ -2853,3 +2853,3 @@
 # echo "Looking for $srcdir/$ac_config_fragdir"
-if test -d "$srcdir/$ac_config_fragdir"; then
+if test -d "$srcdir/$ac_config_fragdir"; then :
   ac_aux_dir=
```

### DIFF
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -13,11 +13,7 @@ fi
 ac_topdir=$srcdir/$ac_reltopdir
 ac_config_fragdir=$ac_reltopdir/config
 # echo "Looking for $srcdir/$ac_config_fragdir"
-if test -d "$srcdir/$ac_config_fragdir"; then
-  AC_CONFIG_AUX_DIR(K5_TOPDIR/config)
-else
-  AC_MSG_ERROR([can not find config/ directory in $ac_reltopdir])
-fi
+AC_CONFIG_AUX_DIR(K5_TOPDIR/config)
 ])dnl
 dnl
 dnl Version info.


frozencemetery:
Hi, autoconf 2.70 does not appear to exist.  The latest released version is 2.69: ftp://ftp.gnu.org/gnu/autoconf/

I don't see an alpha or beta release for autoconf 2.70 on ftp://alpha.gnu.org/pub/gnu/autoconf/ (since your commit indicates this is "2.70 beta 3"); nor do I see a 2.70 of any sort at https://git.savannah.gnu.org/gitweb/?p=autoconf.git (2.69d seems to have been tagged 3 days ago).

trofi:
My apologies for confusion. I think 2.70_beta3 is a gentoo's internal name. Upstream versions it as 2.69d: ftp://alpha.gnu.org/pub/gnu/autoconf/autoconf-2.69d.tar.xz

greghudson:
I'd like to better understand what's changing in autoconf.  Does AC_CONFIG_AUX_DIR() no longer generate any shell code?

greghudson:
Answer: yes.  autoconf commit 4c59bf27d7083088290219450c81d999431b43f1 changed AC_CONFIG_AUX_DIR() to append to an m4 variable rather than generate shell code.

Looking at the output and surrounding code, I think we can just call AC_CONFIG_AUX_DIR() unconditionally.

trofi:
Made the call to `AC_CONFIG_AUX_DIR` unconditional.

PR #1135, opened 2020-11-18
Mark shared libraries executable during installation

frozencemetery:
Based on a patch by Nalin Dahyabhai.

[The full patch also hardcodes PIE support and a couple other things, which I think should no longer be necessary as the distro will pass it in build flags.  Full patch for completeness: [https://github.com/frozencemetery/krb5/commit/c06693e5a17daf0fd585e608e8bfd1eb3eef447c](https://github.com/frozencemetery/krb5/commit/c06693e5a17daf0fd585e608e8bfd1eb3eef447c) ]

### DIFF
--- a/src/config/shlib.conf
+++ b/src/config/shlib.conf
@@ -22,8 +22,10 @@ SHLIBVEXT=.so.v-nobuild
 SHLIBSEXT=.so.s-nobuild
 # Most systems support profiled libraries.
 PFLIBEXT=_p.a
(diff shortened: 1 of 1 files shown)


greghudson:
There's an analagous HPUX case that reads:

        INSTALL_SHLIB='$(INSTALL)'

which I think will work fine (`install` uses mode 755 by default).

We probably also want a comment.

frozencemetery:
Race condition with your email reply.  That suggestion makes sense; updated.

PR #1136, opened 2020-11-19
ccselect: should not return cc_get_principal errors

fihuer:
Before this commit, `ccselect_realm` would return the last error encountered by `krb5_cc_get_principal` ("No credential cache found", typically).
Moreover, krb5_cccol_cursor_free ret code is ignored.

With this commit, only the krb5_cccol_cursor_free return code is taken into account here.

### DIFF
--- a/src/lib/krb5/ccache/cc_kcm.c
+++ b/src/lib/krb5/ccache/cc_kcm.c
@@ -981,8 +981,10 @@ kcm_ptcursor_next(krb5_context context, krb5_cc_ptcursor cursor,
         k5_buf_add_len(&req.reqbuf, id, KCM_UUID_LEN);
         ret = kcmio_call(context, data->io, &req);
         /* Continue if the cache has been deleted. */
-        if (ret == KRB5_CC_END)
+        if (ret == KRB5_CC_END || ret == KRB5_FCC_NOFILE) {
+            ret = 0;
             continue;
+        }
         if (ret)
             goto cleanup;
         ret = kcmreq_get_name(&req, &name);


greghudson:
krb5_cccol_cursor_free() should have been written to return void, although we can't easily change that now.  It always returns 0.  Ignoring its return value is normal.

After the loop ends, ret is the value of the final call to krb5_cccol_cursor_next() when it sets cache to NULL, which is 0.  An error from krb5_cc_get_principal() cannot propagate to the end of the loop.

fihuer:
We have clearly a use-case where we're getting an error here `ccselect module realm yielded error -1765328189/No credential cache found for server principal ...`
And this patch, fixes it.
Any idea of what could cause that ?

greghudson:
What type of ccache is the default collection?  (FILE, DIR, KEYRING, etc.)

fihuer:
We're using KCM.



Le jeu. 19 nov. 2020 à 20:17, Greg Hudson <notifications@github.com> a
écrit :

greghudson:
What's on the server side?  Red Hat's sssd, or is this on macOS?

fihuer:
Heimdal's kcm server.

Le ven. 20 nov. 2020 à 00:24, Greg Hudson <notifications@github.com> a
écrit :

greghudson:
In krb5_cccol_cursor_new(), a list of uuids for the caches in the collection is retrieved.  krb5_cccol_cursor_next() tries each of the uuids in sequence, continuing to the next entry if it gets a KRB5_CC_END reply.  This matches the behavior of Heimdal's krb5_cc_cache_next() for the KCM type.

If one of the uuids in the cache doesn't exist by the time the client gets to it, sssd will reply with KRB5_CC_END, while Heimdal kcm (in either upstream Heimdal or macOS) appears to reply with KRB5_FCC_NOFILE.  This seems like a bug in Heimdal on one end or the other--why have a special case for KRB5_CC_END in kcm_get_cache_next() if the server won't return KRB5_CC_END for that kind of request?  (From the code, it seems like KRB5_CC_END can only be returned for RETRIEVE and GET_CRED_BY_UUID operations, and not for GET_CACHE_BY_UUID operations.)

On our side, my inclination is to add a check in kcm_ptcursor_next() for KRB5_FCC_NOFILE as well as KRB5_CC_END.  This would be better than ignoring the iteration error in ccselect_realm, because it would allow the iteration to continue on to other caches in the uuid list.

greghudson:
I pushed a replacement commit.  Please test it and let me know if it fixes the problem.

fihuer:
Many thanks ! I get back to you ASAP.

frozencemetery (approved):
I haven't tested this, but it seems desirable to me even if only for the `ret` chage.

PR #1138, opened 2020-11-20
code does not compile on Solaris (developer studio 12.6)

Sashan:
change below makes C compiler happy again

### DIFF
--- a/src/tests/softpkcs11/main.c
+++ b/src/tests/softpkcs11/main.c
@@ -2073,6 +2073,177 @@ C_GenerateRandom(CK_SESSION_HANDLE hSession,
 }
 
 
+#define	no_C_InitPIN			\
+    (CK_RV(*)(CK_SESSION_HANDLE, 	\
+    unsigned char *, unsigned long))	\
+	func_not_supported
+
+#define	no_C_SetPIN			\
+    (CK_RV(*)(CK_SESSION_HANDLE,	\
+    unsigned char *, unsigned long,	\
+    unsigned char *, unsigned long))	\
+	func_not_supported
+
+#define	no_C_GetOperationState		\
+    (CK_RV(*)(CK_SESSION_HANDLE,	\
+    unsigned char *, unsigned long *))	\
+	func_not_supported
+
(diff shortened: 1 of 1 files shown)


frozencemetery (commented):
Can you elaborate on why it doesn't compile here?  The `void *` casts should preclude any type issues, which it looks like is what you're trying to address?

greghudson:
As I understand it, the C standard doesn't guarantee that `void *` is compatible with function pointer types.  The POSIX standard does (because of dlsym()), but I don't think it requires the compiler to allow implicit casting from `void *` to function pointer types.

I don't really like this workaround, as it's full of wordy preprocessor declarations.  Other options I can think of are:

* Define an actual stub for each unsupported function.  Even more wordy than the proposed workaround, but more elegant and also more correct.  With the current code, if one of these unsupported functions is invoked by a caller, the right thing probably happens in practice, but it's likely undefined behavior since the called function doesn't have the same parameter list as the caller used.

* Cast to `(void (*)(void))` instead of `(void *)` if that makes the Solaris compiler happy and doesn't generate warnings with gcc/clang.  I don't think this will pan out, unfortunately.

greghudson:
Another option is to replace those entries with NULL.  I initially disregarded this option since NULL is also a data pointer (and not guaranteed to be a valid function pointer in the C standard), but we set vtable functions to NULL all over the place, so it must be acceptable to compilers.

Of course, NULL will have different behavior; a caller will definitely crash if it tries to invoke one of the nulled function pointers.  That has no impact on the tests (I just checked), but it diverges soft-pkcs11 as present in our tests from how it was upstream.  Probably not an issue since there is no longer a soft-pkcs11 upstream, but we'd want to add a comment.

Sashan:
Let's close this pull request. The changed is needed for solarisstudio12.4, which is a legacy compiler for 11.3. developerstudio12.6 is happy with 1.18.3 code as-is. I was too fast on _pull _request_ trigger. Sorry for extra noise.

kaduk:
> on further checking I think NULL actually is a valid function pointer value.

Yes, NULL is a valid function pointer value.

PR #1141, opened 2020-11-24
Document -k option in kvno(1) synopsis

frozencemetery:
becd1ad6830b526d08ddaf5b2b6f213154c6446c attempted to unify the
synopsis, option descriptions, and xusage(), but missed one option.


greghudson:
We also forgot --out-ccache in the synopsis.

PR #1142, opened 2020-11-27
update t_rcache.c w.r.t. to replay cache changes

Sashan:
tell t_rcache.c that commit
    dcb853ac32779b173f39e19c0f24b0087de8577
has modernized and simplified replay cache.

### DIFF
--- a/src/lib/krb5/libkrb5.exports
+++ b/src/lib/krb5/libkrb5.exports
@@ -173,6 +173,7 @@ k5_plugin_register_dyn
 k5_rc_close
 k5_rc_get_name
 k5_rc_resolve
+k5_rc_store
 k5_size_auth_context
 k5_size_authdata
 k5_size_authdata_context
--- a/src/tests/threads/Makefile.in
+++ b/src/tests/threads/Makefile.in
@@ -15,7 +15,7 @@ SRCS=$(srcdir)/t_rcache.c \
 all:
 
 run-t_rcache: t_rcache
-	$(RUN_TEST) ./t_rcache
+	$(RUN_TEST) ./t_rcache file2:test.rcache2
 
 t_rcache: t_rcache.o $(KRB5_BASE_DEPLIBS)
 	$(CC_LINK) -o t_rcache t_rcache.o $(KRB5_BASE_LIBS) $(THREAD_LINKOPTS)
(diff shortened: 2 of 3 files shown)


greghudson:
I made some changes.  Aside from style issues, there were some opportunities for simplification:

* Nothing was done with the key created using krb5_k_create_key().

* They keyblock could be stack-allocated.

* krb5_c_make_random_key() does its own allocations, so initializing the key first is a leak.  (This is unclear in the documentation; I think the docs mean to say that the function allocates and initializes its output key, but it's easy to read instead that the caller should provide one.)

* krb5_c_encrypt() can deal with an overallocated output.  Calling krb5_c_encrypt_length() and blindly setting the buffer length only serves to make the code less safe.

* The thing we're encrypting isn't an authenticator, just some arbitrary string.

Sashan:
Thanks for improving my change. I've run the t_rcache under memorychecker (libumem) and found single memory leak, which is fixed by the most recent updat I've pushed while ago.

greghudson:
That whole block of code doesn't do anything now that there's no API for destroying a replay cache.  And now that krb5_get_server_rcache() ignores the piece argument, we wouldn't really want the test program stomping on the current euid's default replay cache anyway.

This program isn't currently build or run as part of the build.  I think the simplest thing for now is to give it a command-line argument for a replay cache name, and resolve that with k5_rc_resolve().  I will push a followup commit to do that, and squash it before merging to the mainline.  Whoever runs it can be responsible for where the replay cache lives (e.g. `file2:/tmp/something`).

Sashan:
> That whole block of code doesn't do anything now that there's no API for destroying a replay cache. And now that krb5_get_server_rcache() ignores the piece argument, we wouldn't really want the test program stomping on the current euid's default replay cache anyway.

IJust to clarify motivation behind the pull request. 've found that program while testing 1.18 on Solaris (yes, we are lagging behind upstream quite a lot as there are currently more technologies to support than developers to do get job done). Anyway the krb5api test suite we use here keeps copy of the same program, which comes from kerberos 1.14. The part of the step is to build t_rcache.c source code. This was the way how I discovered t_rcache.c is also broken in upstream. I have not reviewed the t_rcache.c that closely.

frozencemetery (approved):
I don't have anything to add except that the comment on line 47 is incorrect and has been that way since the test was added (through multiple refactorings).

greghudson:
That's funny.  I think I will treat that as out of scope, though.

PR #1143, opened 2020-12-15
Improve anonymous cache behavior by adding start_realm support

greghudson:
Currently anonymous caches don't work for referrals or cross-realm TGS requests; they only work as FAST armor or for services in the local realm (and marked as such in the server field).

This PR implements one possible solution, which is to add support for Heimdal's start_realm feature.  Heimdal sets start_realm automatically in the cache dispatch layer, which I do not like; I feel like logic implemented at the ccache dispatch layer generally comes back to bite us later.  Right now the PR is setting it for initial creds acquisition when an output cache is specified, as is done for both kinit -n and kadmin -n.  That's elegant but not totally satisfying; it won't work for "kvno --out-cache=... krbtgt/REALM@REALM" and it won't work if you delegate a different-realm TGT over ssh (as twosigma does).

In general I'm not fond of start_realm because it seems incomplete.  Heimdal more recently implemented a second "anon_pkinit_realm" key within kinit to allow renewal of anonymous caches.  I pointed out that this seemed redundant with start_realm, and Jeff noted that the issuing realm of the TGT isn't necessarily the same as the service realm, as you could have a cache whose primary cred is a cross-TGT.  (Heimdal's ccache dispatch layer only sets start_realm for a local TGT, so it doesn't really support that scenario.  Its get_cred.c implementation of start_realm should work for a cross TGT as it ignores the issuing realm when fetching the initial TGT.  We currently do not, and I'm not sure if we should.)

What I'd really like is for the ccache model to include the concept of a primary cred.  There would be a number of benefits: the caller wouldn't need to identify any particular service principal to renew a cache; a starting cross-TGT would work without needing to ignore the issuing realm; we might not need to scan the ccache in the GSS krb5 mech to determine the cache expiration time.  But that's a big can of worms; every ccache type would need a way to identify the primary cred, and third-party code that interacts with caches might not behave properly without heuristics.

### DIFF
--- a/doc/formats/ccache_file_format.rst
+++ b/doc/formats/ccache_file_format.rst
@@ -174,3 +174,9 @@ refresh_time
     decimal representation of a timestamp at which the GSS mechanism
     should attempt to refresh the credential cache from the client
     keytab.
+
+start_realm
+    This key indicates the realm of the ticket-granting ticket to be
+    used for TGS requests, when making a referrals request or
+    beginning a cross-realm request.  If it is not present, the client
+    realm is used.
--- a/src/clients/kinit/kinit.c
+++ b/src/clients/kinit/kinit.c
@@ -828,7 +828,7 @@ k5_kinit(struct k_opts *opts, struct k5_data *k5)
         if (opts->verbose)
             fprintf(stderr, _("Initialized cache\n"));
 
-        ret = krb5_cc_store_cred(k5->ctx, k5->out_cc, &my_creds);
+        ret = k5_cc_store_primary_cred(k5->ctx, k5->out_cc, &my_creds);
         if (ret) {
             com_err(progname, ret, _("while storing credentials"));
             goto cleanup;
--- a/src/clients/kvno/kvno.c
+++ b/src/clients/kvno/kvno.c
@@ -561,7 +561,10 @@ do_v5_kvno(int count, char *names[], char * ccachestr, char *etypestr,
                 }
                 initialized = 1;
             }
-            ret = krb5_cc_store_cred(context, out_ccache, creds);
+            if (count == 1)
+                ret = k5_cc_store_primary_cred(context, out_ccache, creds);
+            else
+                ret = krb5_cc_store_cred(context, out_ccache, creds);
             if (ret) {
                 com_err(prog, ret, _("while storing creds in output ccache"));
                 exit(1);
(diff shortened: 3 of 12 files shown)


greghudson:
I thought about this a bit more and arrived at something closer to the best solution achievable without major effort.

I undid the refactoring of get_in_tkt.c and instead made an internal API k5_cc_store_primary_cred(), so that start_realm can be supported in four use cases:

* Anonymous tickets (obtained using a caller-specified output cache).
* Delegated tickets over GSS.
* Renewing (or validating) tickets.
* Storing a single ticket with kvno --out-cache.

frozencemetery:
I guess the other approach is to set `start_realm` on the first invocation of `krb5_cc_store_cred()`, or on `krb5_cc_store_cred()` when `start_realm` is unset.  (Is this what you intended with "at the dispatch layer"?  What you have does seem more robust to me.)

frozencemetery (approved):

greghudson:
Right, Heimdal implements setting start_realm within krb5_cc_store_cred().  The logic is a bit complicated, in order to handle cache copies

* The necessary state is kept within the in-memory cache handle, under the assumption that the caller will initialize and store the primary cred using the same handle.  That assumption might not always be true for MIT krb5 because we use krb5_cc_dup() in a couple of places.

* The logic only kicks in for a handle which has seen a krb5_cc_initialize() operation.  (So, properly, "kvno krbtgt/Y@Y" will not write out a start_realm, because a simple kvno doesn't reinitialize the cache.)

* A start_realm entry is implicitly stored when a krbtgt/X@X entry is stored, unless a start_realm entry was already stored implicitly or explicitly.  The start_realm is stored whether or not X is the client realm.

* An attempted store of a start_realm entry is ignored if one was already stored implicitly or explicitly.

I can see a few edge cases which are unlikely to cause problems in practice:

* If a ccache contains krbtgt/X@X (primary cred) and krbtgt/Y@Y (obtained using kvno) and is then copied without preserving order, the incorrect start_realm could be implicitly stored (and the correct one discarded, if there was one).

* For "kinit -S krbtgt/Y user@X", no start_realm is stored, even though in Heimdal a start_realm would make that cache work for TGS requests.  (In our implementation, the reverse is true; we will store a start_realm, but unless we change the code to ignore the initiator realm when looking up the starting TGT, TGS requests will still fail.  If we did primary_service instead of start_realm, this case could work without ignoring the issuing realm, but I don't think that's enough of a reason to diverge from Heimdal by itself.)

* When start_realm is stored implicitly, it is stored after the TGT.  This defeats the Heimdal FILE ccache hack of "don't make the cache globally visible until the first real cred is written" when a start_realm is required for the cache to work, as there is a window when the cache is globally visible but no start_realm value is present in it.

Other than that the logic is correct as far as I know.  But in general I think implementing logic at that layer is asking for trouble, and other logic implemented at that layer has given us trouble in the past.

frozencemetery:
Agreed, thanks for explaining!

PR #1144, opened 2020-12-15
Preserve cred order in memory ccache type

greghudson:
In the memory credential cache type, add new creds to the tail of the
list so that iteration does not reverse the order.

### DIFF
--- a/src/lib/krb5/ccache/cc_memory.c
+++ b/src/lib/krb5/ccache/cc_memory.c
@@ -108,6 +108,7 @@ typedef struct _krb5_mcc_data {
     k5_cc_mutex lock;
     krb5_principal prin;
     krb5_mcc_link *link;
+    krb5_mcc_link **tail;       /* Where to store next added cred */
     /* Time offsets for clock-skewed clients.  */
     krb5_int32 time_offset;
     krb5_int32 usec_offset;
@@ -159,6 +160,7 @@ empty_mcc_cache(krb5_context context, krb5_mcc_data *d)
         free(curr);
     }
     d->link = NULL;
+    d->tail = &d->link;
     d->generation++;
     krb5_free_principal(context, d->prin);
     d->prin = NULL;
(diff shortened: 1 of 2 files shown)


greghudson:
We sometimes discuss possible heuristics within the ccache code which depend on cred store order.  For instance, these days Heimdal's FILE ccache type doesn't actually make the cache globally visible until you store a non-config cred, at which point it calls rename() to make the cache visible atomically.  The assumption is that the first real cred to be stored is the important one, and that other users of the cache will succeed as long as they can see it.  (One problem with this heuristic is that gss-krb5 proxy cred caches contain two important creds.)

A wildcard for this kind of reasoning is that ccaches are sometimes created by copying a memory ccache, such as when delegated creds are stored with gss_store_cred_into().  Currently the MEMORY ccache type reverses the order of creds when you iterate over it.  In many cases there is only one non-config entry in the cache, but at a minimum the reversal would flip whether the config entries (FAST availability, start_realm, etc.) appear before or after the real cred.

The list management required for this change is very simple and the PR passes tests, but I haven't found a good way to explicitly test it yet.

Although copies from memory caches are the most common, it might be useful to know that cred order is preserved for all ccache types.  I don't know if KEYRING preserves order.  KCM order preservation of course depends on the server implementation.

frozencemetery:
For sssd-kcm, it looks like cred insertion happens at the front of an internal list, and UUIDs will be returned in list-order - so it would manifest this same behavior as with MEMORY.  However, the list is already doubly-linked, so this could be fixed easily.

For keyring, Linux does not currently document whether it is order-preserving or not.  Currently they keys interface is implemented on top of assoc_array, which states that "The array can be iterated over.  The objects will not necessarily come out in key order."

frozencemetery (changes requested):
I think `krb5_mcc_remove_cred()` needs to be updated as well.

greghudson:
> I think krb5_mcc_remove_cred() needs to be updated as well.

I looked at it, but it leaves behind empty nodes (so as not to break iteration), so I don't think it needs to be changed.

frozencemetery (approved):
Ah, you're right.  All set then :)

greghudson:
I added a test case.

frozencemetery (approved):
Test case looks good.

PR #1147, opened 2020-12-18
Fix ASN.1 test memory leak

greghudson:
Commit d9443a58cd364349b7d764f4e997f3af7d979a87 introduced a leak in
krb5_decode_test.c with its new decode_fail macro.


DemiMarie:
Looks fine to me.  I wasn’t careful about leaks in unit tests, but they interfere with automated leak checking, so fixing them is still a good idea.

frozencemetery (approved):

PR #1148, opened 2020-12-21
Revert dns_canonicalize_hostname default to true

greghudson:
Field testing of dns_canonicalize_hostname=fallback (ticket 8911)
revealed more disruptive edge cases than anticipated.  Many were fixed
by ticket 8930, but host-based GSS initiator names were recently
discovered to not work, and one other edge case could not be resolved
without a change to external code.

Restore the default to true for now.  Set the value to fallback in the
test suite, to continue testing the desired configuration and to avoid
restoring tests/resolve.

ticket: 8973 (new)
tags: pullup
target_version: 1.19


frozencemetery (approved):
I understand the rationale for this and keeping it in the test suite is appreciated.

PR #1149, opened 2021-01-04
Implement fallback for GSS acceptor names

greghudson:
Commit 3fcc365a6f049730b3f47168f7112c03997c5c0b added fallback support
to krb5_rd_req(), but acquiring acceptor creds for a host-based name
could still fail within check_keytab() in the krb5 mech.

Add an internal libkrb5 API k5_kt_have_match() to check for a matching
keytab entry with canonicalization, and use it in check_keytab().  Add
a library-internal function k5_sname_wildcard_host() to share logic
between rd_req and k5_kt_have_match().

ticket: 8971 (new)

### DIFF
--- a/src/lib/krb5/keytab/ktfns.c
+++ b/src/lib/krb5/keytab/ktfns.c
@@ -31,6 +31,8 @@
 #ifndef LEAN_CLIENT
 
 #include "k5-int.h"
+#include "../krb/int-proto.h"
+#include "../os/os-proto.h"
 
 const char * KRB5_CALLCONV
 krb5_kt_get_type (krb5_context context, krb5_keytab keytab)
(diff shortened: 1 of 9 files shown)


greghudson:
I discovered this problem while investigating host-based initiator name fallback, which I'm still working on.

This approach has a minor performance issue: under common circumstances, canonicalization of the acceptor name will happen twice--once when the keytab is checked for the desired name and again when the GSS context is accepted.  However, a hostbased acceptor name is almost always the local hostname, and in common (though not universal) configurations it can be canonicalized against /etc/hosts, which is cheap.

src/lib/krb5/keytab/ktfns.c
```diff
@@ -129,6 +131,51 @@ krb5_kt_have_content(krb5_context context, krb5_keytab keytab)
...
+    krb5_error_code ret;
+    krb5_keytab_entry ent;
+    krb5_kt_cursor cursor;
+    krb5_boolean match;
+
+    /* Scan the keytab for host-based entries matching accprinc. */
+    ret = krb5_kt_start_seq_get(context, keytab, &cursor);
+    if (ret)
+        return ret;
+    while ((ret = krb5_kt_next_entry(context, keytab, &ent, &cursor)) == 0) {
```
frozencemetery, line 147: `ret` is not read after this assignment - this causes all iteration errors to become `KRB5_KT_NOTFOUND`.

greghudson: Fixed.  Since a double-trinary expression doesn't read well to me, I added an `if (ret) return ret;`.

frozencemetery (approved):
Thanks!

src/lib/krb5/keytab/ktfns.c
greghudson: On self-review I realized we could do:

    return (ret == KRB5_KT_END) ? KRB5_KT_NOTFOUND : ret;

with no reference to `match`, because ret will be 0 if we broke out of the loop with a match and KRB5_KT_END if we finished iteration or encountered an iteration error.  However, the current three-line approach may be more obviously correct and therefore worth the extra verbiage.

PR #1150, opened 2021-01-05
Fix runstatedir makefile substitution

greghudson:
Set localstatedir and runstatedir in config/pre.in so that the default
runstatedir value of ${localstatedir}/run works.  Reported by Mike
Jetzer.


greghudson:
There are some other oddities in how we handle directory substitutions in our makefiles; we seem reluctant to use the standard autoconf variable names, instead using capitalized versions or things like CLIENT_BINDIR and SERVER_BINDIR.  I did a little bit of cleanup directly adjacent to the localstatedir/runstatedir issue, but didn't do a full cleanup pass.

frozencemetery (approved):
This should let me drop some downstream logic, nice.

PR #1151, opened 2021-01-06
Fix all-liblinks when symlinks are unsupported

greghudson:
After building libraries, we make links to a central directory so that
programs and other libraries can be built against them.  These links
were changed from symlinks to hard links by commit
24311fa2b4b95597695560144313debe0211e968, and were reverted to
symlinks by commit 6e0bfbc787509dbaac4947a76e07a2bd6ba7956d.

As part of the first commit, the links to .so and .so.X were made
within the central directory instead of directly to the library build
directory.  However, the dependencies were not updated to ensure that
the link source was created before the link target, so when $(LN_S) is
"cp -pR" and not "ln -s", the build steps could be executed out of
order and fail.  This part of the first commit was not reverted in the
second commit.  Revert it now so that the dependencies match the links.

ticket: 8976


frozencemetery (approved):
Oh autoconf...

PR #1153, opened 2021-01-14
Expose krb5 functions for serialization of credentials

frozencemetery:
In KCM, the responder will receive a serialized krb5_creds object.
Should it wish to perform any operations on this object, it needs to
deserialize and reserialize it.

### DIFF
--- a/doc/appdev/refs/api/index.rst
+++ b/doc/appdev/refs/api/index.rst
@@ -232,6 +232,7 @@ Rarely used public interfaces
    krb5_kt_remove_entry.rst
    krb5_kt_start_seq_get.rst
    krb5_make_authdata_kdc_issued.rst
+   krb5_marshal_credentials.rst
    krb5_merge_authdata.rst
    krb5_mk_1cred.rst
    krb5_mk_error.rst
@@ -285,6 +286,7 @@ Rarely used public interfaces
    krb5_tkt_creds_get_times.rst
    krb5_tkt_creds_init.rst
    krb5_tkt_creds_step.rst
+   krb5_unmarshal_credentials.rst
    krb5_verify_init_creds.rst
    krb5_verify_init_creds_opt_init.rst
    krb5_verify_init_creds_opt_set_ap_req_nofail.rst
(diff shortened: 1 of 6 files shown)


greghudson:
Before we add new APIs we should determine whether krb5_mk_1cred() and krb5_rd_cred() are sufficient for the use case.

They have a certain amount of baggage because the message format is designed to be used within a Kerberos auth context.  However, the implementation specifically allows for a dummy context with no keys or addresses set.  In that case, the "encrypted part" uses ENCTYPE_NULL and the plain text as the ciphertext, as described in RFC 6448.

frozencemetery:
There's no callchain path between those functions and `k5_{un,}marshal_cred`, so if they are, there's technically code duplication.

frozencemetery:
Comparing the we use for cred serialization (~BNF at the top of ccmarshal.c) and the definition of a KRB-CRED in RFC 4120 does not suggest they are compatible to me, but I might be missing something?

greghudson:
I had assumed that the particular marshalling format wasn't important.  krb5_mk_1cred() encodes the cred using ASN.1 DER, not the ccache FILE format.

frozencemetery:
Ah, that makes sense.  The specific use case in mind is the KCM daemon deserializing the stored blobs in order to perform automatic TGT renewal.

greghudson:
Looking back at your IRC message, the particular format is in fact important (since this is for sssd-kcm).  Sorry for the confusion.

greghudson:
For consistency with other APIs, the new APIs should use krb5_data wrappers.  The encoder definitely should not allocate data in the library and tell the caller to free it with free().

frozencemetery:
Hmm, okay.  I'd done it this way because using `krb5_data` runs into trouble with `const`, but I can switch it.

greghudson:
I made some editorial changes.  I also think I want the new APIs to create objects (krb5_data and krb5_credentials) rather than fill in caller-provided structures.  While I'm not as much of a purist about this as I was ten years ago, I do think it makes APIs easier to use, and it's consistent with krb5_decode_ticket().  Old APIs with `krb5_data *` output parameters have also been inconsistent about whether they use caller-allocated storage (krb5_c_decrypt()) or replace the contents with allocated storage (krb5_mk_req()).

frozencemetery:
I could see arguments either way, so I opted for that which entailed fewer allocations.  But I don't disagree that it's easier to work with this way.  Updated accordingly.

Editorial changes make sense, thanks.

PR #1154, opened 2021-01-15
Hostbased initiator name support

greghudson:
Having tried a couple of approaches to this problem, I'm putting this up to keep a record of my design notes.

Hostbased initiator names are not currently supported very well; even without dns_canonicalize_hostname=fallback, they fail if there is no [domain_realm] entry supplying a realm name.  But apparently FreeIPA uses them, and there is no conceptual reason why they shouldn't work on a best-effort basis.

gic_keytab does not support wildcarding like krb5_rd_req() does.  At this time I think that's inherent to the nature of the initiator: we need to pick a concrete principal to authenticate as, and there might be multiple matches for a wildcard in the keytab or cache collection.  By contrast, the acceptor has the ticket encryption key to use as an authority.  For this reason, I think it makes sense that a GSS host-based name like "host" can only match the local host, and if the realm can't be determined with [domain_realm] then we can only match the default realm.

The current implementation approach is to rewrite cred->name->princ after we match a ccache in the collection or acquire creds from the client keytab.  This is much simpler than adding a new field to the cred (we create credentials and use cred->name->princ in many places), but unfortunately it creates a corner case for GSS_C_BOTH credentials, because cred->name->princ->data[1] is now post-canonicalization but kg_rd_req() thinks it is pre-canonicalization.

The mere possibility of this situation is a result of the krb5_sname_to_principal() fallback design.  When we see a host-based krb5_principal, we currently have to deduce from the call graph whether it's pre-canonical or post-canonical.  Heimdal uses a fake name type (KRB5_NT_SRV_HST_NEEDS_CANON) to make this explicit, but that leaves open the possibility of the fake name type leaking out onto the wire.
  
I currently think the safe solution is to create a new (internal) "principal description" type which can contain a service, hostname, iteration state, and (once known) canonical principal name.  (For non-hostbased names it would just contain a fixed canonical principal name.)  I will investigate that approach.

The PR currently reuses reuses krb5_kt_have_match() to match against the client keytab when checking whether we can acquire initial creds.  This is a bit overly permissive because that function uses wildcarding rules and obeys ignore_acceptor_hostname.  Being a little too permissive isn't necessarily a big problem, though.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2406,4 +2406,13 @@ void k5_change_error_message_code(krb5_context ctx, krb5_error_code oldcode,
 #define k5_prependmsg krb5_prepend_error_message
 #define k5_wrapmsg krb5_wrap_error_message
 
+/*
+ * Like krb5_principal_compare(), but with canonicalization of sname if
+ * fallback is enabled.  This function should be avoided if multiple matches
+ * are required, since repeated canonicalization is inefficient.
+ */
+krb5_boolean
+k5_sname_compare(krb5_context context, krb5_const_principal sname,
+                 krb5_const_principal princ);
+
 #endif /* _KRB5_INT_H */
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -105,6 +105,9 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
 
 #endif /* DISABLE_TRACING */
 
+#define TRACE_CC_CACHE_MATCH(c, princ, ret)                             \
+    TRACE(c, "Matching {princ} in collection with result: {kerr}",      \
+          princ, ret)
 #define TRACE_CC_DESTROY(c, cache)                      \
     TRACE(c, "Destroying ccache {ccache}", cache)
 #define TRACE_CC_GEN_NEW(c, cache)                                      \
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -683,7 +683,6 @@ kg_accept_krb5(minor_status, context_handle,
     krb5_flags ap_req_options = 0;
     krb5_enctype negotiated_etype;
     krb5_authdata_context ad_context = NULL;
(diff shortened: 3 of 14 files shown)


frozencemetery:
> Hostbased initiator names are not currently supported very well; even without dns_canonicalize_hostname=fallback, they fail if there is no [domain_realm] entry supplying a realm name. But apparently FreeIPA uses them, and there is no conceptual reason why they shouldn't work on a best-effort basis.

It's not clear to me what a better option would be.  Hardcoding the realm is odd because it should be possible to be realm-agnostic.  And while GSS_KRB5_NT_PRINCIPAL_NAME gets around the issue, it's Kerberos-specific.

frozencemetery:
Additionally, hostbased initiator names are what's used in RFC 7546.

greghudson:
> It's not clear to me what a better option would be.

I couldn't find the FreeIPA code in question, so I don't understand the use case.  But I think most initiator code passes GSS_C_NO_NAME for desired_name and lets the mech decide.

frozencemetery:
Huh, I didn't realize that worked.  I suppose that speaks to your point about how uncommon this is.

I have sent a fix to freeipa for the code in question here: https://github.com/freeipa/freeipa/pull/5452

greghudson:
I think there's been a miscommunication.

I think GSS initiator code frequently passes GSS_C_NO_CREDENTIAL for claimant_cred_handle to gss_init_sec_context().  Or, if it needs a credential for whatever reason, it passes GSS_C_NO_NAME for desired_name to gss_acquire_cred().

The linked PR seems to instead create a name, but passes GSS_C_NO_OID for input_name_type to gss_import_name().  That does work for importing a principal name (or other mech default name type) but isn't related to what I said.

frozencemetery:
Ah, okay.  Well, that explains why it was strange at least :)

greghudson:
I ironed out the acceptor name issue by generating the acceptor matching princ early and storing it.  I fixed a couple of other bugs and wrote a proper commit message.  I think this can be merged pending review.

frozencemetery (approved):
Looks (and tested) good.

PR #1155, opened 2021-01-15
Synchronize documentation on our executables

frozencemetery:
First commit is dead code removal; it's included here because otherwise we'd have to decide whether to document "-passwordserver".

Second commit is the cleanup.  I've placed a summary of what needed to change [in a paste](https://gist.github.com/frozencemetery/2409f82db50752558b28046f88cff5dd) as it's longer than the commit itself.

There's no urgency to this, but I've been sitting on it for several months, so I'm submitting it in order to not forget about it again :)


greghudson:
kdb5_util dump -` is probably undocumented intentionally, because it's only supposed to be used by kadmin and not by an admin.  My preference is to remove it from the usage summary.  The same applies to load.

It looks like the kinit usage message had -I added to it twice in the PR.

kpropd -D was added in commit f1c85fbb0ab9e62b2790647b2681aec4d5fa4585, while -d has always existed.  I don't remember why Nico wanted to add a nodaemon option separate from debug, but it's not there for backward compatibility.

frozencemetery:
Thanks, updated.

greghudson:
I checked on the history of krb5-config --deps, and from the very first version of the file it has never done anything.  I will adjust the comment since "deprecated" doesn't really describe that.

PR #1157, opened 2021-01-28
Allow multiple pkinit_identities lines

kenh:
This patch permits the use of multiple pkinit_identities lines in
krb5.conf.  This patch has two components:

Move the call to crypto_load_certs() in pkinit_identity_initialize()
inside of the same loop as process_option_identity().  This will
attempt to load a certificate from each pkinit_identities line in
the configuration, and if the certificate load fails to move to
the next line.

In pkinit_get_certs_pkcs11(), if pkinit_open_session() fails (for
example, if the specified library does not exist) then return an error.
Previously if defer_id_prompts was set an error was NOT returned.
This makes pkinit_get_certs_pkcs11() behave like all other certificate
loading functions called by crypto_load_certs(), and is required if you
want to be able to specify multiple PKCS#11 libraries on pkinit_identities
configuration lines.

Additionally, if pkinit_open_session has defer_id_prompt set, then make
it so it will return 0 (success) if the module finds an ID successfully.
Previously it would not (but this didn't matter because the error was
ignored by the caller when defer_id_prompt was set).

The resulting semantics is that the PKINT plugin will try multiple
identities until it can successfully load a matching identity.  This
seems like a reasonable combination of useful functionality easy
to implement.

This patch also reverts the change to the krb5 documentation for
pkinit_identities made in commit 456e41f848861217ddf5149b9e52e3ba6d42947c.

### DIFF
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -1123,10 +1123,9 @@ PKINIT krb5.conf options
 **pkinit_identities**
     Specifies the location(s) to be used to find the user's X.509
     identity information.  If this option is specified multiple times,
-    the first valid value is used; this can be used to specify an
-    environment variable (with **ENV:**\ *envvar*) followed by a
-    default value.  Note that these values are not used if the user
-    specifies **X509_user_identity** on the command line.
+    each value is attempted in order until certificates are found.
+    Note that these values are not used if the user specifies
+    **X509_user_identity** on the command line.
 
 **pkinit_kdc_hostname**
     The presence of this option indicates that the client is willing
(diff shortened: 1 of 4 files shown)


greghudson:
I think we might be able to avoid repetition by moving the crypto_load_certs() call to the end of process_option_identity() (after adding a principal parameter).

greghudson:
I moved the crypto_load_certs() probe, added a test case, edited the documentation ("and authentication is attempted" still wasn't right), and edited down the commit message.

frozencemetery (approved):

kenh:
Thanks! I was going to work on those changes, but my personal intruded. The changes look perfect to me!

PR #1158, opened 2021-02-04
krb5_parse_name: improve name-type heuristics

frasertweedale:
Although name-type is only a hint, various documents specify
particular values for some use cases.  In particular:

https://tools.ietf.org/html/rfc4120#section-7.3

    The principal identifier of the ticket-granting service shall be
    composed of three parts: the realm of the KDC issuing the TGS ticket,
    and a two-part name of type NT-SRV-INST, with the first part "krbtgt"
    and the second part the name of the realm that will accept the TGT.

https://tools.ietf.org/html/rfc6111#section-3.1

    A new name type KRB_NT_WELLKNOWN is defined for well-known principal
    names.  The Kerberos principal name is defined in Section 6.2 of
    [RFC4120].
 
             KRB_NT_WELLKNOWN                  11
 
    A well-known principal name MUST have at least two or more
    KerberosString components, and the first component MUST be the string
    literal "WELLKNOWN".
 
https://tools.ietf.org/html/rfc8062#section-3

    The anonymous Kerberos principal name is defined as a well-known
    Kerberos principal name based on [RFC6111].  The value of the name-
    type field is KRB_NT_WELLKNOWN [RFC6111], and the value of the name-
    string field is a sequence of two KerberosString components:
    "WELLKNOWN" and "ANONYMOUS".

Update krb5\_parse\_name to set the name type to NT-SRV-INST when the
name being parsed is a TGS principal name, or KBT\_NT\_WELLKNOWN when
the name being parsed is a well-known principal name.  Also add some
unit tests for this feature.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -3450,6 +3450,10 @@ krb5_rd_priv(krb5_context context, krb5_auth_context auth_context,
  * @note The realm in a Kerberos @a name cannot contain slash, colon,
  * or NULL characters.
  *
+ * Beginning with release 1.20, the name type of the principal will be inferred
+ * as @c KRB5_NT_SRV_INST or @c KRB5_NT_WELLKNOWN based on the principal name.
+ * The type will be @c KRB5_NT_PRINCIPAL if a type cannot be inferred.
+ *
  * Use krb5_free_principal() to free @a principal_out when it is no longer
  * needed.
  *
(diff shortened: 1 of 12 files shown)


greghudson:
The strncmp() comparisons will also match prefixes of the fixed strings (like "kr" instead of "krbtgt").  We have an inline helper data_eq_string() in k5-int.h to do this kind of matching correctly.

We might consider doing the same for krb5_build_principal() and its variants.  Heimdal has similar heuristics and they apply to both function families.

frasertweedale:
@greghudson thank you, I will update the PR soon.

frozencemetery (commented):
Everything seems in order code-wise.  Is the commit authorship correct right now?  It looks like Greg is both author and making [annotated changes].

greghudson:
Good catch; I've fixed the author information.

frozencemetery (approved):

PR #1159, opened 2021-02-11
[WIP] PAC ticket signature

frozencemetery:
Largely untested, work in progress, etc. toward fixing https://krbdev.mit.edu/rt/Ticket/Display.html?id=8974

I also have a freeipa branch for these DAL changes https://github.com/freeipa/freeipa/compare/master...frozencemetery:pts


greghudson:
A checklist:

* We should have a KDC realm option not to issue PACs.
* PACs should be computed last (so all of the other authdata is in the signature) but placed first in the authdata list.
* krbtgt tickets shouldn't include ticket signatures in the PAC, assuming that's the Windows behavior.
* Ticket modification requests (like renewal) should work without a PAC, but shouldn't place a PAC in the modified ticket.  If there's a PAC but it isn't valid, it's reasonable to either deny the request or to issue the ticket with no PAC, I think.  If there is a valid PAC and it has a ticket signature, it should be re-signed for the new ticket.  There's probably no need to call a KDB PAC method for these requests.
* S4U2Proxy should not work without a valid PAC in the evidence ticket.  The TGT also needs a valid PAC for cross-realm RBCD requests, but I think for other S4U2Proxy requests it isn't required.
* The PAC should include a PAC_DELEGATION_INFO buffer for S4U2Proxy tickets.  This will require some NDR encoding but I don't think it would be a lot.
* We will likely run into an issue with RBCD requests in intermediate realms.  These requests look superficially like regular cross-realm TGS requests (no special flags to indicate S4U2Proxy) but the PAC is for the impersonated client, so its PAC_CLIENT_INFO won't match the header TGT client.  I assume MS handles this by noticing that there is a PAC_DELEGATION_INFO buffer in the PAC but some research may be required.

src/include/kdb.h
```diff
@@ -747,6 +728,12 @@ void krb5_db_free_authdata_info(krb5_context context, void *ad_info);
 void
 krb5_dbe_sort_key_data(krb5_key_data *key_data, size_t key_data_length);
 
+krb5_error_code
+krb5_db_issue_pac(krb5_context context, unsigned int flags,
+                  krb5_const_principal client_princ, krb5_db_entry *client,
+                  krb5_timestamp authtime, krb5_boolean is_cross_realm,
+                  krb5_pac *pac_in, krb5_pac *pac_out);
```
greghudson, line 735: Do we want pac_in/pac_out or should the method just modify the PAC in place with krb5_pac_add_buffer()?

frozencemetery: Currently the IPA code always makes a new PAC.  The worry seems to be that resigning with a different algorithm will take up a different amount of space, which would cause problems.

greghudson: But IPA wouldn't be responsible for signing the PAC with the new interface, would it?

frozencemetery: This is true.  We could move that logic into krb5.

iboukris:
Thanks for picking it up, haven't look closely but generally what I had in mind was to suppress the krb5_db_get_authdata_info() kdb method and replace it with an internal kdc function, mostly using the content of test_get_authdata_info(), possibly deferring some checks for later at test_sign_authdata() kdb method (you'd notice the checks are already duplicated there iirc).

> krbtgt tickets shouldn't include ticket signatures in the PAC, assuming that's the Windows behavior.

I had a draft where this was implemented in krb5_kdc_sign_ticket(), by simply checking if ticket->server is krbtgt and not adding the ticket-signature in that case, similarly in krb5_kdc_verify_ticket() if the ticket-signature is there then we check it, otherwise we error out if ticket->server server *isn't* krbtgt.

>  I assume MS handles this by noticing that there is a PAC_DELEGATION_INFO buffer in the PAC but some research may be required.

Yeah, I recall doing some tests which indicated the PAC_DELEGATION_INFO was being checked, do not recall the details but tired to implement some of it in kdb_test.c.

iboukris:
> similarly in krb5_kdc_verify_ticket() if the ticket-signature is there then we check it, otherwise we error out if ticket->server server isn't krbtgt.

Actually, it may be wiser to just drop the ticket-signature buffer if found in a krbtgt ticket. I think this may happen when a non-updated Windows KDC would issue a cross-realm referral based on a 2nd ticket issued from an updated Windows KDC. In this case the ticket-signature will mismatch but we could just ignore it.

src/include/kdb.h
frozencemetery: A larger, currently undocumented, problem is that some fields need modified and we don't have a way to do that.   For instance, the transited services KRB5_PAC_DELEGATION_INFO: it needs to be pulled from the old PAC, updated, and shoved into the new one.  So it seems to me that we either need to either make reissuing the responsibility of the plugin, or we need field update and deletion methods for krb5_pacs.  I don't really have a preference between the two, but the second is more work so I'll defer it until later.

greghudson: Ideally delegation info would be the responsibility of the KDC, not the KDB module.

greghudson: More generally, from context I think you're envisioning that pac_in is the PAC from the header TGT or evidence ticket.  The design I understood from Isaac is that the KDC creates a minimal PAC for the KDB module to extend.  If the KDB module does nothing, the KDC still issues the minimal PAC.  The PAC from an incoming ticket can be a separate parameter.

iboukris:
btw, i realize the authtime can also be taken from the ticket, so no need to pass it. The principal is also in the ticket except for some S4U cases, but maybe NULL could be passed for the default.

frozencemetery:
Isaac, in your code, `krb5_kdc_{sign,verify}_ticket()` are public (`krb5_`).  Was this just for symmetry with `krb5_pac_{sign,verify}_ext()`?  I'm wondering if they should actually be exposed at all beyond for testing.

iboukris:
That wasn't deliberate, I probably just wanted the test to compile, it would be nice however to export the base ticket-signature-verify function for wireshark, so it can show if the signature is ok when provided with a keytab.

frozencemetery:
I've dropped `authtime` per your suggestion.  I prefer the current behavior with the principal though - since we do care about s4u, I think explicit is better than implicit in that regard.

src/kdc/kdc_authdata.c
```diff
@@ -792,6 +377,256 @@ get_auth_indicators(krb5_context context, krb5_enc_tkt_part *enc_tkt,
...
+
+    if (!krb5_pac_parse(context, res->contents, res->length, &pac))
+        return NULL;
+
+    return pac;
+}
+
+/* Check if princ is of the form krbtgt/A@B where A != B. */
+static inline krb5_boolean
+is_cross_realm_krbtgt(krb5_context context, krb5_const_principal princ)
```
greghudson, line 404: See is_cross_tgs_principal() in kdc_util.c.

greghudson:
Realm flags in the KDC are typically handled through kdc_realm_t rather than a profile lookup at point of use.  That would require passing kdc_active_realm to handle_authdata() rather than just a context.

Ticket modification requests include renewal, proxy, validation, and forwarding (see NON_TGT_OPTION in kdc_util.h).  They should all have the same PAC handling.

greghudson:
ndr.h can be folded into kdc_util.h, which already covers multiple files.

The structure and field names for delegation info should follow krb5 naming style, not the upstream naming style.  (This is one of my big complaints about the PKINIT code.)

iboukris:
> I prefer the current behavior with the principal though - since we do care about s4u, I think explicit is better than implicit in that regard.

It occurs to me that these S4U cases are more specifically cross-realm S4U (both self and rbcd), at the same time we have the last argument 'with_realm' which denotes exactly that, so we could change its name to xrealm_s4u, and only allow the principal to differ from the ticket if the flag is true or something of the sort.

iboukris:
> > I prefer the current behavior with the principal though - since we do care about s4u, I think explicit is better than implicit in that regard.
> 
> It occurs to me that these S4U cases are more specifically cross-realm S4U (both self and rbcd), at the same time we have the last argument 'with_realm' which denotes exactly that, so we could change its name to xrealm_s4u, and only allow the principal to differ from the ticket if the flag is true or something of the sort.

Or drop the 'with_realm' argument and change principal to 'xrealm_s4u_princ', so when it is not NULL it also implies 'with_realm'.

src/include/kdb.h
```diff
@@ -1525,6 +1512,13 @@ typedef struct _kdb_vftabl {
                                void *ad_info);
 
     /* End of minor version 0 for major version 8. */
+
+    krb5_error_code (*issue_pac)(krb5_context context, unsigned int flags,
+                                 krb5_const_principal client_princ,
+                                 krb5_db_entry *proxy,
+                                 krb5_timestamp authtime,
+                                 krb5_boolean is_cross_realm,
+                                 krb5_pac old_pac, krb5_pac new_pac);
```
greghudson, line 1521: We will need to bump the major DB version for this change, and the new method belongs before the "End of minor version 0" comment for the new major version.

The new method will need a documentation comment, with particular attention to how the module should handle `flags`.

`is_cross_realm` is redundant with KRB5_KDB_FLAG_CROSS_REALM in `flags`, unless it's trying to convey something subtly different.

The fourth parameter should be named `client` instead of `proxy`.

src/kdc/kdc_authdata.c
```diff
@@ -792,8 +376,221 @@ get_auth_indicators(krb5_context context, krb5_enc_tkt_part *enc_tkt,
...
+    krb5_error_code ret;
+    krb5_pac old_pac = NULL, new_pac = NULL;
+    krb5_boolean is_cross;
+    krb5_const_principal client_princ = NULL;
+    krb5_db_entry *krbtgt = local_tgt;
+    krb5_keyblock *krbtgt_key = local_tgt_key;
+
+    /* Select the right krbtgt for the realm. */
+    if (server_key != NULL)
+        krbtgt_key = server_key;
```
greghudson, line 523: This seems concerning.  I don't think server_key will ever be null, and we never want to use it for the KDC signature of the PAC.  We always want to use local_tgt_key.

src/kdc/kdc_authdata.c
```diff
@@ -792,8 +376,221 @@ get_auth_indicators(krb5_context context, krb5_enc_tkt_part *enc_tkt,
...
+    if (subject_server != NULL)
+        krbtgt = subject_server;
+
+    is_cross = is_cross_tgs_principal(krbtgt->princ);
+
+    old_pac = get_pac(context, enc_tkt_reply->authorization_data);
```
greghudson, line 529: enc_tkt_reply is the encrypted part of the ticket we're issuing.  I don't think we'd ever find an old PAC there.

src/kdc/kdc_authdata.c
```diff
@@ -792,8 +376,221 @@ get_auth_indicators(krb5_context context, krb5_enc_tkt_part *enc_tkt,
...
+        goto done;
+    }
+
+    ret = minimal_pac(context, flags & KRB5_KDB_FLAG_CONSTRAINED_DELEGATION,
+                      client, server, old_pac, new_pac);
+    if (ret)
+        goto done;
+
+    /* Skip the KDB for modifications of tickets with valid PACs. */
+    if (old_pac == NULL || !(flags & NON_TGT_OPTION)) {
```
greghudson, line 572: For ticket modification requests, we can only get here we have a valid old PAC, so we only need the second part of this conditional.

src/kdc/kdc_authdata.c
frozencemetery: This (and a couple other things you've flagged) are cruft from the old interface.  Will fix in a bit.

src/include/kdb.h
frozencemetery: I'd been holding off on bumping the KDB version until this is closer to done, but I suppose that doesn't actually help anything.  Will make the change.

frozencemetery: I've added documentation, but I'm not sure what you have in mind for `flags`.  Probably what I have there is not what you meant :)

greghudson:
Currently minimal_pac() copies the old PAC with the exception of certain fields.  I think this will be a problem for cross-realm due to the need for SID filtering; we'll need to rely on the KDB to do this copying if we aren't simply re-signing the PAC (the `flags & NON_TGT_OPTION` case).

Tangentially, I wonder if we can simply re-sign the PAC (skipping the KDB module) for a regular TGS request with a local TGT.  I'm not familiar enough with PAC contents to be sure.

iboukris:
The re-signing was limited in the current pac api, not allowing name changes which occurs when xrealm-s4u pac reaches the final target, when the realm part gets chopped from the name. The kdc could create an empty pac (just pac_init) and pass it as out_pac and the kdb would be responsible for copying and updating all the buffers except for signature buffers and delegation-info, then the kdc simply signs out_pac.

frozencemetery:
We've been having an internal outage (for more than a week) on the infrastructure I would need to test against AD.  I had hoped to start doing that before digging into RBCD, but right now I don't know when/if I'll be able to do that at all.

abbra:
@greghudson @frozencemetery any prospects of this being merged? I'd like to get FreeIPA changes (and Samba AD DC) to be merged as well and they rely on these ones.

greghudson:
This is unfinished work; it's not just waiting to be merged.  I unfortunately do not have an AD test infrastructure set up (or even trivial access to the necessary licenses, as Microsoft has discontinued their grant program), so I can't promise that I can finish up the work any time soon.

abbra:
@greghudson there is Windows Server 2019 available for evaluation purposes https://www.microsoft.com/en-us/evalcenter/evaluate-windows-server-2019, its instances expire after six months but otherwise should be usable for testing. This doesn't solve an issue of getting some hosting capacity, though...

abbra:
@greghudson I asked my Microsoft contacts on getting MIT Kerberos on a similar program to what we have for Samba Team. They promised to discuss this with Windows engineering soon.

frozencemetery:
If it's helpful, there exist vagrant images of (evaluation) Windows server as well.  I have some scripts for that at https://github.com/frozencemetery/ad-testing/ .  I was able to get an AD DC with krb5 client, and do some S4U testing that way; this is where my NDR test blobs came from.

PR #1160, opened 2021-02-12
Restore krb5_set_default_tgs_ktypes()

greghudson:
Samba only uses the correct name (krb5_set_default_tgs_enctypes) if it
cannot find the old one in the library.

### DIFF
--- a/src/lib/krb5/krb/init_ctx.c
+++ b/src/lib/krb5/krb/init_ctx.c
@@ -376,6 +376,17 @@ krb5_set_default_tgs_enctypes(krb5_context context, const krb5_enctype *etypes)
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/krb5/krb/init_ctx.c
```diff
@@ -376,6 +376,14 @@ krb5_set_default_tgs_enctypes(krb5_context context, const krb5_enctype *etypes)
     return 0;
 }
 
+/* Old name for above function.  This is not a public API, but Samba (as of
+ * 2020-02-12) uses this name if it finds it in the library. */
```
frozencemetery, line 380: 2021, I think.

greghudson: Fixed, thanks.

frozencemetery (approved):

PR #1163, opened 2021-02-25
Allow multiple pkinit_anchor and pkinit_pool lines

kenh:
As discussed on krbdev, this allows multiple pkinit_anchor and pkinit_pool lines in krb5.conf.

To expand on this a bit; it is true that the existing code allows multiple anchor/pool lines.  But if any of the entries fail to process (e.g., file not found), then an error is returned from pkinit_identity_prompt() and pkinit ends up failing.

We've found during our PKINIT deployment that it was useful to list multiple locations in pkinit_anchors and pkinit_pool because certificate locations can vary across systems. This change weakens the error checking for the pkinit_anchors and pkinit_pool lines; if those configuration entries are specified, only one working entry is required.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_clnt.c
+++ b/src/plugins/preauth/pkinit/pkinit_clnt.c
@@ -1101,6 +1101,11 @@ pkinit_client_process(krb5_context context, krb5_clpreauth_moddata moddata,
     }
 
     if (processing_request) {
+        if (reqctx->idopts->anchors == NULL) {
+            krb5_set_error_message(context, KRB5_PREAUTH_FAILED,
+                                   _("No pkinit_anchors supplied"));
+            return KRB5_PREAUTH_FAILED;
+        }
         pkinit_client_profile(context, plgctx, reqctx, cb, rock,
                               &request->server->realm);
         /* Pull in PINs and passwords for identities which we deferred
--- a/src/plugins/preauth/pkinit/pkinit_identity.c
+++ b/src/plugins/preauth/pkinit/pkinit_identity.c
@@ -576,8 +576,9 @@ pkinit_identity_prompt(krb5_context context,
                        int do_matching,
                        krb5_principal princ)
 {
-    krb5_error_code retval = EINVAL;
+    krb5_error_code retval = 0;
     const char *signer_identity;
+    krb5_boolean valid;
     int i;
 
     pkiDebug("%s: %p %p %p\n", __FUNCTION__, context, idopts, id_cryptoctx);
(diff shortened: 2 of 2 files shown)


greghudson:
This change sort of adversely affects the behavior of anonymous PKINIT (kinit -n) when the client isn't configured for PKINIT.  I say "sort of" because the current behavior is already pretty bad.

Currently for that case, pkinit_identity_prompt() does essentially nothing and returns EINVAL, because that's what retval is initialized to.  The user either sees a bogus password prompt if WELLKNOWN/ANONYMOUS has any keys, or "kinit: Pre-authentication failed: Invalid argument while getting initial credentials" if it doesn't.

With the changes, pkinit_identity_prompt() returns 0, and the client code gets much further into making a DH request, eventually failing in create_krb5_trustedCertifiers() because id_cryptoctx->trustedCAs is NULL.  The client sees "kinit: Pre-authentication failed: Preauthentication failed while getting initial credentials" or a bogus password prompt.

The fixes for this problem are out of scope, but I will probably look into them before merging this.  (I think we need two changes: one in the preauth framework to avoid trying that can't support anonymous when doing anonymous, and one within PKINIT to explicitly detect the lack of configuration and supply a good error message.)

frozencemetery (commented):
This doesn't quite seem to do what it says on the tin.  The code suggests that only one success is required (in both cases) if any are specified, while the commit message suggests to me that a successful return implies that one successful load will have occurred on return.  From what I can tell, the code is fine here, but I could be misunderstanding.

greghudson:
> The code suggests that only one success is required (in both cases) if any are specified, while the commit message suggests to me that a success return implies that one load will have completed

What's the difference?  Probably the commit message needs to be reworded, but I'm not understanding the distinction.

kenh:
> > The code suggests that only one success is required (in both cases) if any are specified, while the commit message suggests to me that a success return implies that one load will have completed
> 
> What's the difference? Probably the commit message needs to be reworded, but I'm not understanding the distinction.

As I understand Robbie's comment ... the commit message implies AT LEAST ONE successful load. The actual behavior (which I think was agreed on) is, "_If_ `pkinit_anchors` or `pkinit_pool` is configured, at least one of the configured lines must result in a successful load.

Whether or not the commit message is worth amending ... I'll leave that up to others. I'm also willing to re-work the patch to restore the previous behavior of anonymous pkinit if that is helpful.

greghudson:
I still don't understand what was confusing about the commit message, but I reworded it slightly; let me know if it makes more sense.

I decided to consider checking for a lack of pkinit_anchors to be in scope for the commit, and added a check to pkinit_client_process().  That's enough to avoid making the anonymous PKINIT unconfigured behavior any worse.

frozencemetery:
I think this makes more sense, thanks.

I think Ken explained my issue better than I did, but for function returning zero: the code had "if set, one line must load", while the commit had "at least one must have loaded".  The difference is that the commit suggests one can assume a load completed, while the code restricts that to only when one is configured.

I think you're both right that this is probably an unimportant distinction, and if you prefer the old wording that's fine by me.

PR #1166, opened 2021-03-03
Fix some docs warnings

frozencemetery:
The first commit is the impetus for the PR, though I don't have a good solution.  It seems that our docs builds fail when a literal block ends a doxygen section (trace in commit message).

Second commit fixes all (other) warnings from doxygen.  These are visible in CI, but are not currently fatal.

Note that as in 281210909beef4683be3b63bc1ac1e75c2c9c7eb , I have not addressed the sphinx 4.0 deprecations.  Right now, those are:

/home/rharwood/krb5.git/newer-sphinx/src/doc/rst_composite/conf.py:102: RemovedInSphinx40Warning: The app.add_stylesheet() is deprecated. Please use app.add_css_file() instead.
  app.add_stylesheet('kerb.css')
  /usr/lib/python3/dist-packages/docutils/parsers/rst/states.py:2147: RemovedInSphinx40Warning: highlightlang directive is deprecated. Please use highlight directive instead.

### DIFF
--- a/doc/tools/doxybuilder_types.py
+++ b/doc/tools/doxybuilder_types.py
@@ -268,38 +268,36 @@ class DoxyTypes(object):
 
     def _process_paragraph_content(self, node):
 
+        def add_text(l, s):
+            # Add a space if it wouldn't be at the start or end of a line.
+            if l and not l[-1].endswith('\n') and not s.startswith('\n'):
+                l.append(' ')
+            l.append(s)
+
         result = list()
         content = node.xpath(".//text()")
         for e in content:
-            if node is e.getparent():
-                result.append(e.strip())
+            if e.is_tail or node is e.getparent():
+                add_text(result, e.strip())
             elif e.getparent().tag == 'ref':
-                if e.is_tail:
-                    result.append(e.strip())
-                elif e.strip().find('(') > 0:
-                    result.append(':c:func:`%s`' % e.strip())
+                if e.strip().find('(') > 0:
+                    add_text(result, ':c:func:`%s`' % e.strip())
                 elif e.isupper():
-                    result.append(':c:data:`%s`' % e.strip())
+                    add_text(result, ':c:data:`%s`' % e.strip())
                 else:
-                    result.append(':c:type:`%s`' % e.strip())
+                    add_text(result, ':c:type:`%s`' % e.strip())
             elif e.getparent().tag == 'emphasis':
-                if e.is_tail:
-                    result.append(e.strip())
-                else:
-                    result.append('*%s*' % e.strip())
+                add_text(result, '*%s*' % e.strip())
             elif e.getparent().tag == 'computeroutput':
-                if e.is_tail:
-                    result.append(e.strip())
-                else:
-                    result.append('*%s*' % e.strip())
-            elif  e.getparent().tag == 'defname':
-                result.append('%s, ' % e.strip())
+                add_text(result, '*%s*' % e.strip())
+            elif e.getparent().tag == 'defname':
+                add_text(result, '%s, ' % e.strip())
             elif e.getparent().tag == 'verbatim':
-                result.append('\n::\n\n')
-                result.append(textwrap.indent(e, '    ', lambda x: True))
-                result.append('\n')
+                add_text(result, '\n::\n\n')
+                add_text(result, textwrap.indent(e, '    ', lambda x: True))
+                add_text(result, '\n')
 
-        result = ' '.join(result)
+        result = ''.join(result)
 
         return result
 
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -3123,7 +3123,7 @@ krb5_get_credentials(krb5_context context, krb5_flags options,
  * Serialize a @c krb5_creds object.
  *
  * @param [in]  context         Library context
- * @param [in]  creds           The credentials object to serialize
+ * @param [in]  in_creds        The credentials object to serialize
  * @param [out] data_out        The serialized credentials
  *
  * Serialize @a creds in the format used by the FILE ccache format (vesion 4)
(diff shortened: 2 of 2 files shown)


greghudson:
Can I see the contents of KRB5_RESPONDER_QUESTION_PKINIT.rst (from doc_composite) on the system in question?

frozencemetery:
Sure thing.  This is Debian testing:

```rst
.. highlightlang:: c

.. _KRB5-RESPONDER-QUESTION-PKINIT-data:

KRB5_RESPONDER_QUESTION_PKINIT
==============================

..
.. data:: KRB5_RESPONDER_QUESTION_PKINIT
..

PKINIT responder question.

The PKINIT responder question is asked when the client needs a password that's being used to protect key information, and is formatted as a JSON object. A specific identity's flags value, if not zero, is the bitwise-OR of one or more of the KRB5_RESPONDER_PKINIT_FLAGS_TOKEN_* flags defined below, and possibly other flags to be added later. Any resemblance to similarly-named CKF_* values in the PKCS#11 API should not be depended on. 
::

     {
        identity <string> : flags <number>,
        ...
    }
 
 
::

      The answer to the question MUST be JSON formatted:  
 
::

     {
        identity <string> : password <string>,
        ...
    }
 
 
::

       


===================================== ======================
``KRB5_RESPONDER_QUESTION_PKINIT``       ``"pkinit"``
===================================== ======================
```

(The warning points at line 41, which is the last line that isn't mostly equals signs.)

greghudson:
Okay, I get the same output, so it looks like this issue arises from a newer version of python3-sphinx (as opposed to a newer version of doxygen).

There are some mysterious extra spaces there: one before each of the open braces (so they are indented five spaces instead of four) and one on the blank line after each close brace.  These come from our doc/tools/doxybuilder_types.py line 302 joining a list of strings with space separators.

That might or might not be the problem.  The warning points to line 41 in a 37-line file, and the workaround seems to be serving to separate the literal block from the table below, so the problem could be sphinx getting confused and treating the table as appearing inside the literal block.  Can you try removing those four extra spaces in rst_composite/appdev/refs/macros/KRB5_RESPONDER_QUESTION_PKINIT.rst and seeing if sphinx-build still complains?

greghudson:
I have a vague sense of how Sphinx might be getting confused.  Per https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks literal blocks can be indented or quoted.  Quoted blocks often have ">" prefixes but lots of punctuation is valid, including "=".  A blank line doesn't terminate a literal block, but text at the beginning of a line does.  So with no unindented text in between, the contents of the file looks to a literal block parser like some indented lines (with inconsistent indentation), three blank lines, and a quoted line (actually the beginning of a table).  Since literal blocks shouldn't contain a mix of indented and quoted lines, I think the parser should end the literal block at the last indented line, but getting confused and issuing a warning seems within the realm of possibility.  Fixing the inconsistent indentation might or might not avoid the problem.

Unfortunately I cannot reproduce the warning locally.  I tried sphinx 3.5.1 (the latest) and 3.4.3 (to match debian-testing) from pypi, and all I see is the two deprecation warnings.

frozencemetery:
> Can you try removing those four extra spaces in rst_composite/appdev/refs/macros/KRB5_RESPONDER_QUESTION_PKINIT.rst and seeing if sphinx-build still complains?

There are four blank lines between the final `::` and the table beginning.  One has seven spaces on it - removing them does nothing.  Removing all lines causes build to succeed.  However, it no longer renders as a table.

Removing only the final `::` (and leaving all blank / whitespace lines alone) causes the build to succeed, and the table to render correctly, even without fixing the indentation on the final '}'.

Comparing my local HTML output to the web.mit.edu/kerberos output, I observe an additional difference: the text "The answer to the question must be JSON formatted:" is a literal block for me, but is normal text on the web version.  (I assume this is because it's indented.)

> Unfortunately I cannot reproduce the warning locally. I tried sphinx 3.5.1 (the latest) and 3.4.3 (to match debian-testing) from pypi, and all I see is the two deprecation warnings.

It also happens on Fedora (34 and rawhide, both also 3.4.3), though I imagined that would be harder to check :)

greghudson:
> I get the same output, so it looks like this issue arises from a newer version of python3-sphinx (as opposed to a newer version of doxygen).

I was wrong; your output has extra "::" lines in it after each literal block.  These signal that the subsequent content is a literal block, which explains why "The answer to the question MUST be JSON formatted:" is formatted as a literal block and explains the warning for the table.  So I was looking in the wrong place; the problem likely arises from the Doxygen version (as it interacts with our conversion scripts), not the Sphinx version.

I will see about reproducing this locally with a different version of Doxygen, but if you could stash your doc/doxy/xml/krb5_8hin.xml somewhere I can read it, I can work from that.

frozencemetery:
> I will see about reproducing this locally with a different version of Doxygen, but if you could stash your doc/doxy/xml/krb5_8hin.xml somewhere I can read it, I can work from that.

https://rharwood.fedorapeople.org/krb5_8hin.xml

greghudson:
What version of Doxygen is that?

frozencemetery:
This output is from doxygen 1.9.1.

greghudson:
I think we might be running into an lxml bug.  If you run:

    from lxml import etree

    s = '<para>a<verbatim>b</verbatim>c</para>'
    tree = etree.XML(s)
    content = tree.xpath(".//text()")
    for e in content:
        print(e.getparent().tag + ': ' + e)

You get:

    para: a
    verbatim: b
    verbatim: c

when the parent of `c` should be the `para` tag, as I understand things.

The old version of doxygen put `</para><para>` after each `</verbatim>` tag, so we didn't run into this bug; the subsequent content was correctly attached to a paragraph tag.  The new version does not put them there, so "The answer to the question..." is treated as being in a verbatim block, as is a single space after the second (real) verbatim block.  That single-space literal block causes the table to be treated as part of a literal block by the RST parser, leading to the warning.

I will try to submit a bug report, but I think we also need to figure out a workaround.

greghudson:
Reading the tutorial carefully, this is actually deliberate behavior.  lxml considers `c` to be a "tail" of the `verbatim` tag, so its parent is the prior tag rather than the containing `para` tag.  The other cases in the for loop check for `e.is_tail`, but when I added verbatim block handling in commit 281210909beef4683be3b63bc1ac1e75c2c9c7eb, I didn't know to do so (and the case I was replacing didn't need to).

I will replace the first commit with a candidate fix to doxybuilder_types.py.

frozencemetery:
> I will replace the first commit with a candidate fix to doxybuilder_types.py.

This fixes it for me, thanks!

PR #1167, opened 2021-03-04
cc: add krb5_cc_notification_path

pbrezina:
This is a new API call to implement generic notification interface.

The function return a path to the file that can be watched for changes
through file system watchers like inotify or fswatch. The file is
created, touched and destroyed when the ccache is initialized, changed
and destroyed. The caller should setup a watch on the parent directory
and listen for the file events.

Implemented for FILE, DIR and KCM.

SSSD KCM implementation (Work in progress):
https://github.com/pbrezina/sssd/tree/kcm-notifications

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1485,6 +1485,15 @@ if test "x$with_krb5_config" != xno; then
 		: "${DEFCKTNAME=`$with_krb5_config --defcktname`}"
 	fi
 fi
+
+# Setup the notification path
+AC_ARG_WITH([notification-path],
+  [AS_HELP_STRING([--with-notification-path=PATH],
+   [path to the directory with user notification file])])
+if test x"$with_notification_path" != x; then
+  AC_DEFINE_UNQUOTED(NOTIFICATION_PATH, ["$with_notification_path"], [Notification path])
+fi
+
 dnl The outer brackets around the case statement prevent m4 from eating the
 dnl brackets in the glob patterns.
 if test "${DEFCCNAME+set}" != set; then
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -4572,6 +4572,24 @@ krb5_boolean KRB5_CALLCONV
 krb5_cc_support_switch(krb5_context context, const char *type);
 
 /**
+ * Retrieve path to the ccache notification file. The file is touched everytime
+ * the ccache or the collection that the ccache is part of has changed.
+ *
+ * You can setup a file system watch (e.g. inotify) on the path and
+ * watch for the file events (IN_ATTRIB) to monitor activity of the ccache.
+ *
+ * @param [in] context          Library context
+ * @param [in] cache            Credential cache handle
+ * @param [out] path            Notification file path
+ *
+ * @retval 0              Success. Path is set to the notification file path.
+ * @retval KRB5_CC_NOSUPP Notifications are not supported by this ccache.
+ * @return Kerberos error codes
+ */
+krb5_error_code KRB5_CALLCONV
+krb5_cc_notification_path(krb5_context context, krb5_ccache cache, char **path);
+
+/**
  * Find a credential cache with a specified client principal.
  *
  * @param [in]  context         Library context
--- a/src/lib/krb5/ccache/cc-int.h
+++ b/src/lib/krb5/ccache/cc-int.h
@@ -217,6 +217,7 @@ struct _krb5_cc_ops {
     krb5_error_code (KRB5_CALLCONV *lock)(krb5_context, krb5_ccache);
     krb5_error_code (KRB5_CALLCONV *unlock)(krb5_context, krb5_ccache);
     krb5_error_code (KRB5_CALLCONV *switch_to)(krb5_context, krb5_ccache);
+    krb5_error_code (KRB5_CALLCONV *notification_file)(krb5_context, krb5_ccache, char **);
 };
 
 extern const krb5_cc_ops *krb5_cc_dfl_ops;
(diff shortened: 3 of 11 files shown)


greghudson:
My understanding is that this API is intended to support a GNOME credential UI of some kind.

Does GNOME actually want to monitor changes to specific ccaches, or does it want to know about any changes to the default collection?

halfline:
> My understanding is that this API is intended to support a GNOME credential UI of some kind.
> 
> Does GNOME actually want to monitor changes to specific ccaches, or does it want to know about any changes to the default collection?

gnome-online-accounts currently only monitors for changes to the default collection, but upon getting change notification, it refreshes its internal bookkeeping from the entire collection of credential caches returned from `krb5_cccol_cursor_*` apis.

It uses that bookkeeping to monitor for when credentials are expiring and to try renew or refresh.

If we had a per ccache notification mechanism we would use that, I think.

greghudson:
> If we had a per ccache notification mechanism we would use that, I think.

So GNOME would set up a notification file for each cache in the default collection, and check just the one cache when it receives a notification?  If new caches are added to the collection, would anything cause them to be monitored?

Apologies for all the questions.  Once we add an API to libkrb5, we're committed to supporting it forever, so I want to make sure it will remain useful.  So I would like to make sure that this functionality wouldn't be better supported by a DBus contract (or similar) between sssd and GNOME, or by making sssd responsible for renewal/refresh.

pbrezina:
> > If we had a per ccache notification mechanism we would use that, I think.
> 
> So GNOME would set up a notification file for each cache in the default collection, and check just the one cache when it receives a notification? If new caches are added to the collection, would anything cause them to be monitored?

I think we have to distinguish collection cache type (KCM, DIR) and non-collection (FILE). I intentionally left-out other types but we should think about them as well before we merge anything.

* Collection type:
```
/.../ccol/cc1
/.../ccol/cc2
/.../ccol/cc3
```

Caller will call `krb5_cc_notification_path(cc1) -> /.../ccol/cc1` and setup watch for `/.../ccol` so it can receive also create and delete event, not only touch. For collections, we know that `ccol` will only contain files from the collection, so when a file gets created that is not from a known ccache, we know a new ccache was created, then you iterate over the collection to find it. You can watch for kinit like this.

* Non-collection type
```
/.../dir/file_cc
```
Caller will call `krb5_cc_notification_path(file_cc) -> /.../dir/file_cc` and setup watch for `/.../dir`. But it will only act on events from `/.../dir/file_cc` so it does not respond to creation of random files.

We should probably change the declarations to
```c
krb5_cc_notification_path(krb5_context context, krb5_ccache cache, char **path, bool *is_collection);
```

So the caller knows what to do based on `is_collection`.

> Apologies for all the questions. Once we add an API to libkrb5, we're committed to supporting it forever, so I want to make sure it will remain useful.

It is crucial to make things right so please keep asking and suggesting. I value your cooperation.

> So I would like to make sure that this functionality wouldn't be better supported by a DBus contract (or similar) between sssd and GNOME, or by making sssd responsible for renewal/refresh.

There were some attempts to use D-Bus, as well as KCM only file based notifications. But in the end we thought it would be best if we implemented a generic notification API so the user does not have to distinguish between multiple cache types and implement three different solutions (inotify for FILE and DIR, something for KCM and polling for other).

pbrezina:
@greghudson The proposed design had lots of caveat when trying to deal with different ccaches then FILE and KCM. Each ccache has its own specifics which would have to be propagated to the API. I changed it a lot. The notifications are now handled directly from `ccfns.c` and each ccache only returns a filename. The `krb5_cc_notification_path` then stores this file inside a configurable `NOTIFICATION_PATH`. `--with-notification-path=/run/user/%{euid}` should be used on systems that has this available, if it is not set (default) the functionality will be missing.

I added support to FILE, DIR, KEYRING, KCM. I did not add it to MEMORY since notifications has no use there.

What do you think?

greghudson:
I think this design adds too much complexity to libkrb5, and is likely to run into edge cases (such as other implementations, caches shared between uids, and network filesystems).

pbrezina:
We were also thinking about exposing file descriptor instead of a path and dealing with inotify (or keyring notifications) directly in libkrb5. But this would add even more complexity there, so if the current patch is too much then this would be a dead end, I suppose.

Is there any suggestion from you that would solve the issues? Or should we abandon the idea of having a generic interface and implement a KCM specific one?

greghudson:
> Is there any suggestion from you that would solve the issues? Or should we abandon the idea of having a generic interface and implement a KCM specific one?

I think so.  While I get that it's not very satisfying, I'd prefer solving this problem outside of the ccache layer for now.

The original design of the ccache layer is essentially a type-specific data container, making few assumptions about the systems built on top of it.  Because we also provide command-line tools interfacing with ccaches (kinit, klist, kdestroy), the ccache layer is unfortunately the only consistent integration point for building more structured systems.  But it is also an awkward integration point: it lives inside a cross-platform library with a long-term public API, and it is not the only implementation.  The results of this tension have already been seen in the dis-satisfactory history of collection support.

Querying a ccache about notification support is a fairly conservative extension, except that there are a multitude of notification systems and a libkrb5 API can't be expected to talk about most of them.  Restricting the query to filesystem pathnames works to a first order: DIR and FILE kind of naturally support path-based notifications and KCM can be extended to support it (although there might be permissions issues), but KEYRING can't, nor can native types like MSLSA and Apple's XCC.

Actually building a notification system into the ccache layer is a large expansion in scope, and fits only slightly better than building a notification system into a hash table library (which I'm sure has been done).

pbrezina:
Thank you for looking into it. I'm fine implementing this only for KCM, though it would be great to see some generic support and interface in the future.

PR #1168, opened 2021-03-10
let loadConfigFiles() to use opendir(3)/readdir(3) instead of glob(3)

Sashan:
It seems to me `glob(3)` does not meet `GSSAPI` expectations. No matter
what I was trying to do, it makes `libgss` to read plugin configuration files
found in `etc/gss/mech.d` directory. I wrote a simple test to figure out
how `glob(3)` works:
```

int
main(int argc, const char *argv[])
{
	int	rv;
	glob_t	globlist;
	int	i = 0;

	memset(&globlist, 0, sizeof (glob_t));

	if (argc == 1)
		return (1);

	rv = glob(argv[1], 0, NULL, &globlist);

	if (rv == 0) {
		while (globlist.gl_pathv[i] != NULL) {
			printf("%s\n", globlist.gl_pathv[i]);
			i++;
		}
	}

	return (rv);
}
```

trying to run programm above on `Solaris`/`OpenBSD`:
I'm getting output as follows:
```
lumpy$ ./glob /tmp/etc/gss/mech.d/*.conf
/tmp/etc/gss/mech.d/1.conf
lumpy$ ls -1 /tmp/etc/gss/mech.d/*.conf
/tmp/etc/gss/mech.d/1.conf
/tmp/etc/gss/mech.d/2.conf
/tmp/etc/gss/mech.d/3.conf
/tmp/etc/gss/mech.d/4.conf
/tmp/etc/gss/mech.d/5.conf

```
It looks like `glob(3)` yields one entry at most.

Proposed pull request replaces `glob(3)` with `opendir(3)`/`readdir(3)`
to process all plugin configuration files found in `/etc/gss/mech.d/` directory.
It also does not require `.conf` suffix for plugin configuration.

### DIFF
--- a/src/lib/gssapi/mechglue/g_initialize.c
+++ b/src/lib/gssapi/mechglue/g_initialize.c
@@ -42,7 +42,8 @@
 #include <ctype.h>
 #include <errno.h>
 #ifndef _WIN32
-#include <glob.h>
+#include <sys/types.h>
+#include <dirent.h>
 #endif
 
 #define	M_DEFAULT	"default"
@@ -61,7 +62,7 @@
 #ifndef MECH_CONF
 #define	MECH_CONF "/etc/gss/mech"
 #endif
-#define MECH_CONF_PATTERN MECH_CONF ".d/*.conf"
+#define MECH_CONF_DIR MECH_CONF ".d"
 
 /* Local functions */
 static void addConfigEntry(const char *oidStr, const char *oid,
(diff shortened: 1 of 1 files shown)


greghudson:
> lumpy$ ./glob /tmp/etc/gss/mech.d/*.conf

The shell is expanding `/tmp/etc/gss/mech.d/*.conf` to multiple arguments, and then your program is using the first of those.  Put the glob pattern inside double quotes and see if you still only get one result.

Sashan:
> > lumpy$ ./glob /tmp/etc/gss/mech.d/*.conf
> 
> The shell is expanding `/tmp/etc/gss/mech.d/*.conf` to multiple arguments, and then your program is using the first of those. Put the glob pattern inside double quotes and see if you still only get one result.

yes, of course. you are absolutely right. I've completely forgot about expansion done by shell. I'm closing this pull request. It's invalid.

PR #1170, opened 2021-03-15
Use krb5int_open_plugin for PKCS#11 module

kenh:
Instead of calling dlopen() directly, use the krb5 cross-platform interfaces (krb5int_open_plugin()).

The goal here is to eventually support pkinit on Windows; this is just the first small step in that direction. We had talked about this change a while back on krbdev; the reason I am submitting a pull request now is I would like to submit a pull request soon to include a lot of extra trace points in the pkinit plugin, and this is a small piece of that.

This patch also removes the include of dlfcn.h in pkinit_clnt.c and pkinit_identity.c; that file doesn't exist on Windows and I am not sure why those files include that header as they don't call dlopen(). It looks like it's been that way since the beginning.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_clnt.c
+++ b/src/plugins/preauth/pkinit/pkinit_clnt.c
@@ -34,7 +34,6 @@
 #include "k5-json.h"
 
 #include <unistd.h>
-#include <dlfcn.h>
 #include <sys/stat.h>
 
 /**
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -33,7 +33,6 @@
 #include "pkinit_crypto_openssl.h"
 #include "k5-buf.h"
 #include "k5-hex.h"
-#include <dlfcn.h>
 #include <unistd.h>
 #include <dirent.h>
 #include <arpa/inet.h>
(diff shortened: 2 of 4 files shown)


greghudson:
I pushed some changes:

* einfo needs to be freed (can be conditional on non-zero error return).
* It's not conformant to cast `&getflist` in order to pass it to krb5int_get_plugin_func().  (g_initialize.c does this and it shouldn't.)
* There's no reason for the p11_module field to be a `void *` if it's no longer holding a dlopen() handle.
* pkinit_C_UnloadModule() was an unnecessary wrapper and can be eliminated.

Depending on where you're going with the tracing changes, the freeing of einfo might want to be moved.  There can be a separate error path for krb5int_get_plugin_func() and getflist() errors if necessary.

kenh:
> I pushed some changes:

Those changes all look fine to me, BTW. I'll bring them into our code base so future patches are based on this.

>     * einfo needs to be freed (can be conditional on non-zero error return).

Sigh. I **knew** that when I was looking at the errinfo API, but somehow I didn't actually do that.

>     * It's not conformant to cast `&getflist` in order to pass it to krb5int_get_plugin_func().  (g_initialize.c does this and it shouldn't.)

Fair enough; I think I was just trying to save a few steps.  BTW, is that just a style issue or a standard violation? I would just like to know for the future.
 
>     * There's no reason for the p11_module field to be a `void *` if it's no longer holding a dlopen() handle.
> 
>     * pkinit_C_UnloadModule() was an unnecessary wrapper and can be eliminated.

These two both fell under the mantle of, "Make as few changes to PKINIT as possible, because that code has so many unintended side effects.

> Depending on where you're going with the tracing changes, the freeing of einfo might want to be moved. There can be a separate error path for krb5int_get_plugin_func() and getflist() errors if necessary.

Almost certainly will move the freeing of einfo, but in my limited experience all of problems at that stage come from the PKCS#11 library failing at dlopen() rather than getflist() failing.

greghudson:
> Fair enough; I think I was just trying to save a few steps. BTW, is that just a style issue or a standard violation? I would just like to know for the future.

It's a standard violation, specifically a strict aliasing violation.  Per C99 6.5, the effective type of getflist is `CK_RV (*)(CK_FUNCTION_LIST_PTR_PTR)` as that is its declared type.  Its stored value must only be accessed by an lvalue expression meeting one of six properties (a compatible type, a qualified version of a compatible type, etc.).  `(void *)()` doesn't meet any of the properties.

The same applies to functions yielding a polymorphic pointer via an output parameter.  While it would often be more convenient to pass `(realtype **)&var`, conformant code must use an intermediate `void *` variable instead.  The reasoning is that `void *` and `realtype *` don't necessarily have the same representation, and the compiler might have to generate code to convert between them.  Although real modern platforms do use the same representation for all pointer types, compilers have trended towards being increasingly picky about standards violations and breaking nonconformant code.

PR #1171, opened 2021-03-15
gss-server: Avoid calling listen(2) with zero backlog

iv-m:
Calling listen system call with zero backlog makes gss-server execution (and tests) racy on slower and busy systems.

### DIFF
--- a/src/appl/gss-sample/gss-server.c
+++ b/src/appl/gss-sample/gss-server.c
@@ -766,8 +766,6 @@ main(int argc, char **argv)
         int     stmp;
 
         if ((stmp = create_socket(port)) >= 0) {
-            if (listen(stmp, max_threads == 1 ? 0 : max_threads) < 0)
-                perror("listening on socket");
             fprintf(stderr, "starting...\n");
 
             do {
--- a/src/tests/gss-threads/gss-server.c
+++ b/src/tests/gss-threads/gss-server.c
@@ -782,9 +782,6 @@ main(int argc, char **argv)
 
         stmp = create_socket(port);
         if (stmp >= 0) {
-            if (listen(stmp, max_threads == 1 ? 0 : max_threads) < 0)
-                perror("listening on socket");
-
             do {
                 struct _work_plan * work = malloc(sizeof(struct _work_plan));
 


iv-m:
I've seen consistent test failures on Loongson 3A3000 machine I use to build packages for 32-bit MIPS port of ALT Sisyphus:

```
====== TGS ======
*** [9] Starting: /usr/src/RPM/BUILD/krb5-1.19.1/src/appl/gss-sample/gss-server -export -port 61005 host
starting...
*** [9] Started with pid 13266
*** [10] Executing: /usr/src/RPM/BUILD/krb5-1.19.1/src/appl/gss-sample/gss-client -port 61005 -krb5 localhost.localdomain host testmsg
connecting to server: Connection timed out
*** [10] Completed with return code 1
*** Failure: /usr/src/RPM/BUILD/krb5-1.19.1/src/appl/gss-sample/gss-client failed with code 1.
*** Last mark: TGS
*** Last command (#10): /usr/src/RPM/BUILD/krb5-1.19.1/src/appl/gss-sample/gss-client -port 61005 -krb5 localhost.localdomain host testmsg
*** Output of last command:
connecting to server: Connection timed out
*** Failed in test pass: default
```

It turned out that kernel dropped the client packages:


```
TCP: request_sock_TCP: Possible SYN flooding on port 61005. Dropping request.  Check SNMP counters.
```

It turned out that during tests gss-server called `listen` system with 0 backlog parameter. The change I'm proposing here fixed the problem.

greghudson:
I think the offending listen() call can simply be removed.  There is already a `listen(s, 5)` call in create_socket(), and I don't see any reason for the kernel backlog to be related to the number of threads.

(Also questionable, but well outside the scope of this PR: whether we need a second version of the GSS sample program in tests/gss-threads.)

iv-m:
> I think the offending listen() call can simply be removed. There is already a `listen(s, 5)` call in create_socket(), and I don't see any reason for the kernel backlog to be related to the number of threads.

I thought so, but opted for a more conservative solution fist.

Deletion works for me, and looks better. I've updated the PR.

PR #1172, opened 2021-03-17
Add additional KRB5_TRACE points

kenh:
As discussed previously, we are in the process of moving away from our own forked PKINIT plugin to the stock MIT PKINIT plugin. One thing our forked PKINIT plugin could do was conditionally enable the pkiDebug macro (this plugin fork predates the KRB5_TRACE functionality)

After observing the existing KRB5_TRACE output, our customer support staff had communicated to me that the existing trace points did not convey enough information that they could successfully debug PKINIT problems. This pull request adds additional trace points based on their feedback.

They specifically asked for more information about the following things:

- Any errors/messages regarding loading a PKCS#11 plugin
- Information about the certificates being processed from a PKCS#11 plugin
- Certificate matching (and more specifically, details about SAN matching)
- The certificate chain used to validate a selected certificate

All of those things are covered in this change request. I think nearly all of these have a corresponding pkiDebug() statement.

I attempted to conform to the recommended usage of KRB5_TRACE in this patch. Any suggested changes are welcome.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -1964,7 +1964,7 @@ crypto_retrieve_X509_sans(krb5_context context,
     krb5_principal *princs = NULL;
     char **upns = NULL;
     unsigned char **dnss = NULL;
-    unsigned int i, num_found = 0, num_sans = 0;
+    unsigned int i, num_sans = 0;
     X509_EXTENSION *ext = NULL;
     GENERAL_NAMES *ialt = NULL;
     GENERAL_NAME *gen = NULL;
@@ -2047,7 +2047,6 @@ crypto_retrieve_X509_sans(krb5_context context,
                              __FUNCTION__);
                 } else {
                     p++;
-                    num_found++;
                 }
             } else if (upns != NULL &&
                        OBJ_cmp(plgctx->id_ms_san_upn,
@@ -2058,6 +2057,7 @@ crypto_retrieve_X509_sans(krb5_context context,
                 upns[u] = k5memdup0(name.data, name.length, &ret);
                 if (upns[u] == NULL)
                     goto cleanup;
+                u++;
             } else {
                 pkiDebug("%s: unrecognized othername oid in SAN\n",
                          __FUNCTION__);
@@ -2079,7 +2079,6 @@ crypto_retrieve_X509_sans(krb5_context context,
                              __FUNCTION__);
                 } else {
                     d++;
-                    num_found++;
                 }
             }
             break;
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -32,6 +32,7 @@
 #include "k5-int.h"
 #include "pkinit_crypto_openssl.h"
 #include "k5-buf.h"
+#include "k5-err.h"
 #include "k5-hex.h"
 #include <unistd.h>
 #include <dirent.h>
@@ -101,8 +102,6 @@ static krb5_error_code pkinit_login
  CK_TOKEN_INFO *tip, const char *password);
 static krb5_error_code pkinit_open_session
 (krb5_context context, pkinit_identity_crypto_context id_cryptoctx);
-static struct plugin_file_handle *pkinit_C_LoadModule
-(const char *modname, CK_FUNCTION_LIST_PTR_PTR p11p);
 #ifdef SILLYDECRYPT
 CK_RV pkinit_C_Decrypt
 (pkinit_identity_crypto_context id_cryptoctx,
@@ -143,8 +142,8 @@ static int
 wrap_signeddata(unsigned char *data, unsigned int data_len,
                 unsigned char **out, unsigned int *out_len);
 
-static char *
-pkinit_pkcs11_code_to_text(int err);
+static const char *
+pkcs11err(int err);
 
 
 #ifdef HAVE_OPENSSL_CMS
@@ -1195,11 +1194,10 @@ cms_signeddata_create(krb5_context context,
             }
             certstack = X509_STORE_CTX_get1_chain(certctx);
             size = sk_X509_num(certstack);
-            pkiDebug("size of certificate chain = %d\n", size);
             for(i = 0; i < size - 1; i++) {
                 X509 *x = sk_X509_value(certstack, i);
                 X509_NAME_oneline(X509_get_subject_name(x), buf, sizeof(buf));
-                pkiDebug("cert #%d: %s\n", i, buf);
+                TRACE_PKINIT_CERT_CHAIN_NAME(context, (int)i, buf);
                 sk_X509_push(cert_stack, X509_dup(x));
             }
             X509_STORE_CTX_free(certctx);
@@ -1988,20 +1986,16 @@ crypto_retrieve_X509_sans(krb5_context context,
 
     X509_NAME_oneline(X509_get_subject_name(cert),
                       buf, sizeof(buf));
-    pkiDebug("%s: looking for SANs in cert = %s\n", __FUNCTION__, buf);
 
     l = X509_get_ext_by_NID(cert, NID_subject_alt_name, -1);
     if (l < 0)
         return 0;
 
     if (!(ext = X509_get_ext(cert, l)) || !(ialt = X509V3_EXT_d2i(ext))) {
-        pkiDebug("%s: found no subject alt name extensions\n", __FUNCTION__);
+        TRACE_PKINIT_SAN_CERT_NONE(context, buf);
         goto cleanup;
     }
     num_sans = sk_GENERAL_NAME_num(ialt);
-
-    pkiDebug("%s: found %d subject alt name extension(s)\n", __FUNCTION__,
-             num_sans);
 
     /* OK, we're likely returning something. Allocate return values */
     if (princs_ret != NULL) {
@@ -2088,6 +2082,8 @@ crypto_retrieve_X509_sans(krb5_context context,
         }
     }
     sk_GENERAL_NAME_pop_free(ialt, GENERAL_NAME_free);
+
+    TRACE_PKINIT_SAN_CERT_COUNT(context, (int)num_sans, p, u, d, buf);
 
     retval = 0;
     if (princs != NULL && *princs != NULL) {
@@ -3547,34 +3543,47 @@ prepare_enc_data(const uint8_t *indata, int indata_len, uint8_t **outdata,
 
 #ifndef WITHOUT_PKCS11
 static struct plugin_file_handle *
-pkinit_C_LoadModule(const char *modname, CK_FUNCTION_LIST_PTR_PTR p11p)
+load_pkcs11_module(krb5_context context, const char *modname,
+                   CK_FUNCTION_LIST_PTR_PTR p11p)
 {
-    struct plugin_file_handle *handle;
+    struct plugin_file_handle *handle = NULL;
     CK_RV (*getflist)(CK_FUNCTION_LIST_PTR_PTR);
     struct errinfo einfo = EMPTY_ERRINFO;
+    const char *errmsg = NULL;
     void (*sym)();
     long err;
     CK_RV rv;
 
-    pkiDebug("loading module \"%s\"... ", modname);
-    if (krb5int_open_plugin(modname, &handle, &einfo) != 0) {
-        pkiDebug("not found\n");
-        return NULL;
+    TRACE_PKINIT_PKCS11_OPEN(context, modname);
+    err = krb5int_open_plugin(modname, &handle, &einfo);
+    if (err) {
+        errmsg = k5_get_error(&einfo, err);
+        TRACE_PKINIT_PKCS11_OPEN_FAILED(context, errmsg);
+        goto error;
     }
 
     err = krb5int_get_plugin_func(handle, "C_GetFunctionList", &sym, &einfo);
-    k5_clear_error(&einfo);
-    if (!err) {
-        getflist = (CK_RV (*)())sym;
-        rv = (*getflist)(p11p);
+    if (err) {
+        errmsg = k5_get_error(&einfo, err);
+        TRACE_PKINIT_PKCS11_GETSYM_FAILED(context, errmsg);
+        goto error;
     }
-    if (err || rv != CKR_OK) {
-        krb5int_close_plugin(handle);
-        pkiDebug("failed\n");
-        return NULL;
+
+    getflist = (CK_RV (*)())sym;
+    rv = (*getflist)(p11p);
+    if (rv != CKR_OK) {
+        TRACE_PKINIT_PKCS11_GETFLIST_FAILED(context, pkcs11err(rv));
+        goto error;
     }
-    pkiDebug("ok\n");
+
     return handle;
+
+error:
+    k5_free_error(&einfo, errmsg);
+    k5_clear_error(&einfo);
+    if (handle != NULL)
+        krb5int_close_plugin(handle);
+    return NULL;
 }
 
 static krb5_error_code
@@ -3631,7 +3640,7 @@ pkinit_login(krb5_context context,
                                        (u_char *) rdat.data, rdat.length);
 
         if (r != CKR_OK) {
-            pkiDebug("C_Login: %s\n", pkinit_pkcs11_code_to_text(r));
+            TRACE_PKINIT_PKCS11_LOGIN_FAILED(context, pkcs11err(r));
             r = KRB5KDC_ERR_PREAUTH_FAILED;
         }
     }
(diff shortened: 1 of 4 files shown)


greghudson:
I'd like to remove the pkiDebug() calls where we add trace messages, unless there's a strong argument to keep both.

kenh:
> I'd like to remove the pkiDebug() calls where we add trace messages, unless there's a strong argument to keep both.

Fair enough; I'll push a new change to the branch that removes corresponding pkiDebug() calls.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -3557,24 +3557,32 @@ pkinit_C_LoadModule(const char *modname, CK_FUNCTION_LIST_PTR_PTR p11p)
     long err;
     CK_RV rv;
 
-    pkiDebug("loading module \"%s\"... ", modname);
+    TRACE_PKINIT_CLIENT_PKCS11_OPEN(context, modname);
     if (krb5int_open_plugin(modname, &handle, &einfo) != 0) {
-        pkiDebug("not found\n");
+        TRACE_PKINIT_CLIENT_PKCS11_OPEN_FAILED(context, einfo.msg ?
+                                               einfo.msg : "Unknown error");
+        k5_clear_error(&einfo);
```
frozencemetery, line 3564: This is I think correct, but should be called out in the commit message, or possibly in a separate commit.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -3557,24 +3557,32 @@ pkinit_C_LoadModule(const char *modname, CK_FUNCTION_LIST_PTR_PTR p11p)
...
         rv = (*getflist)(p11p);
     }
     if (err || rv != CKR_OK) {
+        if (err) {
+            TRACE_PKINIT_CLIENT_PKCS11_GETSYM_FAILED(context, einfo.msg ?
+                                                     einfo.msg :
+                                                     "Unknown error");
+        } else {
+            TRACE_PKINIT_CLIENT_PKCS11_FUNC_FAILED(context, "C_GetFunctionList",
+                                                pkinit_pkcs11_code_to_text(rv));
```
frozencemetery, line 3580: Indentation here is off

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -3633,6 +3641,8 @@ pkinit_login(krb5_context context,
 
         if (r != CKR_OK) {
             pkiDebug("C_Login: %s\n", pkinit_pkcs11_code_to_text(r));
+            TRACE_PKINIT_CLIENT_PKCS11_FUNC_FAILED(context, "C_Login",
+                                                pkinit_pkcs11_code_to_text(r));
```
frozencemetery, line 3645: Here too.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -3707,8 +3719,8 @@ pkinit_open_session(krb5_context context,
         }
         label_len = cp - tinfo.label;
 
-        pkiDebug("open_session: slotid %d token \"%.*s\"\n",
-                 (int)slotlist[i], (int)label_len, tinfo.label);
+        TRACE_PKINIT_CLIENT_PKCS11_SLOT(context, (int) slotlist[i],
+                                        (size_t) label_len, tinfo.label);
```
frozencemetery, line 3723: `label_len` is already `size_t`.

src/plugins/preauth/pkinit/pkinit_identity.c
```diff
@@ -383,8 +383,7 @@ process_option_identity(krb5_context context,
     int idtype;
     krb5_error_code retval = 0;
 
-    pkiDebug("%s: processing value '%s'\n",
-             __FUNCTION__, value ? value : "NULL");
+    TRACE_PKINIT_CLIENT_IDENTITY_OPTION(context, value ? value : "NULL");
```
frozencemetery, line 386: The trace layer handles NULL values already - they become "(null)".

src/plugins/preauth/pkinit/pkinit_trace.h
```diff
@@ -51,6 +60,27 @@
...
+    TRACE(c, "PKINIT client PKCS#11 module shows no slots with tokens")
+#define TRACE_PKINIT_CLIENT_PKCS11_OPEN(c, name)                        \
+    TRACE(c, "PKINIT client opening PKCS#11 module \"{str}\"", name)
+#define TRACE_PKINIT_CLIENT_PKCS11_OPEN_FAILED(c, errstr)               \
+    TRACE(c, "PKINIT client PKCS#11 module open failed: {str}", errstr)
+#define TRACE_PKINIT_CLIENT_PKCS11_SLOT(c, slot, len, label)            \
+    TRACE(c, "PKINIT client PKCS#11 slotid {int} token {lenstr}",       \
+          slot, len, label)
+#define TRACE_PKINIT_CLIENT_REGEXP_RULE(c, keyword, comp, value)        \
+    TRACE(c, "PKINIT client checking {str} rule '%{str} with value "    \
```
frozencemetery, line 80: The quoting around the second str doesn't look right, and I think it's grown an extra '%'.

frozencemetery (changes requested):
Thank you for doing this!  Some minor stuff inline.

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
kenh: I am unclear exactly what you mean here; do you want the commit message to explicitly mention that we log the error from loading the PKCS#11 module?  I can do that, but I am just trying to understand exactly what you want.

(The rest of the suggestions look fine and I will implement them)

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
kenh: Whoops, I think I meant that it should have been cast to `int` not `size_t`

src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
frozencemetery: Apologies, that wasn't very clear.

I think you fixed a memory leak here by adjusting where `k5_clear_error` is called.

kenh:
I believe the latest push addresses all of the comments

frozencemetery (approved):
Thanks!

greghudson:
Some concerns, most of which I don't have a clear action plan for at the moment:
 
* pkinit_pkcs11_code_to_text() is not thread-safe, though in a mostly trivial way.  Its name is also too long for a static hepler and it should return a const char *.  It's only used in debug calls prior to this PR.  Changing its name requires changing a bunch of debug calls not currently affected by this commit.

* The pkinit_matching.c functions can be used by either the client or KDC, but the trace macros are prefixed with TRACE_PKINIT_CLIENT.

* CLIENT_REGEXP_RULE and CLIENT_REGEXP_MATCH are pretty noisy in the presence of non-trivial matching rules, although they do convey valuable information about the cert field values.  Ideally we would generate one line of trace output per certificate summarizing the matching data, but that doesn't seem practical given the lists.  An improvement on the current proposal would be to generate one trace message, not two.  `code == REG_NOMATCH ? " NOT" : ""` is not the kind of argument I like to see in a trace macro.  We don't currently have a boolean substitution; we could add one, or we could have one trace call for matching and another one for not matching.

* Tracing counts (CERT_NUM_MATCHING, CERT_CHAIN_COUNT, SAN_CERT_COUNT) seems noisy.  CERT_CHAIN_COUNT is immediately followed by n-1 CNs, so the number is pretty clear.  (I don't know if the last cert CN would ever be useful, since don't currently understand why it's not pushed onto the stack.)

* Two of the new messages (CLIENT_IDENTITY_OPTION and CLIENT_CERT_RULE) say "processing" which is a content-free verb.  (It's also not a great verb to use in function names, although of course we do that all over the place.)  I think "loading" works pretty well for identities.  "matching rule {str} against certificates" might work for cert matching rules.

* SAN tracing seems like it could be improved.  CLIENT_SAN_CERT is the only message to convey the KDC cert CN, which is definitely valuable but seems like it maybe belongs elsewhere; aside from that it's a noisy "starting a function call" message.  On the client side caller we display the SANs the client cares about; on the KDC side we instead trace whether no princ/UPN SANs were found or if we find a match.  CLIENT_SAN_ERR doesn't convey any information about the error and seems like it belongs in the function rather than the caller.  (As far as I can tell, the only possible errors are ENOMEM and EINVAL, the latter resulting when the cert contains no SAN values, which possibly shouldn't be an error since it's roughly the same as the cert having only inapplicable SANs.)

kenh:
>  pkinit_pkcs11_code_to_text() is not thread-safe, though in a mostly trivial way.  Its name is also too long for a static hepler and it should return a const char *.  It's only used in debug calls prior to this PR.  Changing its name requires changing a bunch of debug calls not currently affected by this commit.

Regarding the thread-safe issue ... how would you prefer that be solved? Allocating the return string with asnprintf() seems straightforward, but that would complicate memory handling especially in the case where tracing is not enabled.  We could change it so the caller has to pass in an allocated string and length. I'm willing to go through and change the name to something shorter if that's what it takes.

> The pkinit_matching.c functions can be used by either the client or KDC, but the trace macros are prefixed with TRACE_PKINIT_CLIENT.

Oof, I missed that; I'll fix that.

> CLIENT_REGEXP_RULE and CLIENT_REGEXP_MATCH are pretty noisy in the presence of non-trivial matching rules, although they do convey valuable information about the cert field values.  Ideally we would generate one line of trace output per certificate summarizing the matching data, but that doesn't seem practical given the lists.  An improvement on the current proposal would be to generate one trace message, not two.  `code == REG_NOMATCH ? " NOT" : ""` is not the kind of argument I like to see in a trace macro.  We don't currently have a boolean substitution; we could add one, or we could have one trace call for matching and another one for not matching.

So, this was one of the things specifically called out by our customer support people where "the more information, the better". And this actually has been an issue in the past, as our users are typically issued 3 & 4 certificates and we need to make sure the rules only select one, and figuring out what is going on WITHOUT the noise is very difficult.  Regarding `code == REG_NOMATCH` [...]  I guess I was just going on the style in the corresponding pkiDebug() message. I can do two trace macros.

> Tracing counts (CERT_NUM_MATCHING, CERT_CHAIN_COUNT, SAN_CERT_COUNT) seems noisy.  CERT_CHAIN_COUNT is immediately followed by n-1 CNs, so the number is pretty clear.  (I don't know if the last cert CN would ever be useful, since don't currently understand why it's not pushed onto the stack.)

CERT_NUM_MATCHING directly relates to the certificate selection process, and again that has been a particular issue for us. Regarding SAN_CERT_COUNT, I can see where you're coming from, but typically among the pool of certificates available to the client they have different numbers of SANs so that's actually a good hint to help you figure out which certificate you are trying to match (all of the client certificates in our PKI have the same subject name so you typically can't use that).

CERT_CHAIN_COUNT ... fair enough, that is redundant. I'll remove that.

> Two of the new messages (CLIENT_IDENTITY_OPTION and CLIENT_CERT_RULE) say "processing" which is a content-free verb.  (It's also not a great verb to use in function names, although of course we do that all over the place.)  I think "loading" works pretty well for identities.  "matching rule {str} against certificates" might work for cert matching rules.

Fair enough, will clean up.

> SAN tracing seems like it could be improved.  CLIENT_SAN_CERT is the only message to convey the KDC cert CN, which is definitely valuable but seems like it maybe belongs elsewhere; aside from that it's a noisy "starting a function call" message.  On the client side caller we display the SANs the client cares about; on the KDC side we instead trace whether no princ/UPN SANs were found or if we find a match.  CLIENT_SAN_ERR doesn't convey any information about the error and seems like it belongs in the function rather than the caller.  (As far as I can tell, the only possible errors are ENOMEM and EINVAL, the latter resulting when the cert contains no SAN values, which possibly shouldn't be an error since it's roughly the same as the cert having only inapplicable SANs.)

In fairness, I was focusing on the client behavior.  And I understand the concerns about the noise, but this is a tough one; one things our customer support people mentioned was they had understood what was going on with the existing pkiDebug() messages in the above-mentioned problem areas, so I had to try to replicate the messages that had meaning during the processing, and SAN matching is very important to us. I am of course open to suggestions to improve SAN tracing, but I would like to keep CLIENT_SAN_CERT (without the client part, since it can be called by the KDC) if possible.

And I don't think I wrote the CLIENT_SAN_ERR macro?

greghudson:
> Regarding the thread-safe issue ... how would you prefer that be solved?

Lots of options, none of them perfect.

* Don't solve the problem right now; it's very unlikely to pose an issue at runtime.  It might pose a problem for a static analyzer at some point, or come up in someone's audit, but if so we can address it then.

* Return a fixed string on an unknown error.  Kind of terrible if it happens, but I'm guessing it's very unlikely.

* Use a caller-supplied buffer (strlcpy from the table or snprintf), return void, and put each call prior to the pkiDebug or trace call.

* Add a {pkcs11err} trace substitution.  PKCS11 errors are thematically out of scope for the krb5 library, but it's just a table lookup.  Of course with a mix of pkiDebug() and trace calls referencing PKCS11 errors, we'd have to have two copies of the table, which isn't great.

* As a variation of the above, create a PKINIT trace function which can specifically compose a string (no substitutions) and a PKCS11 error code, calling krb5int_trace() with "{str}" and the composed result.  The relevant PKINIT_TRACE macros would call that function instead of krb5int_trace().  The table and existing converter would move to pkinit_lib.c under this option.

Allocating the result string is probably too much code on the caller side.  Even a caller-supplied buffer is pretty wordy.

> Regarding SAN_CERT_COUNT, I can see where you're coming from, but typically among the pool of certificates available to the client they have different numbers of SANs so that's actually a good hint to help you figure out which certificate you are trying to match (all of the client certificates in our PKI have the same subject name so you typically can't use that).

Aha, so we're specifically concerned about this output when we call crypto_cert_get_matching_data() from pkinit_cert_matching().  This output would appear for each cert after the first CLIENT_CERT_RULE trace (because of the delayed call, which is kind of dumb because we don't really expect invalid rules) but before any per-cert regexp_match() trace output.  I guess the SAN counts can then be hand-correlated with the order of the regexp traces?  That seems kind of painstaking; we could facilitate it by passing a cert index down through component_match() to regexp_match() and including it in the trace output.

One simple improvement is to combine CLIENT_SAN_CERT and CLIENT_SAN_CERT_COUNT, so we see one line containing the subject and number of SANs.  (And then throw the cert name into CLIENT_SAN_CERT_NONE, if we're going to continue treating that as an EINVAL error.)

> And I don't think I wrote the CLIENT_SAN_ERR macro?

You're right, that one is my fault from 2012.

kenh:
I have pushed a new change that I believe addresses all of the points raised previously.

Regarding `pkinit_pkcs11_code_to_text()`, I renamed it to `pkinit_pkcs11_error()`, changed it to return `const char *`, and modified it to return a fixed string when it encountered an unknown error code. In my experience there is only a small number of possible error codes returned from PKCS#11 modules, so I think the odds of getting an unknown one is small.

greghudson:
I made a few changes:

* The TRACE_PKINIT_CLIENT_ prefix was previously only used in pkinit_clnt.c.  I renamed the new trace points to maintain that, and to avoid the really long TRACE_PKINIT_CLIENT_PKCS11 prefix.

* As I discussed previously,  I added a certificate index to the regexp trace points.  I made the displayed index one-based by adding one in the trace macros, on the assumption that it would be more intuitive.  Also as I discussed previously, I made regexp_match() only generate one trace message instead of two.

* I restructured the PKCS11 module loading function for more straightforward error handling.  Since we need to modify its signature for tracing anyway, I gave it a more krb5-appropriate identifier.

* Trace points should not be used more than once (from k5-trace.h: "A new trace macro must be defined in this file for each new location to be traced").  I split PKCS11_FUNC_FAILED() into two trace points since it was used twice.

* If we're renaming the PKCS11 error message helper we may as well make it really short; I changed it to pkcs11err().

I did a quick manual check using a pkinit_cert_match rule to make sure the tracing output looked reasonable to me, but you might want to do a recheck as well.

kenh:
I reviewed your changes, and I think the changes you made to the trace messages are clearer than my original ones, so they look great to me.

I did make one change and just pushed it now. I realized that it wasn't clear exactly WHAT SANS were being returned by `crypto_retrieve_X509_sans()`, so I modified SAN_CERT_COUNT to give the counts of the various types. In the process of doing that I discovered there was a long-standing bug where multiple UPNs in a certificate weren't being returned, only the last one was (the upn array index was never incremented), so I fixed that and noted it in the commit message. I admit I am not sure why you would want to put multiple UPNs in a certificate, but certainly it is possible.

greghudson:
> In the process of doing that I discovered there was a long-standing bug where multiple UPNs in a certificate weren't being returned, only the last one was (the upn array index was never incremented), so I fixed that and noted it in the commit message.

Introduced by me in commit 0f26c1c7504777d6e7bfa1d3dee575c504ab6c05, so this bug first appeared in a 1.16 patch release.  I might decide to pull this fix out into a separate commit, although you're right that probably no one will notice it.

As an aside, num_found is unused in that function, and has been since 1.6.

kenh:
> As an aside, num_found is unused in that function, and has been since 1.6.

Yeah, I noticed that after I pushed that commit. Did you want that removed in this change?

greghudson:
I pulled the bugfix out into a separate commit with a separate ticket, and threw the removal of num_found in there.

kenh:
FYI: Everything looks good with these changes from my viewpoint

PR #1175, opened 2021-03-27
Fix gss-krb5 handling of high sequence numbers

greghudson:
Commits abcfdaff756631d73f49103f679cafa7bc45f14e and
41ddaaeb286e8bb1bba64fb557ba0e4cff9b404d incorrectly changed the
interpretation of authenticator sequence numbers in the range
2^31..2^32-1, mapping them to sign-extended 64-bit values.  The major
Kerberos implementations do not generate sequence numbers this large,
so the changed went unnoticed.  Prevent unwanted sign extension by
casting sequence numbers retrieved from auth contexts to uint32_t
before assigning them to uint64_t fields.  Reported by Jake Scott.

### DIFF
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -982,7 +982,7 @@ kg_accept_krb5(minor_status, context_handle,
     {
         krb5_int32 seq_temp;
         krb5_auth_con_getremoteseqnumber(context, auth_context, &seq_temp);
-        ctx->seq_recv = seq_temp;
+        ctx->seq_recv = (uint32_t)seq_temp;
     }
 
     if ((code = krb5_timeofday(context, &now))) {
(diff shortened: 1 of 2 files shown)


greghudson:
There are six cases where we assign to the 64-bit GSS context sequence numbers during context establishment.

* In acceptor, set ctx->seq_recv to initiator's authenticator sequence number.  This was broken.
* In acceptor, set ctx->seq_send to locally generated AP-REP sequence number.  This was working, at least on platforms where long is 64 bits, but it used a strange construct (since 0xffffffffL explicitly asks for a long, but the value might only fit in an unsigned long).  I changed it to use a cast.  The correctness is moot since the locally generated sequence number won't exceed 2^30-1.
* In acceptor, set ctx->seq_send to ctx->seq_recv when not doing mutual authentication.  The source is already uint64_t; obviously the value will be wrong if seq_recv was set wrong.
* In initiator, set ctx->seq_send to the locally generated authenticator sequence number.  This was broken, but it's moot since the number won't exceed 2^30-1.
* In initiator, set ctx->seq_recv to ctx->seq_send when not doing mutual authentication.  The source is already uint64_t.
* In initiator, set ctx->seq_recv to the other party's AP-REP sequence number.  This is correct since the source (ap_rep_data->seq_number) is uint32_t.

So, only one of the three changes is strictly necessary; the other two are for hygiene.

frozencemetery (approved):

PR #1176, opened 2021-03-29
Fix KCM flag transmission for remove_cred

greghudson:
MIT krb5 uses low bits for KRB5_TC flags, while Heimdal uses high bits
so that the same flag word can also hold KRB5_GC flags.  Add a mapping
function and send the Heimdal flag values when performing a
remove_cred operation.


greghudson:
I was initially concerned about this fix creating a problem for sssd, but it looks like sssd's KCM server doesn't yet implement remove_cred support.

frozencemetery (approved):

PR #1177, opened 2021-03-30
kcm: fallback to GET_CRED_UUID_LIST also on KRB5_CC_IO

pbrezina:
sssd-kcm returns KRB5_CC_IO if the operation code is not known,
therefore krb5 needs to accept this error code to fallback properly.

### DIFF
--- a/src/lib/krb5/ccache/cc_kcm.c
+++ b/src/lib/krb5/ccache/cc_kcm.c
@@ -876,7 +876,7 @@ kcm_start_seq_get(krb5_context context, krb5_ccache cache,
         ret = kcmreq_get_cred_list(&req, &creds);
         if (ret)
             goto cleanup;
-    } else if (ret == KRB5_FCC_INTERNAL) {
+    } else if (ret == KRB5_FCC_INTERNAL || ret == KRB5_CC_IO) {
         /* Fall back to GET_CRED_UUID_LIST. */
         kcmreq_free(&req);
         kcmreq_init(&req, KCM_OP_GET_CRED_UUID_LIST, cache);


pbrezina:
@greghudson There was unconditional fallback when I submitted the original `GET_CRED_LIST` PR to always fallback in case of any error. If you prefer to only fallback on specific error code it is needed to add `KRB5_CC_IO` which is returned from sssd-kcm when operation is not implemented.

greghudson:
I thought I checked sssd, but I guess not.  Looking at the source code, I think what really happens is that sssd-kcm drops the connection and kcmio_unix_socket_read() returns KRB5_CC_IO.  This is exactly the kind of error I didn't want to fall back on, but I guess we have to.  I hope that as sssd-kcm evolves it will take more care to behave similarly to the Heimdal version.

pbrezina:
AFAIK sssd-kcm does not drop the connection but really returns KRB5_CC_IO error:

Op code not found returns ERR_KCM_MALFORMED_IN_PKT:
https://github.com/SSSD/sssd/blob/master/src/responder/kcm/kcmsrv_cmd.c#L201

The error propagates to kcm_reply_error:
https://github.com/SSSD/sssd/blob/master/src/responder/kcm/kcmsrv_cmd.c#L566

Which translates it into KRB5_CC_IO
https://github.com/SSSD/sssd/blob/master/src/responder/kcm/kcmsrv_cmd.c#L656

But unfortunately SSSD returns different error then Heimdal. We can fix this in new release, however this patch is required to correctly fallback with already SSSD versions.

PR #1178, opened 2021-03-31
Use KCM_OP_RETRIEVE in KCM client

greghudson:
In kcm_retrieve(), try KCM_OP_RETRIEVE.  Fall back to iteration if the
server doesn't implement it, or if we can an answer incompatible with
KRB5_TC_SUPPORTED_KTYPES.

In kcmserver.py, implement partial decoding for creds and cred tags so
that we can do a basic principal name match.

### DIFF
--- a/src/include/kcm.h
+++ b/src/include/kcm.h
@@ -87,7 +87,7 @@ typedef enum kcm_opcode {
     KCM_OP_INITIALIZE,          /*          (name, princ) -> ()          */
     KCM_OP_DESTROY,             /*                 (name) -> ()          */
     KCM_OP_STORE,               /*           (name, cred) -> ()          */
-    KCM_OP_RETRIEVE,
+    KCM_OP_RETRIEVE,            /* (name, flags, credtag) -> (cred)      */
     KCM_OP_GET_PRINCIPAL,       /*                 (name) -> (princ)     */
     KCM_OP_GET_CRED_UUID_LIST,  /*                 (name) -> (uuid, ...) */
     KCM_OP_GET_CRED_BY_UUID,    /*           (name, uuid) -> (cred)      */
--- a/src/lib/krb5/ccache/cc_kcm.c
+++ b/src/lib/krb5/ccache/cc_kcm.c
@@ -826,9 +826,55 @@ static krb5_error_code KRB5_CALLCONV
 kcm_retrieve(krb5_context context, krb5_ccache cache, krb5_flags flags,
              krb5_creds *mcred, krb5_creds *cred_out)
 {
-    /* There is a KCM opcode for retrieving creds, but Heimdal's client doesn't
-     * use it.  It causes the KCM daemon to actually make a TGS request. */
-    return k5_cc_retrieve_cred_default(context, cache, flags, mcred, cred_out);
+    krb5_error_code ret;
+    struct kcmreq req = EMPTY_KCMREQ;
+    krb5_creds cred;
+    krb5_enctype *enctypes = NULL;
+
+    memset(&cred, 0, sizeof(cred));
+
+    /* Include KCM_GC_CACHED in flags to prevent Heimdal's sssd from making a
+     * TGS request itself. */
+    kcmreq_init(&req, KCM_OP_RETRIEVE, cache);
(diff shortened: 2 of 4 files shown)


greghudson:
I'll add some background here about KRB5_TC_SUPPORTED_ENCTYPES.  This flag ties into gss_krb5_set_allowable_enctypes(), which is an extension design to support NFS clients.  In a typical Kerberos NFS architecture, userland code (using MIT krb5 or Heimdal) is responsible for establishing the GSS context, but the kernel is responsible for the per-message operations.  The GSS context must therefore use an enctype supported by the kernel, which might not support all of our newfangled enctypes.  So the userland NFS daemon communicates the kernel-supported enctypes to the GSS layer, which stores them in the krb5 context before calling krb5_get_credentials().

In the get_creds.c state machine, we go through some contortions to obtain TGTs using the regular enctypes but the final service ticket using one of the kernel-supported enctypes.

In the ccache layer, we also want a way to match a cached ticket with a particular set of enctypes, so that we don't used a cached ticket with a newfangled enctype.  As far as I can tell Heimdal ignores this problem, and gets away with that because it would be unusual to have an nfs/servername ticket of a kernel-unsupported enctype in the cache.  MIT has KRB5_TC_SUPPORTED_ENCTYPES, which is an unpleasant hack.  Because the matching krb5_creds structure can contain only one enctype, we instead match against the context enctypes as reported by krb5_get_tgs_ktypes().

Because Heimdal ignores this problem at the ccache layer, the KCM protocol has no way to communicate the context enctypes to the KCM daemon.  I chose to address this by checking the response and falling back to iteration if it has an unacceptable enctype.  Even in an NFS scenario this should be very rare, for the same reason that Heimdal can get away with not addressing the problem.

frozencemetery (approved):

joakim-tjernlund:
This fails in sssd 2.5/master because sssd returns KRB5_CC_NOSUPP , not KRB5_CC_IO
Did error code change in sssd 2.5 ?
@pbrezina ?

joakim-tjernlund:
Fast pacth:
```
--- ./src/lib/krb5/ccache/cc_kcm.c.org	2021-05-10 19:59:19.668773908 +0200
+++ ./src/lib/krb5/ccache/cc_kcm.c	2021-05-11 13:17:19.228562311 +0200
@@ -841,7 +841,7 @@
     ret = cache_call(context, cache, &req);
 
     /* Fall back to iteration if the server does not support retrieval. */
-    if (ret == KRB5_FCC_INTERNAL || ret == KRB5_CC_IO) {
+    if (ret == KRB5_FCC_INTERNAL || ret == KRB5_CC_IO || ret == KRB5_CC_NOSUPP) {
         ret = k5_cc_retrieve_cred_default(context, cache, flags, mcred,
                                           cred_out);
         goto cleanup;

pbrezina:
It was always like this.

greghudson:
See PR #1186.

(sssd implementing RETRIEVE would be nice, particularly if it can do a fast lookup on the server principal.  But it wouldn't have a big payoff until we address the scan_ccache() part of https://krbdev.mit.edu/rt/Ticket/Display.html?id=9004 .)

pbrezina:
We currently can't lookup by principal since the credentials are stored only in its binary form as we get them from krb5. I opened: https://github.com/SSSD/sssd/issues/5627

pbrezina:
Also is there any particular error code we should be returning in these two cases https://github.com/krb5/krb5/pull/1178#issuecomment-838418470 to avoid errors like this in the future?

greghudson:
sssd's KCM server should strive to behave similarly to Heimdal's KCM server to avoid interop issues.  Its sources are at https://github.com/heimdal/heimdal/tree/master/kcm with protocol.c containing most of the main logic.  Heimdal's KCM server returns KRB5_FCC_INTERNAL if it can't find a handler for the opcode, so I would recommend that sssd do the same.

Also be aware that Apple also has its own fork of Heimdal and makes the corresponding source code available at opensource.apple.com.  Its KCM server has a number of additions relative to upstream Heimdal (although they also seem to have migrated away from KCM to a new daemon-based ccache type as their default).

pbrezina:
https://github.com/SSSD/sssd/pull/5629

PR #1179, opened 2021-03-31
Simplify krb5_cccol_have_content()

greghudson:
For the purposes of determining whether Kerberos credentials are
present, just check for an initialized ccache (as detected by
krb5_cc_get_principal()), not one with credentials in it.  For KCM
caches in particular, this changes avoids the O(n) expense of starting
an iteration.

Also fix a potential memory leak if a cache is found after an error is
saved.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -2667,14 +2667,14 @@ krb5_error_code KRB5_CALLCONV
 krb5_cccol_cursor_free(krb5_context context, krb5_cccol_cursor *cursor);
 
 /**
- * Check if the credential cache collection contains any credentials.
+ * Check if the credential cache collection contains any initialized caches.
  *
  * @param [in]  context         Library context
  *
  * @version New in 1.11
  *
- * @retval 0 Credentials are available in the collection
- * @retval KRB5_CC_NOTFOUND The collection contains no credentials
+ * @retval 0 At least one initialized cache is present in the collection
+ * @retval KRB5_CC_NOTFOUND The collection contains no caches
  */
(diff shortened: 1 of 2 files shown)


greghudson:
This change removes one of the lesser obstacles to O(1) GSS performance for large ccaches with KCM and KEYRING.  For FILE/DIR, the cost should be O(1) because the iteration stops at the first real cred, but it's still more work than necessary.  (And of course retrieval for FILE is O(n) without big extension along the lines of Nico's hashtable-in-a-config-cred proposal.)

frozencemetery:
> just check for an initialized ccache (as detected by krb5_cc_get_principal()), not one with credentials in it.

I'm not sure how important the distinction is, but right now `krb5_cccol_have_content()` is documented as:

```C
/**
 * Check if the credential cache collection contains any credentials.
 *
 * @param [in]  context         Library context
 *
 * @version New in 1.11
 *
 * @retval 0 Credentials are available in the collection
 * @retval KRB5_CC_NOTFOUND The collection contains no credentials
 */
```

So this is a behavior change, right?

greghudson:
It is a behavior change, but I think not an important one.  You'd only find an initialized cache without a credential in a race condition (which I hope we can some day solve, see https://krbdev.mit.edu:444/rt/Ticket/Display.html?id=7707 ), and in those cases it is just as disruptive for krb5 gss_acquire_cred() to return failure as it is for gss_init_sec_context() to perhaps fail afterwards.

We can change the documentation to more precisely describe the behavior; I initially figured that wasn't necessary.

frozencemetery (approved):

PR #1182, opened 2021-04-16
Get CI working again

greghudson:
For a few weeks I started seeing occasional test failures where the first test involving a KDC would fail with a "Cannot contact any KDC for realm" error.  Now it happens all the time.  I assume Github Actions made a change to their runners and started with only some of them.

For the moment this PR switches the Unix build and tests back to Travis, but I will see if I can figure out the root cause and a workaround with Github Actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,9 @@ jobs:
                 sudo apt-get update -qq
                 sudo apt-get install -y bison dejagnu gettext keyutils ldap-utils libcmocka-dev libldap2-dev libkeyutils-dev libresolv-wrapper libsasl2-dev libssl-dev python3-kdcproxy python3-pip slapd tcl-dev tcsh
                 pip3 install pyrad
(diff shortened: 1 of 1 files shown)


frozencemetery (approved):
Mmm, that's frustrating.

greghudson:
I think the proximal cause is that the runners don't have a hostname that resolves to a local address.

If I switch k5test.py to use `127.0.0.1` instead of `$hostname` in the krb5.conf realm configuration, `make check` gets as far as lib/rpc/unit-test where it fails at "./client -t fv-az151-598 53717 host@fv-az151-598 1026" (i.e. another attempt to connect to a port on the local hostname).

frozencemetery (approved):

greghudson:
Github is supposedly rolling out a fix today, so we can probably just wait this out.

greghudson:
Closing this PR as the runners seem to have updated.

PR #1184, opened 2021-04-26
Fix argument type errors on Windows

greghudson:
Fix three Windows-specific argument type errors, including a crash bug
in the default replay cache type.  Change the compiler flags to treat
several argument type warnings as errors.

The replay cache bug was reported by Thomas Wagner.

[Side note, not in the commit message: I had considered flagging "incompatible types" (C4133) as an error, but we have eleven harmless instances of that warning in the build, all stemming from Windows/POSIX disagreements about pointer types in socket-layer functions like setsockopt().]

### DIFF
--- a/src/util/support/dir_filenames.c
+++ b/src/util/support/dir_filenames.c
@@ -87,7 +87,7 @@ k5_dir_filenames(const char *dirname, char ***fnames_out)
         return ENOENT;
 
     do {
-        if (add_filename(&fnames, &n_fnames, &ffd.cFileName) != 0) {
+        if (add_filename(&fnames, &n_fnames, ffd.cFileName) != 0) {
             k5_free_filenames(fnames);
             FindClose(handle);
             return ENOMEM;
--- a/src/clients/kpasswd/kpasswd.c
+++ b/src/clients/kpasswd/kpasswd.c
@@ -37,7 +37,8 @@ get_name_from_passwd_file(char *program_name, krb5_context context,
 }
 #else /* HAVE_PWD_H */
 static void
-get_name_from_passwd_file(krb5_context context, krb5_principal *me)
+get_name_from_passwd_file(char *program_name, krb5_context context,
+                          krb5_principal *me)
 {
     fprintf(stderr, _("Unable to identify user\n"));
     exit(1);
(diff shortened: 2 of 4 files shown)


src/util/support/dir_filenames.c
```diff
@@ -87,7 +87,7 @@ k5_dir_filenames(const char *dirname, char ***fnames_out)
         return ENOENT;
 
     do {
-        if (add_filename(&fnames, &n_fnames, &ffd.cFileName) != 0) {
+        if (add_filename(&fnames, &n_fnames, ffd.cFileName) != 0) {
```
kaduk, line 90: It looks like `WIN32_FIND_DATA.cFileName` is a `CHAR cFileName[PATH_MAX]`, and IIRC Unix compilers have `&` do the same thing as the implicit conversion from array to pointer type.  But the C spec doesn't really seem to support this, saying that the argument to the `&` operator must be "... or an lvalue that designates an object that is not a bit-field".  The array itself is not, IIUC, an lvalue...

kaduk (approved):
I only did a very minimal verification that the MSVC warning numbers seem to match the listed check.

src/util/support/dir_filenames.c
greghudson: The two expressions have the same numeric value, so this bug is harmless, but the former expression has type `char (*)[PATH_MAX]`.  If the same code were compiled in the main build, we would see a warning like this one:

    warning: passing argument 1 of ‘fn’ from incompatible pointer type [-Wincompatible-pointer-types]
    note: expected ‘const char *’ but argument is of type ‘char (*)[100]’

kaduk: I don't see how to get that type from what n1256.pdf says, but I also think we should just merge the PR and not spend more time on this question :)

greghudson: Note that 6.3.2.1 says "A modifiable lvalue is an lvalue that does not have array type [...]".  So ffd.cFileName is an lvalue, just not a modifiable lvalue.

kaduk: Ah, thanks for the pointer (pun intended).

PR #1186, opened 2021-05-11
Fix KCM retrieval support for sssd

greghudson:
Commit 795ebba8c039be172ab93cd41105c73ffdba0fdb added a retrieval
handler using KCM_OP_RETRIEVE, recognizing the same error code as the
previous KCM_OP_GET_CRED_LIST support.  But sssd (as of 2.4) returns
KRB5_CC_NOSUPP instead of KRB5_CC_IO if it recognizes an opcode as
unsupported.  Add a helper function to recognize all possible
unsupported-opcode errors and use it in kcm_retrieve() and
kcm_start_seq_get().

### DIFF
--- a/src/lib/krb5/ccache/cc_kcm.c
+++ b/src/lib/krb5/ccache/cc_kcm.c
@@ -144,6 +144,20 @@ map_tcflags(krb5_flags mitflags)
     return heimflags;
 }
 
+/*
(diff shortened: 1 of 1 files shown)


frozencemetery (approved):
Well, this is unfortunate.

PR #1187, opened 2021-05-19
Modernize softpkcs11's openssl

frozencemetery:
Technically only the first commit is needed to unbreak the CI, but I took care of the deprecation warning while I was there since it's a similar issue.

One thing I'm not sure about is how far back we want to support: softpkcs11 appears to have glue code for `EVP_PKEY_get0_RSA` that I've left for now.

~~I'm not sure whether we'll need to load the legacy provider for running in openssl's CI.  I have support for that in the other branch if it ends up being needed.~~ (It wasn't needed for this, but it's in the other branch because it's needed for other things.)

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1148,6 +1148,7 @@ int i = 1;
 ])], k5_cv_openssl_version_okay=yes, k5_cv_openssl_version_okay=no)])
   old_LIBS="$LIBS"
   AC_CHECK_LIB(crypto, PKCS7_get_signer_info)
+  AC_CHECK_FUNCS(EVP_PKEY_get_bn_param)
   LIBS="$old_LIBS"
 fi
 if test "$k5_cv_openssl_version_okay" = yes && (test "$enable_pkinit" = yes || test "$enable_pkinit" = try); then
--- a/src/tests/softpkcs11/main.c
+++ b/src/tests/softpkcs11/main.c
@@ -413,47 +413,83 @@ add_object_attribute(struct st_object *o,
     return CKR_OK;
 }
 
+#ifdef HAVE_EVP_PKEY_GET_BN_PARAM
+
+/* Declare owner pointers since EVP_PKEY_get_bn_param() gives us copies. */
+#define DECLARE_BIGNUM(name) BIGNUM *name = NULL
+#define RELEASE_BIGNUM(bn) BN_clear_free(bn)
 static CK_RV
-add_pubkey_info(struct st_object *o, CK_KEY_TYPE key_type, EVP_PKEY *key)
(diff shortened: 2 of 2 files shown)


greghudson:
What was `RSA_set_method(rsa, RSA_PKCS1_OpenSSL());` doing?  It isn't addressed in the commit message.

frozencemetery:
Updated commit message.  Functionally, it did nothing - `RSA_PKCS1_OpenSSL()` was the default already.

frozencemetery:
(The linked openssl PR for these commits is passing CI.)

greghudson:
OPENSSL_clear_free() and EVP_PKEY_CTX_free() can be called unconditionally in the cleanup handlers.

I haven't done all of the footwork to verify this, but the checks like `((CK_ULONG)buffer_len + padding_len < ulDataLen)`  seem like they are probably no longer necessary.  In the old code, buffer_len depended only on the RSA key, while in the new code it's the size computed by an EVP function over the key and data.

In the second commit, it would be good to factor out the variant code into a helper function somehow so that there aren't preprocessor directives inside the function.

> One thing I'm not sure about is how far back we want to support

Given how the build system is currently structured, as far back as PKINIT.  It would be okay to skip the softpkcs11 build and corresponding tests in order to require a newer version of OpenSSL, but we don't softpkcs11 to break the build after PKINIT succeeds.  The build system currently doesn't distinguish.

frozencemetery:
>OPENSSL_clear_free() and EVP_PKEY_CTX_free() can be called unconditionally in the cleanup handlers.

Done.

> I haven't done all of the footwork to verify this, but the checks like ((CK_ULONG)buffer_len + padding_len < ulDataLen) seem like they are probably no longer necessary. In the old code, buffer_len depended only on the RSA key, while in the new code it's the size computed by an EVP function over the key and data.

Done.  (My reading of the original intent was to check the function arguments (`ulDataLen` et al.), and we should be fine having OpenSSL do that for us.)

> Given how the build system is currently structured, as far back as PKINIT. It would be okay to skip the softpkcs11 build and corresponding tests in order to require a newer version of OpenSSL, but we don't softpkcs11 to break the build after PKINIT succeeds. The build system currently doesn't distinguish.

Makes sense.  I think it's probably easiest to leave the glue as-is for now, then.

> In the second commit, it would be good to factor out the variant code into a helper function somehow so that there aren't preprocessor directives inside the function.

"Somehow" indeed :)  It's complicated because the type of the bignums isn't the same, but hopefully this is better.

frozencemetery (commented):
Changes look good, thanks.  I updated the openssl PR to match, and CI is passing there.

PR #1188, opened 2021-05-20
Rewrite not-yet-covered dejagnu tests in Python

greghudson:
Remove the dejagnu scripts gssapi.exp, princexpire.exp, sample.exp,
simple.exp, and tcp.exp.

Add server output checking to t_gss_sample.py to match the checks in
gssapi.exp.

Add a test to t_general.py matching the #6428 regression test in
princexpire.exp.

Add new test scripts t_sample.py and t_simply.py for the appl/sample
and appl/simple applications, to match sample.exp and simple.exp.
Both of these test servers exit after one client execution; add a
k5test function await_daemon_exit() to allow the daemon exit status to
be checked without sending a kill signal.

Add a TCP test to t_bigreply.py to match the oversized-TCP-request
test in tcp.exp.  The existing t_bigreply.py test already covers a
successful TCP requests.


greghudson:
Everything left after this commit is already covered.  kprop.exp has a corresponding Python script, and standalone.exp is covered by t_general.py, t_dump.py, t_keytab.py, and t_pkinit.py.  Actually removing the dejagnu machinery will first require moving a few things out of it (like pkinit-certs), so I'll do that in a separate PR.

src/tests/dejagnu/t_inetd.c
```diff
@@ -82,7 +82,8 @@ main(argc, argv)
 
     progname = argv[0];
 
-    if(argc <= 3) usage();
+    if (argc < 3)
```
frozencemetery, line 85: Not sure I understand this change.  Line 139 references `argv[3]` unconditionally, so I think the original is right?

frozencemetery (changes requested):
It looks like your change deletes a few files that are already gone - I think it needs to be rebased on top of current HEAD (8027531caf6911bb07bf13de087da0e6bef5a348).

src/tests/dejagnu/t_inetd.c
greghudson: It turns out this isn't right.  t_inetd.c wants the program name specified twice--once for its path and again for argv[0].  I misunderstood the code; what we actually need is an adjustment to start_in_inetd() in k5test.py.

greghudson:
> It looks like your change deletes a few files that are already gone

I managed to commit with --amend with no prior checkpoint commit, so I combined my changes with the previous commit.  I'll untangle it.

frozencemetery (approved):

PR #1189, opened 2021-05-21
Remove dejagnu

greghudson:
The first commit moves test facilities around and the second commit does the actual removal.

There is still a little bit of Tcl in the tree after this, in util/profile/prof_test1 (relying on 3600 lines of SWIG-generated Tcl bindings).


frozencemetery (changes requested):
This looks good, but there are some vestiges remaining:

- .gitignore mentions nonexistent diretory /src/tests/dejagnu
- GitHub Actions still installs dejagnu (this is for prof_test1 probably?)
- doing_build.rst mentions it as a prerequisite for some of the tests in the "Testing the build" section (same as above?)
- options2configure.rst mentions it in the behavior of `--with-readline`
- the very end of t_inetd.c has a workaround for signal behavior

greghudson:
I've cleaned up most of those.  prof_test1 only needs Tcl development libraries, not dejagnu.

The t_inetd SIGCHLD code doesn't really match the `Don't wait for a child signal` comment.  (Both code and comment date back to the original 1995 commit of the file, which hasn't changed much since.)  t_inetd.c doesn't fork, so it isn't in a position to wait for a child signal.  SIGCHLD being ignored does propagate to the child process, but the only program t_inetd is used with in 1.0 (appl/sample/sserver) also doesn't fork, nor do programs it was used with later.  The test suite passes without this block of code, so I will remove it.

doc/build/doing_build.rst
```diff
@@ -99,24 +99,14 @@ start building Kerberos; see :ref:`do_build`)::
...
-  search path, though. Alternatively, you can configure with
-  **-**\ **-disable-rpath**, which renders the build tree less suitable for
-  installation, but allows testing without interference from
-  previously installed libraries.
+On some operating systems, you have to run ``make install`` before
+running ``make check``, or the test suite will pick up installed
+versions of Kerberos libraries rather than the newly built ones.  You
+can install into a prefix that isn't in the system library search
+path, though.  Alternatively, you can configure with
+**-**\ **-disable-rpath**, which renders the build tree less suitable
```
frozencemetery, line 107: Is this backward?  (To my mind, it doesn't make sense for disabling rpath to make the tree less suitable for installation - files that live on the system should not be using DT_RPATH, right?)

greghudson: If you build with the default prefix (`/usr/local`) or another prefix that isn't in the system default library path, you need rpaths for the binaries to find the libraries.  A distribution building a package for /usr probably does want to use `--disable-rpath`, though.  We could address that somewhere in doing_build.rst, but it's not in scope for this PR (the text here is just being moved out of a bullet list without change) or this section of the file.

frozencemetery (approved):

PR #1191, opened 2021-05-27
Fix doc build for Sphinx 4.0

greghudson:
Use app.add_css_file() to register krb5.css if possible (it was added
in Sphinx 1.8), since the old name app.add_stylesheet() was removed in
Sphinx 4.0.

Use the highlight directive instead of the highlightlang directive,
which was removed in Sphinx 4.0.

Remove two duplicate table of contents entries to fix warnings.


frozencemetery (commented):
Change looks/tests good.

Do we want to be running the newer sphinx in github actions CI?  We already install Cheetah3 from pip during doc build, so could grab sphinx there too.

greghudson:
Okay, that's done.  I was going to have the newer-Sphinx doc build run on Ubuntu 20.04, but that runs into the buggy version of Doxygen that doesn't generate a node for krb5_const_principal (as noted in PR #1119).

frozencemetery (approved):

PR #1193, opened 2021-05-30
Another round of static analyzers

frozencemetery:
Most of these are unimportant leaks, but the tool flags them as "Important", so here we are.  That said, this PR does not touch ksu, sample/test code, or rpc.

The commits are separated for my own convenience when working with them - I'm happy to modify that as needed.

### DIFF
--- a/src/kadmin/dbutil/kadm5_create.c
+++ b/src/kadmin/dbutil/kadm5_create.c
@@ -185,21 +185,24 @@ static int add_admin_princs(void *handle, krb5_context context, char *realm)
 int add_admin_princ(void *handle, krb5_context context,
                     char *name, char *realm, int attrs, int lifetime)
 {
-    char *fullname;
+    char *fullname = NULL;
     krb5_error_code ret;
     kadm5_principal_ent_rec ent;
     long flags;
+    int fret;
 
     memset(&ent, 0, sizeof(ent));
 
     if (asprintf(&fullname, "%s@%s", name, realm) < 0) {
         com_err(progname, ENOMEM, _("while appending realm to principal"));
-        return ERR;
+        fret = ERR;
+        goto cleanup;
     }
     ret = krb5_parse_name(context, fullname, &ent.principal);
     if (ret) {
         com_err(progname, ret, _("while parsing admin principal name"));
-        return(ERR);
+        fret = ERR;
+        goto cleanup;
     }
     ent.max_life = lifetime;
     ent.attributes = attrs;
@@ -210,13 +213,13 @@ int add_admin_princ(void *handle, krb5_context context,
     ret = kadm5_create_principal(handle, &ent, flags, NULL);
     if (ret && ret != KADM5_DUP) {
         com_err(progname, ret, _("while creating principal %s"), fullname);
-        krb5_free_principal(context, ent.principal);
-        free(fullname);
-        return ERR;
+        fret = ERR;
+        goto cleanup;
     }
 
+    fret = OK;
+cleanup:
     krb5_free_principal(context, ent.principal);
     free(fullname);
-
-    return OK;
+    return fret;
 }
--- a/src/kadmin/dbutil/kdb5_create.c
+++ b/src/kadmin/dbutil/kdb5_create.c
@@ -393,7 +393,7 @@ add_principal(context, princ, op, pblock)
     struct realm_info *pblock;
 {
     krb5_error_code       retval;
-    krb5_db_entry         *entry;
+    krb5_db_entry         *entry = NULL;
     krb5_kvno             mkey_kvno;
     krb5_timestamp        now;
     struct iterate_args   iargs;
@@ -410,20 +410,20 @@ add_principal(context, princ, op, pblock)
     entry->expiration = pblock->expiration;
 
     if ((retval = krb5_copy_principal(context, princ, &entry->princ)))
-        goto error_out;
+        goto cleanup;
 
     if ((retval = krb5_timeofday(context, &now)))
-        goto error_out;
+        goto cleanup;
 
     if ((retval = krb5_dbe_update_mod_princ_data(context, entry,
                                                  now, &db_create_princ)))
-        goto error_out;
+        goto cleanup;
 
     switch (op) {
     case MASTER_KEY:
         if ((entry->key_data=(krb5_key_data*)malloc(sizeof(krb5_key_data)))
             == NULL)
-            goto error_out;
+            goto cleanup;
         memset(entry->key_data, 0, sizeof(krb5_key_data));
         entry->n_key_data = 1;
 
@@ -435,7 +435,7 @@ add_principal(context, princ, op, pblock)
         if ((retval = krb5_dbe_encrypt_key_data(context, pblock->key,
                                                 &master_keyblock, NULL,
                                                 mkey_kvno, entry->key_data)))
-            return retval;
+            goto cleanup;
         /*
          * There should always be at least one "active" mkey so creating the
          * KRB5_TL_ACTKVNO entry now so the initial mkey is active.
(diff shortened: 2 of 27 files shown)


greghudson:
Review of the first ten commits:

* krad/remote.c: looks okay; removal only happens on ENOMEM error so the bug is generally harmless
 
* kadm5_create.c: looks okay

* kdb5_create.c: the patch renames the error_out label to "done" in addition to replacing a bare return.  If we're improving the name in addition to fixing the bug, the preferred name is "cleanup".
 
* ktutil_funcs.c: also change the call to krb5_kt_free_entry() to the un-deprecated _contents name for clarity.

* do_tgs_req.c: the cleanup handler unconditionally uses both state and au_state, so this patch won't work.

* kpropd.c: the error label has a specific meaning, so the patch should goto done instead.

* kproplog.c: the patch probably doesn't squash defect because fd is still leaked on success.  Closing it on success obviously wouldn't work since it would also invalidate the memory map.  We will likely need to pass fd back to the caller to make static analyzers happy.

* kg_get_ccache_name(): there actually is a leak here if out_name is NULL.  There's only one call site and it doesn't pass NULL, so this function should just stop conditionalizing.  The cleanup is welcome either way.
 
*  This static analyzer report is a false positive; the memory is only allocated when err is 0.  But the cleanup is welcome regardless.
 
* spnego_mech.c: looks okay

frozencemetery:
Thanks, rebased and updated the first ten accordingly.

greghudson:
Review of the middle ten commits:

* pwqual_dict.c: okay, will need some minor adjustments

* krb5_db_setup_mkey_name(): same

* kdb_convert.c: use k5memdup0() instead of strndup() (which is POSIX.1-2008 and this would be the first use).  Call the cleanup label "cleanup".  The cleanup appears to admit a double-free of ent->princ.  It should be okay to use krb5_db_free_principal() in both cases, and not free ent->princ separately.  (After all, on successful return the caller is going to use that function for both cases.)

* kr_remote_send(): the line numbers in the defect aren't the same, so I think there's a Fedora patch involved here, although it may not affect this function.  In the error handler of the upstream code, r can refer to either an element from rr->list or to an element we allocated and haven't yet added to rr->list.  In the first case I don't think we want to release it, and in the second case request_finish() will begin by trying to remote it from a list it isn't in.

* gic_keytab.c: okay

* s4u_authdata.c: okay

* send_tgs.c: where we free req.padata in the cleanup code, free padata instead, and get rid of the `padata = NULL;`.

* dnssrv.c: okay.  Not sure what I think about the function always returning 0 (and the analagous SRV function doing the same), but changing that is not in scope.

* sendto_kdc.c: okay

* kdc_j_encode.c: use a tiny if-ladder here: `if (!ret) ret = k5_json_array_add(arra, tmpa);`

frozencemetery:
> kr_remote_send(): the line numbers in the defect aren't the same, so I think there's a Fedora patch involved here, although it may not affect this function.

There is (the patch is [this](https://github.com/frozencemetery/krb5/commit/836fdca1ac4bb58498551e1afe8ca6e55d41902d#diff-211275c93b7a19a68df0226273ba411f9dedf5b19c262e22f5f11a6f03ef2be9)), and it does, but the allocation flagged is the `request_new()` call which is below the region the patch touches.  Unfortunately the analyzer suite gets run against RHEL builds, so the line numbers often don't match.

> in the second case request_finish() will begin by trying to remote it from a list it isn't in

I think we do still leak in this case, so updated to call something more appropriate and split the cases into separate variables.

Updated the rest as well.

greghudson:
Review of the last ten commits:

* adb_policy.c: use k5alloc() instead of k5calloc(), and get rid of the memset().

* kdb_db2.c: okay

* ldap_principal.c: okay

* ldap_principal2.c: okay

* pkinit_open_session(): okay-ish.  We're resorting to freeing a resource along four different code paths.  The function should use a cleanup handler and should stop using the same variable for a PKCS11 error and a krb5_error_code, but this is okay if that seems like too much.

* pkinit_get_certs_pkcs11(): same but more so, since we're freeing two resources along four different code paths.  I think the first step for improvement here would be to move some or all of the large for-loop body into another function.

* pkinit_identity.c: I would call the new variable "copy" instead of "alloc".

* prof_get.c: I would go with a single end_list() call in the cleanup handler, passing either NULL or the output parameter depending on retval.  There's also no need to set retval = 0 in this patch.

* client_init.c and server_init.c: a cleaner approach here would be to move the three check-handle macro bodies to static inline functions in the same headers, and then define CHECK_HANDLE() to call the generic check and the client or server check and return either result if non-zero.  init_any() could directly call the generic check function.

greghudson:
Here is how I would like to see these chunked for merging:

* Remove the libkadm5 init fixes from this PR.  (Apologies if you got started on my feedback for them.)  asan is finding a bunch of related leaks in t_kadm5.c, and I want to handle them together in a separate PR.

* The krad use-after-free loop bug, which I will tag for backport.

* The two gssapi_krb5 set_ccache cleanups.  (No actual reachable leak here, so this is more of a code quality improvement.)

* Everything else, unless I missed something that veers substantially away from a leak fix into code cleanup territory.

frozencemetery:
> Remove the libkadm5 init fixes from this PR. (Apologies if you got started on my feedback for them.)

No worries.  I got sidetracked before digging into them.  Pushed the rest of requested changes now, and will chunk and push again in a bit.

frozencemetery:
Squished and updated.  I've left the pkinit_get_certs_pkcs11 commit out of the "leaks" commit because the refactor diff is now large - can move that in too if desired.

greghudson:
Incomplete notes on the gssapi set_ccache cleanups:

* Experimentally these changes seem to introduce a memory leak in libkadm5 client initialization (which makes two calls to gss_krb5int_ccache_name()), I believe because nothing frees old_name.

* In gss_krb5int_ccache_name() when req->out_name is NULL, the old code seems to do a little bit of unnecessary work, fetching K5_KEY_GSS_KRB5_SET_CCACHE_OLD_NAME and re-setting it to an unchanged value.  The new code goes to more unnecessary work, fetching the old value and storing it in the per-thread variable whether or not the caller wants it.  It seems like we should do less work, not more.

* kg_get_ccache_name() never succeeds with a null out_name, so we shouldn't need to catch that and strdup the old value.

* In kg_get_ccache_name(), calling krb5_cc_set_default_name(context, NULL) dates back to when we used to reuse the same krb5_context in the krb5 mech.  We don't need it any more.

greghudson:
Notes on a round of changes:

* In gss_krb5int_ccache_name(), we were still doing more work than needed if the caller didn't want the old name.  Also, if kg_set_ccache_name() fails, the old code did actually handle it (in a roundabout way) and the PR code didn't.

* In spnego_mech.c, generic_gss_release_oid() can't actually fail and other parts of the code rely on that.  So since we're moving that code block I removed the failure check.

* In ulog_conv_2dbentry() I removed one extraneous change and one no-longer-relevant change (we're not longer using is_add in the cleanup handler).

* In PKINIT, I realize that kret was named so that both kret and pret were explicit.  But I prefer using the modern standard name for ta krb5 error code.  Also, pkinit_login() returns a krb5 error code, not a PKCS11 result code.

* In parse_fs_options(), I initialized copy because the rule is that anything freed in the cleanup handler gets initialized, even if it's the first resource to be allocated.

frozencemetery (commented):
Changes look fine, thanks.

greghudson:
I took a close look at kg_get_ccache_name() and made what I think are reasonable simplifications.  If krb5_cc_default_name() returns null, we don't really know what went wrong and ENOMEM isn't a terrible inference (or at least it's not worse than "Invalid argument").   As such we can just use a couple of trinary expressions to set `*out_name` and `err`, and eliminate `name`.

frozencemetery:
Agreed about `ENOMEM` there.  Code looks correct.

PR #1195, opened 2021-06-03
Allow kinit with keytab to defer canonicalization

frozencemetery:
Using `dns_canonicalize_hostname = fallback`, currently:

```
# KRB5_TRACE=/dev/stderr kinit -kt FILE:/etc/krb5.keytab
kinit: Cannot determine realm for host (principal host/kerberos.example.com@)
#
```

With changes:

```
# kinit -kt FILE:/etc/krb5.keytab
[61971] 1622750537.186595: Matching host/kerberos.example.com@ in collection with result: -1765328243/Can't find client principal host/kerberos.example.com@ in cache collection
[61971] 1622750537.186596: Getting initial credentials for host/kerberos.example.com@
[61971] 1622750537.186597: Found entries for host/kerberos.example.com@EXAMPLE.COM in keytab: aes256-cts, aes128-cts, rc4-hmac, camellia256-cts, camellia128-cts
[61971] 1622750537.186601: Sending unauthenticated request
[61971] 1622750537.186602: Sending request (211 bytes) to EXAMPLE.COM
[61971] 1622750537.186603: Resolving hostname kerberos.example.com
[61971] 1622750537.186604: Sending initial UDP request to dgram 127.0.0.1:88
[61971] 1622750537.186605: Received answer (487 bytes) from dgram 127.0.0.1:88
[61971] 1622750537.186606: Sending DNS URI query for _kerberos.EXAMPLE.COM.
[61971] 1622750537.186607: No URI records found
[61971] 1622750537.186608: Sending DNS SRV query for _kerberos-master._udp.EXAMPLE.COM.
[61971] 1622750537.186609: Sending DNS SRV query for _kerberos-master._tcp.EXAMPLE.COM.
[61971] 1622750537.186610: No SRV records found
[61971] 1622750537.186611: Response was not from primary KDC
[61971] 1622750537.186612: Received error from KDC: -1765328359/Additional pre-authentication required
[61971] 1622750537.186615: Preauthenticating using KDC method data
[61971] 1622750537.186616: Processing preauth types: PA-FX-FAST (136), PA-ETYPE-INFO2 (19), PA-SPAKE (151), PA-ENC-TIMESTAMP (2), PA-FX-COOKIE (133)
[61971] 1622750537.186617: Selected etype info: etype aes256-cts, salt "EXAMPLE.COMhostkerberos.example.com", params ""
[61971] 1622750537.186618: Received cookie: MIT1\x00\x00\x00\x01\xd0\x883\xedq\xd3\xb5\xe7\x9e\xcd\xef$\xd8\x0b\xf5\x81D\x99v(\xa3\xe8\x15n\x85B\x0a\xa3sZ\xbaT"\\x14z\x91\xda\x04\xd0c\xca\x9d,\xfe\x9d\x8f\xbd\x9e\x041\xe6\xfd,\x98\x1f}\x86\x08<K\x1dN\x94<\xb1b\x97dw\x15\xda\xa7\xce-\x8a\x12\x80]\xa5P\xae\xf5\xd2\x05\xb8YN\xc4\x82\xde\x1f\xbf\x0f\xf3\x06\xd1\xb3\xa1\x837\x85Y\xd8\x94w\xe8\xe4\xda\x9a}\x8e\xd8\x1c6\xfc\xf7\x1b\x08\xf7\xde\xc9\xaf3B\x10\xefs\xe5\xe3
[61971] 1622750537.186619: Retrieving host/kerberos.example.com@EXAMPLE.COM from FILE:/etc/krb5.keytab (vno 0, enctype aes256-cts) with result: 0/Success
[61971] 1622750537.186620: AS key obtained for encrypted timestamp: aes256-cts/89F3
[61971] 1622750537.186622: Encrypted timestamp (for 1622750537.190262): plain 301AA011180F32303231303630333230303231375AA105020302E736, encrypted 4E70B81D907C0C4B547474DD573C6CC2ACD7D43AD4ABD50C7AD06627AA0DC85512521AE3BC6FF29FA1AC222EA0A4B381258855EDD2D71FC5
[61971] 1622750537.186623: Preauth module encrypted_timestamp (2) (real) returned: 0/Success
[61971] 1622750537.186624: Produced preauth for next request: PA-FX-COOKIE (133), PA-ENC-TIMESTAMP (2)
[61971] 1622750537.186625: Sending request (447 bytes) to EXAMPLE.COM
[61971] 1622750537.186626: Resolving hostname kerberos.example.com
[61971] 1622750537.186627: Sending initial UDP request to dgram 127.0.0.1:88
[61971] 1622750537.186628: Received answer (808 bytes) from dgram 127.0.0.1:88
[61971] 1622750537.186629: Sending DNS URI query for _kerberos.EXAMPLE.COM.
[61971] 1622750537.186630: No URI records found
[61971] 1622750537.186631: Sending DNS SRV query for _kerberos-master._udp.EXAMPLE.COM.
[61971] 1622750537.186632: Sending DNS SRV query for _kerberos-master._tcp.EXAMPLE.COM.
[61971] 1622750537.186633: No SRV records found
[61971] 1622750537.186634: Response was not from primary KDC
[61971] 1622750537.186635: Processing preauth types: PA-ETYPE-INFO2 (19)
[61971] 1622750537.186636: Selected etype info: etype aes256-cts, salt "EXAMPLE.COMhostkerberos.example.com", params ""
[61971] 1622750537.186637: Produced preauth for next request: (empty)
[61971] 1622750537.186638: AS key determined by preauth: aes256-cts/89F3
[61971] 1622750537.186639: Decrypted AS reply; session key is: aes256-cts/C2FE
[61971] 1622750537.186640: FAST negotiation: available
[61971] 1622750537.186641: Initializing KCM:0 with default princ host/kerberos.example.com@EXAMPLE.COM
[61971] 1622750537.186642: Storing host/kerberos.example.com@EXAMPLE.COM -> krbtgt/EXAMPLE.COM@EXAMPLE.COM in KCM:0
[61971] 1622750537.186643: Storing config in KCM:0 for krbtgt/EXAMPLE.COM@EXAMPLE.COM: fast_avail: yes
[61971] 1622750537.186644: Storing host/kerberos.example.com@EXAMPLE.COM -> krb5_ccache_conf_data/fast_avail/krbtgt\/EXAMPLE.COM\@EXAMPLE.COM@X-CACHECONF: in KCM:0
[61971] 1622750537.186645: Storing config in KCM:0 for krbtgt/EXAMPLE.COM@EXAMPLE.COM: pa_type: 2
[61971] 1622750537.186646: Storing host/kerberos.example.com@EXAMPLE.COM -> krb5_ccache_conf_data/pa_type/krbtgt\/EXAMPLE.COM\@EXAMPLE.COM@X-CACHECONF: in KCM:0
#
```

### DIFF
--- a/src/lib/kadm5/clnt/client_init.c
+++ b/src/lib/kadm5/clnt/client_init.c
@@ -221,9 +221,16 @@ init_any(krb5_context context, char *client_name, enum init_type init_type,
         return KADM5_MISSING_KRB5_CONF_PARAMS;
     }
 
+    /*
+     * Parse the client name.  If it has an empty realm, it is almost certainly
+     * a host-based principal using DNS fallback processing or the referral
+     * realm, so give it the appropriate name type for canonicalization.
+     */
     code = krb5_parse_name(handle->context, client_name, &client);
     if (code)
         goto error;
+    if (init_type == INIT_SKEY && client->realm.length == 0)
+        client->type = KRB5_NT_SRV_HST;
 
     /*
      * Get credentials.  Also does some fallbacks in case kadmin/fqdn
--- a/src/clients/kinit/kinit.c
+++ b/src/clients/kinit/kinit.c
@@ -505,17 +505,6 @@ k5_begin(struct k_opts *opts, struct k5_data *k5)
                     _("when creating default server principal name"));
             goto cleanup;
         }
-        if (k5->me->realm.data[0] == 0) {
-            ret = krb5_unparse_name(k5->ctx, k5->me, &k5->name);
-            if (ret == 0) {
-                com_err(progname, KRB5_ERR_HOST_REALM_UNKNOWN,
-                        _("(principal %s)"), k5->name);
-            } else {
-                com_err(progname, KRB5_ERR_HOST_REALM_UNKNOWN,
-                        _("for local services"));
-            }
-            goto cleanup;
-        }
     } else if (k5->out_cc != NULL) {
         /* If the output ccache is initialized, use its principal. */
         if (krb5_cc_get_principal(k5->ctx, k5->out_cc, &princ) == 0)
--- a/src/lib/krb5/os/os-proto.h
+++ b/src/lib/krb5/os/os-proto.h
@@ -85,10 +85,15 @@ struct sendto_callback_info {
 
 /*
  * Initialize with all zeros except for princ.  Set no_hostrealm to disable
- * host-to-realm lookup, which ordinarily happens after canonicalizing the host
- * part.  Set subst_defrealm to substitute the default realm for the referral
- * realm after realm lookup (this has no effect if no_hostrealm is set).  Free
- * with free_canonprinc() when done.
+ * host-to-realm lookup, which ordinarily happens during fallback processing
+ * after canonicalizing the host part.  Set subst_defrealm to substitute the
+ * default realm for the referral realm after realm lookup.  Do not set both
+ * flags.  Free with free_canonprinc() when done.
+ *
+ * no_hostrealm only applies if fallback processing is in use
+ * (dns_canonicalize_hostname = fallback).  It will not remove the realm if
+ * krb5_sname_to_principal() already canonicalized the hostname and looked up a
+ * realm.  subst_defrealm applies whether or not fallback processing is in use.
  */
 struct canonprinc {
     krb5_const_principal princ;
--- a/src/lib/krb5/os/sn2princ.c
+++ b/src/lib/krb5/os/sn2princ.c
@@ -271,15 +271,33 @@ krb5_error_code
 k5_canonprinc(krb5_context context, struct canonprinc *iter,
               krb5_const_principal *princ_out)
 {
+    krb5_error_code ret;
     int step = ++iter->step;
 
     *princ_out = NULL;
 
-    /* If we're not doing fallback, the input principal is canonical. */
-    if (context->dns_canonicalize_hostname != CANONHOST_FALLBACK ||
-        iter->princ->type != KRB5_NT_SRV_HST || iter->princ->length != 2 ||
+    /* If the hostname isn't from krb5_sname_to_principal(), the input
+     * principal is canonical. */
+    if (iter->princ->type != KRB5_NT_SRV_HST || iter->princ->length != 2 ||
         iter->princ->data[1].length == 0) {
         *princ_out = (step == 1) ? iter->princ : NULL;
+        return 0;
+    }
+
+    /* If we're not doing fallback, the hostname is canonical, but we may need
+     * to substitute the default realm. */
+    if (context->dns_canonicalize_hostname != CANONHOST_FALLBACK) {
+        if (step > 1)
+            return 0;
+        iter->copy = *iter->princ;
+        if (iter->subst_defrealm && iter->copy.realm.length == 0) {
+            ret = krb5_get_default_realm(context, &iter->realm);
+            if (ret)
+                return ret;
+            iter->copy = *iter->princ;
+            iter->copy.realm = string2data(iter->realm);
+        }
+        *princ_out = &iter->copy;
         return 0;
     }
 
--- a/src/tests/t_kadmin.py
+++ b/src/tests/t_kadmin.py
@@ -51,4 +51,16 @@ for i in range(200):
     realm.run_kadmin(['addprinc', '-randkey', 'foo%d' % i])
 realm.run_kadmin(['listprincs'], expected_msg='foo199')
 
+# Test kadmin -k with the default principal, with and without
+# fallback.  This operation requires canonicalization against the
+# keytab in krb5_get_init_creds_keytab() as the
+# krb5_sname_to_principal() result won't have a realm.  Try with and
+# without without fallback processing since the code paths are
+# different.
+mark('kadmin -k')
+realm.run([kadmin, '-k', 'getprinc', realm.host_princ])
+no_canon_conf = {'libdefaults': {'dns_canonicalize_hostname': 'false'}}
+no_canon = realm.special_env('no_canon', False, krb5_conf=no_canon_conf)
+realm.run([kadmin, '-k', 'getprinc', realm.host_princ], env=no_canon)
+
 success('kadmin and kpasswd tests')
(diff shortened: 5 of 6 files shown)


greghudson:
I noticed a variant of this problem a long time ago, where you have `dns_canonicalize_hostname = true` but haven't configured [domain_realms].

I had expected this patch would fix both problems, but unfortunately it only helps with fallback.  It seems that the subst_defrealm flag in iteration doesn't apply without fallback, because k5_canonprinc() short-circuits to the sn2princ result if dns_canonicalize_hostname is true or false.  This limitation also affects host-based GSS initiator names via krb5_cc_cache_match() and k5_sname_compare().  I will look into making subst_defrealm work without fallback canonicalization.

I also note that kadmin -k doesn't work with fallback, without a specified principal name.  I don't think it's the same reason as kinit -k.

frozencemetery:
> I also note that kadmin -k doesn't work with fallback, without a specified principal name. I don't think it's the same reason as kinit -k.

Doesn't seem to be.  In `krb5_init_creds_set_keytab()`, we call `k5_canonprinc()` clearly expecting it to add the default realm if needed (`.subst_defrealm = TRUE`).  However, because `iter->princ->type == KRB5_NT_PRINCIPAL`, we don't actually do any canonicalization including adding the realm.

greghudson:
In the kadmin -k case, kadmin.c calls krb5_sname_to_principal() and then krb5_unparse_name() to generate a string, which is then passed to kadm5_init_with_skey(), which then re-parses it.  This series of operations of course loses track of the name type.

It will be a little tricky to handle this.  A libkrb5 function should normally treat a KRB5_NT_PRINCIPAL principal as one entered by a user; in particular, "kadmin -k -p host/myhostname" shouldn't do hostname canonicalization.  Practically speaking, the lack of a realm is perhaps a sufficient hint to kadm5_init_with_skey() that the string was originally an sn2princ result and should be assigned a KRB5_NT_SRV_HST name type.  (Of course, sn2princ does sometimes produce a realm, but in those cases we don't need any further canonicalization.)  The more expansive option is to create a new version of kadm5_init_with_skey() which accepts a proper principal parameter, but both the kadmin.c machinery and the client_init.c machinery would have to change to allow that.

greghudson:
I inventoried the k5_canonprinc() call sites, and I don't think there are any regressions to the flags limitation--that is, I don't think we're relying on the subst_defrealm or no_hostrealm flags for non-fallback uses.  All of the uses of no_hostrealm are via the GSS mech, which never passes a realm in the input principal for a host-based name (kg_acceptor_princ() doesn't put one in the principal, and init_sec_context.c:get_credentials() removes the realm for host-based names near the start of the function).  All of the uses of subst_defrealm formerly used exact matching.

I have a candidate patch to apply flags in k5_canonprinc(), but I want to think about whether it's possible to share the realm-fixup code with canonicalize_princ().

greghudson:
I pushed three commits, the second amending Robbie's commit with tests.  The first commit still takes the straightforward approach with duplicated realm processing--that is, there are separate pieces of code for handling the realm flags depending on whether we see them during fallback iteration.  I still plan to see if I can factor that out.

src/lib/kadm5/clnt/client_init.c
```diff
@@ -221,9 +221,16 @@ init_any(krb5_context context, char *client_name, enum init_type init_type,
...
+    /*
+     * Parse the client name.  If it has a realm, it is almost certainly a
+     * host-based principal using DNS fallback processing or the referral
+     * realm, so give it the appropriate name type for canonicalization.
+     */
     code = krb5_parse_name(handle->context, client_name, &client);
     if (code)
         goto error;
+    if (client->realm.length == 0)
+        client->type = KRB5_NT_SRV_HST;
```
frozencemetery, line 233: Does this belong in `k5_infer_principal_type()`?

greghudson: An interesting question.  I think the answer is no: in libkadm5clnt we know a bit about the inputs (although not with super-high confidence, since nothing stops someone from running "kadmin -k -p ghudson@"), but more importantly, we know the name is for an AS client principal, so it can't possibly work with an empty realm.  So it's safe to assume that it's a host-based principal from sn2princ, because if it's not it wasn't going to work anyway.  Putting the logic in k5_infer_principal_type() would be much less bounded.

Put another way: this is a hack because we don't want to redesign the libkadm5 initialization APIs, and we can show that it's a safe hack, but it doesn't belong at so low a level as the library functions for building or parsing a principal name.

greghudson: A related question is whether we should also conditionalize on `init_type == INIT_SKEY`, to better scope the hack.  I'm leaning towards yes, but practically it shouldn't matter much either way.

greghudson:
I may wind up scoping the first commit down to just the subst_defrealm flag, and documenting in os-proto.h that no_hostrealm won't remove a realm filled in by krb5_sname_to_principal().

The problem is that canonicalize_princ() preserves a realm in the input principal if there is one.  This is a deliberate decision to support kprop and kpropd, which use krb5_sname_to_principal() but perform a strictly local operation in a known realm.  If we apply the realm flags after canonicalize_princ() has already been run (the non-fallback cases), we can't distinguish between a realm produced by a host-to-realm lookup and a realm filled in by the caller.

We could get rid of all of this complexity by making krb5_sname_to_principal() defer canonicalization and realm lookup for all values of dns_canonicalize_hostname, not just fallback.  We've been hesitant to do this because we can't know what third-party code does with krb5_sname_to_principal() results.  Of course we're potentially breaking that code with dns_canonicalize_hostname=fallback, and that becomes more of a potential issue if and when fallback becomes the default.  But at least as things currently stand, people can set dns_canonicalize_hostname=true and restore the old behavior.

(We also could of course create a successor interface to krb5_sname_to_principal(), perhaps internal, and use it in our own code.)

greghudson:
Okay, first commit re-scoped to just subst_defrealm and documentation added for no_hostrealm.

I tried factoring out the subst_defrealm manipulation and it just increased the line count and made the flow of canonicalize_princ() slightly more awkward.  Without no_hostrealm it's a really small amount of code.

frozencemetery (commented):
Changes look/test good, thanks.

PR #1196, opened 2021-06-11
Set SSL_OP_IGNORE_UNEXPECTED_EOF for OpenSSL contexts

frozencemetery:
Starting in OpenSSL 3, connection termination without a close_notify
alert causes SSL_read() to return SSL_ERROR_SSL.  Set
SSL_OP_IGNORE_UNEXPECTED_EOF so that we get the more familiar
SSL_ERROR_ZERO_RETURN instead so that we continue working with kdcproxy.

Remove the call to SSL_CTX_get_options() since SSL_CTX_set_options()
doesn't clear existing options.

[This is not detected by openssl CI currently because openssl CI doesn't run the proxy tests.]

### DIFF
--- a/src/plugins/tls/k5tls/openssl.c
+++ b/src/plugins/tls/k5tls/openssl.c
@@ -433,7 +433,7 @@ setup(krb5_context context, SOCKET fd, const char *servername,
       char **anchors, k5_tls_handle *handle_out)
 {
     int e;
-    long options;
+    long options = SSL_OP_NO_SSLv2;
     SSL_CTX *ctx = NULL;
     SSL *ssl = NULL;
     k5_tls_handle handle = NULL;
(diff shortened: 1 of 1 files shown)


greghudson:
Is kdcproxy skipping a step at the end of writing the response?  For that matter, is our client skipping a step when it's done?  We seem to just kill the connection once we receive a reponse.

The change seems fine, but we should have a comment about why we're not concerned about truncation attacks.  Basically: we only use this module for the Kerberos and change-password protocols, and in both cases we read one response beginning with a length, so a truncated response will fail to decode.

frozencemetery:
Good question.

kdcproxy itself is just a WSGI application, so it's not responsible for the TLS layer as I understand it.  I think the problem is in python's ssl module: a `SSLSocket` object (e.g., created by `wrap_socket()`) doesn't have a means to close at the TLS layer - it just hangs up the socket, to my reading.

> a truncated response will fail to decode

My first pass ignored the error entirely and just relied on this behavior from the decoder.  I think setting the flag is cleaner, though, since as you say there aren't truncation concerns.

greghudson:
On investigation, it looks like the prior behavior was to return SSL_ERROR_SYSCALL with errno 0 and no error on the stack.  I changed the commit message and comment accordingly.

frozencemetery (commented):
Your changes look fine to me.

PR #1197, opened 2021-06-20
Using locking in MEMORY krb5_cc_get_principal()

greghudson:
Without locking, the principal pointer could be freed out from under
krb5_copy_principal() by another thread calling krb5_cc_initialize()
or krb5_cc_destroy().

[This seems likely to be the cause of the crash mentioned in https://mailman.mit.edu/pipermail/krbdev/2021-May/013462.html now that I have looked at a stack trace provided by Thomas.]

### DIFF
--- a/src/lib/krb5/ccache/cc_memory.c
+++ b/src/lib/krb5/ccache/cc_memory.c
@@ -578,12 +578,17 @@ krb5_mcc_get_name (krb5_context context, krb5_ccache id)
 krb5_error_code KRB5_CALLCONV
 krb5_mcc_get_principal(krb5_context context, krb5_ccache id, krb5_principal *princ)
 {
-    krb5_mcc_data *ptr = (krb5_mcc_data *)id->data;
-    if (!ptr->prin) {
-        *princ = 0L;
-        return KRB5_FCC_NOFILE;
-    }
-    return krb5_copy_principal(context, ptr->prin, princ);
+    krb5_error_code ret;
+    krb5_mcc_data *d = id->data;
+
+    *princ = NULL;
+    k5_cc_mutex_lock(context, &d->lock);
+    if (d->prin == NULL)
+        ret = KRB5_FCC_NOFILE;
+    else
+        ret = krb5_copy_principal(context, d->prin, princ);
+    k5_cc_mutex_unlock(context, &d->lock);
+    return ret;
 }
 
 krb5_error_code KRB5_CALLCONV


frozencemetery (approved):

ssainz:
Hi @greghudson - found a crash call stack in `krb5_copy_principal()`. It is found by my automated crash reporting tool (I do not have the code of the code that called kerberos lib directly). Do you know if this crash is the same as the one reported in this defect? (if so , it could be fixed by upgrading to the krb that has this fix)

> [0] /opt/mstr/install/lib/libkrb5.so.3.3!krb5_copy_principal [/3rdparty/krb5-1.15.1/src/lib/krb5/krb/copy_princ.c : 43 + 0]

[1] /opt/mstr/install/lib/libgssapi_krb5.so.2.2!krb5_gss_import_name [/3rdparty/krb5-1.15.1/src/lib/gssapi/krb5/import_name.c : 181 + 0x12]
[2] /opt/mstr/install/lib/libgssapi_krb5.so.2.2!gssint_import_internal_name [/3rdparty/krb5-1.15.1/src/lib/gssapi/mechglue/g_glue.c : 392 + 0x11]
[3] /opt/mstr/install/lib/libgssapi_krb5.so.2.2!gss_init_sec_context [/3rdparty/krb5-1.15.1/src/lib/gssapi/mechglue/g_init_sec_context.c : 167 + 0x15]
[4] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x1029809
[5] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x106c94a
[6] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x101fa00
[7] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe8dd90
[8] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x101c3dc
[9] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x101c626
[10] /usr/lib64/libc-2.17.so!__GI_freeaddrinfo + 0x28
[11] 0x7f126c687000
[12] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe8fcc1
[13] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x2784196
[14] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe90bb3
[15] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe91eca
[16] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe903f0
[17] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x322da2
[18] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe8e5df
[19] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0xe89abb
[20] /opt/cloudera/impalaodbc/lib/64/libclouderaimpalaodbc64.so + 0x274b603

greghudson:
> Hi @greghudson - found a crash call stack in krb5_copy_principal() in version 1.15.1. Do you know if this crash could have similar root cause as the one fixed by this PR? that is, while krb5_copy_principal() was being executed, another thread deleted the input_name_buffer by calling gss_release_name() or some other function that clears the input_name_buffer?

I don't think this is necessarily a threading issue.  From the stack trace, it appears that the calling code imported a name of type gss_nt_krb5_principal, which is a legacy name-type.  For this type, the buffer value is not treated as bytes but as a krb5_principal_data structure, which itself contains pointers.  The mechglue makes a shallow copy of the buffer, but cannot make a deep copy because it knows nothing about krb5 name types.  Later, the caller invokes gss_init_sec_context() with that name as the target.  Now that it knows the concrete mechanism, the mechglue calls the krb5 mech's gss_import_name() function on the copied buffer value to create a mechanism name.  But the lifetime of the caller's krb5_principal_data structure may have ended by that point, so the internal pointers might no longer be valid.

I can't confirm this theory without tracking down the calling code.  If I am right, I would recommend switching to GSS_KRB5_NT_PRINCIPAL_NAME and string-form principal names if possible.

ssainz:
Hi @greghudson - thanks a lot ~ Will move on with your theory and recommend the app developer to switch to GSS_KRB5_NT_PRINCIPAL_NAME and string-form principal names.

PR #1200, opened 2021-06-28
otp: add support for otp-challenge mechanism

pbrezina:
otp-challenge flag requires edata to obtain a challenge message from the
server that must be presented to the user in order to authenticate 
correctly.

In order to do so, otp_edata sends a password-less Access-Request and 
expects Access-Challenge in reply. Then reads the challenge from 
Reply-Message attribute and remembers the State attribute inside a cookie.

Then it reads the cookie in otp_verify flow and appends it to the next 
Access-Request together with the authentication token so the RADIUS server
can match the second message with the initial one.

---

This implements external challenge functionality that is required for OAuth2
authentication via RADIUS protocol and FreeIPA as discussed here:

* http://mailman.mit.edu/pipermail/krbdev/2021-June/013469.html

The functionality is not yet implemented in FreeIPA, therefore it requires a special
FreeIPA patch and a mocked RADIUS server for testing. The Kerberos code is completed
and ready for review, I would like to wait hovewer for FreeIPA side to be finished
before merging. Please, review.

## How to test:

On server:
1. Install https://github.com/pbrezina/freeipa/tree/otp (enables challenge flag for tuser)
2. Clone and run https://github.com/pbrezina/mock-radius on IPA server
3. Create RADIUS proxy config “localhost/127.0.0.1/Secret123” in IPA
4. Create user “tuser” with RADIUS proxy set to localhost in IPA
5. In kadmin.local > getstrs tuser > you should see challenge flag set to true

On client:
```
kinit -n @IPA.VM -c FILE:armor.ccache
kinit -T FILE:armor.ccache tuser@IPA.VM
```

Note: challenge=true is set only for tuser, other RADIUS proxy users will have challenge=false

### DIFF
--- a/doc/admin/otp.rst
+++ b/doc/admin/otp.rst
@@ -31,6 +31,7 @@ Token types are defined in either :ref:`krb5.conf(5)` or
             retries = <integer> (default: 3)
             strip_realm = <boolean> (default: true)
             indicator = <string> (default: none)
+            challenge = <boolean> (default: false)
         }
 
 If the server field begins with '/', it will be interpreted as a UNIX
@@ -48,6 +49,10 @@ If an indicator field is present, tickets issued using this token type
 will be annotated with the specified authentication indicator (see
 :ref:`auth_indicator`).  This key may be specified multiple times to
 add multiple indicators.
+
+If the challenge field is true then the a challenge message is obtained from the
+RADIUS server first. This message is presented to the user before asking for an
+authentication token.
 
 
 The default token type
--- a/doc/formats/cookie.rst
+++ b/doc/formats/cookie.rst
@@ -95,3 +95,13 @@ records, including none; a second-factor type need not create a state
 field if it does not need one, and no record is created for SF-NONE.
 For other cookies, there must be exactly one second-factor record
 corresponding to the factor type chosen by the client.
+
+OTP cookie format (version 1)
+-------------------------------
+
+Inside an OTP pre-authentication plugin, a data value of type 142 contains state
+of RADIUS authentication if challenge field of an OTP token type is set to true.
+This data is the concatenation of the following:
+
+* a two-byte big-endian version number with the value 1
+* a RADIUS state (rest of the cookie)
--- a/src/include/krad.h
+++ b/src/include/krad.h
@@ -57,6 +57,16 @@
 #define KRAD_SERVICE_TYPE_CALL_CHECK 10
 #define KRAD_SERVICE_TYPE_CALLBACK_ADMINISTRATIVE 11
 
+#define KRAD_CODE_ACCESS_REQUEST 1
+#define KRAD_CODE_ACCESS_ACCEPT 2
+#define KRAD_CODE_ACCESS_REJECT 3
+#define KRAD_CODE_ACCESS_CHALLENGE 11
+
+#define KRAD_ATTR_USER_NAME 1
+#define KRAD_ATTR_USER_PASSWORD 2
+#define KRAD_ATTR_SERVICE_TYPE 6
+#define KRAD_ATTR_NAS_IDENTIFIER 32
+
 typedef struct krad_attrset_st krad_attrset;
 typedef struct krad_packet_st krad_packet;
 typedef struct krad_client_st krad_client;
--- a/src/lib/krad/t_attr.c
+++ b/src/lib/krad/t_attr.c
@@ -53,26 +53,22 @@ main()
 
     noerror(krb5_init_context(&ctx));
 
-    /* Make sure User-Name is 1. */
-    insist(krad_attr_name2num("User-Name") == 1);
+    insist(krad_attr_name2num("User-Name") == KRAD_ATTR_USER_NAME);
 
-    /* Make sure 2 is User-Password. */
-    tmp = krad_attr_num2name(2);
+    tmp = krad_attr_num2name(KRAD_ATTR_USER_PASSWORD);
     insist(tmp != NULL);
     insist(strcmp(tmp, "User-Password") == 0);
 
     /* Test decoding. */
     in = make_data((void *)encoded, sizeof(encoded));
-    noerror(kr_attr_decode(ctx, secret, auth,
-                           krad_attr_name2num("User-Password"),
+    noerror(kr_attr_decode(ctx, secret, auth, KRAD_ATTR_USER_PASSWORD,
                            &in, outbuf, &len));
     insist(len == strlen(decoded));
     insist(memcmp(outbuf, decoded, len) == 0);
 
     /* Test encoding. */
     in = string2data((char *)decoded);
-    retval = kr_attr_encode(ctx, secret, auth,
-                            krad_attr_name2num("User-Password"),
+    retval = kr_attr_encode(ctx, secret, auth, KRAD_ATTR_USER_PASSWORD,
                             &in, outbuf, &len);
     insist(retval == 0);
     insist(len == sizeof(encoded));
@@ -80,9 +76,9 @@ main()
 
     /* Test constraint. */
     in.length = 100;
-    insist(kr_attr_valid(krad_attr_name2num("User-Password"), &in) == 0);
+    insist(kr_attr_valid(KRAD_ATTR_USER_PASSWORD, &in) == 0);
     in.length = 200;
-    insist(kr_attr_valid(krad_attr_name2num("User-Password"), &in) != 0);
+    insist(kr_attr_valid(KRAD_ATTR_USER_PASSWORD, &in) != 0);
 
     krb5_free_context(ctx);
     return 0;
--- a/src/lib/krad/t_attrset.c
+++ b/src/lib/krad/t_attrset.c
@@ -55,24 +55,24 @@ main()
 
     /* Add username. */
     tmp = string2data((char *)username);
-    noerror(krad_attrset_add(set, krad_attr_name2num("User-Name"), &tmp));
+    noerror(krad_attrset_add(set, KRAD_ATTR_USER_NAME, &tmp));
 
     /* Add password. */
     tmp = string2data((char *)password);
-    noerror(krad_attrset_add(set, krad_attr_name2num("User-Password"), &tmp));
+    noerror(krad_attrset_add(set, KRAD_ATTR_USER_PASSWORD, &tmp));
 
     /* Encode attrset. */
     noerror(kr_attrset_encode(set, "foo", auth, buffer, &encode_len));
     krad_attrset_free(set);
 
     /* Manually encode User-Name. */
-    encoded[len + 0] = krad_attr_name2num("User-Name");
+    encoded[len + 0] = KRAD_ATTR_USER_NAME;
     encoded[len + 1] = strlen(username) + 2;
     memcpy(encoded + len + 2, username, strlen(username));
     len += encoded[len + 1];
 
     /* Manually encode User-Password. */
-    encoded[len + 0] = krad_attr_name2num("User-Password");
+    encoded[len + 0] = KRAD_ATTR_USER_PASSWORD;
     encoded[len + 1] = sizeof(encpass) + 2;
     memcpy(encoded + len + 2, encpass, sizeof(encpass));
     len += encoded[len + 1];
@@ -87,7 +87,7 @@ main()
 
     /* Test getting an attribute. */
     tmp = string2data((char *)username);
-    tmpp = krad_attrset_get(set, krad_attr_name2num("User-Name"), 0);
+    tmpp = krad_attrset_get(set, KRAD_ATTR_USER_NAME, 0);
     insist(tmpp != NULL);
     insist(tmpp->length == tmp.length);
     insist(strncmp(tmpp->data, tmp.data, tmp.length) == 0);
--- a/src/lib/krad/t_client.c
+++ b/src/lib/krad/t_client.c
@@ -74,33 +74,31 @@ main(int argc, const char **argv)
 
     tmp = string2data("testUser");
     noerror(krad_attrset_new(kctx, &attrs));
-    noerror(krad_attrset_add(attrs, krad_attr_name2num("User-Name"), &tmp));
+    noerror(krad_attrset_add(attrs, KRAD_ATTR_USER_NAME, &tmp));
 
     /* Test accept. */
     tmp = string2data("accept");
-    noerror(krad_attrset_add(attrs, krad_attr_name2num("User-Password"),
-                             &tmp));
-    noerror(krad_client_send(rc, krad_code_name2num("Access-Request"), attrs,
+    noerror(krad_attrset_add(attrs, KRAD_ATTR_USER_PASSWORD, &tmp));
+    noerror(krad_client_send(rc, KRAD_CODE_ACCESS_REQUEST, attrs,
                              "localhost", "foo", 1000, 3, callback, NULL));
     verto_run(vctx);
 
     /* Test reject. */
     tmp = string2data("reject");
-    krad_attrset_del(attrs, krad_attr_name2num("User-Password"), 0);
-    noerror(krad_attrset_add(attrs, krad_attr_name2num("User-Password"),
-                             &tmp));
-    noerror(krad_client_send(rc, krad_code_name2num("Access-Request"), attrs,
+    krad_attrset_del(attrs, KRAD_ATTR_USER_PASSWORD, 0);
+    noerror(krad_attrset_add(attrs, KRAD_ATTR_USER_PASSWORD, &tmp));
+    noerror(krad_client_send(rc, KRAD_CODE_ACCESS_REQUEST, attrs,
                              "localhost", "foo", 1000, 3, callback, NULL));
     verto_run(vctx);
 
     /* Test timeout. */
     daemon_stop();
-    noerror(krad_client_send(rc, krad_code_name2num("Access-Request"), attrs,
+    noerror(krad_client_send(rc, KRAD_CODE_ACCESS_REQUEST, attrs,
                              "localhost", "foo", 1000, 3, callback, NULL));
     verto_run(vctx);
 
     /* Test outstanding packet freeing. */
-    noerror(krad_client_send(rc, krad_code_name2num("Access-Request"), attrs,
+    noerror(krad_client_send(rc, KRAD_CODE_ACCESS_REQUEST, attrs,
                              "localhost", "foo", 1000, 3, callback, NULL));
     krad_client_free(rc);
     rc = NULL;
@@ -108,11 +106,9 @@ main(int argc, const char **argv)
     /* Verify the results. */
     insist(record.count == EVENT_COUNT);
     insist(record.events[0].error == FALSE);
-    insist(record.events[0].result.code ==
-           krad_code_name2num("Access-Accept"));
+    insist(record.events[0].result.code == KRAD_CODE_ACCESS_ACCEPT);
     insist(record.events[1].error == FALSE);
-    insist(record.events[1].result.code ==
-           krad_code_name2num("Access-Reject"));
+    insist(record.events[1].result.code == KRAD_CODE_ACCESS_REJECT);
     insist(record.events[2].error == TRUE);
     insist(record.events[2].result.retval == ETIMEDOUT);
     insist(record.events[3].error == TRUE);
--- a/src/lib/krad/t_code.c
+++ b/src/lib/krad/t_code.c
@@ -34,19 +34,19 @@ main()
 {
     const char *tmp;
 
-    insist(krad_code_name2num("Access-Request") == 1);
-    insist(krad_code_name2num("Access-Accept") == 2);
-    insist(krad_code_name2num("Access-Reject") == 3);
+    insist(krad_code_name2num("Access-Request") == KRAD_CODE_ACCESS_REQUEST);
+    insist(krad_code_name2num("Access-Accept") == KRAD_CODE_ACCESS_ACCEPT);
+    insist(krad_code_name2num("Access-Reject") == KRAD_CODE_ACCESS_REJECT);
 
-    tmp = krad_code_num2name(1);
+    tmp = krad_code_num2name(KRAD_CODE_ACCESS_REQUEST);
     insist(tmp != NULL);
     insist(strcmp(tmp, "Access-Request") == 0);
 
-    tmp = krad_code_num2name(2);
+    tmp = krad_code_num2name(KRAD_CODE_ACCESS_ACCEPT);
     insist(tmp != NULL);
     insist(strcmp(tmp, "Access-Accept") == 0);
 
-    tmp = krad_code_num2name(3);
+    tmp = krad_code_num2name(KRAD_CODE_ACCESS_REJECT);
     insist(tmp != NULL);
     insist(strcmp(tmp, "Access-Reject") == 0);
 
--- a/src/lib/krad/t_packet.c
+++ b/src/lib/krad/t_packet.c
@@ -62,22 +62,20 @@ make_packet(krb5_context ctx, const krb5_data *username,
     if (retval != 0)
         goto out;
 
-    retval = krad_attrset_add(set, krad_attr_name2num("User-Name"), username);
+    retval = krad_attrset_add(set, KRAD_ATTR_USER_NAME, username);
     if (retval != 0)
         goto out;
 
-    retval = krad_attrset_add(set, krad_attr_name2num("User-Password"),
-                              password);
+    retval = krad_attrset_add(set, KRAD_ATTR_USER_PASSWORD, password);
     if (retval != 0)
         goto out;
 
-    retval = krad_packet_new_request(ctx, "foo",
-                                     krad_code_name2num("Access-Request"),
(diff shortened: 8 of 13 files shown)


greghudson:
Out of curiosity, did github email you about the CI failure?

I'm still working on internalizing the design, but I notice that the cb field of request_st was made polymorphic, so that it can contain either an otp_cb for a verification request or an otp_challenge_cb for a challenge request.  For both the existing otp_state_verify() and the new otp_state_challenge(), there is only one call site and therefore only one possible value for the callback.  So we're passing continuations when we don't need to.   I can sort of see why Nathaniel might have preferred that design, but I would rather remove the continuation passing (so otp_state.c just calls back into main.c) than sacrifice type safety.

We have a helper empty_data() in k5-int.h which should be used in preference to {0} or (krb5_data){0}.

pbrezina:
> Out of curiosity, did github email you about the CI failure?

Yes, I did. It just slipped my attention due to internal changes in our e-mail server. Hopefully, it will be green now.

> I'm still working on internalizing the design, but I notice that the cb field of request_st was made polymorphic, so that it can contain either an otp_cb for a verification request or an otp_challenge_cb for a challenge request. For both the existing otp_state_verify() and the new otp_state_challenge(), there is only one call site and therefore only one possible value for the callback. So we're passing continuations when we don't need to. I can sort of see why Nathaniel might have preferred that design, but I would rather remove the continuation passing (so otp_state.c just calls back into main.c) than sacrifice type safety.

Hmm, I'm not sure what do you mean by one call site since otp_state_verify is called from otp_verify, otp_state_challenge is called from otp_edata and it requires to pass little more information (state and message) then just an error code. Can you explain it in more detail?

> We have a helper empty_data() in k5-int.h which should be used in preference to {0} or (krb5_data){0}.

Thanks, I was looking for that. Done.

greghudson:
> Can you explain it in more detail?

Sure.  In the existing code, when otp_state.c:callback() is done, it invokes req->cb(...).  But req->cb is always on_reponse() from main.c, because otp_state_verify() is only called in one place.  With a little adjustment in declarations, we could eliminate the cb field of request and just call the main.c function by name from callback().

Similarly, in the new code, when on_challenge() is done, it invokes req->cb (after casting).  Again, this callback is always otp_edata_challenge_done() because otp_state_challenge() is only called in one place.  So again, we could just call the (different) function by name.

pbrezina:
I see. I did this change in a new commit.

pbrezina:
@greghudson Hi Greg, would you be able to find some time for a review? Thank you.

greghudson:
I pushed some changes to further increase type-safety.  I restructured the commits so that the structural changes to the existing code are in one commit and the functional additions are in another.

I identified some TBD items (which I can help with):

* I think the OTP cookie value should have a version number, similar to how we have one in the SPAKE cookie.  Otherwise if we want to add something to the cookie value, we have to worry about a 1.20 KDC treating the whole blob as a RADIUS state.

* Documentation in doc/admin/otp.rst

* Documentation in doc/admin/admin_commands/kadmin_local.rst for set_string (maybe).

* Documentation of the OTP cookie semantics in doc/formats/cookie.rst

* We take the challenge flag off the first token in the `otp` string attribute, but potentially send the password-less request to multiple token servers.  Should we perhaps error out if challenge is set and there is more than one token in the array?  It doesn't seem like we can support a mix of challenge and non-challenge tokens, or multiple different tokens with their own challenges (since we still only send one OTP-TOKENINFO in the PA-OTP-CHALLENGE).

greghudson:
Also, I ran into a bit of a mystery while working on the aforementioned changes.  I initially neglected to include the code to fetch the RADIUS state from the coookie and include it in the verification request.  This should have caused the test RADIUS server to send another Access-Challenge response and the KDC to fail the preauth attempt.  And kinit did report a "Preauthentication failed" error.  But it happened with a mysterious delay; kinit had to retransmit the request over UDP (after trying over TCP and getting a connection reset) before receiving the error.  Also, t_otp.py went into some kind of deadlock after the failure, when I would have expected an assertion failure in verify().  I'd like to investigate this and figure out what went awry.

pbrezina:
> I pushed some changes to further increase type-safety. I restructured the commits so that the structural changes to the existing code are in one commit and the functional additions are in another.

Thank you for your work.

> I identified some TBD items (which I can help with):
> 
>     * I think the OTP cookie value should have a version number, similar to how we have one in the SPAKE cookie.  Otherwise if we want to add something to the cookie value, we have to worry about a 1.20 KDC treating the whole blob as a RADIUS state.

Do you mean this one? https://github.com/krb5/krb5/blob/master/src/plugins/preauth/spake/spake_kdc.c#L138

>     * Documentation in doc/admin/otp.rst
>     * Documentation in doc/admin/admin_commands/kadmin_local.rst for set_string (maybe).
>     * Documentation of the OTP cookie semantics in doc/formats/cookie.rst

Ok, I'll work on these.

>     * We take the challenge flag off the first token in the `otp` string attribute, but potentially send the password-less request to multiple token servers.  Should we perhaps error out if challenge is set and there is more than one token in the array?  It doesn't seem like we can support a mix of challenge and non-challenge tokens, or multiple different tokens with their own challenges (since we still only send one OTP-TOKENINFO in the PA-OTP-CHALLENGE).

Enforce single server is a good idea. We could enforce to use the same server in both edata and verify, but perhaps it is an overkill?

greghudson:
> Do you mean this one? https://github.com/krb5/krb5/blob/master/src/plugins/preauth/spake/spake_kdc.c#L138

Yes.  Just a 16-bit number with value 1 before the RADIUS state, and we return KRB5KDC_ERR_PREAUTH_FAILED if we see a cookie with a different version number.

> Enforce single server is a good idea. We could enforce to use the same server in both edata and verify, but perhaps it is an overkill?

Can you elaborate?  I'm not sure what you mean by enforcing the same server between edata and verify.

pbrezina:
> > Do you mean this one? https://github.com/krb5/krb5/blob/master/src/plugins/preauth/spake/spake_kdc.c#L138
> 
> Yes. Just a 16-bit number with value 1 before the RADIUS state, and we return KRB5KDC_ERR_PREAUTH_FAILED if we see a cookie with a different version number.

I added new commit for this so you can see the diff. We can squash it later.

Is it possible to get and set the cookie on the client? What I would like to do in SSSD it to get and remember the cookie in responder if challenge string from krb5_responder_otp_get_challenge is not empty and then set the cookie again when krb5_init_creds_init is called.


> > Enforce single server is a good idea. We could enforce to use the same server in both edata and verify, but perhaps it is an overkill?
> 
> Can you elaborate? I'm not sure what you mean by enforcing the same server between edata and verify.

Nevermind. The idea was to store the server that was used to obtain the challenge in cookie and then use this server in otp_verify, but there would be too many corner cases. It is better to support only single server if challenge is enabled.

greghudson:
> I added new commit for this so you can see the diff. We can squash it later.

This is fine but unnecessary.  When github reports a force-push, one can click on the word "force-pushed" to see a diff between the total patch-set before and after the change.  (This is less useful if a rebase on top of the master branch is included in the force-push, which is why I try to do that separately.)

> Is it possible to get and set the cookie on the client? What I would like to do in SSSD it to get and remember the cookie in responder if challenge string from krb5_responder_otp_get_challenge is not empty and then set the cookie again when krb5_init_creds_init is called.

It isn't, and cookies only remain valid for a few minutes to mitigate the risk of cookie replays by an attacker.  Can you explain the benefit or necessity here?

pbrezina:
> > Is it possible to get and set the cookie on the client? What I would like to do in SSSD it to get and remember the cookie in responder if challenge string from krb5_responder_otp_get_challenge is not empty and then set the cookie again when krb5_init_creds_init is called.
> 
> It isn't, and cookies only remain valid for a few minutes to mitigate the risk of cookie replays by an attacker. Can you explain the benefit or necessity here?

There is a long way between PAM module and krb5 code in SSSD through multiple processes: pam module (pam_sss) -> responder (sssd_pam) -> backend (sssd_be) -> krb5 client (krb5_child). krb5_child performs kinit for the user who's being logged in, returns error code and terminates. At this moment, there does not exist any way to send prompt from krb5 to pam_sss. The change would be rather straightforward if we would be able to remember the state cookie and restore it in subsequent call so I don't need to deal with keeping the krb5_child process alive. But this is just to simplify my work, it is certainly not required.

Is the cookie lifetime configurable or hard coded?

greghudson:
The cookie lifetime is hardcoded at ten minutes.

I'm not sure if getting and setting the cookie would be enough to allow a general init_creds operation to take place across multiple processes.  We might need a way to serialize and deserialize a krb5_init_creds_context, including the modreq object in the clpreauth interface.  That would be a substantial amount of work.

pbrezina:
I added a documentation, squashed patches and rebased.

I have a functional PoC in SSSD so I think we can go on with this pull request.

pbrezina:
SSSD PR: https://github.com/SSSD/sssd/pull/5762

greghudson:
I had second thoughts about krad_packet_check_code(), since it's not common in our tree to do lookups from string constants to find numeric constants.  I instead added krad.h constants for the packet codes and attribute types we use, and switched most of the attr2num and code2num calls to use those instead.

pbrezina:
I agree, it is better this way.

While the sssd implementation is working, it currently uses a mocked challenge values before we agree on a challenge format. We would like to only send oauth2 information in the challenge so it can be used in a translatable message on the client. This is however not user-friendly for kinit since it would not get any human message.

Would you agree to patch kinit to support specific challenge response? If yes, what format do you prefer? We are currently thinking about something as `ipa-oath2 $message` where $message is either json string or base64 encoded blob that holds version info, oauth2 url and oauth2 pin.

greghudson:
I'm uncomfortable with that because I think it further diverges from the specification of otp-challenge in RFC 6560.  In that document, I think the author envisioned the otp-challenge field being used by a hardware token with a (probably binary) challenge-response function.  There is nothing in the text about presenting otp-challenge to the user (whereas it does talk about presenting otp-vendor to the user), and it is defined in the ASN.1 sequence as an OCTET STRING rather than a UTF8String.  Our current client plugin module diverges from this apparent vision when using the prompter, as it displays the challenge in the prompt.  So, already we have a mismatch between the protocol document and the only implementation.  Building a further subprotocol into the otp-challenge field adds an additional tangle.

In addition, thinking about this led me to a possibly much larger concern about the design of this authentication flow.  Looking at the FreeIPA architectural document, it appears that the user is tasked with visiting the provided URL and entering the provided user code, then verifying the authorization request.  There is no mention of then providing the user with an OTP value to by typed into the Kerberos client.  If, after user approval of the authorization request, the KDC issues a ticket based on the cookie alone, I don't think that is secure.  Cookies are bound to a client principal, but not to a particular FAST armor ticket.  An authenticated (as anybody) attacker who can view the network could attach a legitimate user's cookie to their own FAST-armored request, receiving a ticket with the session key encrypted in the attacker's FAST armor key.

I guess we could fix that problem by adding some kind of binding to the armor ticket within the cookie value.  But finding the concern this late makes me worried that in gluing together mismatched components, we might be building a fundamentally insecure authentication protocol.

pbrezina:
> I'm uncomfortable with that because I think it further diverges from the specification of otp-challenge in RFC 6560. In that document, I think the author envisioned the otp-challenge field being used by a hardware token with a (probably binary) challenge-response function. There is nothing in the text about presenting otp-challenge to the user (whereas it does talk about presenting otp-vendor to the user), and it is defined in the ASN.1 sequence as an OCTET STRING rather than a UTF8String. Our current client plugin module diverges from this apparent vision when using the prompter, as it displays the challenge in the prompt. So, already we have a mismatch between the protocol document and the only implementation. Building a further subprotocol into the otp-challenge field adds an additional tangle.

If I understand it correctly, the otp-challenge should be used by the device to calculate OTP. In this case there is no real device, so we can consider the kinit/sssd as a device and it parses the challenge and then as a way to calculate the OTP in presents some information to the user. The challenge for our purpose is really not a human readable string and we would like to convert it to a human readable string.

We could even fill the vendor, but I did not find any RADIUS attribute that could be used for that.

> In addition, thinking about this led me to a possibly much larger concern about the design of this authentication flow. Looking at the FreeIPA architectural document, it appears that the user is tasked with visiting the provided URL and entering the provided user code, then verifying the authorization request. There is no mention of then providing the user with an OTP value to by typed into the Kerberos client. If, after user approval of the authorization request, the KDC issues a ticket based on the cookie alone, I don't think that is secure. Cookies are bound to a client principal, but not to a particular FAST armor ticket. An authenticated (as anybody) attacker who can view the network could attach a legitimate user's cookie to their own FAST-armored request, receiving a ticket with the session key encrypted in the attacker's FAST armor key.

So what you're saying is that the cookie is not encrypted and can be read and used by anyone?

Is the tokeninfo encrypted?

> I guess we could fix that problem by adding some kind of binding to the armor ticket within the cookie value. But finding the concern this late makes me worried that in gluing together mismatched components, we might be building a fundamentally insecure authentication protocol.

Ipa-otpd talks oauth2 with external identity provider, it returns url and pin for the user. This is transmitted in the challenge. User visits the url, enters the pin and authenticates. Then press enter to send a reply. ipa-otpd then talks to the identity provider to check that the authentication was successful.

Currently, sssd sends the PIN as the password, this can be used in ipa-otpd as a verification mechanism (out and in pin are the same).

greghudson:
The cookie is encrypted; see http://web.mit.edu/kerberos/krb5-latest/doc/formats/cookie.html .  But being encrypted is never the end of the story; in the absence of cryptographic bindings, encrypted blobs can be taken from one protocol object and placed in another.

On further reflection I was probably wrong about this attack; for a FAST-armored AS request, I believe the request padata (including the cookie) are contained in the FAST inner body, which is protected by (and therefore bound to) the FAST armor key.  I still feel like we're using mismatched components, though--specifically, using OTP when the KDC never verifies a one-time password.

> Is the tokeninfo encrypted?

It should also be encrypted in the FAST armor key (RFC 6113 section 5.4.4).

> If I understand it correctly, the otp-challenge should be used by the device to calculate OTP. In this case there is no real device, so we can consider the kinit/sssd as a device and it parses the challenge and then as a way to calculate the OTP in presents some information to the user.

I don't think I can convince myself to view the combination of kinit and a human using a web browser as "a device", and as I understand things we aren't calculating a one-time password to present to the KDC at all.

pbrezina:
> The cookie is encrypted; see http://web.mit.edu/kerberos/krb5-latest/doc/formats/cookie.html . But being encrypted is never the end of the story; in the absence of cryptographic bindings, encrypted blobs can be taken from one protocol object and placed in another.
> 
> On further reflection I was probably wrong about this attack; for a FAST-armored AS request, I believe the request padata (including the cookie) are contained in the FAST inner body, which is protected by (and therefore bound to) the FAST armor key.

This is good news.

> I still feel like we're using mismatched components, though--specifically, using OTP when the KDC never verifies a one-time password.

So, where does this leave us? Do you have objections only to modifying kinit or the whole PR?

@sumit-bose Is it possible to get some otp from the oauth2 authentication that the user enter?

> > Is the tokeninfo encrypted?
> 
> It should also be encrypted in the FAST armor key (RFC 6113 section 5.4.4).
> 
> > If I understand it correctly, the otp-challenge should be used by the device to calculate OTP. In this case there is no real device, so we can consider the kinit/sssd as a device and it parses the challenge and then as a way to calculate the OTP in presents some information to the user.
> 
> I don't think I can convince myself to view the combination of kinit and a human using a web browser as "a device", and as I understand things we aren't calculating a one-time password to present to the KDC at all.

abbra:
my understanding is that an application from the URL that a user would visit and where it would enter the value from the text of the prompt, would instruct the user to re-enter a unique code that we treat as a password for the response. Validation of that code/password is done by ipa-otpd on behalf of KDC similar to any other RADIUS conversation, so I am interested how this is different from a RADIUS auth for KDC? KDC does not validate this value in RADIUS auth case either.

sumit-bose:
> my understanding is that an application from the URL that a user would visit and where it would enter the value from the text of the prompt, would instruct the user to re-enter a unique code that we treat as a password for the response. Validation of that code/password is done by ipa-otpd on behalf of KDC similar to any other RADIUS conversation, so I am interested how this is different from a RADIUS auth for KDC? KDC does not validate this value in RADIUS auth case either.

Hi,

no, it is a @pbrezina, described it. With the device authorization scheme the user gets back a URL of the IdP provider and some random number which should be typed in after opening the URL. If this user is already authenticated on the IdP just a confirmation page is shown to allow access, otherwise the user has to authenticate to the IdP before getting the confimation page. Now everything the user has to do is done. All this happens in the browser.

In the meantime in our case ipa-otpd will pull in regular intervals a special URL on the IdP which will return a "authorization_pending" error as long as the user has not finished the steps in the browser. If the steps in the browser are done the IdP will return a user and an access token.

HTH

bye,
Sumit

abbra:
Right, thanks  for correcting me. An example flow can be played with at https://www.oauth.com/playground/device-code.html. If user interaction with the URL of the IdP provider is successful and there is nothing more to enter at kinit prompt, would an empty response be enough for the client-KDC communication to continue?

greghudson:
> So, where does this leave us? Do you have objections only to modifying kinit or the whole PR?

I think it's worth considering better-fitting alternatives than FAST OTP.  The most obvious alternative I can think of is a new preauth type:

* Like FAST OTP it would be a FAST factor, meaning it would require an armor ticket to use.
* The code could conceivably live within the FreeIPA code base as clpreauth and kdcpreauth modules, allowing for easier control by Red Hat.  In this case the clpreauth module would still work with kinit/etc. but only on clients with FreeIPA installed.
* It could use the FAST OTP code as a basis (just because it's a FAST factor using RADIUS on the KDC), but could discard all of the code for handling the OTP protocol data and non-challenge RADIUS requests.
* The new preauth protocol could send exactly the right information (URL and auth code) in the edata.  It wouldn't have to use ASN.1.
* A principal could be configured to allow either OTP or OAUTH, unlike the current approach.

I don't think the FAST OTP approach is fatally flawed, but when you get into the weeds like properly localizing the user instructions for the IdP, it's clear that it's not a perfect fit.

If we do go ahead with the FAST OTP approach, the best compromise I can think of for the kinit user experience is to design a human-and-machine-readable format for the otp-challenge.

pbrezina:
Creating a plugin (say ipa-oauth2) sounds reasonable to me. @abbra @sumit-bose what do you think?

I just want to clarify if I get the plugin architecture right:

- In preauth, we save our date in pa_data and then read it in clpreath. We would not be bound to any specific format and we'll have complete control over it?
- In clpreath:
  - in `prep_question` we'll parse padata and `calls cb->ask_responder_question`, otp uses json format for challenge but it can be arbitrary string?
  - in `process` we'll get the answer with `cb->get_responder_answer`

kinit seems to use krb5_prompter and not the responder interface, how is the challenge presented during kinit?

abbra:
I was considering a separate preauth plugin initially but thought we could avoid it by adding this logic to the FAST OTP plugin. So creating a separate plugin is not against an original plan. ;)

greghudson:
@pbrezina: yes to all of those questions.  If the caller doesn't implement a responder (like kinit) or doesn't answer because it doesn't know the type of question, the clpreauth process method has to fall back to the prompter (or give up if prompter is NULL).  There's some documentation of the prompter callback here: http://web.mit.edu/kerberos/krb5-latest/doc/appdev/init_creds.html#prompter-callback

pbrezina:
@greghudson I'm moving this to custom plugins inside SSSD code. I have few questions.

* What number should I use as padata type?
* In OTP, padata is set in otp_edata. What is the purpose of `otp_return_padata`? Why is the key switched to the FAST key?
* Why is the nonce encrypted? Isn't the whole padata already encrypted?

Thanks, Pavel.

greghudson:
> What number should I use as padata type?

Pick a negative number for now (RFC 4120 section 5.2.7: "Negative values of padata-type are reserved for unregistered use".  To get a registered value, we will need to submit a request to IANA, which will then be subject to expert review, probably by Sam Hartman.  We'll need a writeup for him to review.  I can help with this.  (RFC 6113 section 7.1 lays out the rules of the padata type registry.)

> In OTP, padata is set in otp_edata. What is the purpose of otp_return_padata? Why is the key switched to the FAST key?

The return method can provide additional padata to be sent to the client along with the issued ticket, and can change the reply key.  The OTP preauth method does send a padata message with the issued ticket, but it must change the reply key.  This is laid out in RFC 6560 section 3.6; note that we only implement the variant where the otp-value is presented in the .

The "reply key" is the encryption key for the encrypted part of the KDC-REP message, which contains the ticket session key.  The default value of the reply key is the chosen long-term key of the principal--typically either a password-derived key or a random key which the authorized entity has stored in a keytab.

The OTP preauth mech changes the reply key to the armor key because it is intended to be used as a replacement for Kerberos long-term keys, not in concert with them.  This is an inelegant solution for a couple of reasons.  First, it means that authentication of the KDC to the client hinges upon the armor ticket, and that any holder of the armor ticket, not just the authenticating user, can know the session key of the issued ticket.  Second, the FAST framework envisions the possibility of combining preauth mechs (though this has never been implemented), and having some mechs make the negative statement "this mech replaces the use of long-term keys" runs counter to that vision.  But, it is what it is, and I recommend that you adopt the same solution for the oauth mech.

> Why is the nonce encrypted? Isn't the whole padata already encrypted?

Short answer: you should not need anything like this in the oauth mech.

RFC 6560 was designed to address two scenarios:

1. The KDC can use some back-end to determine whether a candidate OTP value is valid.
2. The KDC can retrieve the current valid OTP value (or perhaps a short list of them, to be tried in sequence).

Scenario 2 has better security properties because the OTP value can be used as a shared secret.  But most OTP infrastructure only allows for scenario 1, and that is the only one implemented in MIT krb5.

In scenario 2 the client does not send the OTP value in the PA-OTP-REQUEST; instead, it uses the OTP value to compute a derivative of the armor key, called the "client key".  The encData field of the PA-OTP-REQUEST proves knowledge of the client key.  In scenario 1 this is (to the best of my understanding) superfluous because the client key is the same as the armor key, which the PA-OTP-REQUEST is already encrypted in.  So if OTP only covered scenario 1 that whole element of the protocol would probably be removed.

greghudson:
Just to clarify, I think the protocol data units in the oauth mech should be:

* KDC padata to client in hint list: just the URL and code, in whatever format is easiest for you.  Also set a cookie containing the RADIUS state.

* Client padata to KDC: an empty message.

And then set the reply key to the armor key like OTP does.

pbrezina:
Thank you for the thorough explanation.

> Just to clarify, I think the protocol data units in the oauth mech should be:
> 
>     * KDC padata to client in hint list: just the URL and code, in whatever format is easiest for you.  Also set a cookie containing the RADIUS state.

Is there any reason to not send RADIUS state directly in padata and let it be returned by the client?

greghudson:
> Is there any reason to not send RADIUS state directly in padata and let it be returned by the client?

That would work, although I think using the cookie is more in keeping with the RFC 6113 architecture.

pbrezina:
Hi Greg, thank you. I put state inside a cookie. I have functional code for the plugins [see here](https://github.com/pbrezina/sssd/tree/otp/src/krb5_plugin/oauth2). It is not yet fully complete since it lacks some glue (currently the oauth2 preauth is available to all principals) and I need to improve data encoding but I think it is in enough shape to discuss our next steps.

- What do we need to provide to register the PA TYPE?
- Should we use prefixed names for the plugin name, question and auth indicator (`sssd_oauth2`) or should we use just generic `oauth2`?

- Shouldn't `pa` be freed in the end of edata?
https://github.com/krb5/krb5/blob/e89abc2d4ea1fea1ec28d470f297514b828e4842/src/plugins/preauth/otp/main.c#L250-L262

Thank you, Pavel.

greghudson:
> Shouldn't pa be freed in the end of edata?

No; see the documentation comment for edata_respond_fn in kdcpreauth_plugin.h.  (I get this is confusing; responsibility for memory gets tricky in async C programming and we didn't necessarily do a great job of it here.)

> Should we use prefixed names for the plugin name, question and auth indicator (sssd_oauth2) or should we use just generic oauth2?

I don't have an opinion here.

> What do we need to provide to register the PA TYPE?

First, you will need to make a writeup of the preauthentication protocol.  Don't focus on KDC implementation details (such as the use of RADIUS).  Ask me to take a look at the writeup before proceeding.

Then, you will submit a request at https://www.iana.org/form/protocol-assignment .  This will ask four questions:

* I'm not sure what "what type of assignment" is asking; perhaps just "numeric".
 
* For "which registry": Kerberos Pre-authentication and Typed Data (https://www.iana.org/assignments/kerberos-parameters/kerberos-parameters.xhtml#pre-authentication)

* For "additional information" give a URL to the writeup.

kaduk:
> I'm not sure what "what type of assignment" is asking; perhaps just "numeric".

some registries make a distinction between "permanent" and "provisional" registrations (and temporary registrations are sometimes also possible, though not, I think, here).  I can try to get the form's instructions clarified.

pbrezina:
Thank you for all the support. I'm closing this PR in favor of https://github.com/SSSD/sssd/pull/5762. Now we just need to finish the padata type registration (in progress) and we are good to go.

PR #1206, opened 2021-08-03
Add more dump.c bounds checks

greghudson:
Although dump files are privileged inputs, the code to read them
should not admit integer overflows.  Add bounds checks for several
fields which are used as allocation lengths or are assigned to
structure fields of smaller size and different signedness.  Reported
by Sharwan Ram and Kihong Keo.


frozencemetery (approved):
Sure, this is fine.

PR #1207, opened 2021-08-06
Atomic ccache refresh

greghudson:
Here is some work on ticket 7707, which is the long-standing problem that refreshing a ccache creates a window of time where krb5_get_credentials() will fail.  As Michael Osipov observed, this can create a problem for concurrent use of a ccache with a client keytab.

The approach here is the fourth alternative described in ticket 7707: make krb5_cc_move() atomic when possible, and make kinit-like activities write to a memory ccache.  Various complications, notes, and TBD items:

* There were some locking calls in krb5_cc_move(), but they just lock process-global mutexes, so I'm not sure what they're supposed to accomplish; anything that could happen to a ccache concurrently within a process could happen within another process.  I didn't preserve them.

* I included a test program to exercise the refresh issue using two subprocesses.  With FILE it fails pretty readily without the changes, and passes with them.  This program won't work for testing MEMORY.  I experimented with adding a threaded mode, but couldn't produce a failure with MEMORY, perhaps because there isn't much thread yielding going on without I/O operations.

* Right now there's just atomic replace support for FILE and (trivially) DIR.  MEMORY should be pretty easy, and could be important for a multithreaded process using a client keytab and a memory ccache (which is a configuration described in appl_servers.rst).  KCM support should be fairly straightforward, with the minor wrinkle that we'll need to call back into the non-atomic move implementation (currently not exported) if the KCM server doesn't support the new opcode.  KEYRING unfortunately might be difficult or impossible.

* On the kinit-like-operations side of things, right now I've only hit get_in_tkt.c which should cover kinit and client keytabs.  Other possible targets include gss_store_cred(), krb5_get_renewed_creds(), and kvno --out-cache.

* For FILE, the current strategy is to serialize the whole ccache and write it out with a single write().  I think that's fine, but we could switch to writing out each cred separately, perhaps using a stdio wrapper.

* krb5_cc_move() obeys the source-before-destination convention before copying.  However, ccache methods ordinarily put the to-be-operated-on ccache just after the context.  So I gave atomic_replace() the opposite ccache parameter order from krb5_cc_move().  Neither option feels particularly safe or right.

* I can see an argument that we should have a new API krb5_cc_replace() that doesn't delete the source ccache.  That would work better for gss_store_cred().  We could leave krb5_cc_move() alone or reimplement it on top of krb5_cc_replace().  Not sure if it's worth increasing the API surface area.
 
* The ccache vtable had an entry for move, but it wasn't used, so I repurposed that slot for atomic_replace.

### DIFF
--- a/src/lib/krb5/ccache/ccbase.c
+++ b/src/lib/krb5/ccache/ccbase.c
@@ -348,50 +348,103 @@ krb5int_cc_typecursor_free(krb5_context context, krb5_cc_typecursor *t)
     return 0;
 }
 
+krb5_error_code
+k5_nonatomic_replace(krb5_context context, krb5_ccache ccache,
+                     krb5_principal princ, krb5_creds **creds)
+{
+    krb5_error_code ret;
+    int i;
+
+    ret = krb5_cc_initialize(context, ccache, princ);
+    for (i = 0; !ret && creds[i] != NULL; creds++)
+        ret = krb5_cc_store_cred(context, ccache, creds[i]);
+    return ret;
+}
+
+static krb5_error_code
+read_creds(krb5_context context, krb5_ccache ccache, krb5_creds ***creds_out)
+{
+    krb5_error_code ret;
+    krb5_cc_cursor cur = NULL;
+    krb5_creds **list = NULL, *cred = NULL, **newptr;
+    int i;
+
+    *creds_out = NULL;
+
+    ret = krb5_cc_start_seq_get(context, ccache, &cur);
+    if (ret)
+        goto cleanup;
+
+    /* Allocate one extra entry so that list remains valid for freeing after
+     * we add the next entry and before we reallocate it. */
+    list = k5calloc(2, sizeof(*list), &ret);
+    if (list == NULL)
+        goto cleanup;
+
+    i = 0;
+    for (;;) {
+        cred = k5alloc(sizeof(*cred), &ret);
+        if (cred == NULL)
+            goto cleanup;
+        ret = krb5_cc_next_cred(context, ccache, &cur, cred);
+        if (ret == KRB5_CC_END)
+            break;
+        if (ret)
+            goto cleanup;
+        list[i++] = cred;
+        list[i] = NULL;
+        cred = NULL;
+
+        newptr = realloc(list, (i + 2) * sizeof(*list));
+        if (newptr == NULL) {
+            ret = ENOMEM;
+            goto cleanup;
+        }
+        list = newptr;
+        list[i + 1] = NULL;
+    }
+    ret = 0;
+
+    *creds_out = list;
+    list = NULL;
+
+cleanup:
+    if (cur != NULL)
+        (void)krb5_cc_end_seq_get(context, ccache, &cur);
+    krb5_free_tgt_creds(context, list);
+    free(cred);
+    return ret;
+}
+
 krb5_error_code KRB5_CALLCONV
 krb5_cc_move(krb5_context context, krb5_ccache src, krb5_ccache dst)
 {
-    krb5_error_code ret = 0;
+    krb5_error_code ret;
     krb5_principal princ = NULL;
+    krb5_creds **creds = NULL;
 
     TRACE_CC_MOVE(context, src, dst);
-    ret = k5_cccol_lock(context);
-    if (ret) {
-        return ret;
-    }
-
-    ret = k5_cc_lock(context, src);
-    if (ret) {
-        k5_cccol_unlock(context);
-        return ret;
-    }
 
     ret = krb5_cc_get_principal(context, src, &princ);
-    if (!ret) {
-        ret = krb5_cc_initialize(context, dst, princ);
-    }
-    if (ret) {
-        k5_cc_unlock(context, src);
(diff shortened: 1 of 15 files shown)


greghudson:
> krb5_cc_move() obeys the source-before-destination convention before copying. However, ccache methods ordinarily put the to-be-operated-on ccache just after the context. So I gave atomic_replace() the opposite ccache parameter order from krb5_cc_move(). Neither option feels particularly safe or right.

One option here is for ccbase.c to take responsibility for reading the credentials from the source cache and marshalling them into an array.  Then the replace method would take (context, cache, princ, creds) arguments, and would be substantially simplified because it would have four fewer error checks.  This would create additional churn if we run into trouble with keeping all of the source creds in memory, but I think that's unlikely.

greghudson:
I coded up the idea above, and added a MEMORY implementation of the replace method.  I think I like it on balance.

greghudson:
I added a KCM implementation, though not going to be of very much use unless sssd implements a back end for it.

Note: this would be the first KCM call where a marshalled principal (not cred) doesn't appear at the end of the arguments or returned data.  When we did that for credentials (in OP_GET_CRED_LIST), we put a length before each credential, and I did the same here for the credential list in the arguments.  Without that, the lack of a remainder return in krb5_unmarshal_credentials() would get in the way of sssd parsing the credentials.

We haven't added APIs for marshalled principals, and they are simpler than credentials (though not trivial; it's {nametype, ncomps, {realmlen, realm}, ncomps*{complen, comps}).

greghudson:
I've hit most of the other cache refresh operations (deferring gss_store_cred() due to ticket 8010).  I think this is now ready for review and merge, so I've given it an appropriate commit message and removed the WIP tag.

src/lib/krb5/ccache/ccbase.c
```diff
@@ -348,50 +348,101 @@ krb5int_cc_typecursor_free(krb5_context context, krb5_cc_typecursor *t)
...
+        ret = krb5_cc_next_cred(context, ccache, &cur, cred);
+        if (ret == KRB5_CC_END)
+            break;
+        if (ret)
+            goto cleanup;
+        list[i++] = cred;
+        list[i] = NULL;
+        cred = NULL;
+
+        newptr = realloc(list, (i + 2) * sizeof(*list));
```
frozencemetery, line 396: The way this is written, on success, list is always one element longer than it needs to be.  (That is, it ends with two NULLs.)  Very likely this isn't worth worrying about - but I don't see a reason we need it, either.

src/lib/krb5/krb/get_in_tkt.c
```diff
@@ -1604,6 +1604,61 @@ warn_des3(krb5_context context, krb5_init_creds_context ctx,
...
+        ret = krb5_cc_set_config(context, mcc, ctx->cred.server,
+                                 KRB5_CC_CONF_FAST_AVAIL, &yes);
+        if (ret)
+            goto cleanup;
+    }
+
+    ret = save_selected_preauth_type(context, mcc, ctx);
+    if (ret)
+        goto cleanup;
+    ret = save_cc_config_out_data(context, mcc, ctx);
```
frozencemetery, line 1642: Missing blank line here?

src/lib/krb5/krb/get_in_tkt.c
```diff
@@ -1604,6 +1604,61 @@ warn_des3(krb5_context context, krb5_init_creds_context ctx,
...
+    ret = krb5_cc_move(context, mcc, out_ccache);
+    if (ret)
+        goto cleanup;
+    mcc = NULL;
+
+cleanup:
+    if (mcc != NULL)
+        krb5_cc_destroy(context, mcc);
+    return ret;
+
```
frozencemetery, line 1659: Extra whitespace here.

src/tests/kcmserver.py
```diff
@@ -151,13 +153,13 @@ def op_retrieve(argbytes):
...
+    cprinc, rest = extract_princ(rest[8:])
+    sprinc, rest = extract_princ(rest)
     cache = get_cache(name)
     for cred in (cache.creds[u] for u in cache.cred_uuids):
-        cred_cprinc, rest = unmarshal_princ(cred)
-        cred_sprinc, rest = unmarshal_princ(rest)
-        if cred_cprinc == cprinc and cred_sprinc == sprinc:
+        cred_cprinc, rest = extract_princ(cred)
+        cred_sprinc, rest = extract_princ(rest)
+        if (princ_eq(cred_cprinc, cprinc) and princ_eq(cred_sprinc, sprinc)):
```
frozencemetery, line 162: Outermost parentheses aren't needed.

frozencemetery (changes requested):
Minor stuff inline.  Otherwise looks good to me.

src/lib/krb5/ccache/ccbase.c
greghudson: If the realloc() fails, we need `list` to still be valid for freeing with krb5_free_tgt_creds(), which means we need the terminating null to already be present.  This probably deserves a comment; I can add one.

frozencemetery (approved):
Thanks!

PR #1208, opened 2021-08-10
Fix verification of PAC KDC signature issued by RODC

iboukris:
We'll need the same for PAC ticket-signature, we could also check that they have the same checksum type and the same RODCIdentifier since the same key is supposed to be used.


iboukris:
This is useful as is for wireshark to check the kdc signature correctly, unwiped.

PR #1213, opened 2021-08-26
conf_ldap.rst: Update links OpenLDAP 2.4 → 2.5

dilyanpalauzov:


greghudson:
How about this instead?

dilyanpalauzov:
Okay.

PR #1218, opened 2021-09-20
Restore OpenSSL 1.0 compatibility in softpkcs11

greghudson:
Commit 00de1aad7b3647b91017c7009b0bc65cd0c8b2e0 used
OPENSSL_clear_free(), which was added in OpenSSL 1.1.  Use
OPENSSL_cleanse() and OPENSSL_free() instead.


frozencemetery (approved):
Sorry about that.

PR #1219, opened 2021-09-29
Always use platform PRNG

greghudson:
Remove the fortuna and device PRNG modules and PRNG modularity, and
move the prng_os implementation into prng.c.  Remove the crypto_mod.h
requirement to implement failure-free AES256 and SHA256.  Deprecate
krb5_c_random_add_entropy() and krb5_c_random_os_entropy() and remove
their call sites.  Deprecate and ignore the -W (weak random) option to
kadmind and kdb5_util create, and stop using it in the test suite.


frozencemetery (approved):

src/lib/crypto/krb/prng.c
```diff
@@ -127,8 +124,30 @@ k5_get_os_entropy(unsigned char *buf, size_t len, int strong)
         return TRUE;
 #endif /* defined(__linux__) && defined(SYS_getrandom) */
 
-    device = strong ? "/dev/random" : "/dev/urandom";
-    return read_entropy_from_device(device, buf, len);
+    return read_entropy_from_device("/dev/urandom", buf, len);
```
kaduk, line 127: I confess I've lost track of which versions of linuxx will suffer the "/dev/urandom returns non-random bits on startup until the RNG is seeded" and don'to have the getrandom syscall ... maybe getrandom is old enough now that it's not a real concern.

frozencemetery: Since we don't use the glibc interface, that'll be true for Linux < 3.17.  For RHEL, this is RHEL 7+ (iirc due to backports); for Debian that's Stretch onward (possibly not Jessie); for Ubuntu it's 15.04+ (so possibly not 14.04 LTS).

greghudson: We don't (and can't) really solve this problem in the current code.  We have the option to read from /dev/random, but we use it so sparingly (because /dev/random also blocks when "entropy runs out" which is almost always) that we don't get anything approaching secure operation in the hypothetical case.

Since modern Linux systems provide a sane option and other Unix-like platforms have generally been more reasonable historically, my preference is to treat the problem as a platform issue and stop trying to take responsibility for it.  This is also the opinion I've seen from most educated people in the space (see https://www.2uo.de/myths-about-urandom/#experts ).

PR #1221, opened 2021-09-30
Support for MacOS X Native credential cache

kenh:
As discussed previously on krbdev, this is a new credential cache implementation for MacOS X that dynamically loads the MacOS X Kerberos framework and makes calls directly into the credential cache functions.  This is necessary to interoperate with the operating system Kerberos implementation on Big Sur and above since the default MacOS X Kerberos cache has switched to using a different IPC mechanism than the KCM-compatible protocol which uses Mach ports.

I will admit that this code has a few implementation warts, but I believe it works well and we have deployed to our user base. Some notes:

- It uses a credential cache name of `APPLE`. I will be the first to admit this is not wonderful. I didn't quite feel comfortable taking over `API` since that was already in use. Suggestions welcome here.
- It works best if you set `DEFCCNAME=APPLE:` when running `configure`
- There doesn't seem to be a way to say "don't enable this credential cache implementation" if initialization fails (like if you can't load the OS Kerberos framework). At least it wasn't obvious to me.

I am open to feedback and suggestions for modifications.


greghudson:
From a look at the Heimdal fork at opensource.apple.com, Apple's new ccache type is called XCACHE, and (as built on Snow Leopard) is also called API.  So, if you resolve API: in the native implementation and use it, it doesn't go through the ccapi.

From the conversation in March, I hadn't expected to see code going through ccapi; I had expected to go through the Heimdal libkrb5 and translate the resulting error codes and data structures.  I can see how using ccapi is more appealing; it's a stable cross-platform ABI (though it's kind of alien to either MIT krb5 or Heimdal and it was designed when krb4 was still a thing) and we already have the declarations for its data types in the tree.  Also, if it's what you get by loading Kerberos.framework/Kerberos then perhaps it's harder to get at the Heimdal libkrb5 functions.  But we do want to be aware that Apple could remove the ccapi surface without breaking their native Kerberos and GSS implementation.

A few entirely incomplete technical notes:

* There's no problem using dlopen() directly on single-platform code, and therefore no need for a comment.  (Using macOS-specific interfaces would also be fine, but only if it carries a benefit.)  If Apple has RTLD_NODELETE, it might be a good idea to use it as it might minimize leakage when our libkrb5 is repeatedly loaded and unloaded.

* In new code I prefer to use C99 stdint.h type names for fixed-width types even when there's a legacy fixed-width type name, so int32_t instead of cc_int32 or krb5_int32.  (Though I haven't started using uint32_t instead of OM_uint32 in GSS code.)

* Inside libkrb5 there is no reason to strdup() strings returned by libkrb5 just so we can safely free them with free().  That's what krb5_free_unparsed_name() (or whatever) would do anyway.

* We have implementations of some of these data translations already inside lib/krb5/ccache/ccapi (like copy_krb5_creds_to_cc_cred_union() which is like mit2applecreds()), although we currently only build that whole directory on Windows.  (We did build it for KfM, back in the day.)  Ideally I'd like to avoid duplication, but you can let me work out the details of that.

kenh:
> From a look at the Heimdal fork at opensource.apple.com, Apple's new ccache type is called XCACHE, and (as built on Snow Leopard) is also called API. So, if you resolve API: in the native implementation and use it, it doesn't go through the ccapi.

I had looked at that, but it was ... a bit complicated to follow everything?

> From the conversation in March, I hadn't expected to see code going through ccapi; I had expected to go through the Heimdal libkrb5 and translate the resulting error codes and data structures. I can see how using ccapi is more appealing; it's a stable cross-platform ABI (though it's kind of alien to either MIT krb5 or Heimdal and it was designed when krb4 was still a thing) and we already have the declarations for its data types in the tree. Also, if it's what you get by loading Kerberos.framework/Kerberos then perhaps it's harder to get at the Heimdal libkrb5 functions. But we do want to be aware that Apple could remove the ccapi surface without breaking their native Kerberos and GSS implementation.

I guess that was never explicitly discussed, but I had always in my mind planned on using the ccapi. One big reason is that the Heimdal krb5 doesn't have all of the ccapi functionality like krb5_cc_switch() (or if it does it is non-obvious to me). And I'm aware that the ccapi functions are officially deprecated, but so are all of the krb5 API functions. I'm not sure there is a wonderful long-term solution right now short of implementing XCACHE.

>     * There's no problem using dlopen() directly on single-platform code, and therefore no need for a comment.  (Using macOS-specific interfaces would also be fine, but only if it carries a benefit.)  If Apple has RTLD_NODELETE, it might be a good idea to use it as it might minimize leakage when our libkrb5 is repeatedly loaded and unloaded.

I think a MacOS-specific interface would require a CoreFoundation dependency which didn't seem like a great idea. RTLD_NODELETE is supported, so I can add that.

>     * In new code I prefer to use C99 stdint.h type names for fixed-width types even when there's a legacy fixed-width type name, so int32_t instead of cc_int32 or krb5_int32.  (Though I haven't started using uint32_t instead of OM_uint32 in GSS code.)
>     * Inside libkrb5 there is no reason to strdup() strings returned by libkrb5 just so we can safely free them with free().  That's what krb5_free_unparsed_name() (or whatever) would do anyway.

Fair enough. Will modify.

>     * We have implementations of some of these data translations already inside lib/krb5/ccache/ccapi (like copy_krb5_creds_to_cc_cred_union() which is like mit2applecreds()), although we currently only build that whole directory on Windows.  (We did build it for KfM, back in the day.)  Ideally I'd like to avoid duplication, but you can let me work out the details of that.

I was a LITTLE hesitant to use that code since it at most supported ccapi v3, and it seems like MacOS X is up to ccapi v7.

greghudson:
(Sorry for the long delay while I worked on the PAC stuff.)

> I was a LITTLE hesitant to use that code since it at most supported ccapi v3, and it seems like MacOS X is up to ccapi v7.

After looking into this in more detail, I think ccapi up to v7 is actually reflected in MIT's code base.  For example, include/CredentialsCache.h has `ccapi_version_max = ccapi_version_7`, and stdcc.c has a `gCCVersion >= ccapi_version_5` case.  It's just that the v2/v3 boundary seems to have been a much more traumatic change, and therefore has much more obvious versioning.

My earlier assumption had been that I should pull the stdcc_util.c conversion functions up a directory and use them in both stdcc.v and cc_apple.c.  But after I started work on that, I found that even the main cc_apple.c logic duplicates stdcc.c in all respects except the use of dlopen().  (Well, and coding practices, which have experienced a lot of drift since that code was last visited.)

This discovery makes me curious: if we resurrect our own CCAPI client library on modern macOS, will it talk to the system CCAPI server?  I actually have no idea how the KfM IPC code worked or if Apple changed that.  I can put together a candidate commit for that if it seems worth testing.  (I am also looking into options for getting my own test environment.  Unfortunately it's a lot harder to obtain remote macOS shell access than it is for Linux.)

kenh:
> (Sorry for the long delay while I worked on the PAC stuff.)

No problem, I've been too busy to get back to this as well.

> After looking into this in more detail, I think ccapi up to v7 is actually reflected in MIT's code base. For example, include/CredentialsCache.h has `ccapi_version_max = ccapi_version_7`, and stdcc.c has a `gCCVersion >= ccapi_version_5` case. It's just that the v2/v3 boundary seems to have been a much more traumatic change, and therefore has much more obvious versioning.

Fair enough, this was not clear to me.  Also, as you note that code has a number of #ifdefs and it seemed like a bit of a mess and not something I wanted to chew up to get working.

> My earlier assumption had been that I should pull the stdcc_util.c conversion functions up a directory and use them in both stdcc.v and cc_apple.c. But after I started work on that, I found that even the main cc_apple.c logic duplicates stdcc.c in all respects except the use of dlopen(). (Well, and coding practices, which have experienced a lot of drift since that code was last visited.)

Weeelll ... I discovered that there a few weirdnesses with the current ccapi implementation on MacOS X that maybe make this not impossible, but will definitely result in more #ifdefs. For example, see my comments in apple_cc_initialize().

> This discovery makes me curious: if we resurrect our own CCAPI client library on modern macOS, will it talk to the system CCAPI server? I actually have no idea how the KfM CCAPI IPC code worked or if Apple changed that. I can put together a candidate commit for that if it seems worth testing. (I am also looking into options for getting my own test environment. Unfortunately it's a lot harder to obtain remote macOS shell access than it is for Linux.)

My only goal is to get a credential cache working with native MacOS X Kerberos. If that's how it happens, that is fine with me.

greghudson:
I didn't have a lot of luck figuring the old KfM build system, but I do have the API cache type half-workingby simply building the ccapi subdir and then linking against Apple's with "-framework Kerberos" in the libkrb5 link line.  I'll need to check if there are any unwanted namespace side-effects to doing that, but it seems simpler than using dlopen().

> I discovered that there a few weirdnesses with the current ccapi implementation on MacOS X that maybe make this not impossible, but will definitely result in more #ifdefs. For example, see my comments in apple_cc_initialize().

I confirmed this issue, and it is slightly thorny as the workaround seems likely to break on Windows if not ifdef'd (based on my read of our CCAPI library and server code).  The other weirdness I see a comment for is in apple_gen_new(), where you parse X-DUMMY-USER, unparse it, and pass that to cc_create_new_ccache().  The existing stdcc.c passes "" here, and that seems to work as well as X-DUMMY-USER on macOS.

Heimdal's approach (in both upstream and the Apple fork) is to defer setting the ccache name in gen_new(), and instead generate the name at initialize time with cc_context_create_new_ccache().  If krb5_cc_initialize() is called on an already existing cache, it manually reinitializes it by deleting all of its creds and resetting the default principal.  If the caller invokes krb5_cc_get_name() before initializing the cache, it creates a new ccache with the default principal name (generally the system username).  That's not terribly elegant and I don't think I want to migrate to that design, but it does mean the native Apple ccache type never uses cc_create_ccache() and therefore doesn't run into the unexpected cache name semantics.

I will attempt to iron out the remaining issues I'm aware of, submit an alternate PR, and ask you to test it.

greghudson:
I had previous missed the issue of getting the default name.  On macOS, ccapi is implemented in MITKerberosShim, which translates the CCAPI operations back to krb5_cc operations using a krb5 context.  Instead of returning the default name for the API collection, cc_get_default_ccache_name() may return the configured default ccname (with "API:" removed if that was the prefix) as set by KRB5CCNAME or [libdefaults] default_cc_name.

The PR works around this by temporarily unsetting KRB5CCNAME in the environment.  That approach isn't thread-safe, might leak memory[1], and doesn't address default_cc_name (although, due to a historical botch on my part, this happens to be a different variable than MIT krb5 uses for the same purpose, default_ccache_name).  Unfortunately there aren't any easy alternatives.  Even if we were calling into the Heimdal API and not the CCAPI, there appears to be no way to get at a ccache type's default_name method without first checking the environment and krb5.conf.

To do this right we might have to duplicate a slice of the XCACHE implementation--just enough to make the RPC to get the default ccache name.  This might not be a lot of code since it uses relatively simple types.  I will look into it.  This might be enough additional variance to justify a separate ccache implementation from stdcc.c, although I would still want to share the CCAPI/MIT credential conversion functions (having already done previous work in that direction).

I also saw a problem where stdcc.c's resolve method tries to open the cache immediately; if it doesn't already exist, cc_context_open_ccache() returns KRB5_CC_END instead of the expected ccErrCCacheNotFound, so the caller gets back KRB5_FCC_INTERNAL.  The PR code doesn't run into this problem because it defers opening the cache.

[1] Historically, implementations of setenv() and unsetenv() can leak memory if the same variable is assigned multiple times, because libc doesn't know if the previous setting was heap-allocated by a previous setenv() call, placed there by putenv(), or pre-existing at process startup.  Some libc implementations have worked around this problem, such as by copying the whole process environment to heap-allocated memory on the first setenv() call.  I don't know whether macOS has done that.

kenh:
> I had previously missed the issue of retrieving the default ccache name. On macOS, ccapi is implemented in MITKerberosShim, which translates the CCAPI operations to Heimdal krb5_cc operations. Instead of returning the default name for the API collection, cc_get_default_ccache_name() may return the configured default ccname (with "API:" removed if that was the prefix) as set by KRB5CCNAME or [libdefaults] default_cc_name.
> 
> The PR works around this by temporarily unsetting KRB5CCNAME in the environment. That approach isn't thread-safe, might leak memory[1], and doesn't address default_cc_name (although, due to a historical botch on my part, this happens to be a different variable than MIT krb5 uses for the same purpose, default_ccache_name). Unfortunately there aren't any easy alternatives. Even if we were calling into the Heimdal API and not the CCAPI, there appears to be no way to get at a ccache type's default_name method without first checking the environment and krb5.conf.

Ah, yes, I forgot about that additional wart.  And I hadn't realized that changing the environment wasn't thread-safe (but now that I think about it, it is obvious).

Obviously, whatever direction you want to go is fine with me; my goal is to get a workable Apple-compatible credential cache implementation into MIT Kerberos so we have one less local patch to deal with.

greghudson:
There needs to be a license statement on cc_apple.c, and I can't add that since it's your copyright.  You can use the comment from src/prototype/prototype.c as a basis, with your name in place of MIT's.

greghudson:
I've pushed a bunch of changes.  Please test this (no special build options should be required, similar to how KCM support works by default on older macOS versions), and add a copyright statement to cc_api_macos.c as noted above.

kenh:
My apologies for ghosting you on this; my personal and professional life has been very busy recently. I will work on these things this week.

kenh:
So, I FINALLY got some time to get back around to test this.  Unfortunately, it does not work for me.

I didn't realize at first this is only enabled for Big Sur and above; unfortunately, this does not work in our environment, as we need to build a single kit that works for everything from Mojave and up. I applied some autoconf changes to enable it for earlier systems, but then it fails because it gets the wrong identifier for the default credential cache. So ... I'm not sure how to go forward from there, because I'm not quite sure what get_primary_uuid() is doing.

kenh:
Alright, after digging into this (and re-reading your comments, because I forgot a lot of the details in the intervening time) ... it seems the issue is somewhere the XPC calls are failing and are returning the default UUID, which doesn't work on Catalina because the default cache name is API:%{uid} (if you set KRB5CCNAME to the correct value, it works fine).

So I am wondering if a possible solution is to fall back to using cc_context_get_default_ccache_name() like my original code did if the XPC calls fail or return the default UUID. The man page for setenv() suggests that memory leaks are not an issue. I realize this isn't perfect, but it would really be incredibly useful to have one set of binaries which worked across a wide variety of versions of MacOS X. I'd also be fine if that fallback was autoconf-configurable. I can code that up if you want.

greghudson:
In theory we could fall back to KCM at runtime for macOS 10 and below, right?  (Perhaps by checking if the XPC calls succeed.)  I can try to put together a candidate for that.

kenh:
One thing worth mentioning is that the KCM code isn't quite as functional as the native credential cache code.

Specifically, things like ```kswitch``` don't seem to work right under KCM. I mean, it's not a dealbreaker since it hasn't quite been working for a while, but I guess I was hoping that if the default was the native credential cache functions then we'd gain some functionality that was missing.

kenh:
Fair enough; I acknowledge that the primary goal is to get the credential cache support into MIT Kerberos, and I can live with things not working perfectly with KCM.

greghudson:
Okay, try this.  I realized we only need a KCM fallback for getting the primary cache name, and we can use the CCAPI for everything else.

kenh:
Alright, so ... I realized I was seeing some strange stuff, so I went back and built this code _on_ Catalina but tested the binaries across a bunch of clean VMs all the way from Mojave to Monterey. I then used the MacOS X ```kinit``` command to get a TGT and then tried to use the MIT Kerberos ```klist``` command.

The short answer is ... the code does not work on anything below Big Sur.

The REASON it doesn't is because it always gets a valid UUID from that RPC call, even back in Mojave. I added a line to get_primary_id to print out the result.  Here's what I see on Mojave (the first ```klist``` command is stock OS X Kerberos):

```
$ klist
Credentials cache: API:E2C64ADF-972C-4122-9D40-7D02B7FAC478
        Principal: kenh@CMF.NRL.NAVY.MIL

  Issued                Expires               Principal
May 13 13:24:53 2022  May 13 23:24:49 2022  krbtgt/CMF.NRL.NAVY.MIL@CMF.NRL.NAVY.MIL
test-mojave:src kenh$ /tmp/krb5/bin/klist
Default cache name: C8B30603-9F32-4949-AA41-9D549D2802C6
klist: No credentials cache found
```

And it turns out when I said the default credential cache on Catalina was ```API:%{uid}```, that was incorrect. I think what is happening there is the Kerberos kit we use ends up using that as a default name and then changes the default credential cache to that, that's why I wanted to test on a clean VM.

I changed ```get_primary_name()``` to always use KCM to get the default credential cache name, and that made it work on earlier releases of MacOS X, but that made it fail on post-Catalina:

```
% /tmp/krb5/bin/klist

Default cache name: 501
klist: Internal credentials cache error
```

So, the options aren't WONDERFUL. It does occur to me you could try the KCM call FIRST and if that doesn't return something with a useful prefix, THEN try the XCACHE RPC API call. Or we could figure out a better way to determine that the XCACHE RPC API doesn't return something useful.

greghudson:
I think the best option given what we currently know is to call uname(), parse the major version of the release field with strtol(), and use XPC if that's at least 20 (the Darwin version with macOS 11), KCM if it's less.

I like that better than prioritizing the KCM query because it's faster and probably more robust--something could have set the KCM primary cache to a uuid.

kenh:
If you're happy with that, then so am I. I can try to code that up and test it (and add that license to the file).

kenh:
Ok, I've just pushed a change which implemented the code we talked about. I also added a license statement to that file.

If this could get into 1.20 that sure would be great. I fully acknowledge I was the reason this has been delayed for so long so if it can't make it I will totally understand.

kenh:
Oh, some notes. I tested this on everything from High Sierra up to and including Monterey. I used the OS ```kinit``` to get a ticket, verified I could see it with the MIT ```klist```, then I used the MIT ```kinit``` with a different principal to verify it could write to the cache and would correctly create a new one. I then verified I could use the MIT ```kswitch``` to switch to the original credential cache. Worked fine across all OS versions.

greghudson:
I re-squashed the commits, edited some  comments, and did some license statement bookkeeping.  I don't think I changed the code enough to warrant re-testing (the only change is to pull out some inner-scope variable declarations since we don't generally use them).  I expect to merge this soon.

1.20 is due out in a week, roughly six months behind what was the normal schedule.  I can't reasonably add a change of this size this close to the release or further delay the release.

bukka:
@kenh @greghudson Not sure if you are aware but this change seems to caused some issues to Mac users running krb5 in forked processes. In our case it is a usual mode how to run PHP which appens when PHP-FPM is used for example. In this case the krb5 is used by pglib client that is used by PHP child process and then crashes because krb5 most likely calls some of the not unsafe functions. Specifically this error is emitted:
```
+[__NSPlaceholderDate initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```
You can find more info in the discussions linked above.

Would you have some suggestion how to best resolve this issue? Is there anything that you can do on your side (e.g. not using some fork unsafe calls)?

greghudson:
> Not sure if you are aware but this change seems to caused some issues to Mac users running krb5 in forked processes.

I reviewed the linked threads and put my thoughts in https://github.com/php/php-src/issues/11818#issuecomment-1694570346 .

kenh:
As a followup; I agree with everything Greg said in that linked issue. Here are some meta-thoughts:

- If you really desire Apple credential cache integration ... well, there are unfortunately no good answers. This code was put into place because on Big Sur and above the Apple credential cache implementation changed the RPC protocol and there isn't really a way of dealing with this aside from calling the Apple Kerberos.framework code (which is what we do) or implementing a pile of RPC stuff ourselves (which I suspect would also require libdispatch).
- I'm unclear what the exact triggering scenario is; if a simple process calls `fork()` and then does libdispatch stuff, does that trigger this issue?
- As Greg mentioned, if you do NOT require Apple credential cache integration then using a different credential cache type (like FILE if you require credential cache sharing across processes, or MEMORY if you do not) is always an option.

bukka:
@kenh Thanks for the follow up. It' just a simple `fork`. PHP-FPM is just creating worker processes that serve that fcgi requests. So it needs to use `fork` for that. The problem for us (PHP) is that krb5 is accessed indirectly through libpq. I'm not sure if it would be a good idea to change the credential cache globally (meaning let brew to always disable it) as I guess there are some use cases for that when used by other apps, right? But maybe having some way to disable in the client should be sufficient. Then if libpq got extended to let us change it as well, it should resolve the issue eventually.

PR #1222, opened 2021-10-04
include:gssrpc: Fix prototype definitions

cryptomilk:
```
gssrpc/svc.h:296:1: error: function declaration isn’t a prototype
[-Werror=strict-prototypes]
  296 | extern void rpctest_service();
```

The Samba developer build is setting `-Werror=strict-prototypes` and fails to build with mit-krb5 master.


cryptomilk:
The problem also exists in xdr.h for `xdrproc_t`. I think to fix this you would need to define more prototypes and use a union.

greghudson:
Upon revisiting this, I noticed that the two declarations changed in the PR were for unused and unimplemented symbols.  I found some other GSSRPC symbols in that category (by searching for everything defined in rename.h) and made an alternative commit to remove them.

cryptomilk:
:+1: Thanks!

PR #1229, opened 2021-11-08
kr_attrset_decode: make sure length byte is unsigned

sumit-bose:
To avoid ambiguity the length byte in the Radius attribute is should be treated as unsigned explicitly. Otherwise it might be possible that attributes longer than 125 characters will be rejected with EBADMSG.

A 253 characters long NAS-Identifier attribute is added to the tests to make sure that attributes with the maximal number of characters are working as expected.

### DIFF
--- a/src/lib/krad/attrset.c
+++ b/src/lib/krad/attrset.c
@@ -217,7 +217,7 @@ kr_attrset_decode(krb5_context ctx, const krb5_data *in, const char *secret,
 
     for (i = 0; i + 2 < in->length; ) {
         type = in->data[i++];
-        tmp = make_data(&in->data[i + 1], in->data[i] - 2);
+        tmp = make_data(&in->data[i + 1], (uint8_t)in->data[i] - 2);
         i += tmp.length + 1;
 
         retval = (in->length < i) ? EBADMSG : 0;
--- a/src/lib/krad/t_packet.c
+++ b/src/lib/krad/t_packet.c
@@ -57,6 +57,14 @@ make_packet(krb5_context ctx, const krb5_data *username,
     krb5_error_code retval;
     const krb5_data *data;
     int i = 0;
+    krb5_data nas_id;
+
+    nas_id = string2data("12345678901234567890123456789012345678901234567890"
+                         "12345678901234567890123456789012345678901234567890"
+                         "12345678901234567890123456789012345678901234567890"
+                         "12345678901234567890123456789012345678901234567890"
+                         "12345678901234567890123456789012345678901234567890"
+                         "123");
 
     retval = krad_attrset_new(ctx, &set);
     if (retval != 0)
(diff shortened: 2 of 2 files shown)


greghudson:
The first commit exposes a latent bug which can be seen by running the libkrad tests under asan.  This bug is currently hidden from the CI because the krad tests weren't functioning.  The test programs were exiting successfully because failing to start the daemon could just mean that pyrad isn't available; this is unfortunate but I will defer considering that for the moment.  Although this PR's main purpose has nothing to do with the latent bug, I do have to fix the latent bug before merging the first commit or we'll break CI runs.

The bug is manifesting in a call to krad_client_free() from t_client.c.  This bug probably does not affect the KDC since
the KDC doesn't call krad_client_free() during regular operation (and in fact, doesn't call it at all in 1.19; we only recently added a call to it at shutdown time in the mainline).

This krad_client_free() call is explicitly testing the teardown of outstanding requests.  t_client.c queues up a request to "localhost" and frees the client object without first running the verto context, so the request is guaranteed to still be present in the client object.

In a typical client configuration, "localhost" has two addresses (IPv6 and IPv4), so after the krad_client_send() call the client object contains a list of two remote structures, which I will call ipv6remote and ipv4remote.  These are linked from the client object in two ways: first, in the client object's `servers` field (in reverse order because it's a singly-linked list, so {ipv4remote, ipv6remote}), and second, in the callback data for the request in the `remotes` array (in forward order, so {ipv6remote, ipv4remote}).  The request is, in turn, linked from ipv6remote in its `list` field.  ipv4remote's `list` field is empty at this point since it would only be reached after a timeout.

When krad_client_free() runs, it iterates over the `servers` list and frees each remote object.  Freeing a remote object requires iterating over that object's requests, and invoking its callback with an ECANCELLED code.  The callback (always client.c:on_response()) iterates over the callback data's `remotes` array and calls kr_remote_cancel() to remove the packet from that remote's request list.

This recursive processing results in a use after free.  First ipv4server is freed, which is a trivial operation because its `list` field is empty.  Then while freeing ipv6server, we iterate over the `remotes` array {ipv4server, ipv6server} cancelling the packet; this references the already freed ipv4server.

I am not yet sure what change to make.  It seems like a design bug that there can be dangling references to a remote object from a client request's callback data after the remote object is deleted--though in practice, remote objects are only freed at shutdown or if they haven't been referenced for an hour, by which time any referencing requests should have completed or timed out.

sumit-bose:
Hi,

would it be acceptable to replace `localhost` with `127.0.0.1` in `t_client.c` to avoid the issue for the time being and revert the change when the issue is fixed?

bye,
Sumit

greghudson:
I came up with a reasonably conservative fix in PR #1230 (don't free any remotes before cleanup, cancel all requests before freeing any remotes during cleanup).  So we shouldn't need to change the test.

sumit-bose:
> I came up with a reasonably conservative fix in PR #1230 (don't free any remotes before cleanup, cancel all requests before freeing any remotes during cleanup). So we shouldn't need to change the test.

Thanks, so I wait until #1230 is committed and then update this PR to make the checks pass.

bye,
Sumit

greghudson:
I merged #1230 and did the rebase here, so you shouldn't have to take any action.

sumit-bose:
Thanks.

bye,
Sumit

PR #1231, opened 2021-12-02
add PA-REDHAT-IDP-OAUTH2 padata type

pbrezina:
This was just registered. It would be nice if krb5 could translate the number into a string in trace messages.


pbrezina:
I'm not sure what I'm supposed to do to fix the docs.

kaduk:
I think you need to add an entry to doc/appdev/refs/macros/index.rst (the new macro gets a new file generated by doxygen, and sphinx will complain if there is a file that is unreferenced).  What is done for the adjacent macros should provide a template for what is needed.  (Also, given that this is going to show up in the documentation, it might be worth adding a doxygen comment for it.)

pbrezina:
Thank you. Please, see the new version.

PR #1234, opened 2021-12-26
Allow init_sec_context() with no default realm

iboukris:
Needs cleanup.


iboukris:
Didn't find a cleaner way so I've just fixed the indentation.

PR #1235, opened 2021-12-30
Do not pass null to asprintf

nrensen:
Passing null to asprintf as a %s parameter causes noisy libc syslog warnings on OpenBSD.
I couldn't find any documentation indicating that passing null as a %.*s parameter was valid, even if the specified length is zero.

The proposed diff addresses one specific instance that was causing noise in my case. Perhaps it would be better to deal with this within split_trailer() or even in empty_data() by using an empty string rather than NULL.

### DIFF
--- a/src/kadmin/server/server_stubs.c
+++ b/src/kadmin/server/server_stubs.c
@@ -4,7 +4,7 @@
  *
  */
 
-#include <k5-platform.h>
+#include <k5-int.h>
 #include <socket-utils.h>
 #include <gssapi/gssapi.h>
 #include <gssapi/gssapi_krb5.h> /* for gss_nt_krb5_name */
@@ -219,6 +219,7 @@ static gss_name_t acceptor_name(gss_ctx_id_t context)
 static int gss_to_krb5_name(kadm5_server_handle_t handle,
                             gss_name_t gss_name, krb5_principal *princ)
 {
+    krb5_error_code ret;
     OM_uint32 minor_stat;
     gss_buffer_desc gss_str;
     int success;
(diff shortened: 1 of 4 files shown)


greghudson:
I'm pretty sure that passing null to a printf function for a `%.*s` conversion with zero length is undefined.  C99 7.19.6.1 says "the argument shall be a pointer to the initial element of an array of character type", and 7.1.4 says "If an argument to a function has an invalid value (such as... a null pointer...), ..., the behavior is undefined".  This is also true for memset() and memcpy(), which is a common complaint against the C standard--especially since the standard allows malloc(0) to null on success.  I think we unfortunately have a ton of violations of this aspect of the standard in the code base, though probably more often with memcpy() than the printf family.

I don't think we can change empty_data().  We frequently use empty_data() to initialize krb5_data values that own memory, and then unconditionally free() the data pointer (either directly or via krb5_free_data() or krb5_free_data_contents()).

Some projects wrap memcpy() and friends to address this problem (which I don't really like, but might do anyway).  We can't easily do that for printf functions because of the variable argument list, so we're stuck addressing the `%.*s` violations piecemeal, or perhaps reducing our use of asprintf() for concatenating counted strings.

For this particular violation, changing split_trailer() to use `make_data("", 0)` or `string2data("")` instead of empty_data() is probably more elegant than testing the pointer within the asprintf() call.  But I also want to think about alternatives to asprintf() for this operation.

greghudson:
I pushed an alternative commit addressing all of the asprintf() `%.*s` uses that could pass null pointers, and addressing the sn2princ case by changing split_trailer().

PR #1238, opened 2022-01-25
Clarify certauth interface documentation

greghudson:
[I told @cryptomilk I would try to clarify this documentation, particularly in regard to the princ parameter.]

Try to make it clearer that princ is the requested client principal,
not a principal decoded from the certificate, and that the module must
decode the certificate and inspect its attributes.  Document
KRB5_CERTAUTH_HWAUTH_PASS in certauth_plugin.h.


cryptomilk:
Thank you.

Hey @abbra and @sumit-bose you might also want to take a look.

PR #1239, opened 2022-02-02
Fix KDC DB propagation when larger than 4 GiB

jrisc:
When doing a full resync of the KDC database, `kprop` stores and sends the size of the dumped database on a 32 bits integer. As a consequence, any value bigger than 2<sup>32</sup>-1 (ie. >= 4GiB) wraps, resulting in a invalid smaller size.

I am not sure if this behavior could actually result in a truncated database being propagated without any error. It seems the deserialization will just fail if the invalid EOF happens at the middle of a line of the transmitted dumped DB file. But maybe it could happen in case invalid EOF occurs on a newline character. It does not looks like the database dump file contains any information about its length in the header.

This problem has been identified and reported to Red Hat by Keith Young from Morgan Stanley. He provided an initial fix I completed with tests, documentation, and some compatibility additions:

In order to remain compatible with previous versions of `kprop`, in case the size of the dumped KDC database is between 1 bit and 4GiB-1B, it is still encoded on 32 bits. The 0 value (which is impossible in practice) is used as an indicator that the size is stored on the 64 following bits, allowing a much larger size value. Hence a fixed `kprop` sending a > 4GiB dumped database to an older version of `kpropd` will see its size as 0, resulting in a header deserialization error (as mentioned in the additions to the documentation).

With this implementation of the compatibility approach described above, `kpropd` would consider valid an encoding where the first 32 bits integer is 0, but the next 64 bits one has a value smaller or equal to 4GiB. This should never happen given how encoding is implemented on the `kprop` side. Maybe such case should raise an error?

I mention the 4GiB limitation in `kprop` and `kpropd` manuals. Maybe this is not necessary? Please tell me what you think.

Also I wrote the documentation additions as if this fix was to be included into 1.20.0, but if I am being too optimistic, just let me know.

### DIFF
--- a/src/kprop/kprop_util.c
+++ b/src/kprop/kprop_util.c
@@ -92,3 +92,45 @@ sn2princ_realm(krb5_context context, const char *hostname, const char *sname,
     *princ_out = princ;
     return 0;
 }
+
+void
+encode_database_size(uint64_t size, krb5_data *buf)
+{
+    assert(buf->length >= 12);
+    if (size > 0 && size <= UINT32_MAX) {
+        /* Encode in 32 bits for backward compatibility. */
+        store_32_be(size, buf->data);
+        buf->length = 4;
+    } else {
+        /* Set the first 32 bits to 0 and encode in the following 64 bits. */
+        store_32_be(0, buf->data);
+        store_64_be(size, buf->data + 4);
+        buf->length = 12;
+    }
+}
+
+krb5_error_code
+decode_database_size(const krb5_data *buf, uint64_t *size_out)
+{
+    uint64_t size;
+
+    if (buf->length == 12) {
+        /* A 12-byte buffer must have the first four bytes zeroed. */
+        if (load_32_be(buf->data) != 0)
+            return KRB5KRB_ERR_GENERIC;
+
+        /* The size is stored in the next 64 bits.  Values from 1..2^32-1 must
+         * be encoded in four bytes. */
+        size = load_64_be(buf->data + 4);
+        if (size > 0 && size <= UINT32_MAX)
+            return KRB5KRB_ERR_GENERIC;
+    } else if (buf->length == 4) {
+        size = load_32_be(buf->data);
+    } else {
+        /* Invalid buffer size. */
+        return KRB5KRB_ERR_GENERIC;
+    }
+
+    *size_out = size;
+    return 0;
+}
--- a/src/kprop/kprop.c
+++ b/src/kprop/kprop.c
@@ -25,6 +25,7 @@
  */
 
 #include "k5-int.h"
+#include <inttypes.h>
 #include <locale.h>
 #include <sys/file.h>
 #include <signal.h>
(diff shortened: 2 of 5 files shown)


greghudson:
Apologies for the delay on this; I've been working on a larger pull request.

We don't normally document limitations or bugs in previous releases, and we don't normally create unit test programs for very simple functions like encode_database_size() and decode_database_size().

I will likely want to make adjustments to the encode and contracts for better length checking, but the general approach seems okay.

jrisc:
I removed documentation and tests.

I also modified `decode_database_size()` to be more restrictive and rely on `krb5_data.length` rather than the value of the first 4 bytes to infer the DB size type.

I used `KRB5KRB_ERR_GENERIC` as error. Would there be a more relevant one for this case?

greghudson:
I pushed some editorial changes.  I also added a functional test case to t_kprop.py, but disabled it unless an environment variable is set.  This is a first for the test suite, but it's not reasonable to produce, copy, and load a >4GB dump file during "make check" and there's no other way to verify that the code path works.  (The erstwhile unit tests didn't verify that the encoder and decoder were called correctly.)

Apologies again for the slow review.  Pending objections, this should be approximately ready to merge.

src/kprop/kprop_util.c
```diff
@@ -92,3 +92,46 @@ sn2princ_realm(krb5_context context, const char *hostname, const char *sname,
...
+
+krb5_error_code
+decode_database_size(const krb5_data *buf, uint64_t *size_out)
+{
+    uint64_t size;
+
+    size = load_32_be(buf->data);
+    if (buf->length == 12) {
+        /* A 12 bytes size buffer must have the first four bytes zeroed. */
+        if (load_32_be(buf->data) != 0)
```
jrisc, line 120: This call seems redundant with `size = load_32_be(buf->data)` at line 117.

src/kprop/kprop_util.c
```diff
@@ -92,3 +92,46 @@ sn2princ_realm(krb5_context context, const char *hostname, const char *sname,
...
+            return KRB5KRB_ERR_GENERIC;
+
+        /* The actual size is stored in the next 64 bits.  Values from
+         * 1..2^32-1 must be encoded in four bytes. */
+        size = load_64_be(buf->data + 4);
+        if (size > 0 && size <= UINT32_MAX)
+            return KRB5KRB_ERR_GENERIC;
+    } else if (buf->length == 4) {
+        size = load_32_be(buf->data);
```
jrisc, line 129: This case is supposed to be handled by line 117 too. If you prefer to have one block per case in the if/else/if statement, it's probably better to remove the line 117 and have each case block do the full decoding and/or error raising process.

jrisc (commented):
Just a remark about the changes in the `decode_database_size()` function

src/kprop/kprop_util.c
greghudson: Yeah, I meant to remove line 117 in the edits and apparently didn't.

PR #1240, opened 2022-02-07
Use typedef to define bool_t and enum_t

Mizuchi:
Otherwise this conflicts with bool_t and enum_t in fbthrift:
https://github.com/facebook/fbthrift/blob/main/thrift/lib/cpp2/type/Tag.h#L59

### DIFF
--- a/src/include/gssrpc/types.hin
+++ b/src/include/gssrpc/types.hin
@@ -95,8 +95,8 @@ typedef int32_t		rpc_inline_t;
 /* This is for rpc/netdb.h */
 @rpcent_define@
 
-#define	bool_t	int
-#define	enum_t	int
+typedef int bool_t;
+typedef int enum_t;
 #ifndef FALSE
 #	define	FALSE	(0)
 #endif


greghudson:
Apologies for the slow review.

I am fine with the change but am confused about the justification.  `#define int bool_t` definitely conflicts with Tag.h, but a typedef seems like it would also conflict.  I tried compiling this fragment with g++ and clang++ and got errors in both:

    typedef int bool_t;
    struct all_c {};
    struct number_c : all_c {};
    struct integral_c : number_c {};
    struct bool_t : integral_c {};

I did check two other Sun RPC implementations and they both use typedefs (libtirpc typedefs them to int32_t, Illumos typedefs them to int.)

Mizuchi:
> `#define int bool_t` definitely conflicts with Tag.h, but a typedef seems like it would also conflict.

It won't since it's in different namespace. e.g. following code compiles (online version: https://godbolt.org/z/3rq5EMvvs)

```
typedef int bool_t;

namespace apache::thrift {
struct all_c {};
struct number_c : all_c {};
struct integral_c : number_c {};
struct bool_t : integral_c {};
}
```

greghudson:
Thanks; I didn't see the namespace declarations.  I will edit the commit message and merge this.

PR #1246, opened 2022-04-22
Add autoconf parameter for default PKCS#11 module

jrisc:
Allow to override `opensc-pkcs11.so` as the default PKCS#11 module name.

This is useful to make the client configuration library-agnostic by using modules like `p11-kit-proxy.so` as a generic layer in front of other PKCS#11 implementations.

In this case `pkinit_identities = PKCS11:` would be enough to enable the use of smartcards by having p11-kit detecting any PKCS#11 module installed.


greghudson:
I added documentation for the new variable and doc substitutions for its value.

jrisc:
Okay, everything looks good to me.

PR #1249, opened 2022-05-19
Always read the configuration files the first time

simo5:
There is at least one case (with flatpaks) where configuration files
in the special read-only /etc all have (time_t)0 as mtime.

This causes the current code to not read the configuration files even
once because the comparison in check_link_mtime() assumes that a mtime
on disk is always greater than 0 and the initial "high" value
g_confFileModTime is set to 0.

To address this we change the initial high value to the be the "invalid"
value (time_t)-1
Then we special case the check in check_link_mtime() to always,
unconditionally, load the configuration file when the initial '-1' value
is seen.

After the first time the code goes through loadConfigFiles() the value
of g_confFileModTime will be set to the highest mtime found, which will
work correctly even if that "highest" value is the artifical (time_t)0
found in these special readonly filesystems.

### DIFF
--- a/src/lib/gssapi/mechglue/g_initialize.c
+++ b/src/lib/gssapi/mechglue/g_initialize.c
@@ -93,7 +93,7 @@ static void free_mechSet(void);
 static gss_mech_info g_mechList = NULL;
 static gss_mech_info g_mechListTail = NULL;
 static k5_mutex_t g_mechListLock = K5_MUTEX_PARTIAL_INITIALIZER;
-static time_t g_confFileModTime = (time_t)0;
+static time_t g_confFileModTime = (time_t)-1;
 static time_t g_confLastCall = (time_t)0;
 
 static gss_OID_set_desc g_mechSet = { 0, NULL };
@@ -469,9 +469,9 @@ load_if_changed(const char *pathname, time_t last, time_t *highest)
 	mtime = check_link_mtime(pathname, &mtime);
 	if (mtime == (time_t)-1)
 		return;
-	if (mtime > *highest)
+	if (mtime > *highest || *highest == (time_t)-1)
 		*highest = mtime;
-	if (mtime > last)
+	if (mtime > last || last == (time_t)-1)
 		loadConfigFile(pathname);
 }
 
@@ -482,7 +482,7 @@ static void
 loadConfigFiles()
 {
 	glob_t globbuf;
-	time_t highest = 0, now;
+	time_t highest = (time_t)-1, now;
 	char **path;
 	const char *val;
 
@@ -522,7 +522,8 @@ updateMechList(void)
 
 #if defined(_WIN32)
 	time_t lastConfModTime = getRegConfigModTime(MECH_KEY);
-	if (g_confFileModTime >= lastConfModTime)
+	if (g_confFileModTime >= lastConfModTime &&
+	    g_confFileModTime != (time_t)-1)
 		return;
 	g_confFileModTime = lastConfModTime;
 	loadConfigFromRegistry(HKEY_CURRENT_USER, MECH_KEY);


greghudson:
I guess the special-case checks are a guard against time_t being unsigned?  I don't think that's the case anywhere, but I also don't see anything in the POSIX spec to prevent it.

src/lib/gssapi/mechglue/g_initialize.c
```diff
@@ -467,11 +467,13 @@ load_if_changed(const char *pathname, time_t last, time_t *highest)
 	time_t mtime;
 
 	mtime = check_link_mtime(pathname, &mtime);
+        if (mtime == (time_t)-1 && last == (time_t)-1)
+            mtime = 0;
```
greghudson, line 471: What's this check for?  mtime == -1 still indicates an error happened even if last is also -1.

simo5: It is to ignore the mtime error on the first invocation (when last == -1), because we do not care what mtime is in that case.
After the first time it behaves like the previous code.

greghudson: Are we expecting an lstat() or stat() error, and do we want to go ahead and try to read the file in that case?  I thought we were just expecting an mtime of 0.

simo5:
> I guess the special-case checks are a guard against time_t being unsigned? I don't think that's the case anywhere, but I also don't see anything in the POSIX spec to prevent it.

Exactly, I just wanted it to be correct in case time_t is unsigned on some architecture.

src/lib/gssapi/mechglue/g_initialize.c
```diff
@@ -467,11 +467,13 @@ load_if_changed(const char *pathname, time_t last, time_t *highest)
...
 	if (mtime == (time_t)-1)
 		return;
 	if (mtime > *highest)
```
greghudson, line 474: I think this also needs a special case for `*highest == (time_t)-1`, assuming we are guarding against unsigned time_t at all.

simo5: I think you are right, fixing

src/lib/gssapi/mechglue/g_initialize.c
simo5: I think we should try to read the file even if lstat/stat fails, worst case we'll get an error in loadConfigFile() when trying to fopen(), and then we'll just return, best case we still can read the config file even though the stat failed.

Are you concerned about the possibility of a transient lstat()/stat() error that may resolve in the future?

I think a transient error of that kind gives no guarantees a future reload will work either, as a future stat may return an mtime lower or equal than other config files present and therefore cause the code to skip the file entirely.

To support recovery from transient errors we'd really need to keep a cache of config files with individual timestamp, and perhaps, for filesystems that support it, a file hash. But I think thisis not a failure case worth optimizing for at this time.

greghudson: > Are you concerned about the possibility of a transient lstat()/stat() error that may resolve in the future?

No, I wasn't intending to raise that possibility.  I just don't see how this check is necessary.  Do we expect to encounter a filesystem where lstat() or stat() fails but the file can be read?  I thought we were just trying to deal with synthetic mtimes of 0.

simo5: Should I just ignore the case when mtime == -1 and leave the code to return immediately like it was doing before ?

greghudson: That would be my preference.

simo5: done

simo5:
Manually tested with  modified flatpaks.
Including systemd user socket activation.

Merging now and will fix any followups later.

simo5:
@greghudson why set highest to -1 ?

greghudson:
I think we're looking at this problem slightly differently.

The way I see it, the bug was that we weren't treating 0 as a realistic mtime for a file to have.  Now that we know it's realistic, we should make sure that our sentinel values are treated as older than 0.  highest should be set to -1 so that, if we find no config files, g_confFileModTime remains at -1; then, if a file later appears with mod time 0, we will read it.  That corner case isn't really expected from a synthetic read-only filesystem, but it's trivial (and in my mind elegant) to accommodate.

Your proposed fix was to make sure that we always attempt to read the config file on the first call.  Under that approach, if we can't find any config file mod times on the first call, it makes sense to set g_confFileModTime to 0 because it's no longer the first call.  You had code to ignore stat errors on the first call, because we should try reading the config file on the first call regardless.  That corner case isn't expected either, and I'm less interested in accommodating that corner case because it requires extra code (and isn't in my mind elegant, because it involves resetting a variable to make the code behave as if we read a mod time of 0).

simo5:
Ah makes sense seen under this light, thanks!

PR #1252, opened 2022-05-26
net-server.c: add AI_NUMERICSERV definition, fixes PPC build

barracuda156:
This PR essentially brings existing Macports patch for PPC systems into the source: https://github.com/macports/macports-ports/blob/master/net/kerberos5/files/patch-src-lib-apputils-net-server.c.diff

It fixes `kerberos5` build for 10.4–10.5.8 and 10.6 PPC.


barracuda156:
P. S. An alternative would be:

#if defined(__APPLE__) && defined(__POWERPC__) /* 10.4–10.5 and 10.6 PPC does not have it */
#define AI_NUMERICSERV 0
#endif

barracuda156:
Thanks!

PR #1258, opened 2022-06-17
pkinit pre-auth fails to load libpkcs11.so on Solaris

Sashan:
the issue has been hit by `pam-krb5` test suite on Solaris.
one of the testcases uses `pkcs11` pre-authentication.
the krb5.conf in question reads as follows:
```
[libdefaults]
        default_realm = ACME.COM

[realms]
        ACME.COM = {
pkinit_identity = FILE:/usr/tmp/ca/kdc.crt,/usr/tmp/ca/kdc.key
pkinit_anchors = FILE:/usr/tmp/ca/ca.crt
pkinit_identities = PKCS11:
                kdc = kdc.acme.com
                admin_server = kdc.acme.com
        }

[domain_realm]
        .acme.com = ACME.COM

[logging]
        default = FILE:/var/krb5/kdc.log
        kdc = FILE:/var/krb5/kdc.log
        kdc_rotate = {

                period = 1d

                versions = 10
        }

[appdefaults]
        gkadmin = {
                help_url = help_url
        }
```

After building a kerberos with `DEBUG` option enabled I could obtain
a log as follows from our `pamtest` binary:
```
Calling pam_authenticate
stat(/usr/lib/krb5/plugins/preauth/spake.so): No such file or directory
pkinit_init_plg_crypto: initializing openssl crypto context at 8069ca8
pkinit_client_plugin_init: returning plgctx at 8064c98
pkinit_init_req_crypto: returning ctx at 80781c0
pkinit_init_identity_crypto: returning ctx at 8078d58
pkinit_client_req_init: returning reqctx at 8078d20
pkinit_client_profile 8065170 8064c98 8078d20 8078e74
pkinit_identity_initialize: 8065170 8078e10 8078d58
process_option_identity: idtype is PKCS11
stat(libpkcs11.so): No such file or directory
pkinit_identity_initialize returned -1765328360 (Preauthentication failed)
pkinit_client_prep_questions: not asking responder question
pkinit_client_prep_questions returning 0
pkinit_client_process 8065170 8064c98 8078d20 80657a8
processing KRB5_PADATA_PK_AS_REQ

```

note a call to `stat(libpkcs11.so)` which fails. the function `krb5int_open_plugin()`
assumes the plugin must be in current working directory. However the
`modload()` however `dlopen()` may use `RT_PATH` to find desired plugin.

### DIFF
--- a/src/util/support/plugins.c
+++ b/src/util/support/plugins.c
@@ -189,9 +189,10 @@ long KRB5_CALLCONV
 krb5int_open_plugin (const char *filepath, struct plugin_file_handle **h, struct errinfo *ep)
 {
     long err = 0;
-    struct stat statbuf;
     struct plugin_file_handle *htmp = NULL;
     int got_plugin = 0;
+#if defined(USE_CFBUNDLE) || defined(_WIN32)
+    struct stat statbuf;
 
     if (!err) {
         if (stat (filepath, &statbuf) < 0) {
@@ -201,6 +202,7 @@ krb5int_open_plugin (const char *filepath, struct plugin_file_handle **h, struct
                          filepath, strerror(err));
         }
     }
+#endif
 
     if (!err) {
         htmp = calloc (1, sizeof (*htmp)); /* calloc initializes ptrs to NULL */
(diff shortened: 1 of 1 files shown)


greghudson:
We can get rid of the USE_CFBUNDLE stuff (it was used in the now-defunct Kerberos for Macintosh, but nothing defines it today), and I think we can get rid of the regular-file check for Windows so that there's no need to call stat().

However, for backport we'll want something as minimal as possible, and I think the current PR branch is roughly that.  (An alternative is reverting commit c5c11839e02c7993eb78f2c94c75c10cf93f2195 on the release branches, but that's not trivial.)  So I will prepare a follow-up commit to simplify the code, and only mark the first commit for backport.

greghudson:
I pushed the follow-up commit, and a small Windows fix to the original commit.  I will need to do some manual testing, since I touched directory loading (not covered by the CI, I think) and Windows.

PR #1259, opened 2022-07-11
Remove pre-CMS PKINIT compatibility code

greghudson:
CMS support is present in OpenSSL 1.0, which is the earliest supported
version.

I looked into changing cms_signeddata_create() to use CMS APIs rather than PKCS7 APIs, but it doesn't look easy.  In the PKCS11 case we need to generate the signature ourselves, and the CMS APIs appear to require that OpenSSL generate the digest and signature internally (with the caller providing the private key).  It's possible that we could do something with engines and providers for OpenSSL 1.x and 3.x respectively.

LibreSSL support isn't a priority for us, but I checked and they added CMS support in 2019.


theoparis:
This broke building with BoringSSL/[aws-lc](https://github.com/aws/aws-lc) - can it be added back?

PR #1266, opened 2022-09-09
Use unsigned 32-bit integer for password expiry delta

jrisc:
During interactive password pre-authentication, a warning is printed in case the password expiry time delta is less than one week. However this time delta is stored as a signed 32-bit integer, while the current and expiry date times are stored as unsigned 32-bit integers (even if their declared type is signed).

As a consequence, an expiry date that is more than 2^31 - 1 seconds in the future will cause this time delta to wrap. It results in the expiry warning message "Your password will expire in less than one hour" to be displayed, while the actual expiry date is more than 68 years in the future.

This commit fixes this issue by casting the delta as unsigned after the expiry date is confirmed to be in the future.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2324,6 +2324,15 @@ ts_delta(krb5_timestamp a, krb5_timestamp b)
     return (krb5_deltat)((uint32_t)a - (uint32_t)b);
 }
 
+/* Return (end - start) as an unsigned 32-bit value, or 0 if start > end. */
+static inline uint32_t
+ts_interval(krb5_timestamp start, krb5_timestamp end)
+{
+    if ((uint32_t)start > (uint32_t)end)
+        return 0;
+    return (uint32_t)end - (uint32_t)start;
+}
+
 /* Increment a timestamp by a signed 32-bit interval, without relying on
  * undefined behavior. */
 static inline krb5_timestamp
--- a/src/lib/gssapi/krb5/accept_sec_context.c
+++ b/src/lib/gssapi/krb5/accept_sec_context.c
@@ -353,8 +353,8 @@ kg_accept_dce(minor_status, context_handle, verifier_cred_handle,
         *mech_type = ctx->mech_used;
 
     if (time_rec) {
-        *time_rec = ts_delta(ctx->krb_times.endtime, now) +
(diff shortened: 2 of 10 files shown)


greghudson:
I'd like to review the uses of krb5_deltat and ts_delta() and see if there's a more categorical solution to this problem.

There is an analagous problem with the ptrdiff_t type in C.  ptrdiff_t generally has the same size as size_t but can represent fewer positive values because it is signed.  If a byte array is allocated consuming more than half of the address space (which is apparently possible in practice, at least on 32-bit platforms), pointer subtractions between the far reaches of that array can be undefined.  With a time machine and a standards-body-mind-control ray, this problem could be traded for a different problem if ptrdiff_t were removed from the language and pointer subtraction instead yielded an unsigned result (of type size_t).  But of course, then pointer subtraction would have to require the first operand to be equal to or larger than the second.  That's true in almost all pointer subtraction scenarios, but perhaps not so exhaustively as to be worth the trade.

Returning to the problem at hand, if all uses of ts_delta() are like most uses of pointer subtraction, perhaps it should be redefined to have an unsigned result, and we wouldn't need ugly casts in the places like this one where the result might exceed INT32_MAX.

greghudson:
From a review of ts_delta() usage, we definitely want a signed delta in a few cases (computing KDC time offsets), and would like it to be signed in some other cases (computing ticket lifetimes; although it seems like a ticket lifetime should be a non-negative interval, it can actually be negative, and negative-lifetimes tickets are even sometimes useful for testing).  On the flip side, it would be more natural to have non-negative interval results within the GSSAPI implementation, where we sometimes need to compute a remaining context or cred lifetime and put it in an OM_uint32.  There is a halfway case in kadmin/server/misc.c where we compute the time since the last policy change (seems like it should be unsigned) but then compare the result to a policy field which happens to be signed.

So, I'm thinking of defining a new helper to return an unsigned result, and calling it ts_interval().  (It would also check if the start of the interval is after the end of the interval and return 0.)  I think this is a fairly natural way of thinking about things, with the wrinkle that it would want its parameters reversed from ts_delta()--ts_interval(end, start) seems strictly less intuitive than ts_interval(start, end).

greghudson:
Does this seem okay as an alternative fix?

PR #1267, opened 2022-09-10
Canonicalize user realm for S4U2Self queries

greghudson:
Based on a trace provided by Jacob Shivers from Red Hat, Active Directory 2016 does not canonicalize the user realm in S4U2Self queries, and adds a transited field to a cross-realm S4U2Self ticket if the user realm is an alias, causing MIT krb5 to fail a transited check on the ticket.

In krb5_get_self_cred_from_kdc(), after obtaining a TGT for the user realm, replace the realm for the S4U2Self request with the realm in the TGT.

ticket: 9070


greghudson:
I think this won't work; get_creds.c puts the requested server name (with the user realm alias, in the problematic scenario) in the krb5_creds result rather than the ticket server.  I think s4u_creds.c would need to decode and examine the ticket to get the canonical realm name.  I will think on this further.

PR #1269, opened 2022-10-04
Make principal and realm optional in setpw structure

josephsutton1:
This matches Heimdal and the Windows KDC.

### DIFF
--- a/src/lib/krb5/asn.1/asn1_k_encode.c
+++ b/src/lib/krb5/asn.1/asn1_k_encode.c
@@ -939,8 +939,8 @@ static const struct atype_info *pa_enc_ts_fields[] = {
 DEFSEQTYPE(pa_enc_ts, krb5_pa_enc_ts, pa_enc_ts_fields);
 
 DEFFIELD(setpw_0, struct krb5_setpw_req, password, 0, ostring_data);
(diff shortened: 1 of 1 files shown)


greghudson:
Is there a specific motivation for this change?  And has any analysis been done on what the existing caller code (kadmin/server/schpw.c) will do if one or both fields are not specified?

josephsutton1:
This ASN.1 discrepancy was noticed during testing for [bug 15047](https://bugzilla.samba.org/show_bug.cgi?id=15047) in Samba, for which we implemented a workaround in Samba's own kpasswd service implementation. It's useful to be able to omit the principal and realm in a setpw request, because in Samba/AD, including them introduces a requirement for possessing rights to the target account, while no such rights are required when the target account is implicit.

I haven't yet tested what happens in process_chpw_request() when either field is omitted. Although there is at least a NULL check there already, it looks like some further checks will be needed to avoid a possible crash when calling krb5_unparse_name().

What we do in Samba is call krb5_unparse_name_flags() (with KRB5_PRINCIPAL_UNPARSE_NO_REALM) and krb5_princ_realm() to get the name and realm. We require them to be either both present or both NULL, and fail the request if only one or the other is specified. I guess similar logic could be added here.

PR #1271, opened 2022-11-09
po: Add Georgian translation

NorwayFun:
Hello,
Here's full translation of latest git pot file (regenerated from source several days ago) of Kerberos to Georgian.


yangxianrui:
你发的邮件已经收到！
Thanks with Best regards,

PR #1272, opened 2022-11-17
Fix syntax error in aclocal.m4

jrisc:
The block argument of the `AC_CACHE_VAL` call ends before the `if` instruction can reach the corresponding `fi` one. This `fi` is definitely meant to be part of this block.

### DIFF
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -409,8 +409,8 @@ else
       [[struct sockaddr_in6 in;
         AF_INET6;
         IN6_IS_ADDR_LINKLOCAL(&in.sin6_addr);]])],
-    [krb5_cv_inet6=yes], [krb5_cv_inet6=no])])
-fi
+    [krb5_cv_inet6=yes], [krb5_cv_inet6=no])
+fi])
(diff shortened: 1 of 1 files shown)


greghudson:
Looking at the history, this mistake has been present since 2000.  I think the only practical consequence is that the AC_CACHE_VAL cleanup `rm -f core conftest.err conftest.$ac_objext conftest.$ac_ext` winds up being conditional upon `$ac_cv_func_inet_ntop" = "yes"`.  How did you notice it?  (I will accept the patch, of course, but whether it gets a ticket and backport depends on whether it has a more significant practical consequence in some environment.)

jrisc:
This problem was identified by [Frédéric Bérat](mailto:fberat@redhat.com) while testing the upcoming version 2.72 of Autoconf on the Fedora build of krb5.

The error message is:
```
checking for IPv6 compile-time support without -DINET6... ./configure: line 10559: syntax error near unexpected token `;;'
./configure: line 10559: `rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext ;;'
```

PR #1281, opened 2022-12-13
Fix PKINIT CMS error checking for older OpenSSL

greghudson:
[@jrisc please verify that this seems okay for Fedora.  Checking specifically for an invalid signature seems much easier than checking for algorithm issues, as the error code is singular and stable whereas algorithm issues can manifest along lots of different code paths.  An invalid signature also seems less likely than an algorithm issue, so we're less likely to return the wrong thing if we assume that almost all error codes indicate algorithm issues.]

Commit 70f61d417261ca17efe3d60d180033bea2da60b0 updated the CMS_verify() error code checks, using two error codes new to OpenSSL 3.0 (RSA_R_DIGEST_NOT_ALLOWED and CMS_R_UNKNOWN_DIGEST_ALGORITHM). This change broke the build for OpenSSL 1.0 and 1.1.

Instead of looking for codes indicating an algorithm issue and assuming that everything else is an invalid signature, check for the code indicating an invalid signature and assume that everything else is an algorithm issue.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -2061,18 +2061,10 @@ cms_signeddata_verify(krb5_context context,
             goto cleanup;
         out = BIO_new(BIO_s_mem());
         if (CMS_verify(cms, NULL, store, NULL, out, flags) == 0) {
-            unsigned long err = ERR_peek_last_error();
-            switch(ERR_GET_REASON(err)) {
-            case RSA_R_DIGEST_NOT_ALLOWED:
-            case CMS_R_UNKNOWN_DIGEST_ALGORITHM:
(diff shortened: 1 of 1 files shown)


jrisc:
Sorry about the late answer.

Tests are passing for Fedora and CentOS. I would like to run some manual tests regarding unsupported digest/signature algorithms. I should be able to confirm it works properly before the end of the week.

jrisc:
I confirm the right error is raised in multiple scenarios, from the client and the KDC.

I just hope there will be no new error code introduced by OpenSSL in addition to `CMS_R_VERIFICATION_FAILURE`. It would result in invalid digest error to be logged, while it might be completely unrelated. The signature verification error message had the merit to be a bit more generic in comparison.

But I am not aware of other error codes that would be returned in case of a failure to verify the signature (and the documentation is not very helpful on that regard). So I think it is fine to merge it, and hope there will be no breaking change too soon.

PR #1282, opened 2022-12-13
Enable LFS on 32-bit platforms

mattst88:
* required to support 64-bit filesystems (inodes) on 32-bit platforms: https://flameeyes.blog/2010/12/15/another-good-reason-to-use-64-bit-installations-large-file-support-headaches/

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -27,6 +27,8 @@ PKG_PROG_PKG_CONFIG
 AC_CHECK_HEADER([stdint.h], [],
   [AC_MSG_ERROR([stdint.h is required])])
 
+AC_SYS_LARGEFILE
+
 AC_CACHE_CHECK([whether integers are two's complement],
   [krb5_cv_ints_twos_compl],
   [AC_COMPILE_IFELSE(


greghudson:
Thinking through the possible concerns:

* We don't use off_t in any APIs, so ABI issues shouldn't be a concern.

* It's important to include the autoconf header (`autoconf.h` for us) before any system headers when using AC_SYS_LARGEFILE.  We have some files which use off_t and do not do this: kdb_log.c and everything in src/plugins/kdb/db2/libdb2.

* fseek() and ftell() do not benefit from large-file support; code must use fseeko() and ftello() instead.  We use fseek() in cc_file.c, kt_file.c, ucgendat.c, xdr_stdio.c, kdb5_ldap_services.c (but only with offset 0), and prof_parse.c.  These files probably do not need to operate on >2GB files.

I'm guessing the primary motivation here is KDB files.  Has this been tested with large KDBs on a 32-bit platform?

mattst88:
> I'm guessing the primary motivation here is KDB files. Has this been tested with large KDBs on a 32-bit platform?

Truth be told, I have not runtime tested this. I don't expect KDB files to be >2GB, but they could definitely be on a file system with 64-bit inodes.

My primary motivation is ChromeOS's new policy of requiring largefile support. I guess you could imagine how aggravating it could be to have somewhat random looking failures depending on a file's inode number.

greghudson:
Do you have more details on this?  I'm not familiar with how large file support interacts with inode numbers.

mattst88:
I don't know a ton about it, but the blog post reference in the first comment explains a bit:

> [...] on testing what software was still using the old `stat()` calls with the data structure with a 32-bit inode number field. It’s not about the single file size, it’s a matter of huge filesystem support.

As I understand it, some file systems use 64-bit inodes, and if your program uses 32-bits for file system operations things will fail.

mattst88:
Maybe @vapier can explain better than I can :)

vapier:
yes, file systems that have 64-bit inodes and whose inode number exceeds 32-bits will break applications that stat().  which krb5 def does.  the insidious thing is that it can happen seemingly at random -- everything works fine with inode numbers that fit in 32-bits, but then fail on files whose inode numbers use the upper 32-bits.  and maybe deleting/recreating the file gets you a new inode number that works.

some filesystems seem more likely to use larger inode numbers than others.  i think XFS, at least for a while?, was more actively utilizing the full space.  they might have backed off from it because it found a lot of bad apps.

> fseek() and ftell() do not benefit from large-file support; code must use fseeko() and ftello() instead. ... These files probably do not need to operate on >2GB files.

i wouldn't look at it in terms of "these don't need to access large files" as that requires a lot of thought & effort, and easy to get wrong.  it's a lot easier to simply say "always use fseeko/ftello".  there is no realistic runtime cost or overhead to using ftello/fseeko over ftell/fseek.  in glibc at least, it's always 64-bit internally, which means there's actually a little bit more code to use the 32-bit interface because it has to truncate & check for overflow.

greghudson:
> i wouldn't look at it in terms of "these don't need to access large files" as that requires a lot of thought & effort, and easy to get wrong. it's a lot easier to simply say "always use fseeko/ftello". there is no realistic runtime cost or overhead to using ftello/fseeko over ftell/fseek.

Overhead isn't a concern, but portability is.  We unfortunately don't have the luxury of assuming POSIX.1-2008 support, mostly because of Windows (we only require Visual Studio 2013, and a Google search suggests these functions aren't present in VS 2014).  The portability glue for ftello/fseko is simple enough, but leaving this code alone is simpler and is unlikely to cause problems.

As for the PR in general: enabling large-file support is fine, but we'd need to modify the files that include system headers before autoconf.h (enumerated in my first comment) and we'd need to test this.  I can run the test suite on a 32-bit buildbot worker once the other concern is accounted for.

Downstreams can of course configure with CPPFLAGS=-D_FILE_OFFSET_BITS=64 without upstream support.

vapier:
autoconf provides a AC_FUNC_FSEEKO macro which sets up `HAVE_FSEEKO`, so you don't have to setup your own tests.  you would need to add the redirect from ftell/fseek to ftello/fseeko, but that shouldn't be too hard since it looks like a common k5-int.h already exists.  imo this doesn't seem like too much source effort, especially if you need to do it even only once, and is still easier than trying to reason about "will these ever interact with files in a way that would cause problems".

PR #1268, opened 2022-09-23
Use non-RFC 8009 enctypes to sign KDC checksum in PAC structure

abbra:
Active Directory does not support RFC 8009 encryption types and [MS-PAC] 2.8.2 spec outlines how cryptographic system has to be selected to produce a KDC checksum. FreeIPA 4.9.10+ defaults to RFC 8009 enctypes, resulting in a PAC buffer with a checksum rejected by Active Directory.

If a local TGT key is one of RFC 8009 encryption types, downgrade the key to use for checksum to either ENCTYPE_AES128_CTS_HMAC_SHA1_96 or ENCTYPE_AES256_CTS_HMAC_SHA1_96 if they available.

Discussed with Microsoft dochelp@ in
https://lists.samba.org/archive/cifs-protocol/2022-September/003761.html

Signed-off-by: Alexander Bokovoy <abokovoy@redhat.com>

PR #1284, opened 2022-12-22
Add PAC full signatures

greghudson:
Here is an initial implementation of PAC full signatures, which is half of Microsoft's response to https://i.blackhat.com/EU-22/Thursday-Briefings/EU-22-Tervoort-Breaking-Kerberos-RC4-Cipher-and-Spoofing-Windows-PACs-wp.pdf .

I am hoping that Samba developers can help me with testing.  Currently in t_pac.c we have a DER-encoded service ticket, and its associated server key and privsvr (krbtgt) key, provided by Isaac as part of his work on ticket checksum support.  This needs to be updated to a ticket whose PAC contains a full signature.  I could of course decrypt and re-sign the ticket using the code in this PR (or with a bit more work, Samba Heimdal's code), but it would be better to use a captured ticket from Windows.


abbra:
@greghudson I will look into producing a trace and a keytab with keys to use that. Will probably take a bit of time due to holidays.

metze-samba:
https://git.samba.org/?p=metze/wireshark/wip.git;a=shortlog;h=refs/heads/ws-metze-current has code to
verify the full signature, see
https://git.samba.org/?p=metze/wireshark/wip.git;a=commitdiff;h=7c4b98b2e17de76ef6f0729d383132604edf661f
https://git.samba.org/?p=metze/wireshark/wip.git;a=commitdiff;h=c03de9f5a1c4b289a0ac7da1f8c1b648869491f4

greghudson:
Per some of my thoughts in PR #1268, it might make sense to skip verify the privsvr checksum if we verify a full checksum (easily accomplished in verify_pac_checksums() by just changing the `if (privsvr != NULL)` to an `else if`), to avoid being the squeaky wheel if privsvr checksums are ever phased out.

The full checksum does not protect the server checksum (it is computed when the server checksum buffer is still zeroed; and the server checksum covers the full checksum instead).  But this seems irrelevant to security, as it protects everything the server checksum protects, and the server checksum is separately verified by the KDC.

greghudson:
Thanks to a trace provided by Stefan Metzmacher, I was able to update t_pac.c and test that the PR code can verify a ticket produced by AD with a full checksum.

I also made the change discussed in my previous comment.

src/lib/krb5/krb/pac.c
```diff
@@ -556,64 +557,75 @@ verify_checksum(krb5_context context, const krb5_pac pac, uint32_t buffer_type,
...
+        /* Zero the full checksum buffer in the copy and verify the full
+         * checksum over the copy with all three checksums zeroed. */
+        ret = zero_signature(context, pac, KRB5_PAC_FULL_CHECKSUM, &copy);
+        if (ret)
+            goto cleanup;
+        ret = verify_checksum(context, pac, KRB5_PAC_FULL_CHECKSUM, privsvr,
+                              KRB5_KEYUSAGE_APP_DATA_CKSUM, &copy);
+        if (ret)
+            goto cleanup;
+    } else if (privsvr != NULL) {
```
metze-samba, line 608: I think this 'else' should be removed and we should check both signatures,
I know this is not strictly needed, but it will catch bugs early and seems to
be more consistent.

greghudson: My reasoning is that the privsvr checksum is vestigial when the full checksum is present and maybe MS will want to phase it out at some point.

We could tolerate a missing privsvr checksum but check it if one is there, if that seems better to you.

metze-samba: Yes, that would be better.

greghudson: The code for this got kind of messy (I realised that the zero_checkum() call for the privsvr checksum would also throw an error if it is missing), so I'm just going to put it back the way it was and continue requiring the privsvr checksum to be present.

abartlet:
Documentation for this in MS-PAC is still via the errata process, but here is a preview with the differences highlighted, thanks to the MS interop team: https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-PAC/%5bMS-PAC%5d-20221212-diff.pdf

greghudson:
I noticed one problem if you can forward it on: The section 2.4 description of the new checksum type says "Extended KDC (privilege server) checksum (section 2.8). PAC structures MUST contain one buffer of this type. Additional KDC checksum buffers MUST be ignored." which I think is an insufficiently modified copy of the existing KDC checksum description.  The new checksum is only included in service tickets (as stated in section 2.8.4) and the third sentence is confusing because this isn't the only KDC checksum type.

PR #1289, opened 2023-02-02
KDB: Set key data pointer to NULL to avoid double free

jrisc:
The `krb5_dbe_def_encrypt_key_data()` function is [freeing `key_data->key_data_contents[0]` in case the call to `krb5_c_encrypt()` failed](https://github.com/krb5/krb5/blob/krb5-1.20.1-final/src/lib/kdb/encrypt_key.c#L111). However, since a06945b4 a failure return code from `krb5_dbe_encrypt_key_data()` in `add_principal()` will [result in a call to `krb5_db_free_principal()`](https://github.com/krb5/krb5/blob/krb5-1.20.1-final/src/kadmin/dbutil/kdb5_create.c#L476). This function [loops over `key_data_contents`, freeing all non-`NULL` pointers](https://github.com/krb5/krb5/blob/krb5-1.20.1-final/src/lib/kdb/kdb5.c#L124-L129).

```
# valgrind --track-origins=yes --keep-stacktraces=alloc-and-free -- /usr/sbin/kdb5_util create -s -P <password> -k arcfour-hmac
[...]
==3559== Invalid free() / delete / delete[] / realloc()
==3559==    at 0x48470E4: free (vg_replace_malloc.c:872)
==3559==    by 0x48826B9: krb5_dbe_free_key_data_contents (kdb5.c:127)
==3559==    by 0x4883B06: krb5_db_free_principal (kdb5.c:844)
==3559==    by 0x1183F5: add_principal.constprop.0 (kdb5_create.c:476)
==3559==    by 0x118837: kdb5_create (kdb5_create.c:292)
==3559==    by 0x10E869: main (kdb5_util.c:344)
==3559==  Address 0x67a7260 is 0 bytes inside a block of size 42 free'd
==3559==    at 0x48470E4: free (vg_replace_malloc.c:872)
==3559==    by 0x48811C8: krb5_dbe_def_encrypt_key_data (encrypt_key.c:111)
==3559==    by 0x1184B4: add_principal.constprop.0 (kdb5_create.c:427)
==3559==    by 0x118837: kdb5_create (kdb5_create.c:292)
==3559==    by 0x10E869: main (kdb5_util.c:344)
==3559==  Block was alloc'd at
==3559==    at 0x484486F: malloc (vg_replace_malloc.c:381)
==3559==    by 0x4881105: krb5_dbe_def_encrypt_key_data (encrypt_key.c:92)
==3559==    by 0x1184B4: add_principal.constprop.0 (kdb5_create.c:427)
==3559==    by 0x118837: kdb5_create (kdb5_create.c:292)
==3559==    by 0x10E869: main (kdb5_util.c:344)
[...]
```

This commit sets `key_data->key_data_contents[0]` to `NULL` after `free()` in `krb5_dbe_def_encrypt_key_data()` to keep it from being freed again in `krb5_dbe_free_key_data_contents()`.

### DIFF
--- a/src/lib/kdb/decrypt_key.c
+++ b/src/lib/kdb/decrypt_key.c
@@ -1,7 +1,7 @@
 /* -*- mode: c; c-basic-offset: 4; indent-tabs-mode: nil -*- */
 /* lib/kdb/decrypt_key.c */
 /*
- * Copyright 1990,1991 by the Massachusetts Institute of Technology.
+ * Copyright 1990,1991,2023 by the Massachusetts Institute of Technology.
  * All Rights Reserved.
  *
  * Export of this software from the United States of America may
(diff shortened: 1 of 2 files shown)


greghudson:
I think we have the same problem again later on in the function, although only on malloc failure.

jrisc:
Indeed, I assigned `NULL` there too.

greghudson:
I edited the commit message, and also added a second commit to modernize this function and the corresponding decryption function.

The encryption function had the unusual (for this project) behavior of freeing and nulling pre-existing pointers in key_data on entry.  None of the current callers rely on this behavior.  External callers of this function are unlikely, so I didn't preserve that behavior in the rewrite.

src/lib/kdb/decrypt_key.c
```diff
@@ -52,88 +52,84 @@
...
+            salt.type = KRB5_KDB_SALTTYPE_NORMAL;
         }
     }
 
-    return retval;
+    *dbkey_out = kb;
+    if (keysalt_out != NULL)
+        *keysalt_out = salt;
+    memset(&kb, 0, sizeof(kb));
+    memset(&salt, 0, sizeof(salt));
```
jrisc, line 128: Here

src/lib/kdb/encrypt_key.c
```diff
@@ -53,83 +53,63 @@
...
+            kd.key_data_contents[1] = k5memdup(keysalt->data.data,
+                                               keysalt->data.length, &ret);
+            if (kd.key_data_contents[1] == NULL)
+                goto cleanup;
         }
     }
 
-    return retval;
+    *key_data_out = kd;
+    memset(&kd, 0, sizeof(kd));
```
jrisc, line 109: And here

jrisc (commented):
Shouldn't `zap()` be used when zeroing memory to keep the compiler from optimizing `memset()` out?

greghudson:
There's nothing sensitive directly in the keyblock or key_data structures directly (just enctypes and pointers to the actual key contents or encrypted key contents).  Those memset()s are just there to prevent double-frees on successful completion, like in https://k5wiki.kerberos.org/wiki/Coding_style/Practices#Output_parameter_handling except we can't simply assign a pointer to null for these values.

However, we should use krb5_free_keyblock_contents() and krb5_free_key_data() in the cleanup handlers to properly scrub the key contents on failure.  I'll make that change.

PR #1291, opened 2023-02-24
Use k5input for SPNEGO token processing

greghudson:
OSS-Fuzz sound several low-impact read overruns in SPNEGO token processing.  They were fixed in commit 47c2a12830dbd7fb8e13c239ddc0ac74129a91f6.  This PR rewrites the SPNEGO token processing code to use k5input for improved safety.


greghudson:
I decided to go further and add a small header-only library for non-libkrb5 DER processing, and use it throughout the GSS library.

PR #1295, opened 2023-03-17
Allow to set PAC ticket signature as optional

jrisc:
[MS-PAC states that](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-pac/76c10ef5-de76-44bf-b208-0d8750fc2edd) *"The ticket signature SHOULD be included in tickets that are not encrypted to the krbtgt account"*. However, the implementation of `krb5_kdc_verify_ticket()` will [require the ticket signature to be present](https://github.com/krb5/krb5/blob/krb5-1.20.1-final/src/lib/krb5/krb/pac.c#L672-L694) in case the target of the request is a service principal.

In gradual upgrade environments, it results in S4U2Proxy requests against a 1.20 KDC using a service ticket generated by an older version KDC to fail (more details [here](https://bugzilla.redhat.com/show_bug.cgi?id=2178298#c0)).

This commit adds a `krb5_kdc_verify_ticket_ext()` function with an extra switch parameter to tolerate the absence of ticket signature in this scenario. If the ticket signature is present, it has to be valid, regardless of this parameter.

This parameter is set based on the `optional_pac_tkt_chksum` string attribute of the TGT KDB entry.

### DIFF
--- a/doc/admin/admin_commands/kadmin_local.rst
+++ b/doc/admin/admin_commands/kadmin_local.rst
@@ -649,6 +649,13 @@ KDC:
     is in the same format as those used by the **pkinit_cert_match**
     option in :ref:`krb5.conf(5)`.  (New in release 1.16.)
 
+**optional_pac_tkt_chksum**
+    Boolean value defining the behavior of the KDC in case an expected
+    ticket checksum signed with one of this principal keys is not
+    present in the PAC. This is typically the case for TGT or
+    cross-realm TGT principals when processing S4U2Proxy requests.
+    (New in release ...)
+
 This command requires the **modify** privilege.
 
 Alias: **setstr**
--- a/doc/appdev/refs/api/index.rst
+++ b/doc/appdev/refs/api/index.rst
@@ -225,6 +225,7 @@ Rarely used public interfaces
    krb5_is_referral_realm.rst
    krb5_kdc_sign_ticket.rst
    krb5_kdc_verify_ticket.rst
+   krb5_kdc_verify_ticket_ext.rst
    krb5_kt_add_entry.rst
    krb5_kt_end_seq_get.rst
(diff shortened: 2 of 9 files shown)


greghudson:
We made the explicit decision to have flag-day upgrades with respect to evidence tickets in 1.20 and 1.21.  I believe Samba made the same decision.  I would prefer not to bear a permanent cost in the API and feature set to support gradual upgrades for the new checksum types.  The cost would need to be compounded to handle gradual upgrades to 1.21 (which introduces the full KDC checksum).  Also, these changes would not be sufficient to allow gradual upgrades from 1.19 to 1.20 in vanilla upstream krb5, as 1.19 without a Samba or IPA KDB module uses AD-SIGNEDPATH.

Red Hat can of course bear the burden of maintaining patches like these locally, but I'll suggest an alternative approach: Fedora could release package upgrades for older versions which add but do not verify the new checksum buffers.

greghudson:
> And since, as far as I understand, ticket signature generation requires access to the ticket's encrypted part, which is not provided by the sign_authdata() callback, we will remain unable to support PAC ticket signature there.

My suggestion is to backport ticket and full checksum generation to libkrb5 in the krb5 1.18 package in RHEL 8.  I might be able to assist with patch preparation (although not immediately) if that's an option.

> BTW, I think the naming of the is_service_tkt variable is invalid in krb5_kdc_verify_ticket(), because the condition for ticket PAC to be present is that the ticket is a TGT. At least it is what k5_pac_should_have_ticket_signature() seems to be testing.

k5_pac_should_have_ticket_signature() returns false for a krbtgt or kadmin/changepw principal, true otherwise.  [MS-PAC] 2.8.3 says "The ticket signature SHOULD be included in tickets that are not encrypted to the krbtgt account (including the change password service) or to a trust account."

PR #1296, opened 2023-03-28
Revise certauth interface

greghudson:
Increment the certauth interface major version to 2, without backward compatibility.  Add the realm names list to the init method.  Add an hwauth_out parameter to the authorize method to make that output independent of the error code.

@kenh: Please let me know if this meets your requirements, and if the major version bump would be a burden.  My sene is that there aren't other users of the certauth interface or that they are in similar situations (site administrators rather than downstream software).  (@jrisc you aren't aware of any Red Hat package uses of this interface, right?)

If the major version bump is a burden, we could instead add a second init function to provide the realm list, similar to the kdcpreauth interface's loop() method, and defer the authorize() cleanup.

In the mail thread about this Nico suggested that PKINIT could provide the whole cert chain to the authorize method.  I don't know how to do that without using OpenSSL types in the interface, so it isn't included in this revision.

### DIFF
--- a/src/include/krb5/certauth_plugin.h
+++ b/src/include/krb5/certauth_plugin.h
@@ -35,7 +35,7 @@
  *
  * The certauth pluggable interface currently has only one supported major
  * version, which is 1.  Major version 1 has a current minor version number of
- * 1.
+ * 2.
  *
  * certauth plugin modules should define a function named
  * certauth_<modulename>_initvt, matching the signature:
@@ -77,6 +77,13 @@ struct _krb5_db_entry_new;
 typedef krb5_error_code
 (*krb5_certauth_init_fn)(krb5_context context,
                          krb5_certauth_moddata *moddata_out);
+
+/*
+ * Optional: Initialize module data.  Supersedes init if present.
+ */
+typedef krb5_error_code
+(*krb5_certauth_init_ex_fn)(krb5_context context, const char *const *realmlist,
+                            krb5_certauth_moddata *moddata_out);
 
 /*
  * Optional: Clean up the module data.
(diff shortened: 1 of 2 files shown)


kenh (commented):
I have reviewed these changes and they all look good to me; it looks straightforward for me to switch my plugins to the new interface and will greatly simplify my code.

I saw the email thread where Nico mentioned passing the whole certificate chain. You probably COULD pass in an array of unsigned char * where each element of the array was the DER serialization of each certificate in the chain. I do not recall if I responded to that or not, but at least in the case of verifying OCSP responses I am not sure that is sufficient. In my limited experience if you want to verify a complete chain via OCSP you might get a response that is signed by someone not in the certificate chain (but is still one of your trusted certificates).

greghudson:
There actually is a FreeIPA certauth module now that I check.  I am guessing that would make the major version bump kind of burdensome, so I may go with the alternate approach.

greghudson:
That is no more complicated than what I had in mind, so sure.  I've pushed a revision with the more conservative version of the feature.

kenh:
I looked at your changes, and they look fine to me.

kenh:
> Looks good from my side as well. The only thing we would need to do for FreeIPA is to rebuild it against newer krb5 version. FreeIPA runs with a single realm only so we don't really need this information at the moment.

So how do you determine that one realm?  That's one of the problems I ran into.  Calling `krb5_get_default_realm()` is an option but it isn't guaranteed that's the same realm the KDC is using.

abbra:
It is easier in IPA case because we only expect this KDC to run IPA KDB driver and our certauth plugin has access to the KDB driver structure. E.g. `ipacontext = ipadb_get_context(kcontext)` allows us to get the proper DB context which then has DB-specific realm information.

PR #1297, opened 2023-03-29
Add PA-REDHAT-PASSKEY padata type

pbrezina:
This adds a newly registered 153 as PA-REDHAT-PASSKEY.


greghudson:
Is there a design document I can look at?

pbrezina:
Hi Greg, yes. It is similar to OAuth2 autentication.

This is the document I sent to IANA, it contains links to the SSSD designs:
https://docs.google.com/document/d/1cFcEdm1dv6a0NWAoL3zJKH2Bpc9D3TxZS_QGEAJn-2k/edit?usp=sharing

greghudson:
Thanks.

It's not a blocker for this PR, but I am concerned that the design has the KDC providing a domain to the client, and nothing about the client validating the domain before passing it to the FIDO token (I assume as the rpId field of the authenticatorGetAssertion method).  FIDO tokens are intended to serve multiple relying parties of varying levels of trust, and un-phishability between parties requires that clients not blindly accept a relying party's claim of identity.  In web terms, you don't want evilcorp.com's web site to be able to request an authentication assertion for establishedbank.com that the site can then use to authenticate to establishedbank.com on behalf of the user.

I think the client should be choosing the rpId based on the Kerberos realm, rather than receiving it from the KDC.

pbrezina:
Thank you, this is a sound comment. I have forwarded it to the main people behind this effort.

PR #1299, opened 2023-04-06
Update CI for newer Ubuntu and fix associated warnings

greghudson:
I'm still unsure of whether resolv-wrapper is expected to work on Ubuntu 22.04 (see https://bugzilla.samba.org/show_bug.cgi?id=14830 ).  I may need to find an alternative way to test URI discovery.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,17 +14,17 @@ jobs:
                 name: [linux-clang, linux-clang-openssl, linux-gcc]
                 include:
                     - name: linux-clang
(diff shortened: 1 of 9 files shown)


greghudson:
The libresolv-wrapper issue appears to be transient (the package just needs to be rebuilt against the new libc), so I shouldn't need to rewrite the URI discovery tests.  I filed this bug report in hopes of getting the package rebuilt:

https://bugs.launchpad.net/ubuntu/+source/resolv-wrapper/+bug/2015570

but failing that we can just wait for 24.04.

PR #1300, opened 2023-04-25
Defer primary KDC lookups

greghudson:
Add an internal variant of krb5_sendto_kdc() which records the answering KDC in a history object.  Callers can check the history object for replica KDC use after the success or failure of the KDC exchange is determined, using only DNS query per realm and transport when DNS queries are needed.

Preserve the current signature of krb5_sendto_kdc() (it is used within the tree outside of libkrb5, and might be used by other software despite being non-public), but remove the behavior of setting *use_primary.

See https://krbdev.mit.edu/rt/Ticket/Display.html?id=7721 for more information.  This solution is a little higher on complexity than I would like, and I need to think about whether I can add any tests for it.  The best alternative design I could think of (the second candidate) would bleed some complexity into the future locate interface by adding the concept of an "only if it's cheap" primary KDC lookup, and is more approximate.

### DIFF
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -203,8 +203,6 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
     TRACE(c, "Attempting password change; {int} tries remaining", tries)
 #define TRACE_GIC_PWD_EXPIRED(c)                                \
     TRACE(c, "Principal expired; getting changepw ticket")
-#define TRACE_GIC_PWD_PRIMARY(c)                        \
-    TRACE(c, "Retrying AS request with primary KDC")
 
 #define TRACE_GSS_CLIENT_KEYTAB_FAIL(c, ret)                            \
     TRACE(c, "Unable to resolve default client keytab: {kerr}", ret)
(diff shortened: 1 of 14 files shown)


yangxianrui:
你发的邮件已经收到！
Thanks with Best regards,

greghudson:
I realized that the PR changes could break the krb5_get_init_creds_password() password-change logic.  The new code does not set use_primary if we happen to contact the primary KDC on the first attempt, so in that case it does not restrict the post-password-change AS-REQ to the primary KDC.  The new k5_kdclist_any_replicas() does not distinguish between primary KDCs and KDCs in a realm with no primary KDCs, so there is no longer a foolproof way to set this flag based on the first attempt.

My fix is to move the fallback to k5_get_init_creds(), so that the post-password-change AS-REQ falls back if necessary.  (This also adds fallback to a few legacy functions and k5_identify_realm(), which should be harmless or modestly beneficial.)  We considered moving the fallback to the get_in_tkt layer long ago in https://krbdev.mit.edu/rt/Ticket/Display.html?id=1505 .  The value of that move has diminished today because gic_keytab.c no longer uses k5_get_init_creds(), but I think it's still a step in the right direction.

PR #1301, opened 2023-05-12
PKINIT ECDH support

greghudson:
I've been wanting to implement ECC PKINIT support (RFC 5349) for a long time, and some concerns about OpenSSL's finite-field DH support gave me the impetus to put some work into an alternative.  Currently I have a proof of concept, but need to work out some details of the user experience before it will be ready for merging.

The biggest problem is how we decide when to use ECDH on the client, how ECDH support interacts with the pkinit_dh_min_bits variable, and how ECDH support interacts with TD-DH-PARAMETERS negotiation.  Some notes on that:

* Heimdal uses ECDH if the client cert has an EC public key.  This is nice in that it puts the decision under control of the client CA, but it doesn't address anonymous PKINIT, and support for EC certs doesn't actually imply KDC support for ECDH (the current MIT PKINIT code works fine with EC certs, empirically, but does not support ECDH).  Heimdal does not implement TD-DH-PARAMETERS.

* The current TD-DH-PARAMETERS negotiation code in the client does not allow downgrades in bit strength.  (It does allow DH parameters that aren't the normally-used Oakley groups, as long as they check out with EVP_PKEY_param_check()/DH_check().)

* RFC 5349 gives a rough bit-strength correspondence table between ECC curve sizes and RSA key sizes.  I'm not sure how accurate that table is and whether it also applies to finite-field DH key sizes, but one option would be to do some additional research and assign FFDH bit equivalents to P-256, P-384, and P-521.  P-256 would probably land in between the 2048-bit and 4096-bit Oakley groups.

* Bit strength is not a great configuration knob.  My intuition is that P-256 is the most secure group out of any available option (except X25519, which has no PKINIT standard) because it has had the most effort focused on its OpenSSL implementation and its bit strength is more than sufficient.

There are also some OpenSSL bug concerns, specifically for the client operation where we combine the KDC public DH key with the client's key parameters to produce an EVP_PKEY object (compose_dh_pkey()).  OpenSSL 1.1 had a bug prior to 1.1.1b where d2i_PublicKey() didn't work for EC keys, but there is a wordier equivalent using EC_KEY functions.  OpenSSL 3.0.0 reintroduced that bug, and in 3.0 there is (I think) no working equivalent that doesn't involve deprecated functions.  Fortunately, this bug was fixed just a few months later in 3.0.1, and we might be able to get away with being optimistic that few people would link against 3.0.0 specifically (as Heimdal appears to).  The PR currently uses the alternative EC_KEY sequence for 1.0/1.1 and d2i_PublicKey() for 3.0.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_constants.c
+++ b/src/plugins/preauth/pkinit/pkinit_constants.c
@@ -320,6 +320,33 @@ static const uint8_t o4096[] = {
     0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF
 };
 
+/* Named curve prime256v1 (1.2.840.10045.3.1.7) as parameters for RFC 3279
+ * section 2.3.5 id-ecPublicKey */
+static const uint8_t p256[] = {
+    0x06, 0x08, 0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x03, 0x01, 0x07
+};
+
+/* Named curve secp384r1 (1.3.132.0.34, from RFC 5480 section 2.1.1.1) as
+ * parameters for RFC 3279 section 2.3.5 id-ecPublicKey */
+static const uint8_t p384[] = {
+    0x06, 0x05, 0x2B, 0x81, 0x04, 0x00, 0x22
+};
+
+/* Named curve secp521r1 (1.3.132.0.35, from RFC 5480 section 2.1.1.1) as
+ * parameters for RFC 3279 section 2.3.5 id-ecPublicKey */
+static const uint8_t p521[] = {
+    0x06, 0x05, 0x2B, 0x81, 0x04, 0x00, 0x23
+};
+
 const krb5_data oakley_1024 = { KV5M_DATA, sizeof(o1024), (char *)o1024 };
 const krb5_data oakley_2048 = { KV5M_DATA, sizeof(o2048), (char *)o2048 };
 const krb5_data oakley_4096 = { KV5M_DATA, sizeof(o4096), (char *)o4096 };
+const krb5_data ec_p256 = { KV5M_DATA, sizeof(p256), (char *)p256 };
+const krb5_data ec_p384 = { KV5M_DATA, sizeof(p384), (char *)p384 };
+const krb5_data ec_p521 = { KV5M_DATA, sizeof(p521), (char *)p521 };
+
+/* RFC 3279 section 2.3.3 dhpublicnumber (1.2.840.10046.2.1) */
+const krb5_data dh_oid = { 0, 7, "\x2A\x86\x48\xce\x3e\x02\x01" };
+
+/* RFC 3279 section 2.3.5 id-ecPublicKey (1.2.840.10045.2.1) */
+const krb5_data ec_oid = { 0, 7, "\x2A\x86\x48\xCE\x3D\x02\x01" };
--- a/doc/admin/conf_files/kdc_conf.rst
+++ b/doc/admin/conf_files/kdc_conf.rst
@@ -768,8 +768,11 @@ For information about the syntax of some of these options, see
     be specified multiple times.
 
 **pkinit_dh_min_bits**
-    Specifies the minimum number of bits the KDC is willing to accept
-    for a client's Diffie-Hellman key.  The default is 2048.
+    Specifies the minimum strength of Diffie-Hellman group the KDC is
+    willing to accept for key exchange.  Valid values in order of
+    increasing strength are 1024, 2048, P-256, 4096, P-384, and P-521.
+    The default is 2048.  (P-256, P-384, and P-521 are new in release
+    1.22.)
 
 **pkinit_allow_upn**
     Specifies that the KDC is willing to accept client certificates
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -1128,9 +1128,10 @@ PKINIT krb5.conf options
         option is not recommended.
 
 **pkinit_dh_min_bits**
-    Specifies the size of the Diffie-Hellman key the client will
-    attempt to use.  The acceptable values are 1024, 2048, and 4096.
-    The default is 2048.
+    Specifies the group of the Diffie-Hellman key the client will
+    attempt to use.  The acceptable values are 1024, 2048, P-256,
+    4096, P-384, and P-521.  The default is 2048.  (P-256, P-384, and
+    P-521 are new in release 1.22.)
 
 **pkinit_identities**
     Specifies the location(s) to be used to find the user's X.509
--- a/src/plugins/preauth/pkinit/pkinit.h
+++ b/src/plugins/preauth/pkinit/pkinit.h
@@ -59,6 +59,10 @@
 
 #define PKINIT_DEFAULT_DH_MIN_BITS  2048
 #define PKINIT_DH_MIN_CONFIG_BITS   1024
+/* Rough finite-field bit strength equivalents for the elliptic curve groups */
+#define PKINIT_DH_P256_BITS         3072
+#define PKINIT_DH_P384_BITS         7680
+#define PKINIT_DH_P521_BITS         15360
 
 #define KRB5_CONF_KDCDEFAULTS                   "kdcdefaults"
 #define KRB5_CONF_LIBDEFAULTS                   "libdefaults"
@@ -100,8 +104,6 @@ static inline void pkiDebug (const char *fmt, ...) { }
     (k5d)->length = (pad)->length; (k5d)->data = (char *)(pad)->contents;
 #define OCTETDATA_TO_KRB5DATA(octd, k5d) \
     (k5d)->length = (octd)->length; (k5d)->data = (char *)(octd)->data;
-
-extern const krb5_data dh_oid;
 
 /*
  * notes about crypto contexts:
--- a/src/plugins/preauth/pkinit/pkinit_clnt.c
+++ b/src/plugins/preauth/pkinit/pkinit_clnt.c
@@ -774,7 +774,7 @@ pkinit_client_profile(krb5_context context,
                       const krb5_data *realm)
 {
     const char *configured_identity;
-    char *eku_string = NULL;
+    char *eku_string = NULL, *minbits = NULL;
 
     pkiDebug("pkinit_client_profile %p %p %p %p\n",
              context, plgctx, reqctx, realm);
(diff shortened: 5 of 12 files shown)


yangxianrui:
你发的邮件已经收到！
Thanks with Best regards,

greghudson:
I decided to add the strings "P-256", "P-384", and "P-521" as valid values for pkinit_dh_min_bits, and to assign them rough finite-field bit strength equivalents.  Based on these bit strengths the client can download from P-256 to 4096-bit DH if the KDC doesn't implement ECC PKINIT.  For now the client's default is still 2048-bit DH.  The KDC proposes P-256 > P-384 > P-521 > 2048-bit DH > 4096-bit DH > 1024-bit DH (as pruned by dh_min_bits) when constructing TD-DH-PARAMETERS.

I still need to add test cases and update the documentation.

greghudson:
CMS algorithms are another part of RFC 5349, which I iintend to handle in a separate PR.  I should be more careful about calling this PR and commit "ECDH support" and not "ECC support"; I'll fix that.

I believe the FFDH strength equivalents in the PR are consistent with FIPS1402IG.  The NIST document is a little harder to check as it separately translates finite-field and elliptic curve algorithms to ranges of brute-force bit strengths.

jrisc:
I tried to force use of the EC groups by setting `pkinit_dh_min_bits` accordingly, but AD still responds with SHA-1-digested CMS data (`invalid digest` because SHA-1 is not allowed by OpenSSL):
```
$ KRB5_TRACE=/dev/stderr kinit -X X509_anchors=FILE:cacert.pem -X X509_user_identity=FILE:user.pem,userkey.pem user </dev/null
[...]: Matching user@ACTIVE.DIRECTORY in collection with result: -1765328243/Can't find client principal user@ACTIVE.DIRECTORY in cache collection
[...]: Getting initial credentials for user@ACTIVE.DIRECTORY
[...]: Sending unauthenticated request
[...]: Sending request (210 bytes) to ACTIVE.DIRECTORY
[...]: Resolving hostname 1.1.1.1
[...]: Sending initial UDP request to dgram 1.1.1.1:88
[...]: Received answer (201 bytes) from dgram 1.1.1.1:88
[...]: Response was not from primary KDC
[...]: Received error from KDC: -1765328359/Additional pre-authentication required
[...]: Preauthenticating using KDC method data
[...]: Processing preauth types: PA-PK-AS-REQ (16), PA-PK-AS-REP_OLD (15), PA-ETYPE-INFO2 (19), PA-ENC-TIMESTAMP (2)
[...]: Selected etype info: etype aes256-cts, salt "ACTIVE.DIRECTORYuser", params ""
[...]: PKINIT loading identity FILE:/tmp/user.pem,/tmp/userkey.pem
[...]: PKINIT loading CA certs and CRLs from FILE /tmp/cacert.pem
[...]: PKINIT client computed kdc-req-body checksum 14/23C70AE4273FCA76F870F94D8FB7C3CB3B4E097E
[...]: PKINIT client making DH request
[...]: PKINIT using P-521 key exchange group
[...]: PKINIT chain cert #0: /DC=directory/DC=active/CN=Users/CN=user
[...]: Preauth module pkinit (16) (real) returned: 0/Success
[...]: Produced preauth for next request: PA-PK-AS-REQ (16)
[...]: Sending request (2970 bytes) to ACTIVE.DIRECTORY
[...]: Resolving hostname 1.1.1.1
[...]: Initiating TCP connection to stream 1.1.1.1:88
[...]: Sending TCP request to stream 1.1.1.1:88
[...]: Received answer (4264 bytes) from stream 1.1.1.1:88
[...]: Terminating TCP connection to stream 1.1.1.1:88
[...]: Response was not from primary KDC
[...]: Processing preauth types: PA-PK-AS-REP (17)
[...]: PKINIT OpenSSL error: Failed to verify CMS message
[...]: PKINIT OpenSSL error: error:03000098:digital envelope routines::invalid digest
[...]: PKINIT client could not verify DH reply
[...]: Preauth module pkinit (17) (real) returned: -1765328304/Failed to verify CMS message: invalid digest
...
```

If AD takes `supportedCMSTypes` into account, it should be possible to push it to use [one of the X9.62 signature types](https://www.rfc-editor.org/rfc/rfc5349.html#section-3). I will make more tests in that direction.

To be mentioned that RFC5349 removes the ambiguity that was debated in the [OpenSSL issue](https://github.com/openssl/openssl/issues/18729), regarding CMS digest and signature types. Since here it is clearly stated that `digestAlgorithm` and `signatureAlgorithm` MUST match:
```
Namely, id-sha1 MUST be used in conjunction with ecdsa-with-Sha1,
id-sha256 MUST be used in conjunction with ecdsa-with-Sha256,
id-sha384 MUST be used in conjunction with ecdsa-with-Sha384, and
id-sha512 MUST be used in conjunction with ecdsa-with-Sha512.
```

src/plugins/preauth/pkinit/pkinit_constants.c
```diff
@@ -320,6 +320,33 @@ static const uint8_t o4096[] = {
...
+ * 2.3.5 id-ecPublicKey */
+static const uint8_t p384[] = {
+    0x06, 0x05, 0x2B, 0x81, 0x04, 0x00, 0x22
+};
+
+/* Named curve secp521r1 (1.3.132.0.35) as parameters for RFC 3279 section
+ * 2.3.5 id-ecPublicKey */
+static const uint8_t p521[] = {
+    0x06, 0x05, 0x2B, 0x81, 0x04, 0x00, 0x23
+};
```
jrisc, line 339: These 2 curves are actually defined in [RFC5480](https://datatracker.ietf.org/doc/html/rfc5480#section-2.1.1.1). Only `prime256v1` is defined in [RFC3279](https://datatracker.ietf.org/doc/html/rfc3279#page-23)

greghudson: I'll add an RFC 5480 reference.  They are still here as parameters for id-ecPublicKey from RFC 3279 (which just means they have a DER OBJECT IDENTIFIER tag and length rather than being the untagged encoding).

jrisc:
Please do not pay attention to my message about the digest type. I made a confusion, this is unrelated since it depends of the certificate type.

I tested each curve with Active Directory using a RSA certificate and OpenSSL 3.0.7. All of them are working.

I made the same tests using an EC certificate, but they result in a `KRB_AP_ERR_MODIFIED` error:
```
$ KRB5_TRACE=/dev/stderr kinit -X X509_anchors=FILE:/tmp/cacert.pem -X X509_user_identity=FILE:/tmp/Administrator.pem,/tmp/Administratorkey.pem Administrator </dev/null
[1784] 1688056674.599459: Matching Administrator@WIN2019-O0GI.TEST in collection with result: -1765328243/Can't find client principal Administrator@WIN2019-O0GI.TEST in cache collection
[1784] 1688056674.599460: Getting initial credentials for Administrator@WIN2019-O0GI.TEST
[1784] 1688056674.599462: Sending unauthenticated request
[1784] 1688056674.599463: Sending request (210 bytes) to WIN2019-O0GI.TEST
[1784] 1688056674.599464: Resolving hostname 1.1.1.1
[1784] 1688056674.599465: Sending initial UDP request to dgram 1.1.1.1:88
[1784] 1688056674.599466: Received answer (201 bytes) from dgram 1.1.1.1:88
[1784] 1688056674.599467: Response was not from primary KDC
[1784] 1688056674.599468: Received error from KDC: -1765328359/Additional pre-authentication required
[1784] 1688056674.599471: Preauthenticating using KDC method data
[1784] 1688056674.599472: Processing preauth types: PA-PK-AS-REQ (16), PA-PK-AS-REP_OLD (15), PA-ETYPE-INFO2 (19), PA-ENC-TIMESTAMP (2)
[1784] 1688056674.599473: Selected etype info: etype aes256-cts, salt "WIN2019-O0GI.TESTAdministrator", params ""
[1784] 1688056674.599474: PKINIT loading identity FILE:/tmp/Administrator.pem,/tmp/Administratorkey.pem
[1784] 1688056674.599475: PKINIT loading CA certs and CRLs from FILE /tmp/cacert.pem
[1784] 1688056674.599476: PKINIT client computed kdc-req-body checksum 14/06A3C002AFA4EB08D2190F1BE7F787C0619E8CA5
[1784] 1688056674.599478: PKINIT client making DH request
[1784] 1688056674.599479: PKINIT using P-256 key exchange group
[1784] 1688056674.599480: PKINIT chain cert #0: /DC=test/DC=win2019-o0gi/CN=Users/CN=Administrator
[1784] 1688056674.599481: Preauth module pkinit (16) (real) returned: 0/Success
[1784] 1688056674.599482: Produced preauth for next request: PA-PK-AS-REQ (16)
[1784] 1688056674.599483: Sending request (2508 bytes) to WIN2019-O0GI.TEST
[1784] 1688056674.599484: Resolving hostname 1.1.1.1
[1784] 1688056674.599485: Initiating TCP connection to stream 1.1.1.1:88
[1784] 1688056674.599486: Sending TCP request to stream 1.1.1.1:88
[1784] 1688056674.599487: Received answer (106 bytes) from stream 1.1.1.1:88
[1784] 1688056674.599488: Terminating TCP connection to stream 1.1.1.1:88
[1784] 1688056674.599489: Response was not from primary KDC
[1784] 1688056674.599490: Received error from KDC: -1765328343/Message stream modified
[1784] 1688056674.599491: Retrying AS request with primary KDC
[1784] 1688056674.599492: Getting initial credentials for Administrator@WIN2019-O0GI.TEST
[1784] 1688056674.599494: Sending unauthenticated request
[1784] 1688056674.599495: Sending request (210 bytes) to WIN2019-O0GI.TEST (primary)
kinit: Message stream modified while getting initial credentials
```

I think it indicates a checksum verification is failing somewhere. It happens with both ECDH and MODP DH.

@greghudson You mentioned in the PR description that PKINIT pre-auth works with EC certificates. May I be missing something? (This is not directly related to this PR, so if there is no trivial solution I should probably open a dedicated issue).

greghudson:
> You mentioned in the PR description that PKINIT pre-auth works with EC certificates. May I be missing something? 

It works with an MIT client and KDC, in my tests.  I don't have an immediate guess as to why a Windows KDC wouldn't like it.  I think with an EC certificate we will use ecdsa-with-sha256 in the SignedData SignerInfo, based on create_signature() and my understanding of OpenSSL.  That seems like it should be okay.

Thanks for testing the ECDH support against Windows.

greghudson:
I did find this in [MS-PKCA] section 1.5:
```
PKCA assumes the following, in addition to any assumptions specified in [MS-KILE]:

1. The key distribution center (KDC) has an X.509 public key certificate [X509], issued by a
certificate authority (CA) and trusted by the clients in the Kerberos realm. For ECC support,
the KDC has an ECC public key certificate issued by a CA and trusted by clients in the Kerberos
realm. The issuing of these [X509] certificates is not addressed in this protocol specification.
```
So it's possible that the Windows KDC failure arises from using an RSA KDC cert and an ECC client cert?  (I'm not sure if that's actually what you tested.)

jrisc:
It may be the reason, yes. I will make another test where I make sure the AD KDC is configured with an EC certificate. I opened another PR #1305 to add EC signatures to `supportedCMSTypes`. I will report what I can find there.

I think the present PR is ready to be merged.

PR #1302, opened 2023-05-14
Get arm64-windows builds working

geraldcombs:
Update various checks to include "ARM64" and "_M_ARM64".

Remove a /BASE flag which fixes an "ARM64 image cannot have base address below 4GB" error and which arguably shouldn't be used in modern times since it prevents using dynamic base addresses.

### DIFF
--- a/src/config/win-post.in
+++ b/src/config/win-post.in
@@ -63,7 +63,7 @@ check-windows: check-recurse
 !endif # WINSUBDIRS
 
 # Use 64-bit LIBNAME and OBJFILE on 64-bit platforms, if defined.
-!if ("$(CPU)" == "IA64") || ("$(CPU)" == "AMD64") || ("$(CPU)" == "ALPHA64")
+!if ("$(CPU)" == "IA64") || ("$(CPU)" == "AMD64") || ("$(CPU)" == "ALPHA64") || ("$(CPU)" == "ARM64")
 !if defined(WIN64LIBNAME)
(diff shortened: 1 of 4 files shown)


geraldcombs:
I haven't tested this beyond linking with Wireshark and verifying that it passes our Kerberos decryption unit test, but everything seems to work OK.

greghudson:
The brg_endian.h change concerns me slightly because there's an upstream we try not to deviate from, and because ARM supports big-endian and little-endian execution modes (although little-endian is by far the most common).  I don't have a better alternative suggestion, so I guess we can make that change anyway.

The removal of the -base flag also concerns me slightly, as I don't know why it's there.  It was added long ago (1996) without a specific explanation.  Do you have any links to documentation on what this linker flag does?

geraldcombs:
Thanks for reviewing this. As far as I can tell, `_M_ARM64` is only defined by Visual Studio and Visual Studio only generates little-endian code. Maybe `_MSC_VER` would be a better check?

`/BASE` is documented at https://learn.microsoft.com/en-us/cpp/build/reference/base-base-address. If the code in question was added in 1996, maybe it was to handle a DLL rebasing issue on Windows 95? https://devblogs.microsoft.com/oldnewthing/?p=36953

greghudson:
I think that information addresses both of my concerns (no need to change the AES test).  I will merge this soon if I don't find any other issues.

PR #1303, opened 2023-05-23
Enable PKINIT plugin if at least 1 of the 3 well-known groups is available

jrisc:
OpenSSL may no longer allow to decode non-well-known Diffie-Hellman groups as EVP_PKEY objects in FIPS mode. However, OpenSSL does not know about the MODP well-known group 2 (1024-bit), which is [considered as a custom group](https://github.com/openssl/openssl/issues/18981). As a consequence, the PKINIT plugin failed to load systematically in FIPS mode because the 3 groups are required.

This pull request allows initialization of the PKINIT plugin if at least one of the MODP well-known groups succeed to load. So the PKINIT plugin will continue to work with groups 14 (2048-bit) and 16 (4096) in FIPS mode.

### DIFF
--- a/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
+++ b/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
@@ -47,7 +47,8 @@
 static krb5_error_code pkinit_init_pkinit_oids(pkinit_plg_crypto_context );
 static void pkinit_fini_pkinit_oids(pkinit_plg_crypto_context );
 
-static krb5_error_code pkinit_init_dh_params(pkinit_plg_crypto_context );
+static krb5_error_code pkinit_init_dh_params(krb5_context,
+                                             pkinit_plg_crypto_context);
 static void pkinit_fini_dh_params(pkinit_plg_crypto_context );
 
 static krb5_error_code pkinit_init_certs(pkinit_identity_crypto_context ctx);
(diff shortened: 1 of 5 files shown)


src/plugins/preauth/pkinit/pkinit_crypto_openssl.c
```diff
@@ -39,15 +39,16 @@
 #include <arpa/inet.h>
 
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
-#include <openssl/core_names.h>
+#include <openssl/evp.h>
```
greghudson, line 42: Why this change?

(There is some accumulated debt here because two of these includes are also done in pkinit_crypto_openssl.h.  And that header should probably go away because it's only included by this one source file.  But I don't think either of those problems is related to this PR.)

jrisc: I remove the inclusion of `openssl/evp.h`, it was already imported in `pkinit_crypto_openssl.c` indeed. It is the same for `openssl/core_names.h`, so I removed it too.

jrisc: Maybe you prefer include operations to remain untouched in this PR. It that's the case, just let me know and I will leave it as it was.

greghudson: Let's leave them alone for now (assuming I'm correct that they aren't needed for the PR), and I'll plan to do a cleanup PR (folding the header file into the source file) after this and the ECC work are done.

greghudson:
I pushed some small edits to the PR branch.

PR #1304, opened 2023-06-10
Switch to modern (ANSI) function prototypes

kenh:
The latest clang on MacOS X has started to report these warnings:

../../include/kadm5/kadm_rpc.h:390:15: warning: a function declaration without a
 prototype is deprecated in all versions of C and is not supported in C2x [-Wdep
recated-non-prototype]

This is an attempt to address this problem (I did confirm that C2x is officially dropping support for K&R prototypes).  This does not address the isses in the RPC code; unfortunately it seems xdrproc_t sometimes takes two arguments and sometimes takes three so I am unsure how to fix that.  This should address all missing prototypes that are not RPC-related. I was surprised there were so many!

### DIFF
--- a/src/appl/gss-sample/gss-client.c
+++ b/src/appl/gss-sample/gss-client.c
@@ -75,7 +75,7 @@ static gss_OID_desc gss_spnego_mechanism_oid_desc =
 {6, (void *)"\x2b\x06\x01\x05\x05\x02"};
 
 static void
-usage()
+usage(void)
 {
     fprintf(stderr, "Usage: gss-client [-port port] [-mech mechanism] "
             "[-spnego] [-d]\n");
@@ -359,9 +359,7 @@ client_establish_context(int s, char *service_name, OM_uint32 gss_flags,
 }
 
 static void
-read_file(file_name, in_buf)
-    char   *file_name;
-    gss_buffer_t in_buf;
+read_file(char *file_name, gss_buffer_t in_buf)
 {
     int     fd, count;
     struct stat stat_buf;
@@ -431,21 +429,10 @@ read_file(file_name, in_buf)
  * verifies it with gss_verify.  -1 is returned if any step fails,
  * otherwise 0 is returned.  */
 static int
-call_server(host, port, oid, service_name, gss_flags, auth_flag,
-            wrap_flag, encrypt_flag, mic_flag, v1_format, msg, use_file,
-            mcount, username, password)
-    char   *host;
-    u_short port;
-    gss_OID oid;
-    char   *service_name;
-    OM_uint32 gss_flags;
-    int     auth_flag, wrap_flag, encrypt_flag, mic_flag;
-    int     v1_format;
-    char   *msg;
-    int     use_file;
-    int     mcount;
-    char    *username;
-    char    *password;
+call_server(char *host, u_short port, gss_OID oid, char *service_name,
+            OM_uint32 gss_flags, int auth_flag, int wrap_flag,
+            int encrypt_flag, int mic_flag, int v1_format, char *msg,
+            int use_file, int mcount, char *username, char *password)
 {
     gss_ctx_id_t context = GSS_C_NO_CONTEXT;
     gss_buffer_desc in_buf, out_buf;
@@ -774,9 +761,7 @@ worker_bee(void *unused)
 }
 
 int
-main(argc, argv)
-    int     argc;
-    char  **argv;
+main(int argc, char **argv)
 {
     int     i;
 
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -545,7 +545,7 @@ if test "$GCC" = yes ; then
     TRY_WARN_CC_FLAG(-Wno-format-zero-length)
     # Other flags here may not be supported on some versions of
     # gcc that people want to use.
-    for flag in overflow strict-overflow missing-format-attribute missing-prototypes return-type missing-braces parentheses switch unused-function unused-label unused-variable unused-value unknown-pragmas sign-compare newline-eof error=uninitialized no-maybe-uninitialized error=pointer-arith error=int-conversion error=incompatible-pointer-types error=discarded-qualifiers error=implicit-int ; do
+    for flag in overflow strict-overflow missing-format-attribute missing-prototypes return-type missing-braces parentheses switch unused-function unused-label unused-variable unused-value unknown-pragmas sign-compare newline-eof error=uninitialized no-maybe-uninitialized error=pointer-arith error=int-conversion error=incompatible-pointer-types error=discarded-qualifiers error=implicit-int error=strict-prototypes; do
       TRY_WARN_CC_FLAG(-W$flag)
     done
     #  old-style-definition? generates many, many warnings
--- a/src/appl/gss-sample/gss-misc.c
+++ b/src/appl/gss-sample/gss-misc.c
@@ -157,10 +157,7 @@ read_all(int fildes, void *data, unsigned int nbyte)
  * if an error occurs or if it could not write all the data.
  */
 int
-send_token(s, flags, tok)
-    int     s;
-    int     flags;
-    gss_buffer_t tok;
+send_token(int s, int flags, gss_buffer_t tok)
 {
     int     ret;
     unsigned char char_flags = (unsigned char) flags;
@@ -230,10 +227,7 @@ send_token(s, flags, tok)
  * and -1 if an error occurs or if it could not read all the data.
  */
 int
-recv_token(s, flags, tok)
-    int     s;
-    int    *flags;
-    gss_buffer_t tok;
+recv_token(int s, int *flags, gss_buffer_t tok)
 {
     int     ret;
     unsigned char char_flags;
@@ -303,10 +297,7 @@ recv_token(s, flags, tok)
 }
 
 static void
-display_status_1(m, code, type)
-    char   *m;
-    OM_uint32 code;
-    int     type;
+display_status_1(char *m, OM_uint32 code, int type)
 {
     OM_uint32 min_stat;
     gss_buffer_desc msg;
@@ -344,10 +335,7 @@ display_status_1(m, code, type)
  * followed by a newline.
  */
 void
-display_status(msg, maj_stat, min_stat)
-    char   *msg;
-    OM_uint32 maj_stat;
-    OM_uint32 min_stat;
+display_status(char *msg, OM_uint32 maj_stat, OM_uint32 min_stat)
 {
     display_status_1(msg, maj_stat, GSS_C_GSS_CODE);
     display_status_1(msg, min_stat, GSS_C_MECH_CODE);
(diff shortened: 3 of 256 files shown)


src/appl/gss-sample/gss-client.c
```diff
@@ -75,7 +75,7 @@ static gss_OID_desc gss_spnego_mechanism_oid_desc =
 {6, (void *)"\x2b\x06\x01\x05\x05\x02"};
 
 static void
-usage()
+usage(void)
```
greghudson, line 78: Does this specific definition actually cause a warning?  Although empty parens in a declaration makes the difference between a prototype and an old-style declaration, in a definition I think it's just a stylistic difference.

kenh: That does not cause a warning, that was more of "well, since I'm in there ..." kind of thing.  Also, I did notice that the one build failed, but I am not quite sure why; it looks like it failed one of the unit tests but I'm not sure how these changed would have caused it.

greghudson: I will look into the build failure and see if it's related.

greghudson:
We had a test build failure in the linux-clang build (which also uses asan).  The output is:
```
LD_LIBRARY_PATH=`echo -L../../../lib | sed -e "s/-L//g" -e "s/ /:/g"` KRB5_CONFIG=../../../config-files/krb5.conf LC_ALL=C  ./t_cc
Testing miscellaneous error conditions
second remove (second pass): Credentials cache I/O operation failed 
Starting test on /tmp/cctest.14450
Test on /tmp/cctest.14450 passed
Starting test on KEYRING:process:/tmp/cctest.14450
```
t_cc.c uses a mix of stdout and stderr messages, and we're seeing the stderr messages first.  Were the messages in order, I think the "second remove" message would have come last, after "Starting test on KEYRING...".  The error indicates that one of several krb5_cc_remove_cred() operations failed with KRB5_CC_IO.  Looking at krcc_remove_cred() and the three instances of KRB5_CC_IO in cc_keyring.c, I think that means krcc_start_seq_get() called keyctl_read_alloc() and got an error.  We can't tell what errno was because we only have a DEBUG_PRINTF() there (we should have an extended error message instead).

I haven't seen this build failure before.  It seems unlikely to be related to the PR, and unlikely to be specific to the asan build given where the error originates.  I will restart the build job and keep an eye out for it happening again.

greghudson:
I looked into how other implementations handle xdrproc_t.  The most recent public Solaris code still uses an old-style type declaration, but libtirpc uses `typedef bool_t (*xdrproc_t)(XDR *, ...);` and glibc uses `typedef bool_t (*xdrproc_t) (XDR *, void *,...);`.

Most RPC implementations do sometimes invoke xdrproc_t functions with three arguments--but as far as I can tell, the third argument is always LASTUNSIGNED (2^32-1) and the functions in question probably always take only two arguments--for instance, xdr_array() calls `(*elproc)(xdrs, target, LASTUNSIGNED)` for each element, but every use of xdr_array() can find uses a two-argument elproc (like xdr_int()).  There appear to be platform ABI assumptions that you can pass an extra argument to a function and it will be ignored, and that you can call through a function type prototyped with `...` when the actual function has a fixed number of arguments. There are some three-argument xdr_* functions in the library such as xdr_opaque(), but I don't think those are ever used as xdrproc_t values.  I think the three-argument form probably has something to do with the use of RPC in the Solaris kernel.  libtirpc appears to have removed these third arguments; its xdr_array.c just calls `(*elproc)(xdrs, target)`.  glibc has not.

Even if we could simplify out the third arguments (and we probably can, followig libtirpc's lead), we would run into the problem that xdr_* functions do not use the same pointer type for their second parameters.  xdr_int() takes an `int *`, xdr_keyblock() takes a `krb5_keyblock *`, etc..  There appears to be another platform ABI assumption that pointer types are interchangeable without compiler-time typecasts.  In our version of the RPC library, any modern declaration of xdrproc_t would result in warnings whenever a concrete xdr_* function is used as an xdrproc_t value, such as in authunix_prot.c where xdr_int is passed to xdr_array().  The solution in other implementations appears to be to put in `(xdrproc_t)` casts whenever this is done.

Assuming we can't ignore the problem, the best solution is probably to follow the lead of glibc/libtirpc by prototype-declaring xdrproc_t and adding a ton of typecasts.  Optionally we could take out those LASTUNSIGNED third arguments and use a two-parameter typecast without a `...` (I think libtirpc could do this even though it doesn't).  The only other solution I can think of is the nuclear option of ditching the RPC API; Heimdal is a proof of concept for this (it has some MIT kadmin wire compatibility built around krb5_storage_*, I believe).  But of course Heimdal never had a libgssrpc and therefore didn't have to worry that someone might be using it besides kadmin.

greghudson:
I pushed a second commit doing what I said above.  Is there more to do in lib/rpc besides xdrproc_t?

kenh:
I just committed a change that included a few extra missing prototypes in lib/rpc that I think should be non-controversial (there were existing prototypes in header files, it's just that some functions didn't use them).

There is one additional thing that might be a small mess.  In ```kadm_rpc_svc.c```, ```rpc_test_svc.c```, and ```ipropd_svc.c```, there is a declaration for ```local``` that triggers the warning:

```
     bool_t (*local)();
```

(I am aware there are also declarations for ```xdr_argument``` and ```xdr_result``` as function pointers without prototyped arguments, no idea why those don't trigger warnings). The first argument of "local" is a union of a bunch of things like ```chrand_arg```, and all of the kadm functions just assume that they take ```chrang_arg``` (or whatever) as the first argument.  I would suspect that the proper way of fixing this is changing local so it either takes a ```void *``` as the first argument or a pointer to the actual union as the first argument and then doing the appropriate thing in each RPC service function. I could do one of those, or if you prefer we could just defer fixing that at this time (the unit test has a slightly different issue, but it's similar).

kenh:
Sigh, I wasn't paying attention that you had fixed up the xdrproc_t stuff (even though you said you did that). So I went back and tracked down the rest of those warnings. Aside from the ones I mentioned in my previous message, the biggest one was a bunch of non-argument function prototypes in ```kadm_rpc.h```. But the functions were completely prototyped in ```admin_xdr.h``` (no idea why that was). I ended up removing the definitions from ```kadm_rpc.h``` and adding an #include for ```admin_xdr.h``` for the few files that didn't include both header files.

greghudson:
> it might be a lot of work but would it be possible to declare every concrete xdr_* function as only having two arguments, the second argument being void * and doing a typecast inside of the concrete functions to the correct type?

This is probably how XDR would work if it were designed to be conformant to ANSI C.  Getting there is tough because type-specific XDR functions are either part of a public API with multiple implementations or are generated by rpcgen.  (Well, in theory.  A lot of ours are hand-rolled.)  The rpcgen-generated code could be updated gradually, at least.

> `bool_t (*local)();`

I will see if any rpcgen implementations have a solution for this.

kenh:
> This is probably how XDR would work if it were designed to be conformant to ANSI C. Getting there is tough because type-specific XDR functions are either part of a public API with multiple implementations or are generated by rpcgen. (Well, in theory. A lot of ours are hand-rolled.) The rpcgen-generated code could be updated gradually, at least.

Fair enough; is the idea that if a new RPC function was added that you'd use rpcgen to generate the appropriate files? Or would the new RPC just be hand-rolled into the existing code? I see a RPC template file for iprop and the unit test, but nothing for kadmin.

greghudson:
The modern rpcgen standard for service boilerplate seems to be `xdrproc_t xdr_argument, xdr_result;` `bool_t (*local)(char *, void *, struct svc_req *);` `%s = (bool_t (*)(char *, void *, struct svc_req *))fname;`.  (You'd think a typedef would make that more readable, but I didn't find any versions of rpcgen that did so.)

kenh:
Hm. Alright, so ... do you want to fix this or defer it to later? If you want to fix it now, what would you prefer? I think the "most correct" solution would be something like:
```
bool_t
create_principal3_2_svc(void *generic_arg, generic_ret *ret,
                        struct svc_req *rqstp)
{
             cprinc3_arg *arg = generic_arg;
```

Simpler would be to declare ```local``` as ```(void *, void *, struct svc_req *)``` and just cast the function pointer to that. I know you said that the standard boilerplate has the first argument as ```char *``` but that doesn't seem correct, given we want a generic pointer?

greghudson:
I would leave the concrete service functions alone; the goal for now is to have a clean build under recent clang (or maybe clang as configured on macOS, not sure which), not to remove the platform ABI assumptions in the RPC library.

I don't know exactly why rpcgen implementations use `char *` for the first parameter and `void *` for the second, when neither will match the concrete  function's parameter anyway.  There are actually two variations depending on whether you use the -M (multithreading) flag; the original version is `char *(*local)(char *, struct svc_req *);`.  Prior to commit 01d2bd6ab3f39163f6185ac4c69f6f87535212b4 we had stubs matching that pattern (except for pointer types).  Perhaps the thinking was that the return parameter is new with the multithreading flag and should use the ANSI C concept.  I would go along with the rpcgen implementations' declarations and casts even if they don't seem consistent.

kenh:
Alright, change pushed. I did a typedef for the service function pointer because I didn't want to repeat a long cast. Sadly, sometimes ```local``` takes two arguments, other times it takes three.

In the "clean build" vein, there are other warnings that might be more complicated, mostly related to alignment issues. But at least now there are much fewer.

greghudson:
While revising this for merge I thought to add -Werror=strict-prototypes to the build, after finding https://reviews.llvm.org/D122895 which is probably the source of of the macOS build warnings.  This flag catches a bunch of additional non-prototype function pointer types, which I'm still working through.

gcc's -Wstrict-prototypes (at least in 11.3) does not like definitions like `static void usage()` (as opposed to `static void usage(void)`), although it is okay with it for non-static functions.  I think this is a little unfortunate: C2x is removing support for non-prototype function declarations so that it can make `int fn();` mean a function with no arguments, but the compiler warning behavior makes us put that extra `void` in more places than is required currently.

greghudson:
With more research I found https://stackoverflow.com/a/24740582 which indicates that there actually _is_ a practical difference between `()` and `(void)` in the parameter list of a function definition--specifically, gcc will not enforce passing zero arguments to a function defined using `()` because it doesn't violate a "constraint" in the C spec, even though the compiler can easily detect undefined behavior.  (clang will issue a warning.)  That lends more support for using `(void)` for static function definitions.

(When reading the above, note carefully the difference between a function's declaration and its definition.  For declarations of functions with no parameters, it is more obviously required to use `(void)` for any version of the language spec before C2x, as `()` is treated as an old-style declaration giving no information about the function's parameters.)

kenh:
> With more research I found https://stackoverflow.com/a/24740582 which indicates that there actually _is_ a practical difference between `()` and `(void)` in the parameter list of a function definition--specifically, gcc will not enforce passing zero arguments to a function defined using `()` because it doesn't violate a "constraint" in the C spec, even though the compiler can easily detect undefined behavior. (clang will issue a warning.) That lends more support for using `(void)` for static function definitions.

That was my vague understanding as well (but the subtlety between a declaration and a definition I definitely did not know about).

PR #1306, opened 2023-07-13
Fix memory leaks

stevegrubb:
This patch fixes a number of memory leaks. Most are in the utilities but there are a few in the libraries.

### DIFF
--- a/src/clients/ksu/ccache.c
+++ b/src/clients/ksu/ccache.c
@@ -40,6 +40,18 @@ copies the default cache into the secondary cache,
 
 ************************************************************************/
 
+static void
+free_creds_list(krb5_context context, krb5_creds **list)
+{
+    size_t i;
+
+    if (list == NULL)
+        return;
+    for (i = 0; list[i]; i++)
+        krb5_free_creds(context, list[i]);
+    free(list);
+}
+
 void show_credential(krb5_context, krb5_creds *, krb5_ccache);
 
 /* modifies only the cc_other, the algorithm may look a bit funny,
@@ -53,20 +65,19 @@ krb5_ccache_copy(krb5_context context, krb5_ccache cc_def,
                  krb5_boolean restrict_creds, krb5_principal primary_principal,
                  krb5_boolean *stored)
 {
-    int i=0;
     krb5_error_code retval=0;
     krb5_creds ** cc_def_creds_arr = NULL;
     krb5_creds ** cc_other_creds_arr = NULL;
 
     if (ks_ccache_is_initialized(context, cc_def)) {
-        if((retval = krb5_get_nonexp_tkts(context,cc_def,&cc_def_creds_arr))){
-            return retval;
-        }
+        retval = krb5_get_nonexp_tkts(context, cc_def, &cc_def_creds_arr);
+        if (retval)
+            goto cleanup;
     }
 
     retval = krb5_cc_initialize(context, cc_target, target_principal);
     if (retval)
-        return retval;
+        goto cleanup;
 
     if (restrict_creds) {
         retval = krb5_store_some_creds(context, cc_target, cc_def_creds_arr,
@@ -79,22 +90,9 @@ krb5_ccache_copy(krb5_context context, krb5_ccache cc_def,
                                       cc_other_creds_arr);
     }
 
-    if (cc_def_creds_arr){
-        while (cc_def_creds_arr[i]){
-            krb5_free_creds(context, cc_def_creds_arr[i]);
-            i++;
-        }
-    }
-
-    i=0;
-
-    if(cc_other_creds_arr){
-        while (cc_other_creds_arr[i]){
-            krb5_free_creds(context, cc_other_creds_arr[i]);
-            i++;
-        }
-    }
-
+cleanup:
+    free_creds_list(context, cc_def_creds_arr);
+    free_creds_list(context, cc_other_creds_arr);
     return retval;
 }
 
@@ -184,32 +182,29 @@ krb5_get_nonexp_tkts(krb5_context context, krb5_ccache cc,
 {
 
     krb5_creds creds, temp_tktq, temp_tkt;
-    krb5_creds **temp_creds;
+    krb5_creds **temp_creds = NULL;
     krb5_error_code retval=0;
     krb5_cc_cursor cur;
     int count = 0;
     int chunk_count = 1;
 
-    if ( ! ( temp_creds = (krb5_creds **) malloc( CHUNK * sizeof(krb5_creds *)))){
-        return ENOMEM;
-    }
-
-
+    temp_creds = xcalloc(CHUNK, sizeof(*temp_creds));
     memset(&temp_tktq, 0, sizeof(temp_tktq));
     memset(&temp_tkt, 0, sizeof(temp_tkt));
     memset(&creds, 0, sizeof(creds));
 
     /* initialize the cursor */
-    if ((retval = krb5_cc_start_seq_get(context, cc, &cur))) {
-        return retval;
-    }
+    retval = krb5_cc_start_seq_get(context, cc, &cur);
+    if (retval)
+        goto cleanup;
 
     while (!(retval = krb5_cc_next_cred(context, cc, &cur, &creds))){
 
         if (!krb5_is_config_principal(context, creds.server) &&
             (retval = krb5_check_exp(context, creds.times))){
+            krb5_free_cred_contents(context, &creds);
             if (retval != KRB5KRB_AP_ERR_TKT_EXPIRED){
-                return retval;
+                goto cleanup;
             }
             if (auth_debug){
                 fprintf(stderr,"krb5_ccache_copy: CREDS EXPIRED:\n");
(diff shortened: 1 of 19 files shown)


src/lib/krb5/ccache/cc_kcm.c
```diff
@@ -972,8 +972,10 @@ kcm_start_seq_get(krb5_context context, krb5_ccache cache,
     if (ret == 0) {
         /* GET_CRED_LIST is available. */
         ret = kcmreq_get_cred_list(&req, &creds);
-        if (ret)
+        if (ret) {
+            free_cred_list(creds);
             goto cleanup;
+        }
```
greghudson, line 978: This is not a leak; kcmreq_get_cred_list() only allocates creds if it returns successfully.  kcmreq_get_cred_list() attempts to make this clear to static analysis tools by setting *creds_out to NULL on entry and only setting it to a non-null value just before returning 0.

src/plugins/kdb/ldap/libkdb_ldap/ldap_realm.c
```diff
@@ -275,8 +275,10 @@ krb5_ldap_delete_realm (krb5_context context, char *lrealm)
                     krb5_parse_name(context, values[i], &principal);
                     if (principal_in_realm_2(principal, lrealm) == 0) {
                         st=krb5_ldap_delete_principal(context, principal);
-                        if (st && st != KRB5_KDB_NOENTRY)
+                        if (st && st != KRB5_KDB_NOENTRY) {
+                            krb5_free_principal(context, principal);
                             goto cleanup;
+                        }
```
greghudson, line 281: `values` also leaks here.  We try to avoid repeating free statements in error handling blocks;  I will take a closer took at this function.

src/lib/krb5/keytab/kt_file.c
```diff
@@ -462,6 +462,8 @@ krb5_ktfile_start_seq_get(krb5_context context, krb5_keytab id, krb5_kt_cursor *
         /* Wrapped?!  */
         KTITERS(id)--;
         KTUNLOCK(id);
+        free(fileoff);
+        fileoff = NULL;
```
greghudson, line 466: Here, `fileoff` has already been set as the value of `*cursorp`; freeing it would leave a dangling pointer in `*cursorp` for the caller.  (Setting `fileoff = NULL` does nothing, as that is a local variable about to go out of scope.)  I'm not sure how this was flagged as a leak, although we do generally try to avoid making the caller free stuff on errors.

greghudson (commented):
I looked at the library leaks so far.  I don't think any of them are likely to happen more than once in a process, although at least two of them should be fixed regardless.

ksu was written under the assumption that memory leaks are okay in a short-lived process--which is true, but not very friendly to static analysis tools.  I am amenable to changing this assumption, but only comprehensively.  My recollection is that there are more leaks than are fixed here, but I will have a closer look.

The other utility leak fixes looked okay, except gss-client.c:client_establish_context() needs to get away from freeing everything applicable in each error block (as you did for call_server()).

src/lib/krb5/keytab/kt_file.c
jrisc: My bad, here I meant to set `cursorp` to `NULL`, not `fileoff`. It was to avoid returning the address of a memory block that that already freed.

Not freeing `fileoff` here was causing a leak in [`src/kadmin/cli/keytab.c:remove_principal()`](https://github.com/krb5/krb5/blob/krb5-1.21-final/src/kadmin/cli/keytab.c#L451). In case `krb5_kt_start_seq_get()` fails, we leave the function without freeing the memory `cursor` is pointing too.

But since a cursor is just supposed to be a moving pointer to an underlying resource, freeing a cursor is something that should not happen, since it might not be pointing to the beginning of the allocated memory block.

However, I see that it's what is being done in [`krb5_ktfile_end_get()`](https://github.com/krb5/krb5/blob/krb5-1.21-final/src/lib/krb5/keytab/kt_file.c#L512), so maybe I'm just overthinking this.

src/lib/krb5/ccache/cc_kcm.c
jrisc: I think I misunderstood this case. The leak appends if [this `k5alloc()` call](https://github.com/krb5/krb5/blob/krb5-1.21-final/src/lib/krb5/ccache/cc_kcm.c#L991) fails, because `creds` is not assigned to `cursor` yet.

jrisc:
@greghudson I think you overwrote @stevegrubb 's initial commit when you rebased on `master`.

src/lib/krb5/keytab/kt_file.c
greghudson: Keytab iteration cursors are allocated objects (not just alias pointers) and do need to be freed.  remove_principal() will still leak the cursor multiple other error cases if krb5_ktfile_start_seq_get() is fixed, so that should be addressed as well.

src/lib/krb5/ccache/cc_kcm.c
greghudson: I agree that there is a leak in that case.  uuids also leaks if it was allocated in the `else if` block.

jrisc:
There are a few remaining leaks I have to fix, plus some additional changes based on the comments I am working on. I am also trying to refactor `gss-client.c:client_establish_context()` to gather the release operations in a final cleanup section, as @greghudson recommended.

greghudson:
I pushed some changes.  From past conversations I don't think it's possible to submit a candidate build to Red Hat's static analysis tooling, but if it is possible, please do so.

TBD:

* I still need to review ksu to see if these changes are acceptably comprehensive.  I also need to manually test ksu since it can't be automatically tested.

* I need to squash and possibly re-split the commits.

jrisc:
We don't get any memory leak left reported after my last commit. Which makes me think some cases remain undetected. The leak of `values` in `krb5_ldap_delete_realm()` is one of them, probably because the allocation is operated by an external library.

@greghudson We will run another scan including you last commit. I purposefully did not squash commits in order to keep previous versions between each scan. We should definitely reorganize them once we are done.

greghudson:
In a manual review of ksu for leaks not fixed in this PR I found:

* fcmd_resolve() leaks tmp_fcmd along every error path.

* match_commands() leaks fcmd_arr along all paths.

* get_line() leaks line if realloc() fails.

* krb5_ccache_copy() leaks cc_def_creds_arr (just the outer wrapper; its elements are freed).  (It would similarly leak cc_other_creds_arr but that variable is never set.  Looking at the commented-out code in krb5_store_all_creds() for creds_other, I think this logic was never really finished.)

* krb5_get_nonexp_tkts() leaks temp_creds along every error path.

* krb5_get_login_princ() leaks linebuf and buf_out on error.

* krb5_ccache_overwrite() leaks the outer ccs_creds_arr wrapper along all paths, and its contents along error paths.  It also leaks temp_principal if it allocates it (note that it aliases it to primary_principal when it does not).

* krb5_ccache_filter() leaks the other cc_creds_arr wrapper along all paths, and its contents along error paths.  It also leaks temp_principal along all paths.

* krb5_find_princ_in_cache() leaks creds_list.

* krb5_auth_check() leaks client, tgtq.client, tgtq.server, and tgt along all paths (except when ksu_tgtname() fails, when it frees client but not tgtq.client; this is the only time it frees anything).

* main() leaks everything it allocates, but it never returns (it always uses exit()) so most likely nothing will care about that.

* resolve_target_cache() leaks target when gen_sym() fails (just needs to use `goto cleanup` in that error case).

jrisc:
@greghudson I fixed the first half of the cases in your list, but I will be unavailable for the three next weeks. I will work on fixing the remaining ones once I am back.

greghudson:
Looking over these commits I saw a couple of themes:

* Some code can be simplified by using xmalloc(), xrealloc(), etc.. as they cannot fail (they exit the process on error).  I'm not sure why ksu historically uses these helpers sometimes and not always.

* We prefer to have free functions which tolerate null values and call them unconditionally in cleanup handlers.

I can clean those up myself in a review pass but I thought I would mention them.

jrisc:
@greghudson Sure, feel free to do so and squash the commits in the process, it will probably make editing simpler.

greghudson:
I made edits, and also fixed some bugs discovered by the test script in PR #1315.  get_best_princ_for_target() had the worst bug: it creates an array of alias pointers (`princ_trials`) and then aliases a pointer from that array into `*client`, leading to a double-free.  This could be solved in several ways; since performance isn't important here, I opted to copy from the array into `*client`.

src/clients/ksu/ccache.c
```diff
@@ -64,12 +65,13 @@ krb5_ccache_copy(krb5_context context, krb5_ccache cc_def,
                  krb5_boolean restrict_creds, krb5_principal primary_principal,
                  krb5_boolean *stored)
 {
-    krb5_error_code retval;
+    krb5_error_code retval=0;
```
jrisc, line 68: I am not a fan of setting the return code to "success" by default, especially since we use `goto`. But I guess the fact we reuse `retval` for each sub-call would make the security benefit of the "error by default" approach limited.

greghudson: Unless I made a mistake here, I was just reverting a change that wasn't necessary for the purpose of the PR.  In general I don't initialize `ret` type variables in new code, but if we start doing general cleanup to the ksu source code we will swamp the memory leak fixes.

PR #1307, opened 2023-07-19
Replace ssl.wrap_socket() Python function

jrisc:
The `ssl.wrap_socket()` function was deprecated in Python 3.7 and [is removed in Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#removed). It results in the following error on Fedora Rawhide:

```
PYTHONPATH=../util VALGRIND="" python3 ./t_proxy.py 
*** Failure: /usr/bin/python3 failed to start with code 1.
*** Last mark: untrusted issuer, hostname mismatch
*** Last command (#4): /usr/bin/python3 /tmp/tmp.YfMScvG3zc/krb5-tests/util/wsgiref-kdcproxy.py 61021 /tmp/tmp.YfMScvG3zc/krb5-tests/tests/proxy-certs/proxy-no-match.pem
*** Output of last command:
Traceback (most recent call last):
  File "/tmp/tmp.YfMScvG3zc/krb5-tests/util/wsgiref-kdcproxy.py", line 17, in <module>
    server.socket = ssl.wrap_socket(server.socket, certfile=pem, server_side=True)
                    ^^^^^^^^^^^^^^^
AttributeError: module 'ssl' has no attribute 'wrap_socket'
```

This PR replaces it by the [`ssl.SSLContext.wrap_socket()`](https://docs.python.org/3.12/library/ssl.html#ssl.SSLContext.wrap_socket) method.

I tested this change on Fedora Rawhide and 38, and CentOS Stream 8 and 9.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1157,10 +1157,9 @@ AC_SUBST(PKINIT)
 # for lib/apputils
 AC_REPLACE_FUNCS(daemon)
 
-# For Python tests.  Python version 3.2.4 is required as prior
-# versions do not accept string input to subprocess.Popen.communicate
-# when universal_newlines is set.
-PYTHON_MINVERSION=3.2.4
+# For Python tests.  Python version 3.4 is required for
+# ssl.create_default_context().
+PYTHON_MINVERSION=3.4
 AC_SUBST(PYTHON_MINVERSION)
 AC_CHECK_PROG(PYTHON,python3,python3)
 if test x"$PYTHON" = x; then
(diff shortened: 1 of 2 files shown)


greghudson:
If I'm reading the documentation right, create_default_context is new in Python 3.4.  We can bump the Python version requirement for tests from 3.2.4 to 3.4 to reflect the march of time, but should change configure.ac to do so.

Looking at my notes for Python features, 3.3 and 3.4 would give us:

* shlex.quote(), which could be used in k5test.py to replace _shell_equiv()
* subprocess.check_output() accepts an input parameter, which I think could be used in k5test.py to simplify _run_cmd().

Taking advantage of those is of course out of scope for this PR.

jrisc:
Just mentioning that using `ssl.create_default_context()` is not essential here. I did so because it was the [recommended way in the Python 3.12 documentation](https://docs.python.org/3.12/library/ssl.html#server-side-operation). Using the `ssl.SSLContext` constructor should also work, and it is supported [since Python 3.2](https://docs.python.org/3.2/library/ssl.html#ssl.SSLContext).

However, I have nothing against increasing the Python test requirement to 3.4 if you wish to do so.

PR #1315, opened 2023-08-30
Add ksu test script

greghudson:
ksu can't easily be tested as part of "make check", because it's setuid and uses a secure krb5_context (which means it will always look at /etc/krb5.conf for configuration).  But these aren't serious impediments for the Github Actions CI, so I wrote a test script for that purpose.  It can also be run on a personal machine which can tolerate a short-term replacement of /etc/krb5.conf.

The asan CI build will fail due to memory leaks, so this will need to be merged after PR #1306 (though of course it can be cherry-picked in a working repository to test that PR).  PR #1309 will require a couple of changes to the script where klist doesn't find a ccache.

There might be something fancy we can do with Linux namespaces to make this friendlier to the host environment, or even runnable as part of make check, but I'm not going to gate this PR on that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,12 +41,17 @@ jobs:
                 MAKEVARS: ${{ matrix.makevars }}
                 CONFIGURE_OPTS:  ${{ matrix.configureopts }}
               run: |
+                # For the ksu tests, allow homedir access from other users.
+                umask 022
+                chmod a+rx $HOME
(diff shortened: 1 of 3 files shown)


greghudson:
It turns out that asan's leak detection doesn't work on setuid binaries, and causes them to always exit with status 1.  So PR #1306 is not a precondition for merging this, but we have to disable leak checking in the check-ksu target.

PR #1317, opened 2023-09-21
Do not require AD-SIGNEDPATH if PAC available

jrisc:
MIT krb5 1.20 and newer KDCs do generate a minimal PAC instead of `AD-SIGNEDPATH` (commit a441fbe3). As a consequence, an evidence ticket generated by a 1.20+ KDC would fail to be processed by an older KDC for a constrained delegation request.

This commit modifies this behavior to check the `AD-SIGNEDPATH` whenever it is present in a TGS-REQ, but do not require it in case a PAC is provided. It does so regardless to the fact the constrained delegation request is from a local realm or a cross-realm.

This PR is meant for krb5 1.18, but the 1.19 code is the same. If this PR is accepted it should be applied on 1.19 too.

### DIFF
--- a/src/kdc/kdc_authdata.c
+++ b/src/kdc/kdc_authdata.c
@@ -668,6 +668,13 @@ has_pac(krb5_context context, krb5_authdata **authdata)
     return has_kdc_issued_authdata(context, authdata, KRB5_AUTHDATA_WIN2K_PAC);
 }
 
+/* Return true if the AD-SIGNEDPATH is present in authorization data. */
+static krb5_boolean
+has_ad_signedpath(krb5_context context, krb5_authdata **authdata)
+{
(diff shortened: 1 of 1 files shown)


greghudson:
1.18 and 1.19 are out of maintenance, but I reviewed this change and the history of the pre-PAC KDC logic for S4U2Proxy.  I think this would allow a service to print an evidence ticket containing an invalid PAC and have it work for S4U2Proxy against a KDC that is using a PAC-unaware KDB module (such as the in-tree LDAP module; the in-tree BDB and LMDB  modules don't support S4U2Proxy).

jrisc:
@greghudson Thank you for letting me know about this risk. We will implement a mechanism downstream of CentOS and RHEL relying on the KDB plugin to set a string attribute disabling the enforcement of `AD-SIGNEDPATH` verification only if the KDB plugin enforces the verification of PAC signatures for constrained delegation requests (freeipa/freeipa#7020).

PR #1318, opened 2023-10-05
make TCP waiting time configurable

sumit-bose:
Currently there is a hardcoded 10s waiting time for TCP connections before other KDCs or connection types are tried.

With this patch a new libdefaults option 'tcp_waiting_time' is introduced to make this hardcode timeout configurable to either wait longer for the KDC to reply or switch faster to other options if a KDC is not responsive.

### DIFF
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -362,6 +362,15 @@ The libdefaults section may contain any of the following relations:
     (:ref:`duration` string.)  Sets the default renewable lifetime
     for initial ticket requests.  The default value is 0.
 
+**request_timeout**
+    (:ref:`duration` string.)  Sets the maximum total time for KDC or
+    password change requests.  This timeout does not affect the
+    intervals between requests, so setting a low timeout may result in
+    fewer requests being attempted and/or some servers not being
+    contacted.  A value of 0 indicates no specific maximum, in which
+    case requests will time out if no server responds after several
+    tries.  The default value is 0.  (New in release 1.22.)
+
 **spake_preauth_groups**
     A whitespace or comma-separated list of words which specifies the
     groups allowed for SPAKE preauthentication.  The possible values
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -363,7 +363,7 @@ The libdefaults section may contain any of the following relations:
     for initial ticket requests.  The default value is 0.
 
 **request_timeout**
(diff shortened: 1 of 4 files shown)


greghudson:
I'd like some more context for this.  The 10s waiting period was added (at Red Hat's request) to minimize the likelihood of sending the same OTP-authenticated request to multiple KDCs, since OTP back ends may enforce that a code only be used once and may introduce a brief delay.  The theory was that once a KDC server address has accepted a TCP connection, we know that it is running and network-accessible, so we can wait for its answer.

What scenarios are you running into where this ten-second wait value needs be changed?

sumit-bose:
Hi,

the context is a Radius server behind the KDC which validates the password and then does a second factor authentication by sending a message to a device of the user and the user has to send a reply.

bye,
Sumit

greghudson:
Would you be okay with just bumping up the hardcoded 10s to something considerably longer?  I don't think there's much risk of a bad user experience from waiting once a KDC has accepted a TCP connection.

abbra:
In customer tests we found out that the time to wait varies considerably and depends on a particular networking and notification method they have behind a RADIUS server. Hence making it configurable to allow adjusting to individual cases without recompile.

greghudson:
I'm more amenable to making this particular timeout unlimited than to introducing this as the first configurable sendto_kdc timeout parameter.

I would be okay with a configurable timeout meaning "regardless of where we are in the request strategy, give up on a KDC request if N seconds have elapsed since the request started".  The post-accept TCP timeout would just be however much time remains on that timer, or unlimited if it isn't set.

sumit-bose:
Hi,

I tried to implement your suggestion about "regardless of where we are in the request strategy, ...". If you do not like it I'm fine with just making the timeout unlimited, i.e. very long because I think this would fit easier into the existing code than calling `poll()` or `select()` with negative values.

bye,
Sumit

greghudson:
Does this seem okay as an alternative implementation?  Particularly note the use of a duration string instead of a number of milliseconds; we use that for most other configurable intervals and I don't think sub-section precision is required here.

sumit-bose:
> Hi,
> 
> thank you for the patches, they are working as expected in my tests. I asked the original reporter as well and am waiting for confirmation.

Hi,

jfyi, the original reporter confirmed that it is working as expected for them as well.

bye,
Sumit

greghudson:
Thanks for the confirmation.  I fixed one important bug in my implementation before merging, which would otherwise cause sendto_kdc to wait forever (or until the configured timeout) after making a UDP request.

PR #1321, opened 2023-11-28
Remove PKINIT RSA support

greghudson:
RSA mode is no longer needed for interoperability.  Reduce the attack surface of clients and KDCs by removing support for it.

[Tagging @kenh and @jrisc in case they know of a use case for PKINIT RSA mode that I do not.  MS-PKCA says the last Windows versions that didn't implement DH were XP and Server 2003.]

### DIFF
--- a/doc/user/user_commands/kinit.rst
+++ b/doc/user/user_commands/kinit.rst
@@ -193,10 +193,6 @@ OPTIONS
     **X509_anchors**\ =\ *value*
         specify where to find trusted X509 anchor information
 
-    **flag_RSA_PROTOCOL**\ [**=yes**]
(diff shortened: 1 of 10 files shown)


kenh:
I am not aware of a use case for RSA in PKINIT (honestly, I did not even know that it was an option!). I double checked and even our oldest implementation code still deployed uses DH.

jrisc:
No, we are not aware of any other use case involving RSA method for PKINIT.

Actually, following a Marvin and Bleichenbacher risk analysis, a recommendation was made to remove krb5 RSA-related code for CentOS/RHEL. So I am happy to see it happening upstream.

PR #1323, opened 2024-01-03
Attempt to improve cross-compilation support

JamesWrigley:
These were found when trying to cross-compile to Windows from a Linux host with BinaryBuilder.jl.

Changes:
- `aclocal.m4`: building shared libraries apparently isn't supported on Windows, but passing `--disable-shared --enable-static` still gave an error about shared libraries not being supported. Now the check will only be done if building shared libraries is enabled.
- `k5-platform.h`: compilation failed because the `SecureZeroMemory()` function wasn't found. Fixed by including `windows.h`.
- `win-mac.h`: gcc got confused by the backslashes. AFAIK all compilers support mapping forward slashes to the right paths on Windows though.
- `gmt-mktime.c`: this file uses `uint32_t`, which requires including `stdint.h`.
- `secure_getenv.c`: `k5_secure_getenv_init()` uses functions like `getuid()` which are POSIX-only, which caused compilation errors when building for Windows, even though it's not used on Windows at all (see `k5-platform.h`). I figured the easiest way to disable it is by ifdef-ing the whole thing. Another alternative would be to exclude it from the Makefile on Windows, but I couldn't figure out how to get that to work.
- `threads.c`: `DWORD` requires the `windows.h` header.

Disclaimer: I barely know what I'm doing, or how much krb5 supports cross-compilation :upside_down_face: And cross-compiling still doesn't work, now I'm hitting errors with `libev` from `libverto`.

### DIFF
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -892,7 +892,7 @@ if test "x$enable_rpath" != xyes ; then
 	PROG_RPATH_FLAGS=
 fi
 
-if test "$SHLIBEXT" = ".so-nobuild"; then
+if test "x$enable_shared" = xyes && test "$SHLIBEXT" = ".so-nobuild"; then
    AC_MSG_ERROR([Shared libraries are not yet supported on this platform.])
 fi
 
--- a/src/include/k5-platform.h
+++ b/src/include/k5-platform.h
@@ -1013,6 +1013,7 @@ extern int krb5int_gettimeofday(struct timeval *tp, void *ignore);
  * may have been written into swap space....
  */
 #ifdef _WIN32
+#include <windows.h>
 # define zap(ptr, len) SecureZeroMemory(ptr, len)
 #elif defined(__STDC_LIB_EXT1__)
 /*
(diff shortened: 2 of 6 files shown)


yangxianrui:
你发的邮件已经收到！
Thanks with Best regards,

greghudson:
We have a semi-separate Windows build using nmake and Visual Studio's C compiler, described in src/windows/README.  The same Makefile.in files are used, but autoconf is not, large portions of the code base are excluded, and the resulting DLLs are organized quite differently to the Unix ones.  So cross-compiling would create a very different result to what we usually build for Windows.  I  haven't heard of BinaryBuilder.jl; is there a particular result you're aiming for?

greghudson:
That is probably easier than cross-compiling the whole tree, as significant parts of the tree have never been ported to Windows.  But it might not be trivial, as the build system has no support for "build just this one library and its dependencies".

JamesWrigley:
Alrighty, I might give that a shot then. Thanks for the help! I'll close this PR since it's not needed.

PR #1325, opened 2024-01-15
Add loongarch support

qiangxuhui:
## About Loongarch 

The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

## Update `config.sub` And `src/config/config.guess`

Update `src/config/config.sub` and `src/config/config.guess` with commands:

```sh
cd src/config/
curl -sL -o config.guess 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD'
curl -sL -o config.sub 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD'
```

### DIFF
--- a/src/config/config.guess
+++ b/src/config/config.guess
@@ -1,10 +1,10 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2023 Free Software Foundation, Inc.
+#   Copyright 1992-2024 Free Software Foundation, Inc.
 
 # shellcheck disable=SC2006,SC2268 # see below for rationale
 
-timestamp='2023-01-01'
+timestamp='2024-01-01'
 
 # This file is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


greghudson:
Apologies for being unresponsive on this.  We generally update config.guess and config.sub just  before starting a new release branch, which will probably happen in the next few months.

qiangxuhui:
Thank you for your reply. In order to add support for `loongarch64`, in addition to `config.guess` and `config.sub`, I also modified `ev.c`. I want to know if there is anything else I need to do?

greghudson:
ev.c has an upstream: http://software.schmorp.de/pkg/libev.html

I am happy to take an update for it from the upstream, but would ideally prefer not to modify it more than necessary.

qiangxuhui:
I have undone the changes to `ev.c` because the changes to `ev.c` are not necessary. When the gcc's >= 4.7, `ECB_MEMORY_FENCE` will be defined as `__atomic_thread_fence (__ATOMIC_SEQ_CST)`, and the minimum version of loongarch-gcc is 12.1.

PR #1326, opened 2024-01-22
Remove license for profile_tcl.c

jrisc:
The SWIG-generated `profile_tcl.c` file was removing in commit 4e186b2789b3613362845b126bf386fa89c26709. There is no need for its license in the NOTICE file anymore.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -172,10 +172,10 @@ following copyright:
    * Redistributions of source code must retain the above copyright
      notice, this list of conditions and the following disclaimer.
 
-   * Redistributions in binary form must reproduce the above
-     copyright notice, this list of conditions and the following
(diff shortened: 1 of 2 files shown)


greghudson:
It looks like I removed this license from doc/notice.rst (the authoritative file) in commit 4e186b2789b3613362845b126bf386fa89c26709, but didn't regenerate NOTICE.

greghudson:
I replaced the PR branch with a commit to regenerate NOTICE from notice.rst, and to fix an obstacle to regenerating it.  It looks like the Fortuna license removal also hadn't made it to NOTICE.  (There are also some slight formatting changes; I assume the default fill column was bumped by one or something.)

jrisc:
Okay, looks good to me.

PR #1328, opened 2024-02-23
PKINIT EC cert improvements

greghudson:
Still to do:

* This is only tested against the internal softpkcs11, which didn't previously have ECDSA support.  I will look into migrating the tests to use softhsm if I can make that work, or getting someone to test with a proper token.

* The tracing improvements rely on trace-logging of extended error messages, which normally happens.  But errors in pkinit_open_session() can be masked because it can be called again after an error (the first time from prep_questions, the second time from process) and the second time it just short-circuits out with success if cctx->p11_module is set.  This needs to be ironed out.


p088gll:
I have tested this change with a yubikey and a CA both with public key algorithm secp384r1 and signature algorithm ecdsa-with-SHA256

kinit now does connect  to the kdc, however it gets an error back:
```
[76850] 1708975309.592675: Terminating TCP connection to stream 192.168.3.11:88
[76850] 1708975309.592676: Received error from KDC: -1765328304/Digest in signed-data not accepted
```

The error in the kdc.log is
```
Feb 26 20:21:45 diamant krb5kdc[75976](info): AS_REQ (8 etypes {aes256-cts-hmac-sha1-96(18), aes128-cts-hmac-sha1-96(17), aes256-cts-hmac-sha384-192(20), aes128-cts-hmac-sha256-128(19), DEPRECATED:des3-cbc-sha1(16), DEPRECATED:arcfour-hmac(23), camellia128-cts-cmac(25), camellia256-cts-cmac(26)}) 192.168.3.11: NEEDED_PREAUTH: golla@TESTREALM for krbtgt/TESTREALM@TESTREALM, Additional pre-authentication required
Feb 26 20:21:49 diamant krb5kdc[75976](info): preauth (pkinit) verify failure: Failed to verify CMS message: verification failure
Feb 26 20:21:49 diamant krb5kdc[75976](info): AS_REQ (8 etypes {aes256-cts-hmac-sha1-96(18), aes128-cts-hmac-sha1-96(17), aes256-cts-hmac-sha384-192(20), aes128-cts-hmac-sha256-128(19), DEPRECATED:des3-cbc-sha1(16), DEPRECATED:arcfour-hmac(23), camellia128-cts-cmac(25), camellia256-cts-cmac(26)}) 192.168.3.11: PREAUTH_FAILED: golla@TESTREALM for krbtgt/TESTREALM@TESTREALM, Failed to verify CMS message: verification failure
Feb 26 20:21:49 diamant krb5kdc[75976](info): closing down fd 13
```

greghudson:
Thank you for testing.  It will likely take me a bit of research to figure out why the CKM_ECDSA PKCS11 operation is doing something different from OpenSSL's EVP_PKEY_sign/verify with that key, and to determine which one is correct.

greghudson:
I was able to migrate the tests to SoftHSM, which allowed me to reproduce the failure locally and step through the code to find the differences.

It looks like PKCS11's CKM_ECDSA encodes the ECDSA signature according to section 2.3.1 of the PKCS11 spec, which says to encode the r and s values in octet strings of fixed size and concatenate them together; for P-256 this produces a 64-byte string.  OpenSSL's EVP_PKEY_sign() encodes the signature in ASN.1 DER, producing a 72-byte string for P-256 (usually; there appears to be no padding, so it could occasionally be shorter).

RFC 3278 (ECC algorithms in CMS) section 2.1.1 agrees with OpenSSL: "signature MUST contain the DER encoding (as an octet string) of a value of the ASN.1 type ECDSA-Sig-Value".  This is fortunate, as it means we don't need to change the way PKINIT works with EC certs in the filesystem (aside from fixing the metadata).

We should be able to translate the PKCS11 ECDSA signature to the CMS format.  I will work on this.

greghudson:
I implemented the re-encoding and the PR branch works against SoftHSM.  @p088gll this should be ready for another test with a Yubikey token.

I will still need to iron out the aforementioned pkinit_open_session() situation before merge.

greghudson:
I fixed the pkinit_open_session() problem, so this is just waiting on testing with a Yubikey token and self-review.

p088gll:
Its looking good. I am getting a TGT now :-)

PR #1329, opened 2024-03-01
Allow GSS_C_CHANNEL_BOUND_FLAG on the client

metze-samba:
Allow an application to provide GSS_C_CHANNEL_BOUND_FLAG on the client side and
force KERB_AP_OPTIONS_CBT.

See also https://github.com/heimdal/heimdal/pull/1234


greghudson:
Initial reactions, not comprehensive:

* Is this meaning of GSS_C_CHANNEL_BOUND_FLAG described anywhere, or is it de novo in these PRs?

* Does the target-principal authdata change have anything to do with the channel bindings change?  If not, they should be two PRs.

* Is there reasoning behind limiting the addition of target-principal authdata to calls from GSSAPI?  I do not like the mechanism for detecting whether the call came from GSS.

metze-samba:
> * GSS_C_CHANNEL_BOUND_FLAG

greghudson:
To review and summarize:

* Prior to 1.19, MIT krb5 treated channel bindings as mandatory for the client if specified on the server, which made it hard to gradually deploy channel bindings for a protocol with existing clients.  Heimdal, by contrast, would allow the initiator to send all-zeros channel bindings if the server supplied channel bindings.

* There were several proposals to make channel bindings gradually deployable without changing the MIT krb5 behavior, but in the end we decided to make MIT krb5 permissive like Heimdal and instead just report whether the context is channel-bound via the GSS_C_CHANNEL_BOUND_FLAG output flag.  The more complicated proposals revolved around supplying GSS_C_CHANNEL_BOUND_FLAG as an input flag on the acceptor to enable the more permissive behavior, but this required API extensions as gss_accept_sec_context() doesn't take input flags from the application.

* Microsoft has some protocols, including LDAP, which can be used either inside or outside of TLS.  Outside of TLS they use wrap tokens, while inside of TLS they use channel bindings but no wrap tokens.  This unfortunately allows a ticket and authenticator to be recorded from a non-TLS session and then replayed within a TLS session, unless the acceptor enforces initiator channel bindings within TLS.  To address this, Microsoft created AD-AUTH-DATA-AP-OPTIONS and KERB_AP_OPTIONS_CBT.  If the authenticator contains this bit, the client is asserting that it would have supplied channel bindings if it were operating over TLS, and therefore initiator channel bindings should be enforced if acceptor channel bindings are available.

* MIT krb5 implemented the CBT flag on the server, and kind of half-assedly on the client via a config option.  (I probably shouldn't have accepted the config option, as no one is likely to ever use it.)

* This PR proposes to set GSS_C_CHANNEL_BOUND_FLAG on output if the initiator and acceptor both provide no channel bindings.  While this change provides strictly more information (the acceptor application knows whether it provided channel bindings), it also seems counter-intuitive and possibly dangerous, especially if anyone is using the flag under the current semantics.

* This PR proposes to enforce a _lack_ of channel bindings from the initiator if the acceptor application provided no channel bindings and the authenticator contains the CBT flag.  I don't think there is a security need for this change; there is no attack where an acceptor records an authenticator from inside a TLS session and replays it outside.  I could also see this change causing unnecessary failures, if an acceptor application hasn't been upgraded to provide channel binding information but clients have been upgraded to include the CBT flag.

* This PR proposes to add a new meaning to the GSS_C_CHANNEL_BOUND_FLAG by making it an initiator input flag, with the semantics of "tell the acceptor to strictly enforce channel bindings".  I don't think this has been proposed before, but I don't find anything objectionable about it, aside from it being weird to have exactly one flag that doesn't get put on the wire in the krb5 mech.

metze-samba:
> To review and summarize:
> 
>     * Prior to 1.19, MIT krb5 treated channel bindings as mandatory for the client if specified on the server, which made it hard to gradually deploy channel bindings for a protocol with existing clients.  Heimdal, by contrast, would allow the initiator to send all-zeros channel bindings if the server supplied channel bindings.
> 
>     * There were several proposals to make channel bindings gradually deployable without changing the MIT krb5 behavior, but in the end we decided to make MIT krb5 permissive like Heimdal and instead just report whether the context is channel-bound via the GSS_C_CHANNEL_BOUND_FLAG output flag.  The more complicated proposals revolved around supplying GSS_C_CHANNEL_BOUND_FLAG as an input flag on the acceptor to enable the more permissive behavior, but this required API extensions as gss_accept_sec_context() doesn't take input flags from the application.
> 
>     * Microsoft has some protocols, including LDAP, which can be used either inside or outside of TLS.  Outside of TLS they use wrap tokens, while inside of TLS they use channel bindings but no wrap tokens.  This unfortunately allows a ticket and authenticator to be recorded from a non-TLS session and then replayed within a TLS session, unless the acceptor enforces initiator channel bindings within TLS.  To address this, Microsoft created AD-AUTH-DATA-AP-OPTIONS and KERB_AP_OPTIONS_CBT.  If the authenticator contains this bit, the client is asserting that it would have supplied channel bindings if it were operating over TLS, and therefore initiator channel bindings should be enforced if acceptor channel bindings are available.
> 
>     * MIT krb5 implemented the CBT flag on the server, and kind of half-assedly on the client via a config option.  (I probably shouldn't have accepted the config option, as no one is likely to ever use it.)
> 
>     * This PR proposes to set GSS_C_CHANNEL_BOUND_FLAG on output if the initiator and acceptor both provide no channel bindings.  While this change provides strictly more information (the acceptor application knows whether it provided channel bindings), it also seems counter-intuitive and possibly dangerous, especially if anyone is using the flag under the current semantics.

I rechecked now Windows behaves and it turns out that the current behaviour of gss_accept_sec_context() in MIT (but also
heimdal) doesn't need to change.

>     * This PR proposes to add a new meaning to the GSS_C_CHANNEL_BOUND_FLAG by making it an initiator input flag, with the semantics of "tell the acceptor to strictly enforce channel bindings".  I don't think this has been proposed before, but I don't find anything objectionable about it, aside from it being weird to have exactly one flag that doesn't get put on the wire in the krb5 mech.

This PR now just tries to add a way the application can trigger KERB_AP_OPTIONS_CBT.
I agree that GSS_C_CHANNEL_BOUND_FLAG is strange because it's not put on the wire.
But that special means is already there on the acceptor side.

I'm open to other ways to force KERB_AP_OPTIONS_CBT, but given the already existing  GSS_C_CHANNEL_BOUND_FLAG
it seems logical to use it...

greghudson:
I made some changes:

* In init_sec_context.c, we can avoid putting GSS_C_CHANNEL_BOUND_FLAG in ctx->gss_flags altogether, and just look at it in kg_new_connection().  I added a comment where we set gss_flags to make it clear that the omission is on purpose.

* I changed some names so that we're not calling the CBT enabling flag "channel-bound" anywhere except where we look for it in req_flags.

* I added documentation (including some missing documentation for the old meaning of GSS_C_CHANNEL_BOUND_FLAG and the way channel bindings are handled) and fleshed out the commit message.

PR #1332, opened 2024-03-20
Improve error message for DES kadmin/history key

greghudson:
[Prompted by https://mailman.mit.edu/pipermail/kerberos/2024-March/023101.html ; credit to @kenh for diagnosing the issue before I did.]

If the kadmin/history entry contains an unsupported encryption type, produce a better error message than "Bad encryption type".  Reuse the error code KADM5_BAD_HIST_KEY (unused since release 1.8).  Non-updated kadmin clients will report the message "Password history principal key version mismatch", which at least points in the direction of password history.


kenh:
LGTM

PR #1335, opened 2024-04-05
acquire_cred_from: add support to pass krb5 prompter

abbra:
krb5 preauthentication methods may request additional information from the application. krb5 prompter allows to hook into this process. Without prompters there is no way to handle RADIUS (OTP) or OAUTH2 or passkey-based pre-authentication methods provided by the FreeIPA and SSSD.

Add cred store-based mechanism to communicate an application prompter via GSSAPI krb5 mechanism. An application can use the following cred store entries to pass the callback and its opaque data structure:

  'prompter' -- pass krb5_prompter_fct prompter
  'prompter_data' -- pass opaque (void*) data to be passed to the prompter


greghudson:
I can try to dig up past discussions of how we might handle initial authentication prompting in GSS, but I think passing pointers in GSS buffers isn't the right way.  (Tagging @nicowilliams in case he wants to chime in on this.)

What is the proximal motivation for wanting to do initial authentication via GSS and not libkrb5?

abbra:
One of primary motivations is to get better integration into GSSAPI applications. Initial authentication with `gss_acquire_cred_from()` with cred store specifying a password or `gss_acquire_cred_with_password()` will not work for any pre-authentication method that does require prompts. This means applications pretty much cannot use those methods and their support for new pre-authentication methods becomes limited. As a result, most of them call out for `kinit` and parse its output. This is not really robust. With increasing governmental requirements to switch from password-based to passwordless methods, this workflow complicates usability as well.

I am trying to understand how I can get integration with IAKERB for these new pre-authentication methods. IAKERB requires use of GSSAPI by both client and IAKERB-proxy application. One of my aims is to be able to use OTP or SSSD-provided pre-authentication methods in a KDC running over localhost while clients are connecting from a different host. Basically, doing a variation of `gss-sample/gss-client -user testuser -iakerb` against a `gss-sample/gss-server -iakerb` where only a `gss-server` is in direct sight of the KDC.

greghudson:
I'd be interested in hearing more about what leads you towards IAKERB.  We added it back in 2010 and have heard very little interest in it, to the point where I regretted allowing it to be merged (especially after the code became responsible for a couple of CVEs).  Although Microsoft developers wrote the initial IAKERB draft, MS seems to have gone towards MS-KKDCP instead.

I reviewed what I could find of past discussions around how to do prompting for GSS initial authentication.  Basically, when we added the cred store extensions, there was some thought that the cred store in gss_acquire_cred_from() could be used to supply answers to preauth questions, but the consensus was that it wouldn't be a good design (it would conflate "where the credentials are" with "user inputs needed to acquire initial credentials).  However, we did rename the key-value data type from something cred-store specific to gss_key_value_set_desc so that it could be reused for a future initial credentials API.  I remember Nico sketching out such an API, but I haven't been able to find it.

Complicating matters, there is also the responder interface, added in 2012 at the behest of Red Hat.  This is another addition that added great complexity to the code and seems to have been kind of a dead end.  But it was also designed with a GSS initial creds interface partly in mind; in particular, responder questions were required to be representable as flat byte strings that could be communicated across proxies or language bindings.

abbra:
I changed the PR to provide both `responder` and `prompter` via cred-store. One of them would be accepted at a time, not both, so applications can choose what to support.

greghudson:
I will review the linked Microsoft materials; thanks for those.  (There was actually mail from Steve to the kitten list about this last year [1], but I had forgotten.) MIT's implementation of IAKERB accidentally implemented an older draft of the spec.  I wrote up a plan for updating it without breaking self-compatibility [2], but never implemented it.

I agree that the general case of initial credential acquisition requires user interaction with server-supplied inputs.  The approach in this PR has several problems, though:

1. It isn't easily proxiable.  With the current gss-proxy implementation, the values supplied in the cred store would be invalid pointers in the proxy server and would cause it to crash (or worse).

2. It isn't friendly to language bindings.  It is possible to make callback interfaces work over language bindings, but it's a lot of work for each language.  Anyone using these new keys with existing language binding implementations would crash the process.

3. This particular callback interface uses libkrb5 types, so any caller using these keys is a consumer of libkrb5 (albeit just its headers) as well as GSSAPI.  So it isn't a properly generic GSS interface.  This problem is solvable by creating GSS extension callback types to act as stubs around libkrb5, but the first two problems would remain.

I found and reviewed the IRC conversation you referred to [3].  I don't think it came to a consensus in favor of a callback-style API for user interaction; although you floated that idea a couple of times, Simo and Nico speculated about passing back error codes indicating prompts, or adding a new API.

So, is it possible to make this work without callbacks?  Yes, but it's not an easy undertaking.  We would need (1) a new GSS extension API explicitly designed to handle user interaction and state while acquiring initial credentials, and (2) to rearchitect the current preauth system (including the pluggable interface so that prompts and questions can be communicated back to the caller of krb5_init_creds_step(), rather than serviced synchronously within a step.  That may be a larger project than we currently have resources for.

If we do decide to implement a path of lesser resistance in order to get IAKERB working with all preauth mechs, I would be happier with a new API entry point added to krb5_gss_glue.c than I would be with passing pointers to C data types via the gss_acquire_cred_from() cred store.

[1] https://mailarchive.ietf.org/arch/msg/kitten/VLOAFb4Furo4T4nr88FNrjXG6Gw/
[2] https://datatracker.ietf.org/doc/html/draft-ietf-kitten-iakerb-03#appendix-A
[3] https://colabti.org/ircloggy/krbdev/2020-05-20

abbra:
Thank you for this explanation and research. I want to continue investigations to enable GSSAPI prompting but meanwhile I switched my code to raw krb5 for initial acquisition and will handle prompts there.

I tried IAKerb with a GSS client missing line of sight for KDC. In this case, I ran KDC on the same system as a `gss-server` test program, bound to localhost (127.*) address. The client, `gss-client`, is running on a different machine and only has exposure to the `gss-server`'s port. I get errors:

```
$ ./src/appl/gss-sample/gss-client -iakerb -port 1234 -user localadmin -pass some-password client2.ipa2.test host "foo bar"
Sending init_sec_context token (size=195)...continue needed...reading token flags: 0 bytes read
```
On the server side:
```
./gss-server -port 1234 -verbose -keytab /etc/krb5.keytab -iakerb host
starting...
Received token (size=195): 
60 81 c0 06 06 2b 06 01 05 02 05 05 01 30 0a a1 
08 04 06 56 44 41 2e 4c 49 6a 81 a7 30 81 a4 a1 
03 02 01 05 a2 03 02 01 0a a3 1a 30 18 30 0a a1 
04 02 02 00 96 a2 02 04 00 30 0a a1 04 02 02 00 
95 a2 02 04 00 a4 7c 30 7a a0 07 03 05 00 40 01 
00 10 a1 17 30 15 a0 03 02 01 01 a1 0e 30 0c 1b 
0a 6c 6f 63 61 6c 61 64 6d 69 6e a2 08 1b 06 56 
44 41 2e 4c 49 a3 1b 30 19 a0 03 02 01 02 a1 12 
30 10 1b 06 6b 72 62 74 67 74 1b 06 56 44 41 2e 
4c 49 a5 11 18 0f 32 30 32 34 30 34 32 30 30 37 
30 38 30 35 5a a7 06 02 04 17 69 26 15 a8 14 30 
12 02 01 14 02 01 13 02 01 12 02 01 11 02 01 1a 
02 01 19 
GSS-API error accepting context: Unspecified GSS failure.  Minor code may provide more information
GSS-API error accepting context: ASN.1 structure is missing a required field
```
It looks like there is an incomplete processing?

Both systems are Fedora 39, with krb5 1.21.2 on the server and git master on the client.

abbra:
I'll try to debug it next week. It might be related to different default realms in the server/client configurations.

abbra:
Yes, once I supplied correct `krb5.conf` and forced use of that, everything worked via IAKerb. This is between MIT server and MIT client.

abbra:
Small update. @cryptomilk and I got iakerb working with Samba in this WIP branch: https://git.samba.org/?p=asn/samba.git;a=log;h=refs/heads/asn-iakerb. You can see a small demo at https://www.youtube.com/watch?v=F8YPQlxk0NU.

One thing is to figure out how to trigger and support a realm autodiscovery over IAKerb. This is covered by the draft's section 3.1: https://datatracker.ietf.org/doc/html/draft-ietf-kitten-iakerb-03#section-3.1. A client would need to send an IAKerb header with empty realm. However, MIT Kerberos implementation of iakerb does not handle this. `iakerb_acceptor_step()` does not expect realm in the header to be NULL. So it cannot do auto discovery. An empty realm would not work either because empty realm (`realm->data[0] == 0`) would fail in `k5_locate_server()` which is called from `k5_locate_kdc()` which is called from `k5_sendto_kdc()` which is called from `krb5_sendto_kdc()` that `iakerb_acceptor_step()` uses to send the IAKerb package.

I think it would be instrumental if iakerb implementation handled realm autodiscovery itself without forcing to do so an application which otherwise has zero knowledge about iakerb other than specifying its mech OID. This will be helpful with local KDC cases in Windows 11 where actual realm is not known at all. Right now we have to create an explicit configuration file and pass it upon an app invocation to allow even proceeding with iakerb.

cryptomilk:
JFYI: https://gitlab.com/wireshark/wireshark/-/merge_requests/15542

greghudson:
Realm discovery wasn't in draft-zhu-ws-kerb, which explains the lack of handling.  The target-realm field in the ASN.1 sequence is not optional, so it can't be NULL, but we can and should detect an empty realm and send the response specified in section 3.1.

PR #1338, opened 2024-04-17
gssapi/krb5: add KRB5_AUTHDATA_TARGET_PRINCIPAL for gssapi exchanges

metze-samba:
This is based on https://github.com/krb5/krb5/pull/1329 as it touches the same code area.

See also https://github.com/heimdal/heimdal/pull/1240


greghudson:
My biggest concern for this is the use of KRB5_PRINCIPAL_UNPARSE_DISPLAY, and the edge cases that might create should we use this authdata on the acceptor.

I suspect this is just the result of historical decisions we can't change (I wish Kerberos had some constraints on principal elements to make quoting unnecessary, and I wish Microsoft had used the accepted ASN.1 wire format for a principal in a Kerberos PDU rather than a UTF-16 stringified form), but I want to do some digging to make sure it's as correct as possible.

greghudson:
A few notes:

* The PR defines KRB5_AUTHDATA_KERB_LOCAL and KRB5_AUTHDATA_TOKEN_RESTRICTIONS, but does not use them.  These are out of scope for this PR, and I think the values are also reversed (MS-KILE 3.2.5.7 and 3.4.5.3).

* MS-KILE calls this element KERB_AUTH_DATA_CLIENT_TARGET.  We don't want to be using KERB_ as a prefix in krb5.h, but maybe KRB5_AUTHDATA_CLIENT_TARGET would be truer to the spec than KRB5_AUTHDATA_TARGET_PRINCIPAL.

* MS-KILE 3.2.5.8 says "When the server name is not Krbtgt [it includes authdata extensions]".  I suspect kadmin/changepw is also excepted, but maybe we can just look at the server name and not use a flag to indicate that we're doing a GSS AP-REQ?  Does Microsoft ever make AP-REQs that aren't for krbtgt, kadmin/changepw, or GSS?

* We do not limit principal components to UTF-8, and I wouldn't be shocked to find out about Latin-1 or Shift-JIS components floating around in databases.  The PR would currently error out trying to authenticate to a server principal with invalid UTF-8.  Omitting the authdata element might be a better fallback.

* KRB5_PRINCIPAL_UNPARSE_DISPLAY was added in 1.7.  It suppresses the quoting of tabs, newlines, backspaces, NULs, backslashes, and (most importantly) slashes and at-signs in components.  It's currently used in PAC processing, I think mainly to suppress the quoting of at-signs in enterprise principals.  Do we expect to see enterprise principals as server names in AP-REQs?  Unparsing with this flag is not reversible in the general case, due to the lack of quoting of separators.  So if we wanted to use this element in rd_req_dec.c, we could use it as a filter on keytab entry names, but we couldn't safely just re-parse the name and look up that name in the keytab.

I could see embarking on a larger project to make Unix and Microsoft principal parsing conform: ban slashes and control characters in principal components, ban at-signs in realm names, split on the last at-sign when parsing (so that at-signs don't require quoting in components), perhaps require valid UTF-8.  There might be some collateral damage, though.

metze-samba:
> A few notes:
> 
>     * The PR defines KRB5_AUTHDATA_KERB_LOCAL and KRB5_AUTHDATA_TOKEN_RESTRICTIONS, but does not use them.  > These are out of scope for this PR, and I think the values are also reversed (MS-KILE 3.2.5.7 and 3.4.5.3).

Yes, they can be dropped, I copied them from heimdal where they were wrong.
142 is also called LOOPBACK instead of LOCAL...

>     * MS-KILE calls this element KERB_AUTH_DATA_CLIENT_TARGET.  We don't want to be using KERB_ as a prefix in krb5.h, but maybe KRB5_AUTHDATA_CLIENT_TARGET would be truer to the spec than KRB5_AUTHDATA_TARGET_PRINCIPAL.

Yes, the correct name was only added recently to MS-KILE.

>     * MS-KILE 3.2.5.8 says "When the server name is not Krbtgt [it includes authdata extensions]".  I suspect kadmin/> changepw is also excepted, but maybe we can just look at the server name and not use a flag to indicate that we're doing > a GSS AP-REQ?  Does Microsoft ever make AP-REQs that aren't for krbtgt, kadmin/changepw, or GSS?

Thanks for finding that reference.

I'm not sure for the kadmin/changepw case and other non-gss cases.
I'll ask Andrew Bartlett maybe he knows.

>     * We do not limit principal components to UTF-8, and I wouldn't be shocked to find out about Latin-1 or Shift-JIS components floating around in databases.  The PR would currently error out trying to authenticate to a server principal with invalid UTF-8.  Omitting the authdata element might be a better fallback.

Yes, a fallback to omitting sounds ok.
 
>     * KRB5_PRINCIPAL_UNPARSE_DISPLAY was added in 1.7.  It suppresses the quoting of tabs, newlines, backspaces, NULs, backslashes, and (most importantly) slashes and at-signs in components.  It's currently used in PAC processing, I think mainly to suppress the quoting of at-signs in enterprise principals.  Do we expect to see enterprise principals as server names in AP-REQs?  Unparsing with this flag is not reversible in the general case, due to the lack of quoting of separators.  So if we wanted to use this element in rd_req_dec.c, we could use it as a filter on keytab entry names, but we couldn't safely just re-parse the name and look up that name in the keytab.
> 

I'm not sure about enterprise principals as server principals.

I've also not yet thought about any way on the acceptor to process this.

Not there's also the KERB_AP_OPTIONS_UNVERIFIED_TARGET_NAME, which is related.

Heimdal has some gss_name_attributes to get and set authorization data,
but it at least currently has the problem that it conflicts with the automatically
added things like KRB5_AUTHDATA_GSS_API_ETYPE_NEGOTIATION, KRB5_AUTHDATA_AP_OPTIONS,
as they all want to be in the first KRB5_AUTHDATA_IF_RELEVANT element.



> I could see embarking on a larger project to make Unix and Microsoft principal parsing conform: ban slashes and control characters in principal components, ban at-signs in realm names, split on the last at-sign when parsing (so that at-signs don't require quoting in components), perhaps require valid UTF-8. There might be some collateral damage, though.

Maybe the only way would be for the application to provide the raw utf16 value,
but I guess that would need some gss_name_attribute, but the powerful authorization data plugin api on MIT
was too complex for me to try that path, if it would have been easy I had choosen it also for KERB_AP_OPTIONS_CBT.

PR #1343, opened 2024-05-08
Draft: iakerb: The target-realm in the IAKERB-HEADER should be an utf8string

cryptomilk:
```
IAKERB-HEADER ::= SEQUENCE {
    -- Note that the tag numbers start at 1, not 0, which would
    -- be more conventional for Kerberos.
    target-realm      [1] UTF8String,
       -- The name of the target realm.
    cookie            [2] OCTET STRING OPTIONAL,
       -- Opaque data, if sent by the server,
       -- MUST be copied by the client verbatim into
       -- the next IAKRB_PROXY message.
    ...
}
```

target-realm is a UTF8String since draft-ietf-kitten-iakerb-00. I guess if we just fix this, we would break communication with older MIT Kerberos versions?

### DIFF
--- a/src/lib/krb5/asn.1/asn1_k_encode.c
+++ b/src/lib/krb5/asn.1/asn1_k_encode.c
@@ -1092,7 +1092,7 @@ DEFOPTIONALEMPTYTYPE(opt_ptr_seqof_princ_plus_realm,
                      ptr_seqof_princ_plus_realm);
 
 /* First context tag is 1, not 0. */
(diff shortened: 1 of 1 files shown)


greghudson:
It does look like we botched this when we first implemented IAKERB (in addition to implementing the wrong draft).  This is new information to me.

We could make acceptors retain compatibility with current initiators by trying two different versions of the ASN.1 sequence, and sending the old one if we see the old one.  I don't think we can make initiators retain compatibility with current acceptors while still sending something acceptable to correct implementations.

greghudson:
Closing this as it's been folded into PR #1345.

PR #1346, opened 2024-05-09
fuzzing: oss-fuzz init

pkillarjun:
I'm rewriting everything from scratch; I need your basic review on:

1) Documentation Update `doc/build/options2configure.rst`.
2) Variable Naming in Build System.
3) License of .C Files.
4) Should I use .zip files (some people do not prefer this due to security reasons) or folders (this will make the management messy and hard)?
5) And a simple question: "How these `deps` files are created? I'm unable to figure this out."

PR #1350, opened 2024-06-04
Add cifuzz workflow for PR fuzzing

pkillarjun:
Continuing the fuzzing integration: https://github.com/krb5/krb5/pull/1346.

Adding [cifuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration) for checking build failures and PR fuzzing.
Using the main [example](https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/example_cifuzz.yml) provided by the OSS-Fuzz team, with an updated time of only 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,38 @@
+name: CIFuzz
+
+on:
+    pull_request: {paths: [src/**, .github/workflows/build.yml]}
+
+permissions: {}
+jobs:
+    Fuzzing:
+        runs-on: ubuntu-latest
+        permissions:
+            security-events: write
+        steps:
+        - name: Build Fuzzers
+          id: build
+          uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+          with:
+            oss-fuzz-project-name: 'krb5'
+            language: c
+        - name: Run Fuzzers
+          uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+          with:
+            oss-fuzz-project-name: 'krb5'
+            language: c
+            fuzz-seconds: 300
+            output-sarif: true
+        - name: Upload Crash
+          uses: actions/upload-artifact@v3
+          if: failure() && steps.build.outcome == 'success'
+          with:
+            name: artifacts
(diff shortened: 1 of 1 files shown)


greghudson:
Can I see an example of what the upload step does?

It would be good to have something like this to avoid fuzzing doc-only PRs:
```
on:
    pull_request: {paths: [src/**, .github/workflows/cifuzz.yml]}
```

pkillarjun:
> Can I see an example of what the upload step does?

#### Upload Crash

[Upload Crash](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#understanding-results) uses [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) to save crashes found by fuzzing as artifacts.

```log
Artifacts: When the fuzzer crashes the input file that causes the crash is uploaded as an artifact.
To download the artifact, do the following steps:
```

#### Upload Sarif

There is no proper information available that I could find.
But, when a crash is found, the OSS-Fuzz infrastructure is alerted using [SARIF](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github) files. A nice detailed report is then generated for maintainers with a better UI.

My Souce: [Souce-1](https://github.com/google/oss-fuzz/issues/10090#issue-1668674254), [Souce-2](https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/sarif_utils.py#L22) .

pkillarjun:
> > It would be good to have something like this to avoid fuzzing doc-only PRs:
> > ```
> > on:
> >     pull_request: {paths: [src/**, .github/workflows/cifuzz.yml]}
> > ```
> 
> I don't think it's necessary. [Continuous Integration: How It Works](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#how-it-works).
> 
> ```
> If your project supports [OSS-Fuzz’s code coverage](https://google.github.io/oss-fuzz/advanced-topics/code-coverage), 
> CIFuzz only runs the fuzzers affected by a pull request/commit.
> ```
> 
> But it will definitely try to build the project. I don't mind disabling CIFuzz for certain PRs.

Done, anyway.

PR #1351, opened 2024-06-05
Search OpenSSL using libtool instead of checking for libraries

uilianries:
The OpenSSL project [can use Zlib](https://github.com/openssl/openssl/blob/23b6ef4894679aa0278c93de29007d1e695856ee/crypto/comp/c_zlib.c#L27) as transitive dependency and will result in failure in case not listed to be linked in krb5 as well.

To avoid prone errors, by searching libraries only, the PKG_CHECK_MODULES will search for openssl.pc and load all required libraries instead, including its transitive dependencies.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -295,13 +295,10 @@ AC_ARG_WITH([tls-impl],
   [TLS_IMPL=$withval], [TLS_IMPL=auto])
 case "$TLS_IMPL" in
 openssl|auto)
-  AC_CHECK_LIB(ssl,SSL_CTX_new,[have_lib_ssl=true],[have_lib_ssl=false],
-               -lcrypto)
-  AC_MSG_CHECKING([for OpenSSL])
+  PKG_CHECK_MODULES([OPENSSL], [openssl >= 1.1.0], [have_lib_ssl=true], [have_lib_ssl=false])
   if test x$have_lib_ssl = xtrue ; then
     AC_DEFINE(TLS_IMPL_OPENSSL,1,[Define if TLS implementation is OpenSSL])
-    AC_MSG_RESULT([yes])
-    TLS_IMPL_LIBS="-lssl -lcrypto"
+    TLS_IMPL_LIBS=$OPENSSL_LIBS
     TLS_IMPL=openssl
     AC_MSG_NOTICE([TLS module will use OpenSSL])
   else


greghudson:
Just so I have the context, in what environment does linking against -lssl -lcrypto fail due to an OpenSSL dependency on zlib?  Normally I would expect that to be handled via shared library dependencies.

(Also, nitpick for the PR title: libtool isn't the same as pkg-config.)

uilianries:
@greghudson Thank you for your quick answer!

> Normally I would expect that to be handled via shared library dependencies.

Yes, thereis no big deal when dealing with shared dependencies. But, in my case I also have the scenario where dependencies can static. Is it something not suported by the project? 

Here is a build log using GCC 11 on Linux. All dependencies are static libraries, I'm not consuming from the standard libraries in the system (/usr/lib), but built in /opt.

https://gist.github.com/uilianries/96230baa7688d6bdcde8280aec8186ff

The important part is:

```
+ set -x
+ perl -p -e BEGIN { $SIG{__WARN__} = sub {die @_} }; $e=$ARGV; $e =~ s/OBJS\...$//; s/^/ /; s/ $//; s/ / $e/g; OBJS.SH
+ objlist= openssl.so notls.so
+ gcc -shared -fPIC -Wl,-h,k5tls.so.0 -Wl,--no-undefined -o k5tls.so openssl.so notls.so -L../../../lib -lkrb5 -lkrb5support -lssl -lcrypto -m64 -L/opt/libevent/lib -L/opt/libverto/lib -L/opt/openssl/lib -L/home/uilian/.conan2/p/zlibcdd5270f9ab9f/p/lib -Wl,--version-script binutils.versions
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `zlib_stateful_expand_block':
c_zlib.c:(.text+0x79): undefined reference to `inflate'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `zlib_stateful_compress_block':
c_zlib.c:(.text+0x102): undefined reference to `deflate'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `zlib_stateful_finish':
c_zlib.c:(.text+0x13d): undefined reference to `inflateEnd'
/usr/bin/ld: c_zlib.c:(.text+0x146): undefined reference to `deflateEnd'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `zlib_stateful_init':
c_zlib.c:(.text+0x225): undefined reference to `inflateInit_'
/usr/bin/ld: c_zlib.c:(.text+0x28c): undefined reference to `deflateInit_'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `bio_zlib_ctrl':
c_zlib.c:(.text+0x478): undefined reference to `deflate'
/usr/bin/ld: c_zlib.c:(.text+0x576): undefined reference to `zError'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `bio_zlib_write':
c_zlib.c:(.text+0x820): undefined reference to `deflate'
/usr/bin/ld: c_zlib.c:(.text+0x8c6): undefined reference to `zError'
/usr/bin/ld: c_zlib.c:(.text+0x941): undefined reference to `deflateInit_'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `bio_zlib_read':
c_zlib.c:(.text+0xa5c): undefined reference to `inflate'
/usr/bin/ld: c_zlib.c:(.text+0xaf6): undefined reference to `zError'
/usr/bin/ld: c_zlib.c:(.text+0xb57): undefined reference to `inflateInit_'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `zlib_oneshot_expand_block':
c_zlib.c:(.text+0xbdc): undefined reference to `uncompress'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `zlib_oneshot_compress_block':
c_zlib.c:(.text+0xc5c): undefined reference to `compress'
/usr/bin/ld: /opt/openssl/lib/libcrypto.a(libcrypto-lib-c_zlib.o): in function `bio_zlib_free':
c_zlib.c:(.text+0xcbf): undefined reference to `inflateEnd'
/usr/bin/ld: c_zlib.c:(.text+0xce8): undefined reference to `deflateEnd'
collect2: error: ld returned 1 exit status
```

gcc lists openssl to be linked, but it not listing zlib too, resulting in that error.

greghudson:
Thanks for the context.  There is nothing explicitly unsupported about this kind of environment, although I think it's uncommon.  It's especially unfortunate that this results in a link error--that is, if the configure system isn't going to make something work automatically in a given environment, it's better to give a configure-time error or disable that part of the build than to let the build break.  I guess SSL_CTX_new() alone isn't enough to trigger a link error for lack of zlib.

All that said, I am concerned about disabling MS-KKDCP support in environments which have linkable OpenSSL binaries but don't have pkg-config.  We do use PKG_CHECK macros elsewhere in configure.ac, but I would describe them as lower-stakes decisions.

A potential solution might be a super-macro that uses PKG_CHECK_MODULES if pkg-config is found and AC_CHECK_LIB if not (and possibly also if pkg-config doesn't find the library, in case the user specified LDFLAGS but not PKG_CONFIG_PATH).  I found an ax_check_library macro in autoconf-archive, which isn't this super-macro (it behaves like PKG_CHECK_MODULES but does not use pkg-config) but might be a useful component of one.

(While investigating this I found various degrees of anti-pkg-config and anti-pkg.m4 sentiment such as https://stackoverflow.com/questions/10220946/pkg-check-modules-considered-harmful and the links therein.  I get the impression that that was a largely a battle waged a decade ago and lost, but I'm not sure we've arrived in a day where pkg-config is close to as universal as linkable OpenSSL libraries.)

uilianries:
@greghudson Thank you for you really dedicated investigation. I can see the effort is much bigger than what is expected.

> A potential solution might be a super-macro that uses PKG_CHECK_MODULES if pkg-config is found and AC_CHECK_LIB if not (and possibly also if pkg-config doesn't find the library, in case the user specified LDFLAGS but not PKG_CONFIG_PATH). I found an ax_check_library macro in autoconf-archive, which isn't this super-macro (it behaves like PKG_CHECK_MODULES but does not use pkg-config) but might be a useful component of one.

I guess it's too much. In that case, I just would close PR. You see, it's not a common case, because zlib need to be associated and it's optional. Second, need to link to static libraries, but as far as I see, most of package managers, like AUR and APT, distribute krb5 along with shared libraries.

PR #1354, opened 2024-06-27
Change krb5_get_credentials() endtime behavior

greghudson:
Historically, krb5_get_credentials() uses in_creds->times.endtime both as the TGS request endtime and as a cache lookup criterion.  These uses are in conflict; setting a TGS request endtime can only serve to limit the maximum lifetime of the issued ticket, while a cache lookup endtime limits the minimum lifetime of an acceptable cached ticket. The likely outcome is to never use a cached credentials, leading to poor performance as we add an entry to the cache for each request.

Switch to the Heimdal behavior of using in_creds->times.endtime only as the TGS request endtime.

### DIFF
--- a/src/include/krb5/krb5.hin
+++ b/src/include/krb5/krb5.hin
@@ -3043,10 +3043,10 @@ krb5_free_tgt_creds(krb5_context context, krb5_creds **tgts);
  * session key type is specified in @a in_creds->keyblock.enctype, if it is
  * nonzero.
  *
- * The expiration date is specified in @a in_creds->times.endtime.
- * The KDC may return tickets with an earlier expiration date.
- * If @a in_creds->times.endtime is set to 0, the latest possible
- * expiration date will be requested.
+ * If @a in_creds->times.endtime is specified, it is used as the requested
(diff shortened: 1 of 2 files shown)


greghudson:
I thought about arguments why we might choose the other direction in resolving this conflict--using endtime only as a cache lookup criterion, not a TGS request parameter.  The best argument is strained: it would only make sense if the TGS imposes a fairly strict lifetime policy on the service principal, and at the same time the application has a (reasonably low) minimum lifetime requirement for the service ticket.  In this scenario we would use a cached ticket until it gets too close to being expired, and not fill the cache with one entry per call... until the TGT lifetime falls below the application minimum lifetime, at which point we would we start to fill up the cache without satisfying the application's minimum lifetime requirement.

On the flip side, we would be choosing the opposite meaning as Heimdal, which would be hostile to application developers.  Also, the one known use case of setting in_creds->times.endtime is to limit the maximum lifetime of an AFS service ticket, as service tickets with a lifetime longer than 30 days may be rejected (for historical reasons involving rxkad and krb4).  Using the requested endtime strictly as a minimum would make this impossible.

PR #1357, opened 2024-07-09
Add acceptor-side IAKERB realm discovery

greghudson:
@abbra mentioned IAKERB realm discovery in PR #1335.   https://datatracker.ietf.org/doc/html/draft-ietf-kitten-iakerb-03#section-3.1 says:
```
A client using IAKERB may not have any realm information, even for
the realm of the client's host, or may know that the client host's
realm is not appropriate for a given enterprise principal name.  In
such cases, the client can retrieve the realm of the GSS-API acceptor
as follows: the client returns GSS_S_CONTINUE_NEEDED with the output
token containing an IAKERB message with an empty target-realm in the
IAKERB-HEADER and no Kerberos message following the IAKERB-HEADER
structure.  Upon receipt of the realm request, the GSS-API acceptor
fills out an IAKERB_PROXY response message, filling the target-realm
field with the realm of the acceptor, and returns
GSS_S_CONTINUE_NEEDED with the output token containing the IAKERB
message with the server's realm and no Kerberos message following the
IAKERB-HEADER header.
```
The PR branch currently does the easy part of acceptor support, but there are two tricky problems:

* Getting "the realm of the acceptor".  Right now the PR branch queries the configured default realm, which I think could  suffice but isn't a great fit.  A krb5 acceptor doesn't necessarily need a configured realm, and there could be acceptors for services in different realms on the same host (even in the same process).  I do wonder if the alternatives would just be adding more code without arriving at a robust solution, though.  The verifier credential probably won't have realm information (only if the application used a Kerberos principal as the verifier name; it's much more common to use a host-based name or the default name or the default verifier cred).  We could look at the keytab with k5_kt_get_principal(), but the first entry in the keytab isn't guaranteed to be the one that the client plans to contact.

* If we can't determine the realm name, what should we do?  The draft doesn't specify error behavior, and the two error codes it defines (KRB_AP_ERR_IAKERB_KDC_NOT_FOUND and KRB_AP_ERR_IAKERB_KDC_NO_RESPONSE) aren't applicable, nor is any existing protocol error as far as I can tell.  So I guess we respond with KRB_ERR_GENERIC.

If we want to add initiator support, we need to figure out what circumstances (API calls and environment) should trigger it.  Section 3.1 of the draft says "A client using IAKERB may not have any realm information, even for the realm of the client's host, or may know that the client host's realm is not appropriate for a given enterprise principal name."  I'm not aware of any current API signalling we could use for knowing that the client host's realm is inappropriate, but it seems reasonable to use realm discovery if the application specifies an enterprise name and the client has no default_realm.  Unfortunately, I think currently we won't get as far as IAKERB in this case, as the krb5 gss_import_name() call will fail, so we'd have to start thinking about more wide-ranging changes to the krb5 mechanism.

(Also tagging @cryptomilk)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -463,6 +463,7 @@ local.properties
 /src/tests/gssapi/t_export_cred
 /src/tests/gssapi/t_export_name
 /src/tests/gssapi/t_gssexts
+/src/tests/gssapi/t_iakerb
 /src/tests/gssapi/t_imp_cred
 /src/tests/gssapi/t_imp_name
 /src/tests/gssapi/t_invalid
--- a/src/lib/gssapi/krb5/iakerb.c
+++ b/src/lib/gssapi/krb5/iakerb.c
@@ -283,6 +283,53 @@ cleanup:
     return code;
 }
 
+/* Generate a response to a realm discovery request. */
+static krb5_error_code
+iakerb_acceptor_realm(iakerb_ctx_id_t ctx, gss_cred_id_t verifier_cred,
+                      gss_buffer_t output_token)
+{
+    krb5_error_code ret;
+    OM_uint32 dummy;
+    krb5_gss_cred_id_t cred = (krb5_gss_cred_id_t)verifier_cred;
+    krb5_data realm = empty_data(), reply = empty_data();
+    krb5_error error = { 0 };
+    char *defrealm = NULL;
+
+    /* Get the acceptor realm from the verifier cred if we can; otherwise try
+     * to use the default realm. */
+    if (cred != NULL && cred->name != NULL &&
+        cred->name->princ->realm.length > 0) {
+        realm = cred->name->princ->realm;
+    } else {
+        ret = krb5_get_default_realm(ctx->k5c, &defrealm);
+        if (ret) {
+            /* Generate an error reply if there is no default realm. */
+            error.error = KRB_ERR_GENERIC;
+            ret = krb5_mk_error(ctx->k5c, &error, &reply);
+            if (ret)
+                goto cleanup;
+        } else {
+            realm = string2data(defrealm);
+        }
+    }
+
+    ret = iakerb_make_token(ctx, &realm, NULL, &reply, output_token);
+    if (ret)
+        goto cleanup;
+    ret = iakerb_save_token(ctx, output_token);
+    if (ret)
+        goto cleanup;
+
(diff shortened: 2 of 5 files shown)


abbra:
I think our best option here is to derive realm from the verifier credential, and if that realm is missing, fall back to the default realm. If we cannot have a default realm, respond with KRB_ERR_GENERIC. I guess that importing server cred would handle that for us?

As for the client side, right now we fail if there is no configuration. So missing krb5.conf (default realm there, really) throws us away already. Indeed, this needs to be fixed to allow reaching iakerb at all.

greghudson:
> I guess that importing server cred would handle that for us?

What specific call do you mean for importing server cred?

metze-samba:
Here are some captures between two MacOS hosts, the latest wireshark is able to show some details:
https://www.samba.org/~metze/caps/krb5/iakerb/apple-mac/

The client starts with a single username and WELLKNOWN:COM.APPLE.LKDC as realm
([sysadmin@WELLKNOWN:COM.APPLE.LKDC](mailto:sysadmin@WELLKNOWN:COM.APPLE.LKDC)) and gets a WRONG-REALM redirect to the
specific LKDC:SHA1.F7FA85D015BF6B8BD281E2FA400CA70BD332417F realm,
I guess each host have it's own LKDC.SHA1.* realm...

I think it would be good to wait until we know how a Windows to Windows
authentication flow looks like in order to decide the best strategy.

For now the caller could just pass a client principal like
[sysadmin@WELLKNOWN:COM.APPLE.LKDC](mailto:sysadmin@WELLKNOWN:COM.APPLE.LKDC)...

Maybe we'll need try multiple combinations, but I guess
at least for now, this is something the callers (e.g. samba) could do.

abbra:
> > I guess that importing server cred would handle that for us?
> 
> What specific call do you mean for importing server cred?

I mean you can attempt to derive the realm from the verifier cred which is passed to `gss_accept_sec_context()`.  We don't use it in case of IAKerb token at the moment but maybe we can pass it in and decide to use that one for the defaults in case of the discovery?

abbra:
Thanks! FreeIPA configures every system to have [domain_realm] entry for the host itself. I guess we can mirror this for local KDC setup as well but probably best would be to implement a hostrealm plugin that would dynamically handle realm mapping for the local KDC case?

PR #1358, opened 2024-07-11
Remove unused files

greghudson:
I noticed some of these when searching the code for `()` vs. `(void)` issues, and some when looking at the two affected directories.  I may or may not try to figure out a way to find a comprehensive list of unused source files before merging.


greghudson:
I manually reviewed the file list and found a bunch of additional files to remove.  A few notes on files I considered removing but left alone:

* util/et/ISSUES: files like this are not where we do issue-tracking, but this particular file listed enough issues that I didn't want to summarily remove it (whereas the issues documented for the ancient builtin crypto code are not very significant).

* util/ss/test_ss.c: while I'm not in the habit of keeping around test programs with no build system support, we have no ss tests currently (except indirectly via kadmin tests etc.) and this file looks like it's still a decent starting point.

* kprop/replica-update: there's no build system support or documentation about this file, but it seems possible that it has come in handy as an example.

greghudson:
I tentatively brought back lib/krb5/posix because it's referenced in the Windows build.  I may wind up getting rid of it again (with a corresponding fix to the Windows build), because the only thing in there is a syslog() implementation and I can't find a reason why we would need that inside the krb5 DLL.

PR #1359, opened 2024-07-19
Add unix domain socket support

cryptomilk:
This will add unix domain socket support for the KDC.


cryptomilk:
I've just started with the implementation. So it is work in progress. I haven't tested it yet.

cryptomilk:
We can successfully do a kinit over unix domain sockets now:

```
LD_LIBRARY_PATH=$(pwd)/lib KRB5_CONFIG=/home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/krb5.conf KRB5_TRACE=/dev/stderr ./clients/kinit/kinit asn@LKDC.EXAMPLE.SITE -c ./tmpccache
[430719] 1727433174.569823: Getting initial credentials for asn@LKDC.EXAMPLE.SITE
[430719] 1727433174.569825: Sending unauthenticated request
[430719] 1727433174.569826: Sending request (193 bytes) to LKDC.EXAMPLE.SITE
[430719] 1727433174.569827: Sending TCP request to domain socket localhost:/home/asn/tmp/lkdc.sock
[430719] 1727433174.569828: Received answer (253 bytes) from domain socket localhost:/home/asn/tmp/lkdc.sock
[430719] 1727433174.569829: Terminating TCP connection to domain socket localhost:/home/asn/tmp/lkdc.sock
[430719] 1727433174.569830: Received error from KDC: -1765328359/Additional pre-authentication required
[430719] 1727433174.569833: Preauthenticating using KDC method data
[430719] 1727433174.569834: Processing preauth types: PA-FX-FAST (136), PA-ETYPE-INFO2 (19), PA-ENC-TIMESTAMP (2), PA-FX-COOKIE (133)
[430719] 1727433174.569835: Selected etype info: etype aes256-cts, salt "LKDC.EXAMPLE.SITEasn", params ""
[430719] 1727433174.569836: Received cookie: MIT
Password for asn@LKDC.EXAMPLE.SITE: 
[430719] 1727433176.866083: AS key obtained for encrypted timestamp: aes256-cts/62F7
[430719] 1727433176.866085: Encrypted timestamp (for 1727433176.867747): plain 301AA011180F32303234303932373130333235365AA10502030D3DA3, encrypted 1CBD799A2166E2F582ADE30AA227E8408A96F63570597B8ED2769CB117045AE443C0435A5FCCDE453CC4B9E23E973DBBE814C726F3655E6C
[430719] 1727433176.866086: Preauth module encrypted_timestamp (2) (real) returned: 0/Success
[430719] 1727433176.866087: Produced preauth for next request: PA-FX-COOKIE (133), PA-ENC-TIMESTAMP (2)
[430719] 1727433176.866088: Sending request (288 bytes) to LKDC.EXAMPLE.SITE
[430719] 1727433176.866089: Sending TCP request to domain socket localhost:/home/asn/tmp/lkdc.sock
[430719] 1727433176.866090: Received answer (821 bytes) from domain socket localhost:/home/asn/tmp/lkdc.sock
[430719] 1727433176.866091: Terminating TCP connection to domain socket localhost:/home/asn/tmp/lkdc.sock
[430719] 1727433176.866092: Processing preauth types: PA-ETYPE-INFO2 (19)
[430719] 1727433176.866093: Selected etype info: etype aes256-cts, salt "LKDC.EXAMPLE.SITEasn", params ""
[430719] 1727433176.866094: Produced preauth for next request: (empty)
[430719] 1727433176.866095: AS key determined by preauth: aes256-cts/62F7
[430719] 1727433176.866096: Decrypted AS reply; session key is: aes256-cts/490F
[430719] 1727433176.866097: FAST negotiation: available
[430719] 1727433176.866098: Resolving unique ccache of type MEMORY
[430719] 1727433176.866099: Initializing MEMORY:axZFtjM with default princ asn@LKDC.EXAMPLE.SITE
[430719] 1727433176.866100: Storing config in MEMORY:axZFtjM for krbtgt/LKDC.EXAMPLE.SITE@LKDC.EXAMPLE.SITE: fast_avail: yes
[430719] 1727433176.866101: Storing asn@LKDC.EXAMPLE.SITE -> krb5_ccache_conf_data/fast_avail/krbtgt\/LKDC.EXAMPLE.SITE\@LKDC.EXAMPLE.SITE@X-CACHECONF: in MEMORY:axZFtjM
[430719] 1727433176.866102: Storing config in MEMORY:axZFtjM for krbtgt/LKDC.EXAMPLE.SITE@LKDC.EXAMPLE.SITE: pa_type: 2
[430719] 1727433176.866103: Storing asn@LKDC.EXAMPLE.SITE -> krb5_ccache_conf_data/pa_type/krbtgt\/LKDC.EXAMPLE.SITE\@LKDC.EXAMPLE.SITE@X-CACHECONF: in MEMORY:axZFtjM
[430719] 1727433176.866104: Storing asn@LKDC.EXAMPLE.SITE -> krbtgt/LKDC.EXAMPLE.SITE@LKDC.EXAMPLE.SITE in MEMORY:axZFtjM
[430719] 1727433176.866105: Moving ccache MEMORY:axZFtjM to FILE:./tmpccache
[430719] 1727433176.866106: Destroying ccache MEMORY:axZFtjM
```

cryptomilk:
@greghudson Could you please give me some pointers how to implement tests for this and where?

greghudson:
I would put tests for this feature in tests/t_sendto_kdc.py.  Currently there is only a fallback test there, so there isn't a similar test to look to for boilerplate.  The basic test will need to create a configuration with Unix socket transport, initiate a Kerberos exchange, and probably look at the trace logs to make sure that Unix socket transport is actually being used.

In the long run we might want to flip things around so that the whole test suite runs using Unix socket transport, and TCP/UDP support is tested explicitly in t_sendto_kdc.py, with some way to conditionalize that out.  Then the test suite could be run multiple times in parallel in a single network namespace.  However, there are other obstacles to that goal (the test suite involves protocols other than Kerberos, such as kadmin and appl/gss-sample), so that's not an immediate milestone.

cryptomilk:
@greghudson I've updated documentation and added a test for the KDC. However I also added support for kadmind and kpasswd to listen on a unix socket. Should this be covered by tests too? If yes where should they go?

The doc update gives an error:

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`.
Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/
```

greghudson:
I'll tentatively say that tests for the password change protocol belong in t_changepw.py for the kadmin protocol in t_kadmin.py.  If by kadmind you just mean the password change server that's part of kadmind, then only t_changepw.py.

I will hopefully fix the doc build script soon.

cryptomilk:
@greghudson I'm not sure what I'm doing wrong. I've created a config so that kpasswd_listen is set to:

```
admin_server = magrathea:61001, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kadmin.sock
kpasswd_server = magrathea:61002, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kpasswd.sock
```

in testdir/kdc.conf.kadmin_unix

But it isn't listening to to it, it doesn't look like kadmind is picking up the config. If I add gdb to kadmind:

```
(gdb) p *params                                                                            
$4 = {                                                                                    
  mask = 1594614651,                                                                      
  realm = 0x46b910 "KRBTEST.COM",
  kadmind_port = 61001,
  kpasswd_port = 61002,                                                                   
  admin_server = 0x425260 "magrathea",  
  dbname = 0x42a980 "/var/kerberos/krb5kdc/principal",
  acl_file = 0x46e1b0 "/home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/acl", 
  dict_file = 0x46e160 "/home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/dictfile",
  mkey_from_kbd = 0,  
  stash_file = 0x46e110 "/home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/stash",
  mkey_name = 0x0,
  enctype = 18,                                                                           
  max_life = 86400,
  max_rlife = 0,       
  expiration = 0,                   
  flags = 0,                      
  keysalts = 0x46b9a0,                                                                                                                                                              
  num_keysalts = 2,                                                                                                                                                                 
  kvno = 0,                                                                                                                                                                         
  iprop_enabled = 0,                
  iprop_ulogsize = 1000,            
  iprop_poll_time = 120,          
  iprop_logfile = 0x46b970 "/var/kerberos/krb5kdc/principal.ulog",
  iprop_port = 61004,                                                                                                                                                               
  iprop_resync_timeout = 300,                                                                                                                                                       
  kadmind_listen = 0x0,                                                                                                                                                             
  kpasswd_listen = 0x0,
  iprop_listen = 0x0
}
````

As you can see, the values are NULL. It looks like it just loads a default config, but I'm giving it a different. See top commit.

**Comment by cryptomilk** [comment]
It is really strange, as

```
(gdb) show environment
KRB5_CONFIG=/home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/krb5.conf.kadmin_unix
KRB5_KDC_PROFILE=/home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kdc.conf.kadmin_unix
```

Seems to be correct.

**Comment by greghudson** [comment]
I see the test script setting kdc_unisock_listen, but this value is only picked up by kdc/main.c, not by alt_prof.c, so it's not surprising that it doesn't show up in `*params`.

I see the PR is just calling `loop_add_unix_socket(params->kpasswd_listen);` which I don't think is correct or sufficient.  The existing kpasswd_listen variable is for UDP and TCP addresses/ports.  I'm not sure why it's for both, when the KDC has separate variables for UDP and TCP; I'd have to review the history there.  Regardless, there needs to be a separate new variable for kpasswd Unix socket listening, with code in kadm5_get_config_params() to read it and set it in a new field in kadm5_config_params.

**Comment by cryptomilk** [comment]
As there is was no differentiation between TCP and UDP and kpasswd_listen is a comma seperated list, I just added support so you can have also unix sockets set in that line.

```
kpasswd_server = magrathea:61002, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kpasswd.sock
```

However it looks like the config isn't really parsed. This should set `params->kpasswd_listen` but it doesn't, it is `NULL`. I will debug further today. The config parser seems to not work correctly.

I've wanted to add `kpasswd_unixsock_listen` but the bit field for config parsing is full. I couldn't add it.

**Comment by cryptomilk** [comment]
Looking into krb5_aprof_get_string() which reads the value:

```
 197  
 198      /* Excise the entry we want from the null-terminated list,
 199       * and free up the rest. */
 200      if (uselast) {
 201          *stringp = values[lastidx];
 202          values[lastidx] = NULL;
 203      } else {
 204          *stringp = values[0];
 205          values[0] = values[lastidx];
 206          values[lastidx] = NULL;

(gdb) p values[0]
$44 = 0x4242e0 "magrathea:61001, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kadmin.sock"
(gdb) p values[1]
$45 = 0x423f70 "magrathea:61001"
(gdb) p stringp
$46 = (char **) 0x7fffffffcd30
(gdb) p *stringp
$47 = 0x423f70 "magrathea:61001"
```

`uselast` is always set to `TRUE` so it doesn't pick up the config file value. Where is values[1] actually coming from?

**Comment by greghudson** [comment]
I'm a little lost as to whether the question is about kpasswd_server (the client setting) or kpasswd_listen, but whichever it is: is the setting perhaps appearing in both krb5.conf and kdc.conf?  In that case the values would be read from both.

**Comment by cryptomilk** [comment]
@greghudson I'm not able to add `kpasswd_unixsock_listen` nor `kadmind_unixsock_listen`. The reason is that the config values are read using a bit mask ([here](https://github.com/krb5/krb5/blob/master/src/lib/kadm5/alt_prof.c#L516)). The bitmask [`kadm5_config_params`](https://github.com/krb5/krb5/blob/master/src/lib/kadm5/admin.h#L135) is full. You can't add any additional bits. Well there is one free (commented out), but we would need two.

Because of this, I looked into other ways. As `kadmind_listen` and `kpasswd_listen` are comma separated lists, I decided to add support for the path for the unix domain socket within those lists. Is that fine for you?

If yes, then all which is missing is updating the config options and the PR would be ready for review.

**Comment by greghudson** [comment]
I will look into the details, but that is probably fine since we don't have separate UDP and TCP listening options for kadmind and kpasswd.

**Comment by greghudson** [comment]
For the kadm5_params leak, I think iprop_listen needs to be added as well.  (Leak was created by commit aa91cb5dbbd4356c7a9069f4f52a10f70d91bc00.)

**Comment by greghudson** [comment]
For the kpasswd_sendto_msg_callback(), the current practice would be to move addrs to the top level declarations for the functions, initialized to NULL, and free it in the cleanup handler.  Then `local_kaddr` can just be set to `*addrs[0]` in place of that whole block of field copies, so that `local_kaddr.contents` aliases `addrs[0]->contents` instead of holding an allocated copy.

**Comment by greghudson** [comment]
I pushed the two bug fixes after making an additional simplification to kpasswd_sendto_msg_callback().

I think that function will need additional attention for Unix domain socket support.  That arm of the if-then block is supposed to be a backstop against the platform not providing the local address in getsockname().  Presumably the new tests were getting there because the socket local address was revealed, but was not an IPv4 or IPv6 address.  Sending the host's first local address does not seem correct.

There doesn't appear to be a registered Kerberos address type for Unix domain sockets (per https://github.com/krb5/krb5-assignments/blob/master/addr-type).  But there is a specification for directional addresses (RFC 4120 section 7.1).  We haven't historically done anything with them because there's no easy way to know whether the other party understands them.  (Well, nothing has prevented us from adding receiving support for them aside from lack of attention, I guess.)  Since kpasswd over Unix domain sockets is a new sub-protocol, we could use directional addresses with confidence that the other side will understand them.

**Comment by cryptomilk** [comment]
I've implemented `ADDRTYPE_DIRECTIONAL` support. Also added it to audit and changed the test to check it works correctly.

**Comment by cryptomilk** [comment]
@greghudson I've converted the audit test to unix domain socket, but this doesn't work. After the test jsonwalker.py is called with a dict expecting an the "fromaddr" to includ an "ip".

Either we duplicate both tests, or we call jsonwalker.py inside of the t_audit.py. Not sure which is the preferred way.

**Comment by greghudson** [comment]
The first commit adds documentation for setting kadmind_listen, kdc_listen, and kpasswd_listen to "" to disable listening.  However, I think this is only implemented for kdc_listen (and kdc_tcp_listen, which is already documented).  There is explicit code in kdc/main.c to check for empty strings, not in ovsec_kadmd.c.  loop_add_*_address() appears to treat an empty string as equivalent to no string, so adds the wildcard port.

I would be okay with changing this behavior, probably by changing loop_add_addresses() to do nothing if strtok_r() returns no results and removing the special cases in kdc/main.c.

**Comment by greghudson** [comment]
The commits adding directional address support do not seem consistent with RFC 4120 section 7.1.  Directional addresses are only to be used in KRB-SAFE and KRB-PRIV messages, and the payload is 00 00 00 00 or 00 00 00 01, not a Unix domain socket address.

**Comment by cryptomilk** [comment]
I've changed loop_add_addresses() to do nothing if we can't find a token.

Also Alexander and I stopped using ADDRTYPE_DIRECTORNAL. The first sentence of RFC 4120 section 7.1 is:

> All negative values for the host address type are reserved for local
>   use.  All non-negative values are reserved for officially assigned
>   type fields and interpretations.

So we did the following:

```
/*
 * All negative values for the host address type are reserved for local use.
 *
 * 0x0001 = AF_UNIX or PF_LOCAL
 */
#define ADDRTYPE_UNIXSOCK       (0x8000 | 0x0001)
```

We added it before:

```
/* macros to determine if a type is a local type */
#define ADDRTYPE_IS_LOCAL(addrtype) (addrtype & 0x8000)
```

**Comment by abbra** [comment]
@greghudson could you please explain why in this PR we need to care about the directional addresses if the rest of krb5 is not handling it? We only wanted to propagate UNIX domain socket information where an address information is propagated already. Do we need to do that at all?

Is it mostly about an audit and access control for addressful tickets?

My other concern is a combination of IAKERB and unix domain socket: a TGT obtained via IAKERB could come from a system that uses UNIX domain socket internally to communicate with a KDC. Such TGT then could be presented to a different service that might be using a different transport to talk to a KDC from the same realm. What do we need to care about here?

**Comment by greghudson** [comment]
We need to do three things:

1. Set an appropriate s-address in KRB-SAFE and KRB-PRIV messages when operating over Unix domain transport.  As a matter of correctness I believe directional addresses are the best fit here.  (As a matter of practicality the source addresses generally aren't checked, so we could put in any garbage in there and the changepw and kprop protocols would work.  But I would prefer to be correct.)

2. Log and audit Unix domain socket addresses correctly.  The challenge is that net-server.c communicates addresses to server applications using krb5_fulladdr structures containing a krb5_address and a port.  While we could shoehorn Unix socket address into this form using a local address type, I believe it is cleaner to change the net-server.c interface to communicate raw sockaddrs to applications.

3. Correctly handle addressful tickets during TGS requests.  I don't think we need to represent Unix domain socket addresses in ticket caddr fields, but if a TGS request is made over a Unix domain socket and the header ticket is addressful, we should correctly fail to match any of the ticket addresses.

Because I am asking for a refactoring of net-server.c, I will take responsibility for this part of the PR.  It may take me a few days.

> My other concern is a combination of IAKERB and unix domain socket: a TGT obtained via IAKERB could come from a system that uses UNIX domain socket internally to communicate with a KDC. Such TGT then could be presented to a different service that might be using a different transport to talk to a KDC from the same realm. What do we need to care about here?

Addressful tickets are rarely used.  When they are used, the AS client decides what addresses get stored in the ticket.  The ticket is then restricted to use from one of the listed addresses.  In our client implementation (e.g. if you do "kinit -a"), we construct a list of the local interface IP addresses and asks for those.

If you use IAKERB to make an AS request, then it wouldn't matter how the GSS service communicates with the KDC; the ticket would still contain the addresses requested by the client.  If the client uses the TGT from one of its interface addresses then that would be fine.  However, if the client uses IAKERB to make a TGS request via the GSS service, that won't work, as KDC will see the request as coming from an address not listed.  This is true regardless of what kind of socket (IP or Unix domain) the GSS service uses to communicate with the KDC.  Basically, addressful tickets aren't compatible with making TGS requests over IAKERB, for the same reason that they aren't compatible with NATs.

**Comment by greghudson** [comment]
I pushed changes to the first half or so of the commit chain.  I believe I included everything in the PR commits except for the klist.c change, which (as previously discussed) I don't believe is needed.

There are two uses of krb5_address for auditing in semi-public plugin interfaces.  For those I kept the ADDRTYPE_UNIXSOCK local-use address type.  For protocol uses I switched to directional addresses as previously discussed.

File: src/include/socket-utils.h [hunk 1]
```diff
@@ -159,4 +159,40 @@ sa_socklen(const struct sockaddr *sa)
         abort();
 }
 
+static inline int
+sa_compare(struct sockaddr *sa1, struct sockaddr *sa2)
```
cryptomilk, line 163: We can constify this too now.

cryptomilk:
Shouldn't set_cloexec_fd() be a function avoid overwriting flags?

```c
int set_cloexec_fd(int fd)
{
    int flags = fcntl(fd, F_GETFD, 0);
    if (flags < 0)
        return -1;
    retrun fcntl(fd, F_SETFD, flags|FD_CLOEXEC);
}
```

greghudson:
> Shouldn't set_cloexec_fd() be a function avoid overwriting flags?

Perhaps in theory, but there aren't any file descriptor flags besides FD_CLOEXEC, so it doesn't matter in practice.

greghudson:
I need more information on the relevance of the cloexec helpers to this PR.  Right now the commit does not appear to be in scope.

cryptomilk:
Feel free to drop it and squasch the sa_compare one.

greghudson:
In 028f9d8866897e4d8139392f919b59fa6cc783d7 ("Add support for unix domain sockets in net-server") I'm having trouble interpreting the intent of the loop_add_addresses() change.  From the comment it seems like it's supposed to allow a mix of IPv4/IPv6 and Unix domain socket specifications, but the code doesn't seem to agree.  If type == UNX, bare port specifications are allowed (and ignored), but any IPv4/IPv6 specification with a host or address would error out.

Also, since Unix domain socket addresses don't contain ports, I would do the conditionals before calling 
k5_parse_host_string().  As it stands we are treating ':' as a separator in a Unix domain socket path for no reason.

cryptomilk:
This is needed that you can specify:

```
kadmind_listen = 61001, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kadmin.sock
```

We are calling:

```
socket_paths = "61001, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kadmin.sock"
loop_add_addresses(socket_paths, 0, UNX, NULL); 
```

This means we are adding unix sockets and need to ignore ip addresses. In this case we only have a port and no host specified. So host will be NULL and we need to ignore it.

greghudson:
I get that much.  But if you write "kadmind_listen = 127:0.0.1:61001, /home/asn/workspace/prj/oss/mit-krb5/asn-unixsock/obj/testdir/kadmin.sock" the current code will error out.  Is that intentional, and if so, why?

cryptomilk:
Nope, it isn't intentional. I guess it falls under untested code is broken code :-)

cryptomilk:
Tomorrow I can look into it and add tests.

greghudson:
I pushed changes to that commit.  I didn't add tests, but now that the code isn't looking at the k5_parse_host_string() results for type == UNX I'm not sure we need tests for specific non-pathname listener specifications.

cryptomilk:
Looks good to me now, thanks!

src/lib/krb5/os/changepw.c
```diff
@@ -143,6 +147,17 @@ kpasswd_sendto_msg_callback(SOCKET fd, void *data, krb5_data *message)
...
         local_kaddr.length = sizeof(ss2sin6(&local_addr)->sin6_addr);
         local_kaddr.contents = (krb5_octet *) &ss2sin6(&local_addr)->sin6_addr;
+#ifndef _WIN32
+    } else if (local_addr.ss_family == AF_UNIX) {
+        /*
+         * There is no standard way to represent Unix domain sockets.  Assume
+         * that the receiver will accept a directional address if it verifies
+         * the source address.
+         */
+        local_kaddr.addrtype = ADDRTYPE_DIRECTIONAL;
```
abbra, line 157: I wonder if we simply can copy `addr_directional_init` structure here? It would mean probably moving addr-utils around for linking purposes, though.

greghudson: The best alternative home for addr-utils.c is lib/krb5/os, with k5_ names and prototypes in k5-int.h.  (libkrb5support generally stays closer to the platform, so dealing in krb5_address structures would be an expansion of scope.)  changepw.c still couldn't use sockaddr_to_address() because it has the special requirement of maybe calling krb5_os_localaddr() (unless that's outdated, but we'd need to test on Windows to check).  It looks like print_addr() could possibly be leveraged in localaddr.c.

I previously landed on this placement and treating changepw.c as a special case, but I can't really articulate why I didn't like the libkrb5 private API placement as much.   I will try the experiment of moving it and see if I run into complications.

src/lib/krb5/os/locate_kdc.c
```diff
@@ -288,24 +287,47 @@ locate_srv_conf_1(krb5_context context, const krb5_data *realm,
...
+            case TCP:
+            case UDP:
+                default_port = udpport;
+                break;
+            case HTTPS:
+                default_port = 443;
+                break;
+            default:
+                break;
+            }
```
greghudson, line 320: Why the expansion of the ternary expression to this switch statement?

cryptomilk: I think during development we handle also UNX in the switch statement and later removed it again. I find the switch more readable, but feel free to revert back to the ternary.

greghudson:
I am considering making the following adjustment to improve symmetry between server listening options:

* UNIX domain socket paths can be specified in kdc_listen.
* kdc_unixsock_listen goes away.
* kdc_tcp_listen defaults to the value of kdc_listen instead of to "88".  (UNIX domain socket paths in kdc_tcp_listen are ignored, as naturally happens in the current implementation.)  This part will be in a separate PR as it isn't required for UNIX domain socket support.

The idea is that kdc_tcp_listen becomes a legacy/specialty option to handle cases where people don't want their KDC to handle TCP traffic (so the main use case is `kdc_tcp_listen = ""`).  Aside from that, each service has a single _listen option that can list any kind of address, and IPv4/IPv6 addresses cover both UDP and TCP if the service can do both.

greghudson:
I made a few more changes.  I took out the systemd and sa_compare() commits; can we put them in a separate PR?  (Before or after this one is merged; before means we keep having to rebase on this one, but I'm mostly done with edits I think.)

I am leaning towards removing the audit test.  That functionality isn't very important (it's not a public interface, though it is used by Oracle) and adding the test is awkward due to the unusual structure of the existing test.

cryptomilk:
- Yes, we can put them in a separate PR and merge it after this hit master.
- The audit test was useful during development to get the addresses correctly and make sure we don't break anything when changing something. I would keep them around and drop them right before you merge.

PR #1364, opened 2024-08-01
Set TL_DATA mask flag for "kdb5_util use_mkey"

jrisc:
Since commit c877f13c8985d820583b0d7ac1bb4c5dc36e677e, the `KADM5_TL_DATA` flag is set in `add_new_mkey()` and `kdb5_purge_mkeys()` to make sure KDB modules apply tl-data changes. However, this flag is not set in `kdb5_use_mkey()`. This results in active KVNO change (which is based on tl-data) to be ignored by KDB modules relying on DB entry flags.

The present pull request fixes this issue by setting the `KADM5_TL_DATA` flag in `kdb5_use_mkey()`.

### DIFF
--- a/src/kadmin/dbutil/kdb5_mkey.c
+++ b/src/kadmin/dbutil/kdb5_mkey.c
@@ -510,6 +510,8 @@ kdb5_use_mkey(int argc, char *argv[])
         goto cleanup_return;
     }
 
+    master_entry->mask |= KADM5_TL_DATA;
+
     if ((retval = krb5_db_put_principal(util_context, master_entry))) {
         com_err(progname, retval,
                 _("while adding master key entry to the database"));
@@ -780,7 +782,7 @@ update_princ_encryption_1(void *cb, krb5_db_entry *ent)
         goto fail;
     }
 
-    ent->mask |= KADM5_KEY_DATA;
+    ent->mask |= KADM5_KEY_DATA | KADM5_TL_DATA;
 
     if ((retval = krb5_db_put_principal(util_context, ent))) {
         com_err(progname, retval, _("while updating principal '%s' key data "
--- a/src/kadmin/dbutil/kdb5_util.c
+++ b/src/kadmin/dbutil/kdb5_util.c
@@ -600,6 +600,9 @@ add_random_key(int argc, char **argv)
         exit_status++;
         return;
     }
+
+    dbent->mask |= KADM5_ATTRIBUTES | KADM5_KEY_DATA | KADM5_TL_DATA;
+
     ret = krb5_db_put_principal(util_context, dbent);
     krb5_db_free_principal(util_context, dbent);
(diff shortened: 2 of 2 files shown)


jrisc:
I realized this problem while testing master key upgrade with FreeIPA's KDB module. After creating a new master key `kdb5_util use_mkey [new MKVNO]` was not having any effect.

greghudson:
In the same file, it looks like update_princ_encryption_1() needs this fix.

greghudson:
It also looks like kdb5_util.c:add_random_key() needs to set a mask (with a bunch of values), although I have no idea why we have a "kdb5_util ark" command or who would be using it.

krb5_db_def_rename_principal() also does a put without setting a mask.  This is unlikely to be a practical issue as KDBs using the mask probably have separate rename methods, but I think it's still incorrect.

jrisc:
I added the following flags:

* `src/kadmin/dbutil/kdb5_mkey.c`
  * `add_new_mkey()`:
    * `KADM5_MKVNO`
    * `KADM5_KVNO`
  * `kdb5_use_mkey()`:
    * `KADM5_MOD_TIME`
  * `update_princ_encryption_1()`:
    * `KADM5_MKVNO`
    * `KADM5_MOD_TIME`
  * `kdb5_purge_mkeys()`:
    * `KADM5_MOD_TIME`
* `src/kadmin/dbutil/kdb5_util.c`
  * `add_random_key()`
    * `KADM5_KEY_DATA`
    * `KADM5_KVNO`
    * `KADM5_ATTRIBUTES`
    * `KADM5_LAST_PWD_CHANGE`

I don't think we can add flags in `krb5_db_def_rename_principal()`, since it is defined in `src/lib/kdb/kdb_default.c`, while the mask flags are defined in `src/lib/kadm5/admin.h`.

greghudson:
I don't think we need to be setting the KADM5_KVNO flag for any KDB operations.  It is used by libkadm5 to indicate the presence of or desire for the kadm5_principal_ent_rec kvno field.  There is no equivalent field in krb5_db_entry, just the key data array.  Specifying KADM5_KEY_DATA is enough if the key data is changed.

I tentatively feel the same way about the KADM5_MKVNO flag.  Here again there is an mkvno field in kadm5_principal_ent_rec but not in krb5_db_entry; in a KDB entry the mkvno is represented inside the tl_data list.  Although a KDB module might represent the mkvno separately from tl-data, I think that's an implementation detail below the KDB library layer, and the module shouldn't expect a hint that that specific tl-data value has changed.

abbra:
Attn @cryptomilk for Samba kdb driver...

abbra:
Regarding Samba KDB driver. We did review and it intentionally does not allow any modifications via kadmin.local/kadmin tools. All changes to user and machine accounts would need to go through Samba AD tools. So they aren't affected.

PR #1366, opened 2024-08-10
Fuzzing: oss-fuzz updates

pkillarjun:
## Updates
* Update old existing fuzzers code coverage https://github.com/krb5/krb5/pull/1366/commits/1dbd9240b8aaadb3ce8d1650f3296bc42736262b.
* Added Some new Extra fuzzers https://github.com/krb5/krb5/pull/1366/commits/1120df05f9eff728dd3e555ed2db45db6b6473c9.

## Note
Hi, I did try to split this cf into multiple PRs, so it would be easier to merge and understand.

I can't break https://github.com/krb5/krb5/pull/1366/commits/1dbd9240b8aaadb3ce8d1650f3296bc42736262b and https://github.com/krb5/krb5/pull/1366/commits/1120df05f9eff728dd3e555ed2db45db6b6473c9 into two PRs.
Because https://github.com/krb5/krb5/pull/1366/commits/1120df05f9eff728dd3e555ed2db45db6b6473c9 depends on updates in https://github.com/krb5/krb5/pull/1366/commits/1dbd9240b8aaadb3ce8d1650f3296bc42736262b.

## HOW-to-merge
1. https://github.com/krb5/krb5/pull/1378
2. https://github.com/krb5/krb5/pull/1379
3. https://github.com/krb5/krb5/pull/1366


pkillarjun:
### Memory leak in `fuzz_pac`

### Reproduce
```bash
./tests/fuzzing/fuzz_pac leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip
```

### Fix patch

```patch
diff --git a/src/lib/krb5/krb/pac.c b/src/lib/krb5/krb/pac.c
index 5d1fdf1ba..77adcd272 100644
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -557,9 +557,11 @@ verify_pac_checksums(krb5_context context, const krb5_pac pac,
         ret = k5_pac_locate_buffer(context, pac, KRB5_PAC_SERVER_CHECKSUM,
                                    &server_checksum);
         if (ret)
-            return ret;
-        if (server_checksum.length < PAC_SIGNATURE_DATA_LENGTH)
-            return KRB5_BAD_MSIZE;
+            goto cleanup;
+        if (server_checksum.length < PAC_SIGNATURE_DATA_LENGTH) {
+            ret = KRB5_BAD_MSIZE;
+            goto cleanup;
+        }
         server_checksum.data += PAC_SIGNATURE_DATA_LENGTH;
         server_checksum.length -= PAC_SIGNATURE_DATA_LENGTH;
```

[leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip](https://github.com/user-attachments/files/17220676/leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip)

pkillarjun:
## Avoid using assert while fuzzing

### Reproduce
```bash
./tests/fuzzing/fuzz_pac crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip
```

### Fix patch
```patch
diff --git a/src/lib/krb5/krb/pac.c b/src/lib/krb5/krb/pac.c
index 5d1fdf1ba..e3fe7a17d 100644
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -146,8 +146,16 @@ k5_pac_locate_buffer(krb5_context context, const krb5_pac pac, uint32_t type,
     if (buffer == NULL)
         return ENOENT;
 
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
     assert(buffer->offset < pac->data.length);
     assert(buffer->size <= pac->data.length - buffer->offset);
+#else
+    if (!(buffer->offset < pac->data.length))
+        return EINVAL;
+
+    if (!(buffer->size <= pac->data.length - buffer->offset))
+        return EINVAL;
+#endif
 
     if (data_out != NULL)
         *data_out = make_data(pac->data.data + buffer->offset, buffer->size);
```
[crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip](https://github.com/user-attachments/files/17220894/crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip)

greghudson:
The zip files attached to the last two comments don't seem to be valid when I download them.  They are 716 and 624 bytes, and `file` just identifies them as `data`.

For the assertion failures, if the k5_pac_locate_buffer() assertions can be violated with input to krb5_pac_parse(), there's a bug to be fixed.  My initial guess is that the first assertion should be `assert(buffer->offset <= pac->data.length);` as krb5_pac_parse() only rejects when `buffer->offset > len`.

pkillarjun:
> The zip files attached to the last two comments don't seem to be valid when I download them. They are 716 and 624 bytes, and file just identifies them as data.

Oh, yes, they are data files because GitHub doesn't allow me to upload a file with no extension or with a `.bin` extension.
It's not a `.zip` file, this is a crash file, I forgot to clarify.

```bash
# What i did
mv leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip
mv crash-ec1315e247fe542b4376085207e7e3c6dfcb14da crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip

# What you can do
mv leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac
mv crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip crash-ec1315e247fe542b4376085207e7e3c6dfcb14da
```

pkillarjun:
> For the assertion failures, if the k5_pac_locate_buffer() assertions can be violated with input to krb5_pac_parse(), there's a bug to be fixed. My initial guess is that the first assertion should be `assert(buffer->offset <= pac->data.length);` as krb5_pac_parse() only rejects when `buffer->offset > len`.

### Updated patch
```patch
diff --git a/src/lib/krb5/krb/pac.c b/src/lib/krb5/krb/pac.c
index 77adcd272..909196b8d 100644
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -146,7 +146,7 @@ k5_pac_locate_buffer(krb5_context context, const krb5_pac pac, uint32_t type,
     if (buffer == NULL)
         return ENOENT;
 
-    assert(buffer->offset < pac->data.length);
+    assert(buffer->offset <= pac->data.length);
     assert(buffer->size <= pac->data.length - buffer->offset);
 
     if (data_out != NULL)
```

### Running `fuzz_pac`
```bash
./tests/fuzzing/fuzz_pac ~/Downloads/crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip

INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1480091838
INFO: Loaded 1 modules   (14541 inline 8-bit counters): 14541 [0x8f54f0, 0x8f8dbd), 
INFO: Loaded 1 PC tables (14541 PCs): 14541 [0x86ff48,0x8a8c18), 
./tests/fuzzing/fuzz_pac: Running 1 inputs 1 time(s) each.
Running: /home/brrr/Downloads/crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip
Executed /home/brrr/Downloads/crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip in 2 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
```

Yes, it dose works.

pkillarjun:
Should I break this into two PRs?

- https://github.com/krb5/krb5/pull/1366/commits/1dbd9240b8aaadb3ce8d1650f3296bc42736262b
- https://github.com/krb5/krb5/pull/1366/commits/1120df05f9eff728dd3e555ed2db45db6b6473c9

For that, the second PR needs to be a draft until the first one is merged.
I really don't like that, revising it again and again.

greghudson:
The structure of this PR seems fine; there's no need to split it.

greghudson:
If I understand correctly:

* fuzz_gss.c (old) passes its input to gss_accept_sec_context()
* fuzz_gssapi.c (new) passes its input to krb5int_parse_enctype_list(), gss_oid_to_str(), and gss_str_to_oid().  Any given input will probably only be valid or nearly-valid for one of those, but we've taken the same approach before in fuzz_util.c and probably other harnesses.

If that's correct, then: krb5int_parse_enctype_list() doesn't have much to do with the GSSAPI; it is used primarily by config parsing within libkrb5.  And it doesn't seem great to have two fuzzer harnesses with essentially the same name; we could perhaps rename the new one to fuzz_gss_util.c to differentiate it.

src/tests/fuzzing/fuzz_oid.c
```diff
@@ -0,0 +1,75 @@
...
+
+    gss_oid_to_str(&minor, &oid_desc, &buf);
+    gss_release_buffer(&minor, &buf);
+
+    buf.value = (void *)data;
+    buf.length = size;
+
+    gss_str_to_oid(&minor, &buf, &oid);
+
+    // gss_release_oid() is buggy.
```
pkillarjun, line 67: `gss_release_oid()` does call the `gssint_mechglue_initialize_library`->`gssint_mechglue_init`.
but it never calls the `gssint_mechglue_fini`, so there is a global variable memory leak.

### Foot note

I don't know how to fix this. I tried, though.

src/tests/fuzzing/fuzz_attrset.c
```diff
@@ -0,0 +1,78 @@
...
+
+    data_in = make_data((void *)data, size);
+
+    ret = krb5_init_context(&context);
+    if (ret)
+        return 0;
+
+    ret = kr_attrset_decode(context, &data_in, "f", auth, &set);
+    if (!ret)
+        kr_attrset_encode(set, "f", auth, FALSE, buffer, &encode_len);
```
pkillarjun, line 72: https://github.com/krb5/krb5/commit/871125fea8ce0370a972bf65f7d1de63f619b06c is supported.

src/tests/fuzzing/fuzz_oid.c
greghudson: What mechanism is detecting the memory leak?  If at process exit there is allocated memory that is reachable from global variables, that shouldn't be considered an error (for instance, it happens in libc if you ever call dlopen() with RTLD_NODELETE).

pkillarjun: ### compile and run
```bash
export CC=clang
export CXX=clang++
export CFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize=fuzzer-no-link"
export CXXFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize=fuzzer-no-link"
export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"

mkdir -p build
cd build
./../src/configure CFLAGS="-fcommon $CFLAGS" CXXFLAGS="-fcommon $CXXFLAGS" --enable-static --disable-shared --enable-ossfuzz

# fuzz_attrset will fail here, but that's okay
make -j$(nproc) --ignore-errors
echo "Brrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr" > save.raw
./tests/fuzzing/fuzz_oid save.raw
```

### asan-logs
```c
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 2116037642
INFO: Loaded 1 modules   (24200 inline 8-bit counters): 24200 [0xa8ec10, 0xa94a98), 
INFO: Loaded 1 PC tables (24200 PCs): 24200 [0x9d5288,0xa33b08), 
./tests/fuzzing/fuzz_oid: Running 1 inputs 1 time(s) each.
Running: save.raw
Executed save.raw in 0 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***

=================================================================
==310003==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x500ee0 in realloc (/home/brrr/krb5/build/tests/fuzzing/fuzz_oid+0x500ee0) (BuildId: 1dac8079cd2e6b709ecef5bef948c6033b347e46)
    #1 0x611084 in mecherrmap__pairarray_grow /home/brrr/krb5/build/lib/gssapi/generic/./errmap.h:106:11
    #2 0x60ff2e in mecherrmap_add /home/brrr/krb5/build/lib/gssapi/generic/./errmap.h:205:11
    #3 0x60e8f4 in gssint_mecherrmap_map /home/brrr/krb5/build/lib/gssapi/generic/../../.././../src/lib/gssapi/generic/util_errmap.c:223:11
    #4 0x6100fa in gssint_mecherrmap_map_errcode /home/brrr/krb5/build/lib/gssapi/generic/../../.././../src/lib/gssapi/generic/util_errmap.c:244:12
    #5 0x56365d in gss_str_to_oid /home/brrr/krb5/build/lib/gssapi/mechglue/../../.././../src/lib/gssapi/mechglue/g_oid_ops.c:77:2
    #6 0x541883 in LLVMFuzzerTestOneInput /home/brrr/krb5/build/tests/fuzzing/../.././../src/tests/fuzzing/fuzz_oid.c:65:5
    #7 0x44b994 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/brrr/krb5/build/tests/fuzzing/fuzz_oid+0x44b994) (BuildId: 1dac8079cd2e6b709ecef5bef948c6033b347e46)
    #8 0x4347ea in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) (/home/brrr/krb5/build/tests/fuzzing/fuzz_oid+0x4347ea) (BuildId: 1dac8079cd2e6b709ecef5bef948c6033b347e46)
    #9 0x43a130 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/brrr/krb5/build/tests/fuzzing/fuzz_oid+0x43a130) (BuildId: 1dac8079cd2e6b709ecef5bef948c6033b347e46)
    #10 0x465ea6 in main (/home/brrr/krb5/build/tests/fuzzing/fuzz_oid+0x465ea6) (BuildId: 1dac8079cd2e6b709ecef5bef948c6033b347e46)
    #11 0x7f938100b087 in __libc_start_call_main (/lib64/libc.so.6+0x2a087) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #12 0x7f938100b14a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2a14a) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #13 0x42ee04 in _start (/home/brrr/krb5/build/tests/fuzzing/fuzz_oid+0x42ee04) (BuildId: 1dac8079cd2e6b709ecef5bef948c6033b347e46)

SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

pkillarjun: >  If at process exit there is allocated memory that is reachable from global variables, that shouldn't be considered an error (for instance, it happens in libc if you ever call dlopen() with RTLD_NODELETE).

https://github.com/krb5/krb5/blob/038793c3083f44c4fb62626c12f80c80147029cf/src/lib/gssapi/generic/util_errmap.c#L133-L135

https://github.com/krb5/krb5/blob/038793c3083f44c4fb62626c12f80c80147029cf/src/lib/win_glue.c#L364-L366


`static mecherrmap m;` is the variable where memory is allocated.
`gssint_mechglue_fini` is never called except for `win` `DLL_SHUTDOWN`.

greghudson: I see.  The problem is not that the finalizer isn't running (that's deliberate, see 1bfcf572241a4ec0e44e609e5c6b7c0b11b08eea) or that asan is reporting still-reachable allocations as a leak.  The problem is that in this program, gss_str_to_oid() or gss_oid_to_str() can add an entry the error map _before_ the initializer runs.  When gss_release_oid() triggers the initializer, it re-initializes the error map, leaking the entry contained within.  So even if the finalizer was run there would be a (properly detected) leak.

I think the fix is to make gssint_mecherrmap_map() ensure that the library is initialized.  I tested a proof of concept of that and it squashes the error as expected, but I need to figure out whether there's a more elegant way of doing it.

greghudson: PR #1383 should address this problem.

pkillarjun: > The problem is that in this program, gss_str_to_oid() or gss_oid_to_str() can add an entry the error map before the initializer runs. When gss_release_oid() triggers the initializer, it re-initializes the error map, leaking the entry contained within.

> PR https://github.com/krb5/krb5/pull/1383 should address this problem.

This work,  thanks.

pkillarjun: I had a little concern over this call `gss_release_oid()->gssint_mechglue_initialize_library()`.
But `CALL_INIT_FUNCTION` is handling it great.

https://github.com/krb5/krb5/blob/038793c3083f44c4fb62626c12f80c80147029cf/src/lib/gssapi/mechglue/g_initialize.c#L182-L185

greghudson:
I made small edits, hopefully without breaking anything this time.

fuzz_asn.c has undefined behavior due to function and object type aliasing (C99 6.3.2.8).  While I usually don't like using the preprocessor for anything non-trivial, I would rather see fuzz_asan rewritten as a macro in this case.

pkillarjun:
> fuzz_asn.c has undefined behavior due to function and object type aliasing (C99 6.3.2.8).

Thanks for noticing the bug.

> While I usually don't like using the preprocessor for anything non-trivial, I would rather see fuzz_asan rewritten as a macro in this case.

Can you review this? https://github.com/krb5/krb5/pull/1366/commits/b0993e509b93e02ba6c0fc2187118cb0a51dec4f. I am not very good with macro.

pkillarjun:
I think you can merge (or re-review) it; we will see what happens to https://github.com/krb5/krb5/pull/1386

pkillarjun:
The updated `free_kkdcp_message()` works fine.

PR #1367, opened 2024-08-15
Replace Windows installer FilesInUse dialog text

greghudson:
Windows installers must provide a FilesInUse dialog, with the intent of minimizing the need to restart the system after software installations.  For reasons related to the LSA cache, the KfW installer always schedules a post-installation reboot (see commit 50a3c3cbeab32577fba2b21deb72a64015c48ec7).  Reword the FilesInUse dialog text to recommend clicking Ignore, causing the conflicting files to replaced on the reboot.  (The affected files are likely C runtime DLLs, and the existing versions will almost certainly suffice if KfW programs are run prior to a reboot.)

### DIFF
--- a/src/windows/installer/wix/lang/ui_1033.wxi
+++ b/src/windows/installer/wix/lang/ui_1033.wxi
@@ -360,7 +360,7 @@
         </Control>
         <Control Id="BannerBitmap" Type="Bitmap" X="0" Y="0" Width="374" Height="44" TabSkip="no" Text="[BannerBitmap]" />
         <Control Id="Text" Type="Text" X="20" Y="55" Width="330" Height="30">
-          <Text>The following applications are using files that need to be updated by this setup. Close these applications and then click Retry to continue the installation or Cancel to exit it.</Text>
+          <Text>Some updated files are in use by the following applications.  Click Ignore to continue installation, then reboot the system to replace the affected files.</Text>
         </Control>
         <Control Id="BannerLine" Type="Line" X="0" Y="44" Width="374" Height="0" />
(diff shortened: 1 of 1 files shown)


talregev:
@greghudson I added the krb5 port to vcpkg. and it share the same tag release windows, linux osx and other os.
Can we unify the release tag for both?
Can we also add pc files for windows to detect krb5 in windows as well?
Currently pc files are patch in vcpkg, but it better will come from the original source code. here.

greghudson:
Kerberos for Windows has different version numbering because it is a very different product.  Relative to the baseline, it is missing a lot (like the KDC and the kadmin system), its libraries are organized differently, and it adds a GUI interface and associated library.

To be included in a cross-platform packaging system that includes Windows, I think we would want support for a Windows build that creates something comparable to the baseline build, and we might even want to rewrite the whole build system.  (I am assuming that autoconf is still a non-starter on Windows, though perhaps that's no longer the case.)  That's unfortunately a bigger project than I currently have time for.

kaduk (approved):

talregev:
> Kerberos for Windows has different version numbering because it is a very different product. Relative to the baseline, it is missing a lot (like the KDC and the kadmin system), its libraries are organized differently, and it adds a GUI interface and associated library.
> 
> To be included in a cross-platform packaging system that includes Windows, I think we would want support for a Windows build that creates something comparable to the baseline build, and we might even want to rewrite the whole build system. (I am assuming that autoconf is still a non-starter on Windows, though perhaps that's no longer the case.) That's unfortunately a bigger project than I currently have time for.

Thank you for detail answer.

talregev:
@greghudson Will you consider only add pc files on windows build?

PR #1368, opened 2024-08-22
Minimize mutex assertion errors at process exit

greghudson:
In this PR I want to consider approaches to minimizing the chance of k5_mutex_lock() assertion errors at process exit.  These happen when a krb5 or GSS library function that uses a singleton mutex (like cc_typelist_lock)I is called after library finalization.  I believe we have seen that in the past due to platform bugs with unloading order; recently @nicowilliams stated in a Rust issue that he sees it happen when a multithreaded program exits the process from one thread while another thread is still working.  I don't think we can make that scenario truly safe, but we can probably squash the problem in practice.

Candidate solutions:

1. In the Rust issue, Nico suggested we not destroy mutexes in library finalizers, saying that there is "zero value in doing so".  While that's true for process exit, I don't think it's true for library unload; we can't really say that our libraries should never be unloaded (I am told that PAM frameworks do it) or that it's okay if they leak memory each time they are.
2. We could revert commit 6350fd0c909d84c00200885e722cc902049ada05 ("Assume mutex locking cannot fail").  I would rather not do this as in some functions it can add significantly to the complexity of ensuring consistent states on function return.
3. We could make k5_mutex_lock() return an error code again, but continue to keep k5_mutex_unlock() as returning void, and make it stop checking the error.  This would still be somewhat painful.
4. We could have checked and unchecked (asserting) variants of k5_mutex_lock(), and use the checked variant for singleton mutexes that get destroyed by library finalizers.
5. We could add a k5_mutex_check() and use it along the code paths that lock singleton mutexes.  This leaves open a race window (library is finalized between k5_mutex_check() and k5_mutex_lock()), so probably would not fully solve the problem in practice.
6. We could set a global on library destruction, and check it along the code paths that lock singleton mutexes.  Same race problem as (5), and is maybe less clear anyway as it isn't clearly connected to the mutex resource.  Plus it's itself not thread-safe.
7. We could stop destroying mutexes (or just singleton mutexes) on platforms that we believe don't associate dynamic resources with mutexes.  glibc documents that this is the case for the LinuxThreads implementation, for instance.  This requires layer-crossing assumptions and wouldn't be a comprehensive solution.

Any of the following external changes would make this problem easy, but we can't expect any of them to happen:

* Programs could stop exiting while more than one thread is running.
* Platforms could suspend threads other than the current one when a program begins exiting.  (Could create deadlocks in existing code.)
* Platforms could robustly guarantee that library code is not invoked during or after execution of that library's finalizers.
* Platforms could provide an indication of whether the process is currently exiting.  If it is, our finalizers could decline to destroy mutexes.  (raeburn wanted this back in 2004, based on the comments about PROGRAM_EXITING in k5-threads.h.)
* Platforms could stop implementing library unloading.  (This apparently true for musl libc: https://wiki.musl-libc.org/functional-differences-from-glibc.html#Unloading_libraries )
* Mutex interfaces (pthreads and Windows) could require that mutexes are of constant size and do not require destruction.  A slightly more achievable variant of this is that implementations' pthread_mutex_destroy() could avoid changing the state of the mutex, so that ensuing calls to pthread_mutex_lock() don't fail.  Even that's pretty unlikely, and I don't think there is a similarly minimal change for Windows.

### DIFF
--- a/src/aclocal.m4
+++ b/src/aclocal.m4
@@ -139,6 +139,9 @@ fi
 if test "$use_linker_fini_option" = yes; then
   AC_DEFINE(USE_LINKER_FINI_OPTION,1,[Define if link-time options for library finalization will be used])
 fi
+if test "$lib_unload_prevented" = yes; then
+  AC_DEFINE(LIB_UNLOAD_PREVENTED,1,[Define if library unloading is prevented])
+fi
 ])
 
 dnl find dlopen
--- a/src/config/shlib.conf
+++ b/src/config/shlib.conf
@@ -32,9 +32,15 @@ DYNOBJEXT='$(SHLIBEXT)'
 MAKE_DYNOBJ_COMMAND='$(MAKE_SHLIB_COMMAND)'
 DYNOBJ_EXPDEPS='$(SHLIB_EXPDEPS)'
 DYNOBJ_EXPFLAGS='$(SHLIB_EXPFLAGS)'
-#
+# On some platforms we will instruct the linker to run named functions
+# (specified by LIBINITFUNC and LIBFINIFUNC in each library's Makefile.in) as
+# initializers or finalizers.
 use_linker_init_option=no
 use_linker_fini_option=no
+# Where possible we will prevent unloading of the libraries we build, in which
+# case we can skip running finalizers.  Do not set use_linker_fini_option if
+# setting lib_unload_prevented.
+lib_unload_prevented=no
 
 STOBJEXT=.o
 SHOBJEXT=.so
@@ -280,7 +286,7 @@ mips-*-netbsd*)
 	SHLIBVEXT='.so.$(LIBMAJOR).$(LIBMINOR)'
 	SHLIBSEXT='.so.$(LIBMAJOR)'
 	SHLIBEXT=.so
-	LDCOMBINE='ld -shared -soname $(LIBPREFIX)$(LIBBASE)$(SHLIBSEXT)'
+	LDCOMBINE='ld -shared -soname $(LIBPREFIX)$(LIBBASE)$(SHLIBSEXT) -z nodelete'
 	SHLIB_RPATH_FLAGS='-R$(SHLIB_RDIRS)'
 	SHLIB_EXPFLAGS='$(SHLIB_RPATH_FLAGS) $(SHLIB_DIRS) $(SHLIB_EXPLIBS)'
 	RPATH_FLAG='-Wl,-rpath -Wl,'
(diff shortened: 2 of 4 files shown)


greghudson:
I've put an implementation of candidate solution 4 on the PR branch.  The complexity is substantially less than it was before commit 6350fd0c909d84c00200885e722cc902049ada05 (because we only have to check about 40% of the locks and none of the unlocks).

greghudson:
Further discussion on the Rust thread (I'm avoid linking it as I don't want to create noise for them, but it's issue 126600) indicates that there may be platform features that can better solve this problem for us:

* On ELF platforms, `-Wl,-z,nodelete` should mark a library as non-unloadable.  If we do this, we don't need to run library finalizers.
* On macOS, any use of thread-local storage with the library blocks unloading.  I will need to do more research to see if there are associated costs.
* On Windows, finalization goes through a call to DllMain() with fdwReason=DLL_PROCESS_DETACH.  If the lpvReserved argument is non-null, the process is exiting and we can decline to run library finalizers (https://learn.microsoft.com/en-us/windows/win32/dlls/dllmain).

greghudson:
I replaced the commit on the PR branch with one to prevent library unloading where possible, and to skip finalizers when we can do that.  macOS is still TBD, as I will have to figure out the exact right incantation and see if I can test it.

I tested this on Linux using tests/gssapi/reload and valgrind's leak detection, verifying that finalizers aren't being run but global structures don't leak (though they are still accessible after exit).  I did not test on the three BSDs and Solaris, but did verify that `-z nodelete` is in the linker man page on all of those platforms, so I'm pretty confident about those changes.  I'm also pretty confident about the Windows change, as it is taken almost directly from an example in the above-linked Microsoft documentation.

greghudson:
I regained access to my macOS setup and did some testing.  Adding a "static __thread int dummy;" to libgssapi_krb5 did not seem to prevent it from being unloaded, and the consequences of omitting finalizers when unloading isn't blocked are kind of glaring (assertion failure the second time libgssapi_krb5 is loaded because a per-thread storage key wasn't cleaned up when it was unloaded).  So I will probably leave macOS alone in this PR.

I discovered that OpenSSL has prevented libcrypto from being unloaded since 1.1.  They use `-z nodelete` on Linux, but on other platforms they open another reference to libcrypto at library initialization time (using dladdr() to find the filename).  I don't plan to implement that strategy currently, but it's an option if we hear more reports of finalizer races on non-ELF platforms.  (macOS does support RTLD_NODELETE, but storing the dlopen() handle in a global and never closing it would also work.)

PR #1370, opened 2024-09-05
libkrad: Add Message-Authenticator support

jrisc:
[draft-ietf-radext-deprecating-radius-03](https://datatracker.ietf.org/doc/html/draft-ietf-radext-deprecating-radius-03) describes multiple mitigation measures for recent vulnerabilities discovered in the RADIUS protocol, especially the [BlastRADIUS](https://www.blastradius.fail/attack-details) one (CVE-2024-3596). One of them is the enforcement of the `Message-Authenticator` attribute.

`Message-Authenticator` was initially defined in [RFC2869](https://datatracker.ietf.org/doc/html/rfc2869) (2000) as an extension to improve authenticity and integrity verification.

However the BlastRADIUS vulnerability is now pushing vendors to generalize the use of this attibute. FreeRADIUS is now enforcing `Message-Authenticator` since versions 3.2.5 and 3.0.27. `libkrad` has to support `Message-Authenticator` in order to remain compatible with these implementations.

The present implementation does not take cryptography agility into account because, as explained in [draft-ietf-radext-deprecating-radius-03](https://datatracker.ietf.org/doc/html/draft-ietf-radext-deprecating-radius-03#name-introduction), the plan is to rely on TLS and DTLS rather than updating the algorithms used in the protocol itself:
> The state of MD5 security was again discussed in [[RFC6151](https://www.rfc-editor.org/rfc/rfc6151)], which states in Section 2:
>
> MD5 is no longer acceptable where collision resistance is required such as digital signatures.
>
> That statement led to RADIUS security being reviewed in [[RFC6421](https://www.rfc-editor.org/rfc/rfc6421)], [Section 3](https://rfc-editor.org/rfc/rfc6421#section-3). The outcome of that review was the text in the remainder of [[RFC6421](https://www.rfc-editor.org/rfc/rfc6421)], which created crypto-agility requirements for RADIUS. The main outcome of those requirements was not any change to RADIUS, but instead the definition of RADIUS/TLS in [[RFC6614](https://www.rfc-editor.org/rfc/rfc6614)], and RADIUS over DTLS in [[RFC7360](https://www.rfc-editor.org/rfc/rfc7360)]. The other outcome was a consensus that adding crypto-agility to RADIUS was likely not a good idea, and that standardizing RADIUS over TLS instead was a significantly better path forward.

### DIFF
--- a/src/lib/krad/packet.c
+++ b/src/lib/krad/packet.c
@@ -36,6 +36,7 @@
 typedef unsigned char uchar;
 
 /* RFC 2865 */
+#define MSGAUTH_SIZE (2 + MD5_DIGEST_SIZE)
 #define OFFSET_CODE 0
 #define OFFSET_ID 1
 #define OFFSET_LENGTH 2
@@ -222,6 +223,106 @@ packet_set_attrset(krb5_context ctx, const char *secret, krad_packet *pkt)
     return kr_attrset_decode(ctx, &tmp, secret, pkt_auth(pkt), &pkt->attrset);
 }
 
+/* Determine if a packet requires a Message-Authenticator attribute. */
+static inline krb5_boolean
+requires_msgauth(const char *secret, krad_code code)
+{
+    /* If no secret is provided, assume that the transport is a UNIX socket.
+     * Message-Authenticator is required only on UDP and TCP connections. */
+    if (*secret == '\0')
+        return FALSE;
+
+    /*
+     * Per draft-ietf-radext-deprecating-radius-03 sections 5.2.1 and 5.2.4,
+     * Message-Authenticator is required in Access-Request packets and all
+     * potential responses when UDP or TCP transport is used.
+     */
+    return code == krad_code_name2num("Access-Request") ||
+        code == krad_code_name2num("Access-Reject") ||
+        code == krad_code_name2num("Access-Accept") ||
+        code == krad_code_name2num("Access-Challenge");
+}
+
+/* Check if the packet has a Message-Authenticator attribute. */
+static inline krb5_boolean
+has_pkt_msgauth(const krad_packet *pkt)
+{
+    krad_attr msgauth_type = krad_attr_name2num("Message-Authenticator");
+
+    return krad_attrset_get(pkt->attrset, msgauth_type, 0) != NULL;
+}
+
+/* Return the beginning of the Message-Authenticator attribute in pkt, or NULL
+ * if no such attribute is present. */
+static const uint8_t *
+lookup_msgauth_addr(const krad_packet *pkt)
+{
+    krad_attr msgauth_type = krad_attr_name2num("Message-Authenticator");
+    size_t i;
+    uint8_t *p;
+
+    i = OFFSET_ATTR;
+    while (i + 2 < pkt->pkt.length) {
+        p = (uint8_t *)offset(&pkt->pkt, i);
+        if (msgauth_type == *p)
+            return p;
+        i += p[1];
+    }
+
+    return NULL;
+}
+
+/*
+ * Calculate the message authenticator MAC for pkt as specified in RFC 2869
+ * section 5.14, placing the result in mac_out.  Use the provided authenticator
+ * auth, which may be from pkt or from a corresponding request.
+ */
+static krb5_error_code
+calculate_mac(const char *secret, const krad_packet *pkt,
+              const uint8_t auth[AUTH_FIELD_SIZE],
+              uint8_t mac_out[MD5_DIGEST_SIZE])
+{
+    uint8_t zeroed_msgauth[MSGAUTH_SIZE];
+    krad_attr msgauth_type = krad_attr_name2num("Message-Authenticator");
+    const uint8_t *msgauth_attr, *msgauth_end, *pkt_end;
+    krb5_crypto_iov input[5];
+    krb5_data ksecr, mac;
+
+    msgauth_attr = lookup_msgauth_addr(pkt);
+    if (msgauth_attr == NULL)
+        return EINVAL;
+    msgauth_end = msgauth_attr + MSGAUTH_SIZE;
+    pkt_end = (const uint8_t *)pkt->pkt.data + pkt->pkt.length;
+
+    /* Read code, id, and length from the packet. */
+    input[0].flags = KRB5_CRYPTO_TYPE_DATA;
+    input[0].data = make_data(pkt->pkt.data, OFFSET_AUTH);
+
+    /* Read the provided authenticator. */
+    input[1].flags = KRB5_CRYPTO_TYPE_DATA;
+    input[1].data = make_data((uint8_t *)auth, AUTH_FIELD_SIZE);
+
+    /* Read any attributes before Message-Authenticator. */
+    input[2].flags = KRB5_CRYPTO_TYPE_DATA;
+    input[2].data = make_data(pkt_attr(pkt), msgauth_attr - pkt_attr(pkt));
+
+    /* Read Message-Authenticator with the data bytes all set to zero, per RFC
+     * 2869 section 5.14. */
+    zeroed_msgauth[0] = msgauth_type;
+    zeroed_msgauth[1] = MSGAUTH_SIZE;
+    memset(zeroed_msgauth + 2, 0, MD5_DIGEST_SIZE);
+    input[3].flags = KRB5_CRYPTO_TYPE_DATA;
+    input[3].data = make_data(zeroed_msgauth, MSGAUTH_SIZE);
+
+    /* Read any attributes after Message-Authenticator. */
+    input[4].flags = KRB5_CRYPTO_TYPE_DATA;
+    input[4].data = make_data((uint8_t *)msgauth_end, pkt_end - msgauth_end);
+
+    mac = make_data(mac_out, MD5_DIGEST_SIZE);
+    ksecr = string2data((char *)secret);
+    return k5_hmac_md5(&ksecr, input, 5, &mac);
+}
+
 ssize_t
 krad_packet_bytes_needed(const krb5_data *buffer)
 {
@@ -255,6 +356,7 @@ krad_packet_new_request(krb5_context ctx, const char *secret, krad_code code,
     krad_packet *pkt;
     uchar id;
     size_t attrset_len;
+    krb5_boolean msgauth_required;
 
     pkt = packet_new();
     if (pkt == NULL) {
@@ -274,9 +376,13 @@ krad_packet_new_request(krb5_context ctx, const char *secret, krad_code code,
     if (retval != 0)
         goto error;
 
+    /* Determine if Message-Authenticator is required. */
+    msgauth_required = (*secret != '\0' &&
+                        code == krad_code_name2num("Access-Request"));
+
     /* Encode the attributes. */
-    retval = kr_attrset_encode(set, secret, pkt_auth(pkt), pkt_attr(pkt),
-                               &attrset_len);
+    retval = kr_attrset_encode(set, secret, pkt_auth(pkt), msgauth_required,
+                               pkt_attr(pkt), &attrset_len);
     if (retval != 0)
         goto error;
 
@@ -284,6 +390,13 @@ krad_packet_new_request(krb5_context ctx, const char *secret, krad_code code,
     pkt->pkt.length = attrset_len + OFFSET_ATTR;
     pkt_code_set(pkt, code);
     pkt_len_set(pkt, pkt->pkt.length);
+
+    if (msgauth_required) {
+        /* Calculate and set the Message-Authenticator MAC. */
+        retval = calculate_mac(secret, pkt, pkt_auth(pkt), pkt_attr(pkt) + 2);
+        if (retval != 0)
+            goto error;
+    }
 
     /* Copy the attrset for future use. */
     retval = packet_set_attrset(ctx, secret, pkt);
@@ -307,14 +420,18 @@ krad_packet_new_response(krb5_context ctx, const char *secret, krad_code code,
     krb5_error_code retval;
     krad_packet *pkt;
     size_t attrset_len;
+    krb5_boolean msgauth_required;
 
     pkt = packet_new();
     if (pkt == NULL)
         return ENOMEM;
 
+    /* Determine if Message-Authenticator is required. */
+    msgauth_required = requires_msgauth(secret, code);
+
     /* Encode the attributes. */
-    retval = kr_attrset_encode(set, secret, pkt_auth(request), pkt_attr(pkt),
-                               &attrset_len);
+    retval = kr_attrset_encode(set, secret, pkt_auth(request),
+                               msgauth_required, pkt_attr(pkt), &attrset_len);
     if (retval != 0)
         goto error;
 
@@ -330,6 +447,18 @@ krad_packet_new_response(krb5_context ctx, const char *secret, krad_code code,
     if (retval != 0)
         goto error;
 
+    if (msgauth_required) {
+        /*
+         * Calculate and replace the Message-Authenticator MAC.  Per RFC 2869
+         * section 5.14, use the authenticator from the request, not from the
+         * response.
+         */
+        retval = calculate_mac(secret, pkt, pkt_auth(request),
+                               pkt_attr(pkt) + 2);
+        if (retval != 0)
+            goto error;
+    }
+
     /* Copy the attrset for future use. */
     retval = packet_set_attrset(ctx, secret, pkt);
     if (retval != 0)
@@ -341,6 +470,34 @@ krad_packet_new_response(krb5_context ctx, const char *secret, krad_code code,
 error:
     free(pkt);
     return retval;
+}
+
+/* Verify the Message-Authenticator value in pkt, using the provided
+ * authenticator (which may be from pkt or from a corresponding request). */
+static krb5_error_code
+verify_msgauth(const char *secret, const krad_packet *pkt,
+               const uint8_t auth[AUTH_FIELD_SIZE])
+{
+    uint8_t mac[MD5_DIGEST_SIZE];
+    krad_attr msgauth_type = krad_attr_name2num("Message-Authenticator");
+    const krb5_data *msgauth;
+    krb5_error_code retval;
+
+    msgauth = krad_packet_get_attr(pkt, msgauth_type, 0);
+    if (msgauth == NULL)
+        return ENODATA;
+
+    retval = calculate_mac(secret, pkt, auth, mac);
+    if (retval)
+        return retval;
+
+    if (msgauth->length != MD5_DIGEST_SIZE)
+        return EMSGSIZE;
+
+    if (k5_bcmp(mac, msgauth->data, MD5_DIGEST_SIZE) != 0)
+        return EBADMSG;
+
+    return 0;
 }
 
 /* Decode a packet. */
@@ -394,21 +551,35 @@ krad_packet_decode_request(krb5_context ctx, const char *secret,
                            krad_packet **reqpkt)
 {
     const krad_packet *tmp = NULL;
+    krad_packet *req;
     krb5_error_code retval;
 
-    retval = decode_packet(ctx, secret, buffer, reqpkt);
-    if (cb != NULL && retval == 0) {
+    retval = decode_packet(ctx, secret, buffer, &req);
+    if (retval)
+        return retval;
+
+    /* Verify Message-Authenticator if present. */
+    if (has_pkt_msgauth(req)) {
+        retval = verify_msgauth(secret, req, pkt_auth(req));
+        if (retval) {
+            krad_packet_free(req);
+            return retval;
+        }
+    }
+
+    if (cb != NULL) {
         for (tmp = (*cb)(data, FALSE); tmp != NULL; tmp = (*cb)(data, FALSE)) {
             if (pkt_id_get(*reqpkt) == pkt_id_get(tmp))
                 break;
         }
+
+        if (tmp != NULL)
+            (*cb)(data, TRUE);
     }
 
-    if (cb != NULL && (retval != 0 || tmp != NULL))
-        (*cb)(data, TRUE);
-
+    *reqpkt = req;
     *duppkt = tmp;
-    return retval;
+    return 0;
 }
 
 krb5_error_code
@@ -435,9 +606,17 @@ krad_packet_decode_response(krb5_context ctx, const char *secret,
                 break;
             }
 
-            /* If the authenticator matches, then the response is valid. */
-            if (memcmp(pkt_auth(*rsppkt), auth, sizeof(auth)) == 0)
-                break;
+            /* Verify the response authenticator. */
+            if (k5_bcmp(pkt_auth(*rsppkt), auth, sizeof(auth)) != 0)
+                continue;
+
+            /* Verify Message-Authenticator if present. */
+            if (has_pkt_msgauth(*rsppkt)) {
+                if (verify_msgauth(secret, *rsppkt, pkt_auth(tmp)) != 0)
+                    continue;
+            }
+
+            break;
         }
     }
 
--- a/src/lib/krad/t_packet.c
+++ b/src/lib/krad/t_packet.c
@@ -172,6 +172,9 @@ main(int argc, const char **argv)
     krb5_data username, password;
     krb5_boolean auth = FALSE;
     krb5_context ctx;
+    const krad_packet *dupreq;
+    const krb5_data *encpkt;
+    krad_packet *decreq;
 
     username = string2data("testUser");
 
@@ -184,9 +187,17 @@ main(int argc, const char **argv)
 
     password = string2data("accept");
     noerror(make_packet(ctx, &username, &password, &packets[ACCEPT_PACKET]));
+    encpkt = krad_packet_encode(packets[ACCEPT_PACKET]);
+    noerror(krad_packet_decode_request(ctx, "foo", encpkt, NULL, NULL,
+                                       &dupreq, &decreq));
+    krad_packet_free(decreq);
 
     password = string2data("reject");
     noerror(make_packet(ctx, &username, &password, &packets[REJECT_PACKET]));
+    encpkt = krad_packet_encode(packets[REJECT_PACKET]);
+    noerror(krad_packet_decode_request(ctx, "foo", encpkt, NULL, NULL,
+                                       &dupreq, &decreq));
+    krad_packet_free(decreq);
 
     memset(&hints, 0, sizeof(hints));
     hints.ai_family = AF_INET;
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2404,4 +2404,9 @@ krb5_boolean
 k5_sname_compare(krb5_context context, krb5_const_principal sname,
                  krb5_const_principal princ);
 
+/* Generate an HMAC-MD5 keyed checksum as specified by RFC 2104. */
+krb5_error_code
+k5_hmac_md5(const krb5_data *key, const krb5_crypto_iov *data, size_t num_data,
+            krb5_data *output);
+
 #endif /* _KRB5_INT_H */
--- a/src/lib/crypto/krb/checksum_hmac_md5.c
+++ b/src/lib/crypto/krb/checksum_hmac_md5.c
@@ -92,3 +92,31 @@ cleanup:
     free(hash_iov);
     return ret;
 }
+
+krb5_error_code
+k5_hmac_md5(const krb5_data *key, const krb5_crypto_iov *data, size_t num_data,
+            krb5_data *output)
+{
+    krb5_error_code ret;
+    const struct krb5_hash_provider *hash = &krb5int_hash_md5;
+    krb5_keyblock keyblock = { 0 };
+    krb5_data hashed_key;
+    uint8_t hkeybuf[16];
+    krb5_crypto_iov iov;
+
+    /* Hash the key if it is longer than the block size. */
+    if (key->length > hash->blocksize) {
+        hashed_key = make_data(hkeybuf, sizeof(hkeybuf));
+        iov.flags = KRB5_CRYPTO_TYPE_DATA;
+        iov.data = *key;
+        ret = hash->hash(&iov, 1, &hashed_key);
+        if (ret)
+            return ret;
+        key = &hashed_key;
+    }
+
+    keyblock.magic = KV5M_KEYBLOCK;
+    keyblock.length = key->length;
+    keyblock.contents = (uint8_t *)key->data;
+    return krb5int_hmac_keyblock(hash, &keyblock, data, num_data, output);
+}
--- a/src/lib/crypto/libk5crypto.exports
+++ b/src/lib/crypto/libk5crypto.exports
@@ -103,3 +103,4 @@ krb5_c_prfplus
 krb5_c_derive_prfplus
 k5_enctype_to_ssf
 krb5int_c_deprecated_enctype
+k5_hmac_md5
--- a/src/lib/krad/attr.c
+++ b/src/lib/krad/attr.c
@@ -122,6 +122,23 @@ static const attribute_record attributes[UCHAR_MAX] = {
     {"NAS-Port-Type", 4, 4, NULL, NULL},
     {"Port-Limit", 4, 4, NULL, NULL},
     {"Login-LAT-Port", 1, MAX_ATTRSIZE, NULL, NULL},
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for tunnelling */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for tunnelling */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for tunnelling */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for tunnelling */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for tunnelling */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for tunnelling */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for Apple Remote Access Protocol */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for Apple Remote Access Protocol */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for Apple Remote Access Protocol */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for Apple Remote Access Protocol */
+    {NULL, 0, 0, NULL, NULL}, /* Reserved for Apple Remote Access Protocol */
+    {NULL, 0, 0, NULL, NULL}, /* Password-Retry */
+    {NULL, 0, 0, NULL, NULL}, /* Prompt */
+    {NULL, 0, 0, NULL, NULL}, /* Connect-Info */
+    {NULL, 0, 0, NULL, NULL}, /* Configuration-Token */
+    {NULL, 0, 0, NULL, NULL}, /* EAP-Message */
+    {"Message-Authenticator", MD5_DIGEST_SIZE, MD5_DIGEST_SIZE, NULL, NULL},
 };
 
 /* Encode User-Password attribute. */
--- a/src/lib/krad/attrset.c
+++ b/src/lib/krad/attrset.c
@@ -164,14 +164,42 @@ krad_attrset_copy(const krad_attrset *set, krad_attrset **copy)
     return 0;
 }
 
+/* Place an encoded attributes into outbuf at position *i.  Increment *i by the
+ * length of the encoding. */
+static krb5_error_code
+append_attr(krb5_context ctx, const char *secret,
+            const uint8_t *auth, krad_attr type, const krb5_data *data,
+            uint8_t outbuf[MAX_ATTRSETSIZE], size_t *i)
+{
+    uint8_t buffer[MAX_ATTRSIZE];
+    size_t attrlen;
+    krb5_error_code retval;
+
+    retval = kr_attr_encode(ctx, secret, auth, type, data, buffer, &attrlen);
+    if (retval)
+        return retval;
+
+    if (attrlen > MAX_ATTRSETSIZE - *i - 2)
+        return EMSGSIZE;
+
+    outbuf[(*i)++] = type;
+    outbuf[(*i)++] = attrlen + 2;
+    memcpy(outbuf + *i, buffer, attrlen);
+    *i += attrlen;
+
+    return 0;
+}
+
 krb5_error_code
 kr_attrset_encode(const krad_attrset *set, const char *secret,
-                  const unsigned char *auth,
+                  const uint8_t *auth, krb5_boolean add_msgauth,
                   unsigned char outbuf[MAX_ATTRSETSIZE], size_t *outlen)
 {
-    unsigned char buffer[MAX_ATTRSIZE];
     krb5_error_code retval;
-    size_t i = 0, attrlen;
+    krad_attr msgauth_type = krad_attr_name2num("Message-Authenticator");
+    const uint8_t zeroes[MD5_DIGEST_SIZE] = { 0 };
+    krb5_data zerodata;
+    size_t i = 0;
     attr *a;
 
     if (set == NULL) {
@@ -179,19 +207,21 @@ kr_attrset_encode(const krad_attrset *set, const char *secret,
         return 0;
     }
 
-    K5_TAILQ_FOREACH(a, &set->list, list) {
-        retval = kr_attr_encode(set->ctx, secret, auth, a->type, &a->attr,
-                                buffer, &attrlen);
-        if (retval != 0)
+    if (add_msgauth) {
+        /* Encode Message-Authenticator as the first attribute, per
+         * draft-ietf-radext-deprecating-radius-03 section 5.2. */
+        zerodata = make_data((uint8_t *)zeroes, MD5_DIGEST_SIZE);
+        retval = append_attr(set->ctx, secret, auth, msgauth_type, &zerodata,
+                             outbuf, &i);
+        if (retval)
             return retval;
+    }
 
-        if (i + attrlen + 2 > MAX_ATTRSETSIZE)
-            return EMSGSIZE;
-
-        outbuf[i++] = a->type;
-        outbuf[i++] = attrlen + 2;
-        memcpy(&outbuf[i], buffer, attrlen);
-        i += attrlen;
+    K5_TAILQ_FOREACH(a, &set->list, list) {
+        retval = append_attr(set->ctx, secret, auth, a->type, &a->attr,
+                             outbuf, &i);
+        if (retval)
+            return retval;
     }
 
     *outlen = i;
--- a/src/lib/krad/internal.h
+++ b/src/lib/krad/internal.h
@@ -43,6 +43,8 @@
 #define UCHAR_MAX 255
 #endif
 
+#define MD5_DIGEST_SIZE 16
+
 /* RFC 2865 */
 #define MAX_ATTRSIZE (UCHAR_MAX - 2)
 #define MAX_ATTRSETSIZE (KRAD_PACKET_SIZE_MAX - 20)
@@ -65,10 +67,11 @@ kr_attr_decode(krb5_context ctx, const char *secret, const unsigned char *auth,
                krad_attr type, const krb5_data *in,
                unsigned char outbuf[MAX_ATTRSIZE], size_t *outlen);
 
-/* Encode the attributes into the buffer. */
+/* Encode set into outbuf.  If add_msgauth is true, include a zeroed
+ * Message-Authenticator as the first attribute. */
 krb5_error_code
 kr_attrset_encode(const krad_attrset *set, const char *secret,
-                  const unsigned char *auth,
+                  const uint8_t *auth, krb5_boolean add_msgauth,
                   unsigned char outbuf[MAX_ATTRSETSIZE], size_t *outlen);
 
 /* Decode attributes from a buffer. */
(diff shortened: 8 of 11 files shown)


jrisc:
Tests are failing because `pyrad` seems to not support `Message-Authenticator`.

src/lib/krad/packet.c
```diff
@@ -222,6 +224,175 @@ packet_set_attrset(krb5_context ctx, const char *secret, krad_packet *pkt)
...
 
+/* Determines if request or response requires a Message-Authenticator
+ * attribute. */
+inline static krb5_boolean
+requires_msgauth(const char *secret, krad_code code)
+{
+    /* If no secret provided, make the assumption packets exchange will happen
+     * on UNIX socket. Message-Authenticator is required only on UDP and TCP
+     * connections. */
+    if (!secret || 0 == strlen(secret))
```
jrisc, line 235: FreeIPA uses `libkrad` to exchange radius messages between 2 local daemons using a UNIX socket. In this case we pass an empty string as secret, since data protection is not needed. I use this fact to determine if verification of `Message-Authenticator` is required or not.

The problem is some applications might be using libkrad on UNIX sockets with a secret. In such case, enforcing `Message-Authenticator` does not make sense.

A better approach would be to check if the message was sent via a UDP or TCP connection, but such information is not available in `krad_packet_decode_request()` nor `krad_packet_decode_response()`.

abbra:
> Tests are failing because `pyrad` seems to not support `Message-Authenticator`.

It does have support for `Message-Authenticator`.

I instructed the tests and see the following:
```
LD_LIBRARY_PATH=`echo -L../../lib | sed -e "s/-L//g" -e "s/ /:/g"` KRB5_CONFIG=../../config-files/krb5.conf LC_ALL=C  ./t_remote python3 ./t_daemon.py
Packet AuthPacket({80: [b'\x9a\xee\x80\xd7"{\x07\xaf\x84RW\xfdm\xcc\x13\x04'], 'User-Name': ['testUser'], 'User-Password': [b'\x9fq\x17\xc2\x82\t.\xa7\x95\xc1c\x01?uY\xb7']}) handled. Reply is 2
Packet AuthPacket({80: [b'\x9a\xee\x80\xd7"{\x07\xaf\x84RW\xfdm\xcc\x13\x04'], 'User-Name': ['testUser'], 'User-Password': [b'\x9fq\x17\xc2\x82\t.\xa7\x95\xc1c\x01?uY\xb7']}) handled. Reply is 2
Packet AuthPacket({80: [b'\x9a\xee\x80\xd7"{\x07\xaf\x84RW\xfdm\xcc\x13\x04'], 'User-Name': ['testUser'], 'User-Password': [b'\x9fq\x17\xc2\x82\t.\xa7\x95\xc1c\x01?uY\xb7']}) handled. Reply is 2
Packet AuthPacket({80: [b'\x9a\xee\x80\xd7"{\x07\xaf\x84RW\xfdm\xcc\x13\x04'], 'User-Name': ['testUser'], 'User-Password': [b'\x9fq\x17\xc2\x82\t.\xa7\x95\xc1c\x01?uY\xb7']}) handled. Reply is 2
req: 0x1c25e010, resp: (nil), retval = 110, evt.error = 1, evt.result = 110
Packet AuthPacket({80: [b'<\xef\xb2\x0c\x16N\xbf\xa8z\xcctp\xa8\x8by\x16'], 'User-Name': ['testUser'], 'User-Password': [b'\xb1\x7f\x90\x87\x9b\xfez\n\xde|y\x90\xd7\xa4\x9c\x19']}) handled. Reply is 3
Packet AuthPacket({80: [b'<\xef\xb2\x0c\x16N\xbf\xa8z\xcctp\xa8\x8by\x16'], 'User-Name': ['testUser'], 'User-Password': [b'\xb1\x7f\x90\x87\x9b\xfez\n\xde|y\x90\xd7\xa4\x9c\x19']}) handled. Reply is 3
Packet AuthPacket({80: [b'<\xef\xb2\x0c\x16N\xbf\xa8z\xcctp\xa8\x8by\x16'], 'User-Name': ['testUser'], 'User-Password': [b'\xb1\x7f\x90\x87\x9b\xfez\n\xde|y\x90\xd7\xa4\x9c\x19']}) handled. Reply is 3
Packet AuthPacket({80: [b'<\xef\xb2\x0c\x16N\xbf\xa8z\xcctp\xa8\x8by\x16'], 'User-Name': ['testUser'], 'User-Password': [b'\xb1\x7f\x90\x87\x9b\xfez\n\xde|y\x90\xd7\xa4\x9c\x19']}) handled. Reply is 3
req: 0x1c25f030, resp: (nil), retval = 110, evt.error = 1, evt.result = 110
req: 0x1c260a50, resp: (nil), retval = 125, evt.error = 1, evt.result = 125
req: 0x1c260920, resp: (nil), retval = 110, evt.error = 1, evt.result = 110
req: 0x1c25e010, resp: (nil), retval = 125, evt.error = 1, evt.result = 125
req: 0x1c25f030, resp: (nil), retval = 125, evt.error = 1, evt.result = 125
req: 0x1c260920, resp: (nil), retval = 125, evt.error = 1, evt.result = 125
req: 0x1c2640f0, resp: (nil), retval = 125, evt.error = 1, evt.result = 125
t_remote.c:153: insist failed:
	record.events[0].error == FALSE
```

It looks like in both cases I get either ETIMEDOUT (110) or ECANCELLED (125) as a result.

abbra:
Thank you, now it works. The only issues I see are in documentation generator:
```
Warning, treated as error:
/home/runner/work/krb5/krb5/src/doc/rst_composite/appdev/refs/api/krb5_rfc2104_hmacmd5_checksum.rst:document isn't included in any toctree
```

The other doc test needs github actions upgrade.

greghudson:
I would like to avoid making the new libk5crypto API public, as it isn't in scope for an RFC 3961 library.  It looks like that would solve the documentation build issue, whatever that is.  See k5_sha256() for an example of a non-public libk5crypto API.

greghudson:
Why add krad_packet_decode_response_unsafe()?

jrisc:
I initially added `krad_packet_decode_response_unsafe()` because I thought `pyrad` could not generate Message-Authenticator. We can drop this commit if you think there is no need to provide a way to decode response packets without Message-Authenticator.

greghudson:
If we don't have immediate plans to use then I'd prefer to drop it.

jrisc:
Hello @greghudson,
Is there anything else you think I should change or add in this PR to make it ready for merging?

greghudson:
I have some changes in progress.

I have one discussion point: I'm not sure we need a requires_msgauth() that checks the packet code against a list of codes.  The draft doesn't discuss other RADIUS packet types (maybe because they aren't in common use?), but the reasoning for requiring Message-Authenticator applies to any request or response as far as I can tell.

jrisc:
I had a look at the [freeradius-server implementation](https://github.com/FreeRADIUS/freeradius-server/blob/4312a2df8e0829c87811f42da7591a852350c068/src/protocols/radius/base.c#L367-L386), and it seems they generate Message-Authenticator for a few codes in addition to the ones mentioned in the IETF draft:
* Access-Accept
* Access-Challenge
* Access-Reject
* Access-Request
* Accounting-Request
* Accounting-Response
* CoA-ACK
* CoA-NAK
* CoA-Request
* Disconnect-ACK
* Disconnect-NAK
* Disconnect-Request
* Status-Server

I think libkrad supports all these codes, except the CoA ones. I have no clue why they excluded the other codes like Accounting-Status, Password-*, Status-Client, ... and most importantly Protocol-Error which must have a Message-Authenticator according to the IETF draft.

jrisc:
I've sent a message to freeradius-devel@lists.freeradius.org (with copy to [krbdev](https://mailman.mit.edu/pipermail/krbdev/2024-October/013683.html)) asking about the relevant cases when to generate a Message-Authenticator.

alandekok:
> I had a look at the [freeradius-server implementation](https://github.com/FreeRADIUS/freeradius-server/blob/4312a2df8e0829c87811f42da7591a852350c068/src/protocols/radius/base.c#L367-L386), and it seems they generate Message-Authenticator for a few codes in addition to the ones mentioned in the IETF draft:

Please check the code again.

The server mandates the include of `Message-Authenticator` for Access-* and Protocol-Error, as per draft-ietf-radext-deprecating-radius, and for Status-Server as per RFC 5997.

The server _allows_ `Message-Authenticator` in other packet types, if the administrator requests that it be added to the packet.  However, it is not added automatically, and it is not required.

greghudson:
Thanks for initiating that discussion.  It sounds like I was wrong that Message-Authenticator should be considered equally required in all RADIUS packets.

To a certain extent I feel like libkrad only exists to facilitate Access-Requests, and maybe the code check is unnecessary for that reason.  But I can see a reasonable counterargument that nothing currently prevents libkrad from acting as a generic RADIUS library, and adding implicit assumptions isn't a safe practice.  So we'll leave the code check in for now.

Tangentially: I don't like that libkrad and the OTP preauth module look up constant strings in a constant table so much (e.g. `krad_code_name2num("Access-Request")`).  Although the work it does isn't very performance-sensitive, it's distracting to see that much needless inefficiency.  After this PR is done I will consider introducing some numeric constants to krad.h and make the code use those instead.

abbra:
@greghudson regarding `krad_code_name2num()` usage, FreeIPA code also uses this API (since krad code and the one in FreeIPA was written by the same folks), so I hope the API will still stand for use -- at least for some MIT Kerberos releases. It also was a bit awkward in my experience, replacing it with compile-time constants would be nice.

greghudson:
Okay, some changes pushed.  (It looks like I didn't tidy up one of my XXX comments, relating to wanting to rename "signature" in verify_msgauth().  I'll fix that before merging.)

I am considering three more changes:

* In calculate_msgauth(), have the caller always pass the authenticator (easy to supply pkt_auth(pkt) for request packets) and always search for the Message-Authenticator attribute.  (And always check its available length, I think?  Currently the length is unchecked for Access-Request packets, which means `msgauth_attr + MSGAUTH_SIZE` could be beyond the end of the packet.)  I think this would make the control flow easier to understand.

* Remove clone_attrs_plus_zeroed_msgauth(), and instead add a flag to kr_attrset_encode() to put a zeroed Message-Authenticator attribute in front.  (This would eliminate a current check to make sure the caller didn't put in a Message-Authenticator attribute; we can put that in kr_attrset_encode() perhaps.)

* While we apparently shouldn't treat Message-Authenticator as required except for Access-Request and its replies, is there any reason we shouldn't generate one for all RADIUS packets?  (In which case kr_attrset_encode() wouldn't need the aforementioned flag; it could just do it.)

src/lib/krad/packet.c
```diff
@@ -222,6 +224,162 @@ packet_set_attrset(krb5_context ctx, const char *secret, krad_packet *pkt)
...
+/*
+ * Calculate a message authenticator for pkt as specified in RFC 2869 section
+ * 5.14, placing the result in msgauth_out.  If auth is non-null, use it
+ * instead of the authenticator in pkt, and assume that the first attribute in
+ * pkt is Message-Authenticator.
+ */
+static krb5_error_code
+calculate_msgauth(const char *secret, const krad_packet *pkt,
+                  uint8_t auth[AUTH_FIELD_SIZE],
+                  uint8_t msgauth_out[MD5_DIGEST_SIZE])
```
jrisc, line 292: I think that `signature` was actually a better name here because we are not generating the full Message-Authenticator (code + length + signature/keyed checksum), it's only the signature part.

It's actually how it's called [in the RFC](https://datatracker.ietf.org/doc/html/rfc2869#page-34):
> When the checksum is calculated the signature string should be
> considered to be sixteen octets of zero.

jrisc:
>     * In calculate_msgauth(), have the caller always pass the authenticator (easy to supply pkt_auth(pkt) for request packets) and always search for the Message-Authenticator attribute.  (And always check its available length, I think?  Currently the length is unchecked for Access-Request packets, which means `msgauth_attr + MSGAUTH_SIZE` could be beyond the end of the packet.)  I think this would make the control flow easier to understand.

Since `verify_msgauth()` is always called after `decode_packet()`, which calls `kr_attrset_decode()` where each attribute length is checked, I don't any packet size-related test is necessary.

jrisc:
> * Remove clone_attrs_plus_zeroed_msgauth(), and instead add a flag to kr_attrset_encode() to put a zeroed Message-Authenticator attribute in front.  (This would eliminate a current check to make sure the caller didn't put in a Message-Authenticator attribute; we can put that in kr_attrset_encode() perhaps.)

I thought about this, but I didn't want to do it because I was afraid adding a flag would break ABI compatibility. But now I realize that the `kr_attrset_encode()` function is not mentioned in `libkrad.exports`. So adding a flag parameter, and modify the function to add a zeroed Message-Authenticator in its encoded form directly would be a simpler approach indeed.

I see that in both cases where `kr_attrset_encode()` is used, the input attrset variable is not reused afterwards. And the attrset member of the packet structure is populated based on the packet buffer in `packet_set_attrset()`. So the fact the initial constant attrset argument of `krad_packet_new_request()` and `krad_packet_new_response()` will not match the attributes encoding should not cause any issue.

jrisc:
> * While we apparently shouldn't treat Message-Authenticator as required except for Access-Request and its replies, is there any reason we shouldn't generate one for all RADIUS packets?  (In which case kr_attrset_encode() wouldn't need the aforementioned flag; it could just do it.)

I'm afraid I don't know enough about RADIUS to recommend any change that go beyond the content of the RFCs. All I can say is that if we decide to generalize the use of Message-Authenticator, we should still have some kind of proper switch to disable it for cases where it is irrelevant (e.g. UNIX socket connections).

jrisc:
I will make some tests tomorrow to confirm the latest changes did not affect the compatibility with FreeRADIUS.

alandekok:
On Oct 3, 2024, at 12:02 PM, Greg Hudson ***@***.***> wrote:
> 	• While we apparently shouldn't treat Message-Authenticator as required except for Access-Request and its replies, is there any reason we shouldn't generate one for all RADIUS packets? (In which case kr_attrset_encode() wouldn't need the aforementioned flag; it could just do it.)

  At this time, there is no value in adding it to other RADIUS packets.

  The cryptographers examined the other packet types, and determined that they are not vulnerable.

alandekok:
On Oct 3, 2024, at 12:43 PM, Julien Rische ***@***.***> wrote:
> Since verify_msgauth() is always called after decode_packet(), which calls kr_attrset_decode() where each attribute length is checked, I don't any packet size-related test is necessary.

  The Message-Authenticator and Response Authenticator should be checked before any packet decoding is done.  The processing steps are generally:

* verify that it is a well-formed RADIUS packet.

* see if it's a packet which we accept

* verify Message-Authenticator (and Response Authenticator if relevant)

* only then is it a "good" packet which can be used to match responses to requests, do decoding, etc.

greghudson:
verify_msgauth() calls `krad_packet_get_attr(rsp, msgauth_type, 0);` which unconditionally dereferences `rsp`.  Then it has a conditional on rsp being non-null.  What is the intent there?  It seems like this would crash krad_packet_decode_request() (used by FreeIPA).

greghudson:
I think the latest force-push covers most of the code changes I wanted to make:

* I renamed calculate_msgauth() to calculate_mac() per our conversation about distinguishing the Message-Authenticator payload from the full attribute.  As previously discussed, I changed it to always accept an authenticator (either from the packet or its corresponding request) and always look up the Message-Authenticator attribute location.

* I folded in the fixes to verify_msgauth(), and then simplified its signature to use the same inputs as calculate_mac() (packet and authenticator) instead of one or two packets.

* As previously discussed, I added a flag to kr_attrset_encode() and used that instead of clone_attrs_plus_zeroed_msgauth().

As I understand it, a consequence of this PR is that we will stop interoperating with old RADIUS servers that don't set a Message-Authenticator in replies.   I think we want to return a better error for this case than ENOENT.

jrisc:
Sorry @greghudson , I wanted to force-push changes on my `Fix Message-Authenticator verification for requests` commit, but I haven't realized you made changes in-between. I will restore you changes and rebase on them.

src/lib/krad/packet.c
```diff
@@ -397,18 +562,30 @@ krad_packet_decode_request(krb5_context ctx, const char *secret,
...
 
     retval = decode_packet(ctx, secret, buffer, reqpkt);
-    if (cb != NULL && retval == 0) {
-        for (tmp = (*cb)(data, FALSE); tmp != NULL; tmp = (*cb)(data, FALSE)) {
+    if (retval)
+        return retval;
+
+    /* Verify Message-Authenticator if present. */
+    if (has_pkt_msgauth(*reqpkt)) {
+        retval = verify_msgauth(secret, *reqpkt, pkt_auth(*reqpkt));
```
jrisc, line 570: The reason why the null `rsp` mistake in `verify_msgauth()` was never noticed is because this function was called at the wrong place in `krad_packet_decode_request()` (it was verifying potential duplicated request rather than the one being actually decoded).

src/lib/krad/t_packet.c
```diff
@@ -184,9 +186,13 @@ main(int argc, const char **argv)
 
     password = string2data("accept");
     noerror(make_packet(ctx, &username, &password, &packets[ACCEPT_PACKET]));
+    noerror(krad_packet_decode_request(ctx, "foo", &packets[ACCEPT_PACKET]->pkt,
```
jrisc, line 189: It seems we were missing calls to `krad_packet_decode_request()` in the tests. However, the structure definition of the `krb5_packet` generated by `krad_packet_new_request()` is private to `packet.c`. I had to move it to `internal.h` in order to access its `pkt` member.

jrisc:
> * verify that it is a well-formed RADIUS packet.
> * see if it's a packet which we accept
> * verify Message-Authenticator (and Response Authenticator if relevant)
> * only then is it a "good" packet which can be used to match responses to requests, do decoding, etc.

Does this mean we are suppose to verify the Message-Authenticator before actually decoding the attributes? This is definitely doable when processing a Access-Accept, Access-Reject, Access-Challenge, or Protocol-Error response, because the Message-Authenticator must be the first attribute, so be know its location in the buffer without having to decode the attributes section of the packet.

But for Access-Request, [draft-ietf-radext-deprecating-radius-03](https://datatracker.ietf.org/doc/html/draft-ietf-radext-deprecating-radius-03#section-5.2.1-3) mentions:

> The Message-Authenticator SHOULD be the first attribute in all Access-Request packets. That is, it should be placed immediately after the packet header. Implementations MAY place the Message-Authenticator elsewhere in an Access-Request packet.

So my understanding is we cannot do such an assumption for Access-Request, because we would have to at least decode the length of each of potential attributes preceding it.

To summarize, should we verify the Message-Authenticator before doing any attributes decoding on Access-Accept, Access-Reject, Access-Challenge, or Protocol-Error? And is it okay to search its position in an Access-Request, since contrary to the former packet types, the position of the Message-Authenticator is not pre-defined?

greghudson:
I wouldn't really be in favor of moving Message-Authenticator earlier at the expense of greater complexity in the verification code.  As RADIUS packets do not contain ciphertext (precluding any padding oracle or timing channel attacks), the most compelling argument I can think of for an earlier check is that it would maximize the likelihood that authentication is required to exploit a hypothetical vulnerability in the packet processing code.  But if the verification code becomes more complex as a result of not decoding the packet first, then the likelihood of a vulnerability increases, so it isn't clear that there would be a net gain in the overall safety of the code.

It might not be _that_ much more complicated for MAC verification to operate on the serialized buffer rather than a decoded packet.  It already kind of does, since lookup_msgauth_attr() iterates through the serialized packet rather than the decoded attrset.  But we would have to make a lot of small changes.  Relative to the current code:

* We would want to separate out some of decode_packet() into a well-formedness check that could be performed prior to MAC verification.

* The convenience macros like pkt_auth() and pkt_attr() would have to be rethought; they index into the serialization of a packet, but they require on a pointer to a decoded packet structure as input.

* calculate_mac() would have to check the length of the Message-Authenticator attribute rather than relying on it happening during packet decoding.  (I am assuming that the well-formedness check would reject any attribute lengths that exceed the remaining buffer length, but wouldn't know about specific attribute types.)

alandekok:
@greghudson The issue for RADIUS packets is that they're UDP, and can be forged.  From a strict security point of view, an implementation should take the following steps when processing packets:

* verify src/dst ip/port are expected  / authorized
* verify that the packet is "well formed"
* validate any authenticators
* only then start to allocate memory, decode the packet into internal data structures, etc.

The alternative is to allow an attacker to send large amounts of spoofed packets, which can cause the implementation to allocate large amounts of memory, albeit temporarily.

if the code has some kind of rate limiting, then the attack is less concerning.  A server would have to implement some kind of upper bound on "received but not yet validated" packets.  A client would have to check "we already have a reply for this request, discard any duplicate packets".

The "check for well-formed packet" code is pretty simple.  Maybe 50 lines of code:

```
if packet length is less than 20, discard
if header.code is not as expected, discard
if header.length > packet length discard
if header.length < packet length, truncate packet length to header.length

for each attribute in packet
    if attribute postion + 2 > end of packet, discard
    if attribute position + attribute.length > end of packet, discard
    if attribute.code == message authenticator && attribute.length != 18, discard
    skip ahead to next attribute
```

There's no need to check specific attribute types at this stage.

Once the packet is known to be well-formed, it should have its contents authenticated.  the shared secret is used to validate the Message-Authenticator, Request Authenticator (non-Access-Request packets), and Response Authenticator.

Only then does the implementation have a valid and correct RADIUS packet.  It can then start allocating memory to decode the attributes, etc.

greghudson:
I am not sure that allocating and discarding temporary memory to process a spoofed packet is a big concern for us.  For the KDC, there are probably other ways to incur the same or a greater cost by sending Kerberos packets (which go through ASN.1 DER decoding as a first step).

But I also want to zero in on:

> if header.code is not as expected, discard

libkrad's current decoding APIs (krad_packet_decode_request() and krad_packet_decode_response()) don't have an input that would let us check the code against a short list.

Of course if we skip this check, a spoofer who wants to exercise the decoder can simply use a code that doesn't require a MAC.  The packet wouldn't be discarded based on the code until it has already been decoded and the library caller looks at it.

So if we want to be serious about minimizing the processing cost incurred by spoofed packets, we would need to either (1) amend the API (not impossible; it's explicitly not a stable API), or (2) declare that libkrad is only about processing Access-Request queries and responses, in which case a MAC is always required when the secret is non-empty.

jrisc:
I think splitting  `kr_attrset_decode()` into two different functions (one for validation, and one for attributes allocation) shouldn't too complicated to do. And so is the modification of `calculate_mac()` to accept a packet buffer rather than a `krad_packet`. But since `krad_packet` contains the packet buffer, it is just about passing a member rather than the all structure. And so it is for macros like `pkt_auth()`.

> (2) declare that libkrad is only about processing Access-Request queries and responses, in which case a MAC is always required when the secret is non-empty.

When it comes to removing support for all non Access-* packets, I am worried about the impact it would have because [we are currently accepting a lot of them](https://github.com/krb5/krb5/blob/krb5-1.21.3-final/src/lib/krad/code.c#L34-L86):

* Accounting-Message
* Accounting-Request
* Accounting-Response
* Accounting-Status
* Alternate-Resource-Reclaim-Request
* Change-Filters-Ack
* Change-Filters-Nak
* Change-Filters-Request
* Disconnect-Ack
* Disconnect-Nak
* Disconnect-Request
* Event-Request
* Event-Response
* IP-Address-Allocate
* IP-Address-Release
* NAS-Reboot-Request
* NAS-Reboot-Response
* New-Pin
* Next-Passcode
* Password-Ack
* Password-Expired
* Password-Reject
* Password-Request
* Resource-Free-Request
* Resource-Free-Response
* Resource-Query-Request
* Resource-Query-Response
* Status-Client
* Status-Server
* Terminate-Session

> (1) amend the API (not impossible, as it's explicitly not a stable API)

Given the fact the primary problem here is continuity of service, since new FreeRADIUS versions (against which libkrad is not compatible) were already released on some systems (and more are going to be updated in the incoming weeks), I think we should try to fix the problem as much as possible internally in the library. Any ABI change will delay the adoption of the fix even more. So I would suggest to solve the "expected packet code" problem in a different PR.

alandekok:
On Oct 7, 2024, at 6:05 AM, Julien Rische ***@***.***> wrote:
> I think splitting kr_attrset_decode() into two different functions (one for validation, and one for attributes allocation) shouldn't too complicated to do. And so is the modification of calculate_mac() to accept a packet buffer rather than a krad_packet. But since krad_packet contains the packet buffer, it is just about passing a member rather than the all structure. And so it is for macros like pkt_auth().
> 
> (2) declare that libkrad is only about processing Access-Request queries and responses, in which case a MAC is always required when the secret is non-empty.

  I would say a MAC is always required on _sending_, but is _configurably required_ on reception.  See the IETF draft for details.

> When it comes to removing support for all non Access-* packets, I am worried about the impact it would have because we are currently accepting a lot of them:

  Does the library use these packets for anything?  I've split them below into groups:

> 	• Accounting-Request
> 	• Accounting-Response

  Normal accounting.  Only needed if the client tracks user sessions, and sends periodic updates 

> 	• Change-Filters-Ack
> 	• Change-Filters-Nak
> 	• Change-Filters-Request

  These don't exist.  They are CoA-Request, CoA-ACK, CoA-NAK.

  These packets are sent to the client to change authorization.  i.e. "I gave you 10Mb, but I changed my mind.  You now have 20Mb".

  If the client doesn't track user sessions, these packets can be ignored.

> 	• Disconnect-Ack
> 	• Disconnect-Nak
> 	• Disconnect-Request

 These are sent to the client to ask it to disconnect the user.

  If the client doesn't track user sessions, these packets can be ignored.

> 	• Status-Server

  Sent by the client to the  server, to see if the server is still alive.

> 	• Accounting-Message
> 	• Accounting-Status
> 	• Alternate-Resource-Reclaim-Request
> 	• Event-Request
> 	• Event-Response
> 	• IP-Address-Allocate
> 	• IP-Address-Release
> 	• NAS-Reboot-Request
> 	• NAS-Reboot-Response
> 	• New-Pin
> 	• Next-Passcode
> 	• Password-Ack
> 	• Password-Expired
> 	• Password-Reject
> 	• Password-Request
> 	• Resource-Free-Request
> 	• Resource-Free-Response
> 	• Resource-Query-Request
> 	• Resource-Query-Response
> 	• Status-Client
> 	• Terminate-Session

  These packet types are historical, and no one uses them for anything.  They can be removed from the library without impacting anyone.

alandekok:
On Oct 5, 2024, at 2:28 PM, Greg Hudson ***@***.***> wrote:
> if header.code is not as expected, discard
> 
> libkrad's current decoding APIs (krad_packet_decode_request() and krad_packet_decode_response()) don't have an input that would let us check the code against a short list.

  There are a very small number of valid packet codes.  See my other message.

  Maybe 10 codes are used for anything.  Perhaps 5 for client to server packets, and another 5 for server to client packets.

> Of course if we skip this check, a spoofer who wants to exercise the decoder can simply use a code that doesn't require a MAC. The packet wouldn't be discarded based on the code until it has already been decoded and the library caller looks at it.

  That seems to be a fairly severe API limitation.  UDP packets can be trivially forged.  It's not a good idea to allocate memory based on spoofed, unauthenticated packets.

abbra:
Both FreeIPA and SSSD are using the following packet codes:
- Access-Challenge
- Access-Request
- Access-Accept

They do not expect or send other packet code types.

API-wise, I think we can coordinate rebuilds downstream, where needed.

greghudson:
There are four behavior changes implemented or under discussion for this PR:

1. Send Message-Authenticator as the first attribute when constructing access-related RADIUS packets, if the secret is non-empty.  As I understand the situation, this is required to restore interoperability with FreeRADIUS.
2. Verify Message-Authenticator in received RADIUS packets when one is present.  As I understand the situation, this carries no complications, and prevents the BlastRADIUS attack if the other endpoint implements (1).
3. Reject access-related RADIUS packets that don't have a Messge-Authenticator, if the secret is non-empty.  This has the potential to cause interoperability problems (I wouldn't consider backporting it to 1.21).  The draft recommends this change be configurable; in the PR is currently always on.
4. Leverage Message-Authenticator to limit the maximum work required to reject spoofed RADIUS packets.  This require some API or scope-related changes, and I am dubious that it is required.

As there appears to be a timeliness requirement, I would suggest paring down this PR to implement just (1) and (2)--basically, removing the few lines of code that implement (3).  Then we can open a new PR to implement (3) and perhaps (4).

src/lib/krad/packet.c
```diff
@@ -170,34 +180,24 @@ auth_generate_random(krb5_context ctx, uchar *rauth)
...
-    retval = krb5_c_make_checksum(ctx, CKSUMTYPE_RSA_MD5, NULL, 0, &data,
-                                  &hash);
-    free(data.data);
-    if (retval != 0)
-        return retval;
-
-    memcpy(rauth, hash.contents, AUTH_FIELD_SIZE);
-    krb5_free_checksum_contents(ctx, &hash);
-    return 0;
+    return krb5_k_make_checksum_iov(ctx, CKSUMTYPE_RSA_MD5, NULL, 0, input, 4);
```
jrisc, line 200: I replaced `krb5_c_make_checksum()` by an iov checksum function to avoid allocating memory. I used `krb5_k_make_checksum_iov()` because `krb5_c_make_checksum_iov()` seems to not accept a NULL keyblock (it causes a segfault).

src/lib/krad/packet.c
```diff
@@ -170,34 +180,24 @@ auth_generate_random(krb5_context ctx, uchar *rauth)
 /* Generate a response authenticator field. */
 static krb5_error_code
 auth_generate_response(krb5_context ctx, const char *secret,
-                       const krad_packet *response, const uchar *auth,
-                       uchar *rauth)
+                       const krb5_data *respbuf, const uint8_t *auth,
+                       uint8_t *rauth)
```
jrisc, line 184: I modified `auth_generate_response()` to accept a raw packet buffer because it is needed to find the matching request of a response, which is then used to verify the Message-Authenticator.

src/lib/krad/packet.c
```diff
@@ -591,51 +628,63 @@ krad_packet_decode_response(krb5_context ctx, const char *secret,
...
+    krb5_data validbuf;
     const krad_packet *req = NULL;
-    krad_packet *rsp = NULL;
+    krb5_boolean matched = FALSE;
     krb5_error_code retval;
 
-    retval = decode_packet(ctx, secret, buffer, &rsp);
-    if (retval != 0)
+    if (!cb)
+        return EINVAL;
```
jrisc, line 638: I removed the possibility to call `krad_packet_decode_response()` with a NULL pending requests iterator, because finding the matching request is needed in order to verify the Message-Authenticator.

src/lib/krad/packet.c
```diff
@@ -479,63 +478,92 @@ krad_packet_new_response(krb5_context ctx, const char *secret, krad_code code,
...
+    }
+
+    if (i != pktlen)
+        return EBADMSG;
+
+    if (validated_len)
+        *validated_len = pktlen;
+
     return 0;
 }
```
jrisc, line 545: We verify than the packet is well-formed with this function.

src/lib/krad/packet.c
jrisc: Actually we should still accept it in case the secret is an empty string.

jrisc:
> 4\. Leverage Message-Authenticator to limit the maximum work required to reject spoofed RADIUS packets.  This change requires either an API or library applicability change to do robustly, and I am dubious that it is required.

I started to work on this in my last commit. I will move it to a different PR.

jrisc:
> 2\. Verify Message-Authenticator in received RADIUS packets when one is present.  As I understand the situation, this change carries no complications, and prevents the BlastRADIUS attack if the other endpoint implements (1).

I think requiring Message-Authenticator to be present is also part of BlastRADIUS mitigation:
https://www.blastradius.fail/#mitigation

I am not sure if MD5 collision could be used to remove the Message-Authenticator attribute from the packet. Anyway, we will also have to implement a setting to restrict use of the `Proxy-State` attribute as explained here:
https://www.blastradius.fail/attack-details

jrisc:
I removed `Protocol-Error` from the list of codes requiring a Message-Authenticator, because it is defined in [RFC7930](https://datatracker.ietf.org/doc/html/rfc7930#section-4) which we do not implement.

jrisc:
I've split the code into 2 commits. If we decide to not enforce the presence of Message-Authenticator in `Access-*` packets for now, we can drop the second commit. The first commit generates Message-Authenticator and checks it if present, but does not raise any error if it is missing.

alandekok:
On Oct 7, 2024, at 4:30 PM, Julien Rische ***@***.***> wrote:
> Actually we should still accept it in case the secret is an empty string.

  I'm not sure why the secret would be an empty string.  This is forbidden by RFC 2865.

      The secret MUST NOT be empty (length 0) since this would allow
      packets to be trivially forged.

jrisc:
@alandekok This is a workaround we use in FreeIPA at this point to exchange packet between the MIT krb5 KDC and another local proxy daemon via UNIX socket. This daemon only has access to the secret used for exchanges with the actual RADIUS server.

We rely on this trick to disable the generation of Message-Authenticator in the present commits for cases where encryption and integrity checks are not necessary (UNIX socket and loopback connections). But some changes are definitely required for handling such cases properly (probably by modifying the API to add parameters).

I will add this point to the krb5#1376 description.

greghudson:
> I am not sure if MD5 collision could be used to remove the Message-Authenticator attribute from the packet.

This is covered in the draft in section 5.2.4, and in https://www.blastradius.fail/pdf/radius.pdf section 7, subsection "Attacks on Partial Enforcement".

An MD5 collision cannot be used to remove the Message-Authenticator MAC from the legitimate server reply, as it goes into the response authenticator hash and is not predictable to the attacker (so cannot be part of the chosen prefixes for the collision).  What an attacker can do is append the MAC bytes and any subsequent bytes from the real reply to the fake reply.  Doing this naively would cause the client to reject the packet if it checks the MAC--but the attacker can hide the MAC bytes within another attribute to prevent them from appearing as a Message-Authenticator and being checked.  If the Message-Authenticator appears last, this workaround is good enough for an attack to succeed.  But if the Message-Authenticator appears first, then the "any subsequent bytes" would include the echoed gibberish in the legitimate reply and the attack cannot succeed.

So I still believe that (1) and (2) are enough to prevent BlastRADIUS--in fact, (1) alone (on the server) seems likely to be enough.  (3) does make the protocol substantially less porous (though still not up to modern standards), so I understand why it is included in mitigation advice.

jrisc:
@greghudson I think I see your point, but the problem is [draft-ietf-radext-deprecating-radius-03](https://datatracker.ietf.org/doc/html/draft-ietf-radext-deprecating-radius-03) does not require the Message-Authenticator to be the first attribute in `Accept-Request` (contrary to its possible response packet types).

I think this is because, as explained in the "Attacks on Partial Enforcement" paragraph from https://www.blastradius.fail/pdf/radius.pdf you pointed to, some clients are still generating Message-Authenticator as non-first attribute. So I suppose the idea was to mitigate BlastRADIUS by updating the server side only, while keeping compatibility with the clients implementing Message-Authenticator as defined in [RFC2869](https://datatracker.ietf.org/doc/html/rfc2869) (where there are no constraints regarding its position in the attribute sequence). So `(1)` is not something we can expect to be true in other implementations (at least for the first attribute part).

So at this point, my understanding is we are discussing whether we should include `(3)` or not in this PR. I suppose this would be possible if we fulfil these two conditions:

* libkrad is only meant to be used on client side
    * Because the fact we generate Message-Authenticator as first attribute protects it from being hidden in another attribute
    * We do not require the response to have a Message-Authenticator, leaving this responsibility to the server
* If Message-Authenticator is present, it must be the first attribute, otherwise we must reject the packet
    * This implies we have to search the all attribute sequence to find a potential non-first attribute Message-Authenticator (which would mean the server does not implement [draft-ietf-radext-deprecating-radius-03](https://datatracker.ietf.org/doc/html/draft-ietf-radext-deprecating-radius-03))
    * So if we find a valid one as first attribute, or if we cannot find it anywhere in the sequence, we accept the response, otherwise we reject it

At the moment, this PR accepts Message-Authenticator no matter where it is located in the attribute sequence. But from your last message I understand this is a risk, and we should enforce it to be the first attribute as much as possible (i.e. for `Access-Accept`, `Access-Reject`, and `Access-Challenge` responses).

But in the end, the problem is libkrad provides the API to implement both the client and the server side, as we are doing it in FreeIPA (in a partial way since we don't use it over UDP). So if we enforce a "Message-Authenticator must be first attribute if it exists" approach, libkrad RADIUS server would be safe, as long as it interacts with libkrad clients. But it won't be compatible with clients that do not generate Message-Authenticator as first attribute, which is very likely to exist, since no RFC nor RFC draft enforces this.

However, time is against us in this case, and at this point this is more about continuity of service than security. I propose we consider this PR as just the implementation of "Message-Authenticator as defined by RFC2869", and keep the rest of the BlastRADIUS mitigations for krb5#1376. We can drop the second commit enforcing the presence of Message-Authenticator. AFAIU, just the first commit would make the vulnerability of libkrad relatively low when used on client side (which is probably the most common case by far), since we generate `Access-Request` with Message-Authenticator as first attribute, and so must be the response it will decode if the server implements BlastRADIUS mitigation measures.

alandekok:
On Oct 10, 2024, at 7:53 AM, Julien Rische ***@***.***> wrote:
> @greghudson I think I see your point, but the problem is draft-ietf-radext-deprecating-radius-03 does not require the Message-Authenticator to be the first attribute in Accept-Request (contrary to its possible response packet types).
> 
> I think this is because, as explained in the "Attacks on Partial Enforcement" paragraph from https://www.blastradius.fail/pdf/radius.pdf you pointed to, some clients are still generating Message-Authenticator as non-first attribute. So I suppose the idea was to mitigate BlastRADIUS by updating the server side only, while keeping compatibility with the clients implementing Message-Authenticator as defined in RFC2869 (where there are no constraints regarding its position in the attribute sequence). So (1) is not something we can expect to be true in other implementations (at least for the first attribute part).

  The requirement for M-A be first in replies is because that _in some cases_, it protects clients who have not upgraded, and who don't even verify the contents of Message-Authenticator.

> So at this point, my understanding is we are discussing whether we should include (3) or not in this PR. I suppose this would be possible if we fulfil these two conditions:

  Please do not create "home-grown" solutions to the issue.  The recommendations in draft-ietf-radext-deprecating-radius are the result of multi-month review by both cryptographers, and by multiple RADIUS vendors with decades of experience.  The recommendations are both cryptographically secure, and also are compatible with existing RADIUS implementations

> 	• libkrad is only meant to be used on client side
> 		• Because the fact we generate Message-Authenticator as first attribute protects it from being hidden in another attribute
> 		• We do not require the response to have a Message-Authenticator, leaving this responsibility to the server

  Then the system may be vulnerable, if the server is not upgraded.  It is perhaps a small risk, but still a real one.

> 	• If Message-Authenticator is present, it must be the first attribute, otherwise we must reject the packet

  Do not do this.  The draft-ietf-radext-deprecating-radius document explicitly forbids clients or servers from checking the location of Message-Authenticator.

> At the moment, this PR accepts Message-Authenticator no matter where it is located in the attribute sequence. But from your last message I understand this is a risk, and we should enforce it to be the first attribute as much as possible (i.e. for Access-Accept, Access-Reject, and Access-Challenge responses).

  The draft forbids this behavior.  Do not do this.

  The IETF draft contains pages and pages of explanation about the vulnerability.  Those arguments can't be countered by a few sentences of text.

  Please don't invent things.  Just read the draft, follow the recommendations, and move on.  This shouldn't even be an issue for discussion.  If people have questions about why the mitigations were chosen, or why they work, or what impact they have, please email me.  I can not only explain, I can update the draft to be clearer.

> But in the end, the problem is libkrad provides the API to implement both the client and the server side, as we are doing it in FreeIPA (in a partial way since we don't use it over UDP). So if we enforce a "Message-Authenticator must be first attribute if it exists" approach, libkrad RADIUS server would be safe,

  Do not do this.  You will break compatibility with existing RADIUS servers.

> as long as it interacts with libkrad clients. But it won't be compatible with clients that do not generate Message-Authenticator as first attribute, which is very likely to exist, since no RFC nor RFC draft enforces this.

  Exactly.

> However, time is against us in this case, and at this point this is more about continuity of service than security. I propose we consider this PR as just the implementation of "Message-Authenticator as defined by RFC2869", and keep the rest of the BlastRADIUS mitigations for #1376. We can drop the second commit enforcing the presence of Message-Authenticator. AFAIU, just the first commit would make the vulnerability of libkrad relatively low when used on client side (which is probably the most common case by far), since we generate Access-Request with Message-Authenticator as first attribute, and so must be the response it will decode if the server implements BlastRADIUS mitigation measures.

  If the client checks the value (not location) of Message-Authenticator, then it is protected _so long as the server has been upgraded_.  If the client doesn't _require_ Message-Authenticator to exist in the packet (independent of location), then it is vulnerable when servers have not been upgraded.

  If it is difficult for libkrad to implement a "require Message-Authenticator" configuration flag, then it is only a small risk to omit the flag.  The vulnerability is difficult to exploit, and essentially all RADIUS servers will be updated.

  I would suggest just adding a note to the documentation saying "we don't require it, for the following reasons".

greghudson:
> At the moment, this PR accepts Message-Authenticator no matter where it is located in the attribute sequence. But from your last message I understand this is a risk, and we should enforce it to be the first attribute as much as possible (i.e. for Access-Accept, Access-Reject, and Access-Challenge responses).

I disagree.  All of the added risk (of not mitigating BlastRADIUS) comes from _sending_ Message-Authenticator in a later position, not from receiving it.  If we see a packet with a Message-Authenticator in a later position, we know that the other endpoint is not following the best practices for mitigating BlastRADIUS attacks, but we also know (after verifying the MAC) that this particular packet is authentic.  As Alan notes, the draft (in section 5.2.5) forbids policing the other endpoint in this matter.

> We can drop the second commit enforcing the presence of Message-Authenticator on decoding.

Yes.  I will re-review and merge the first commit, hopefully later today, and mark it for backport.

greghudson:
Sorry, discovered one last thing: I think EBADE is Linux-specific.  It looks like the PR goes to special care to return EBADE instead of EBADMSG if no matching request is found; I think the old code returns 0 with *reqpkt == NULL.  Is there any additional reasoning behind this part of the PR?

jrisc:
I used `EBADE` to distinguish the case where no matching request can be found from the case where request authenticator or Message-Authenticator verification failed (both errors can only apply if a pending requests iterator function is provided). Since the packet being decoded is an unexpected response in such situation, I thought `Bad exchange` was the closest match since, strictly speaking, the problem is not about the packet's encoding. But if this is not portable, I am not against using `EBADMSG` for both.

The problem about providing a pending requests iterator but not enforcing the request to be found is it skips request authenticator and Message-Authenticator verification. Maybe a more balanced way to handle this case would be to fail only if a pending requests iterator is provided AND the matching request was not found AND the response being decoded has a Message-Authenticator. Or in other words: require the Message-Authenticator to be verified if it is present and if a pending requests iterator is provided.

Combining both these changes would result in this function:

```c
krb5_error_code
krad_packet_decode_response(krb5_context ctx, const char *secret,
                            const krb5_data *buffer, krad_packet_iter_cb cb,
                            void *data, const krad_packet **reqpkt,
                            krad_packet **rsppkt)
{
    uchar auth[AUTH_FIELD_SIZE];
    const krad_packet *req = NULL;
    krad_packet *rsp = NULL;
    krb5_boolean msgauth_present;
    krb5_error_code retval;

    retval = decode_packet(ctx, secret, buffer, &rsp);
    if (retval)
        goto cleanup;

    if (cb != NULL) {
        msgauth_present = has_pkt_msgauth(rsp);

        while ((req = (*cb)(data, FALSE)) != NULL) {
            if (pkt_id_get(rsp) != pkt_id_get(req))
                continue;

            /* Response */
            retval = auth_generate_response(ctx, secret, rsp, pkt_auth(req),
                                            auth);
            if (retval)
                goto cleanup;

            /* If the authenticator matches, then the response is valid. */
            if (k5_bcmp(pkt_auth(rsp), auth, sizeof(auth)) != 0)
                continue;

            /* Verify Message-Authenticator if present. */
            if (msgauth_present) {
                retval = verify_msgauth(secret, rsp, pkt_auth(req));
                if (retval)
                    continue;
            }

            break;
        }

        /* If a Mesaage-Authenticator is present, the matching request must be
         * found to verify it. */
        if (msgauth_present && req == NULL) {
            retval = EBADMSG;
            goto cleanup;
        }
    }

    *reqpkt = req;
    *rsppkt = rsp;

cleanup:
    if (cb != NULL && (retval != 0 || req != NULL))
        (*cb)(data, TRUE);
    if (retval != 0)
        krad_packet_free(rsp);

    return retval;
}
```

greghudson:
> The problem about providing a pending requests iterator but not enforcing the request to be found is it skips request authenticator and Message-Authenticator verification.

I think this problem is in scope for PR #1376 but not for a backportable change.

The prior code seems pretty deliberate in returning 0 with no matching request, even if it finds a request with the same ID but failed response authenticator verification.  The (likely) only consumer of the function, remote.c:on_io_request(), just discards the decoded response if there isn't a matching request.

For 1376 I am happy to consider changes to the krad_packet_decode_response() contract to improve API safety or to perform less work on a packet we have reason to believe is spoofed.  But for the backportable PR I think we want to treat failed Message-Authenticator verification the same way as the existing code treats failed response authenticator validation, and return 0 with no matching request.

alandekok:
On Oct 15, 2024, at 10:40 AM, Greg Hudson ***@***.***> wrote:
> The prior code seems pretty deliberate in returning 0 with no matching request, even if it finds a request with the same ID but failed response authenticator verification. The (likely) only consumer of the function, remote.c:on_io_request(), just discards the decoded response if there isn't a matching request.

  I would agree that "no matching request" is not an error condition.  i.e. the code should just return 0, and not EBADE.

  There is no need to bug the application with a problem if the RADIUS API returned no valid data.  This should be for:

* non RADIUS packet sent to the RADIUS socket
* shared secret is wrong
* reply is received late, after the original request has timed out.

  All of these are really "no data", so far as the application is concerned.  They shouldn't be error conditions.

jrisc:
I tested the latest commit with FreeIPA. It works against FreeRADIUS.

PR #1371, opened 2024-09-06
Fix various issues detected by static analysis

jrisc:
Here are fixes for problems of various severity detected by static analysis.

### DIFF
--- a/src/clients/klist/klist.c
+++ b/src/clients/klist/klist.c
@@ -681,7 +681,7 @@ show_credential(krb5_creds *cred, const char *defname)
     krb5_error_code ret;
     krb5_ticket *tkt = NULL;
     char *name = NULL, *sname = NULL, *tktsname, *flags;
-    int extra_field = 0, ccol = 0, i;
+    int extra_field = 0, ccol = 0, i, r;
     krb5_boolean is_config = krb5_is_config_principal(context, cred->server);
 
     ret = krb5_unparse_name(context, cred->client, &name);
@@ -711,11 +711,11 @@ show_credential(krb5_creds *cred, const char *defname)
         fputs("config: ", stdout);
         ccol = 8;
         for (i = 1; i < cred->server->length; i++) {
-            ccol += printf("%s%.*s%s",
-                           i > 1 ? "(" : "",
-                           (int)cred->server->data[i].length,
-                           cred->server->data[i].data,
-                           i > 1 ? ")" : "");
+            r = printf("%s%.*s%s", i > 1 ? "(" : "",
+                       (int)cred->server->data[i].length,
+                       cred->server->data[i].data, i > 1 ? ")" : "");
+            if (r >= 0)
+                ccol += r;
         }
         fputs(" = ", stdout);
         ccol += 3;
--- a/src/kadmin/dbutil/dump.c
+++ b/src/kadmin/dbutil/dump.c
@@ -695,6 +695,11 @@ process_k5beta7_princ(krb5_context context, const char *fname, FILE *filep,
 
     dbentry->len = u1;
     dbentry->n_key_data = u4;
+
+    if (u5 > UINT16_MAX) {
+        load_err(fname, *linenop, _("invalid principal extra data size"));
+        goto fail;
+    }
     dbentry->e_length = u5;
 
     if (kp != NULL) {
--- a/src/kdc/ndr.c
+++ b/src/kdc/ndr.c
@@ -242,7 +242,7 @@ ndr_enc_delegation_info(struct pac_s4u_delegation_info *in, krb5_data *out)
 {
     krb5_error_code ret;
     size_t i;
-    struct k5buf b;
+    struct k5buf b = EMPTY_K5BUF;
     struct encoded_wchars pt_encoded = { 0 }, *tss_encoded = NULL;
     uint32_t pointer = 0;
 
--- a/src/lib/kdb/decrypt_key.c
+++ b/src/lib/kdb/decrypt_key.c
@@ -60,7 +60,7 @@ krb5_dbe_def_decrypt_key_data(krb5_context context, const krb5_keyblock *mkey,
                               krb5_keyblock *dbkey_out,
                               krb5_keysalt *keysalt_out)
 {
-    krb5_error_code ret;
+    krb5_error_code ret = KRB5_CRYPTO_INTERNAL;
     int16_t keylen;
     krb5_enc_data cipher;
     krb5_data plain = empty_data();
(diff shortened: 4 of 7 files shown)


src/clients/klist/klist.c
```diff
@@ -711,11 +711,12 @@ show_credential(krb5_creds *cred, const char *defname)
         fputs("config: ", stdout);
         ccol = 8;
         for (i = 1; i < cred->server->length; i++) {
-            ccol += printf("%s%.*s%s",
-                           i > 1 ? "(" : "",
-                           (int)cred->server->data[i].length,
-                           cred->server->data[i].data,
-                           i > 1 ? ")" : "");
+            r = printf("%s%.*s%s", i > 1 ? "(" : "",
```
jrisc, line 714: If the `printf()` call fails, the negative return code notifying the error breaks the column counter.

src/kadmin/dbutil/dump.c
```diff
@@ -695,6 +695,11 @@ process_k5beta7_princ(krb5_context context, const char *fname, FILE *filep,
 
     dbentry->len = u1;
     dbentry->n_key_data = u4;
+
+    if (u5 > UINT16_MAX) {
+        load_err(fname, *linenop, _("invalid principal extra data size"));
+        goto fail;
+    }
     dbentry->e_length = u5;
```
jrisc, line 703: `u5` is parsed as a 32-bit unsigned integer, while `dbentry->e_length` is a 16-bit unsigned integer.

src/kdc/ndr.c
```diff
@@ -242,7 +242,7 @@ ndr_enc_delegation_info(struct pac_s4u_delegation_info *in, krb5_data *out)
 {
     krb5_error_code ret;
     size_t i;
-    struct k5buf b;
```
jrisc, line 245: `free(b.data)` could be called in the `cleanup` section before `b` is initialized.

src/lib/kdb/decrypt_key.c
```diff
@@ -60,7 +60,7 @@ krb5_dbe_def_decrypt_key_data(krb5_context context, const krb5_keyblock *mkey,
                               krb5_keyblock *dbkey_out,
                               krb5_keysalt *keysalt_out)
 {
-    krb5_error_code ret;
+    krb5_error_code ret = 0;
```
jrisc, line 63: If `kd->key_data_contents[0] != NULL && kd->key_data_length[0] >= 2` and `keysalt_out != NULL` are evaluated to `false`, `ret` will be returned without ever being set. But I am not sure  if this is possible in practice.

src/lib/rpc/svc_auth_gss.c
```diff
@@ -297,7 +297,7 @@ svcauth_gss_validate(struct svc_req *rqst, struct svc_rpc_gss_data *gd, struct r
 	struct opaque_auth	*oa;
 	gss_buffer_desc		 rpcbuf, checksum;
 	OM_uint32		 maj_stat, min_stat, qop_state;
-	u_char			 rpchdr[128];
```
jrisc, line 300: Up to 432 bytes might be written in `rpchdr`, but the array is only 128 bytes long.

src/lib/rpc/svc_udp.c
```diff
@@ -272,12 +272,12 @@ static bool_t svcudp_reply(
      if (xdr_replymsg(xdrs, msg) &&
 	 (!has_args ||
 	  (SVCAUTH_WRAP(xprt->xp_auth, xdrs, xdr_results, xdr_location)))) {
-	  slen = (int)XDR_GETPOS(xdrs);
+	  slen = XDR_GETPOS(xdrs);
 	  if (sendto(xprt->xp_sock, rpc_buffer(xprt), slen, 0,
 		     (struct sockaddr *)&(xprt->xp_raddr), xprt->xp_addrlen)
 	      == slen) {
 	       stat = TRUE;
-	       if (su->su_cache && slen >= 0) {
```
jrisc, line 280: `slen` cannot be negative because `XDR_GETPOS()` returns `u_int`, and its value is not set elsewhere.

src/lib/rpc/svc_udp.c
```diff
@@ -272,12 +272,12 @@ static bool_t svcudp_reply(
      if (xdr_replymsg(xdrs, msg) &&
 	 (!has_args ||
 	  (SVCAUTH_WRAP(xprt->xp_auth, xdrs, xdr_results, xdr_location)))) {
-	  slen = (int)XDR_GETPOS(xdrs);
+	  slen = XDR_GETPOS(xdrs);
 	  if (sendto(xprt->xp_sock, rpc_buffer(xprt), slen, 0,
 		     (struct sockaddr *)&(xprt->xp_raddr), xprt->xp_addrlen)
 	      == slen) {
```
jrisc, line 278: The range of `size_t` is larger or equal to the `u_int` (`unsigned int`) one. So it is safe to pass the return value of `XDR_GETPOS()` to `sendto()`.

The returned type of `sendto()` is `ssize_t`, but afaik C can handle comparison of integers with different signedness safely.

src/kdc/ndr.c
```diff
@@ -242,7 +242,7 @@ ndr_enc_delegation_info(struct pac_s4u_delegation_info *in, krb5_data *out)
...
+    struct k5buf b = { 0 };
```
jrisc, line 245: `free(b.data)` could be called in the `cleanup` section before `b` is initialized.

src/lib/rpc/svc_auth_gss.c
```diff
@@ -297,7 +297,7 @@ svcauth_gss_validate(struct svc_req *rqst, struct svc_rpc_gss_data *gd, struct r
...
+	u_char			 rpchdr[32 + MAX_AUTH_BYTES];
```
jrisc, line 300: Up to 432 bytes might be written in `rpchdr`, but the array is only 128 bytes long.

src/lib/rpc/svc_udp.c
greghudson: C's rules for comparing signed and unsigned values aren't safe in general, and there's a gcc warning for doing so.  If `ssize_t` is larger than `u_int` (64-bit platforms) then the comparison would be safe, but if they're the same size (32-bit platforms) then the signed value would be converted to unsigned.

jrisc: I thought that `ssize_t` was always of an higher order than `size_t` in term of integer type size, but it seems this is not the case indeed. apparently both are using the native system word size. So since `sendto()` returns a value in the range [`-1`, `INT32_MAX`] on 32-bit architectures, the only case that can lead to a bug is if `XDR_GETPOS()` returns `UINT32_MAX`.

greghudson: It's true that this comparison would work for all expected circumstances, but this code will trigger a `-Wsign-compare` warning on 32-bit platforms, so we'll need to do something else--probably assign the sendto() result to a new variable and check for `r >= 0 && (u_int)r == slen`.

src/kdc/ndr.c
jrisc: Done.

src/lib/rpc/svc_udp.c
jrisc: Done.

src/lib/rpc/svc_auth_gss.c
jrisc: Indeed this check should protect against any out-of-bound write operation on `rpchdr`. However, is there a reason to limit the header size to 128 bytes while the credentials size alone can be up to 400 bytes long?

jrisc: According to [RFC2203 section 5.2.2](https://datatracker.ietf.org/doc/html/rfc2203#section-5.2.2):

> The verifier field in the RPC message header is also described by the
> opaque_auth structure.  All creation requests have the NULL verifier
> (AUTH_NONE flavor with zero length opaque data).
>
> Following the verifier are the call data (procedure specific
> parameters).  Note that the proc field of the call_body structure is
> set to NULLPROC, and thus normally there would be zero octets
> following the verifier.  However, since there is no RPC data exchange
> during a context creation, it is safe to transfer information
> following the verifier.  It is necessary to "overload" the call data
> in this way, rather than pack the GSS-API token into the RPC header,
> because RPC Version 2 restricts the amount of data that can be sent
> in the header.  The opaque body of the credential and verifier fields
> can be each at most 400 octets long, and GSS tokens can be longer
> than 800 octets.

I cannot find what is the restricted size of the header for RPC v2 though...

jrisc: Apparently this part of the code was initially implemented without any size checks in 02c51b4f59b6c86c2112cd10a209322ad1cbb0c6 (2004), and it was fixed in 2007 after CVE-2007-3999 in d9d289e5519303478acf1853a89a3e0fbf170463 and e900740e5a92b06d75d8af45d6904e905365cd57.

src/util/support/threads.c
```diff
@@ -151,7 +150,6 @@ int krb5int_pthread_loaded (void)
         || &pthread_mutex_destroy == 0
         || &pthread_mutex_init == 0
         || &pthread_self == 0
-        || &pthread_equal == 0
```
jrisc, line 154: On some glibc versions, [`pthread_equal()` may be defined as an inline function](https://sourceware.org/git/?p=glibc.git;a=blobdiff;f=nptl/sysdeps/pthread/pthread.h;h=f60ecdee18c48559603baec1e5fceb0534f38a8b;hp=5f34302e0c23a2b40301a35d67d2421064519a24;hb=c26ca5e17ded3275872d9fb48776d155ca7285d5;hpb=1b8373f475105307ee3b64d423ffec995ddd6cde) in the pthread header file. It would result in this test to always be true, no matter if libpthread is loaded or not.

PR #1372, opened 2024-09-21
Remove ancient platforms from shlib.conf

greghudson:
Remove the handling of DEC OSF (last release in 1992), SINIX (1995), AIX 4 (1999), BSD/OS (2003), and IRIX (2006).


abbra:
LGTM.  Given AIX prior to AIX 7 is not supported anymore by IBM, it makes sense to unify AIX5+ detection.

PR #1375, opened 2024-10-04
Avoid using locking krb5_crypto_us_timeofday() in krb5int_trace()

alexey-tikhonov:
Logger doesn't need unique timestamps, so expensive mutex lock within krb5_crypto_us_timeofday() makes logger slower for no reason.

New helper - krb5int_us_timeofday() - is merely a wrapper around existing get_time_now() and doesn't need to be exported. A wrapper looks better than pulling 'struct time_now' to a header.

### DIFF
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -698,6 +698,7 @@ krb5_error_code krb5int_c_copy_keyblock_contents(krb5_context context,
                                                  const krb5_keyblock *from,
                                                  krb5_keyblock *to);
 
+krb5_error_code k5_us_timeofday(krb5_timestamp *, krb5_int32 *);
 krb5_error_code krb5_crypto_us_timeofday(krb5_timestamp *, krb5_int32 *);
 
 /*
--- a/src/lib/krb5/os/c_ustime.c
+++ b/src/lib/krb5/os/c_ustime.c
@@ -73,6 +73,21 @@ get_time_now(struct time_now *n)
 
 #endif
 
+krb5_error_code
+k5_us_timeofday(krb5_timestamp *seconds, krb5_int32 *microseconds)
+{
+    struct time_now now;
+    krb5_error_code err;
+
+    err = get_time_now(&now);
+    if (err)
+        return err;
+
+    *seconds = now.sec;
+    *microseconds = now.usec;
+    return 0;
+}
+
 static struct time_now last_time;
 
 krb5_error_code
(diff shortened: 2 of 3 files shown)


greghudson:
The timestamp change seems uncontroversial aside from a nit: we use `k5_ rather than `krb5int_` as the prefix for tree-internal libkrb5 functions now.

Caching getpid() within a library is trickier.  glibc used to do it, but stopped in 2016 because (as I understand it) the PID can change due a fork(); while glibc can intercept calls to fork(), it can't intercept direct calls to the clone system call.  I am not sure if making it a thread-local variable has the desired result of recognizing a PID change when the process forks.  (I assume that's the hope, since the main effect of making it thread-local is unnecessary; all threads within a process have the same PID.)

When glibc stopped caching getpid(), there was some discussion of implementing it as a vDSO symbol.  But the apparent consensus among everyone who had the authority to make that happen is that getpid() should never be a hot spot.  systemd did have a hot spot involving getpid() (also because the getpid() call was within a library, I believe), and implemented its own caching in a somewhat complicated way using pthread_atfork().

I am a little surprised to see performance optimization work around trace logging.  Is sssd doing something that requires tracing to be on all of the time?  Normally we don't worry so much about the performance of facilities that are only used for troubleshooting.

alexey-tikhonov:
> The timestamp change seems uncontroversial aside from a nit: we use `k5_` rather than `krb5int_` as the prefix for tree-internal libkrb5 functions now.

Fixed.

> Caching getpid() within a library is trickier. glibc used to do it, but stopped in 2016 because (as I understand it) the PID can change due a fork(); while glibc can intercept calls to fork(), it can't intercept direct calls to the clone system call. I am not sure if making it a thread-local variable has the desired result of recognizing a PID change when the process forks. (I assume that's the hope, since the main effect of making it thread-local is unnecessary; all threads within a process have the same PID.)

Hope was that `pthread_once()` will handle fork as well but looks like it's indeed much more complicated:
https://sourceware.org/legacy-ml/libc-alpha/2018-02/msg00537.html

Ok, dropped this patch.

> I am a little surprised to see performance optimization work around trace logging. Is sssd doing something that requires tracing to be on all of the time?

SSSD has a feature (enabled by default) that logs everything into in-memory ring buffer even if printing to log files is disabled. Then this buffer is flushed to log file if a "serious" error happens - kind of "back trace".
But I admit that libkrb5 logs constitute a tiny part of all SSSD logs so that's not too important in this case.

PR #1376, opened 2024-10-07
libkrad: Strengthen packet validation and continue to implement BlastRADIUS mitigation

jrisc:
This PR follows krb5#1370 where support for `Message-Authenticator` is being implemented.

As explained by @alandekok, libkrad should follow this process when decoding packets:

* Verify that it is a well-formed RADIUS packet.
* See if it's a packet which we accept
* Verify Message-Authenticator (and Response Authenticator if relevant)
* Only then is it a "good" packet which can be used to match responses to requests, do decoding, etc.

Verifying that the packet is well-formed should be done this way:
```
if packet length is less than 20, discard
if header.code is not as expected, discard
if header.length > packet length discard
if header.length < packet length, truncate packet length to header.length

for each attribute in packet
    if attribute postion + 2 > end of packet, discard
    if attribute position + attribute.length > end of packet, discard
    if attribute.code == message authenticator && attribute.length != 18, discard
    skip ahead to next attribute
```

When decoding a response, the application should check that the incoming packet has a response type matching a pending request type. Complying with this requirement would require to either remove support for `Access-*` packet types, or add a whitelist parameter to `krad_packet_decode_request()` and `krad_packet_decode_response()`.

To complete BlastRADIUS mitigation, libkrad should support `require Message-Authenticator` and `limit Proxy-State` settings.


greghudson:
The Mitigation section of blastradius.fail does not mention Proxy-State, and the FAQ notes that "having the client discard packets with unexpected Proxy-States does not mitigate the vulnerability" because there are other attributes to hide the garbage in.

PR #1378, opened 2024-10-11
Profile: Enable OSS-Fuzz build macro And  Fix: Pac memory leak & Off-by-one  

pkillarjun:
First: https://github.com/krb5/krb5/commit/3911709d968926659ead530a08aaa2bbce640f01

While it is not an issue for fuzz_profile to handle non-existent paths or directories and return an error, when I was looking at code coverage reports https://github.com/google/oss-fuzz/issues/12282, I think there is a problem with the coverage sanitizer.

Second and Third: https://github.com/krb5/krb5/commit/9cbcbf451a8cbebb0e6a41da9816ca562d176d97 https://github.com/krb5/krb5/commit/7fc776398eca8bc8cc1931f762033c85d60d1993

These are new bugs that form PR https://github.com/krb5/krb5/pull/1366
I am making a separate PR, because I think my last PR is bloated.

Reproduce:
```bash
./tests/fuzzing/fuzz_pac leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip
./tests/fuzzing/fuzz_pac crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip
```

[leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip](https://github.com/user-attachments/files/17337272/leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip)
[crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip](https://github.com/user-attachments/files/17337274/crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip)

#### Foot-Note 
- `crash-ec1315e247fe542b4376085207e7e3c6dfcb14da.zip`
-  `leak-e5067f20c8a4c43c783ed8e754fb4608e12cd6ac.zip`
 Are not zip files, I changed the extension because GitHub doesn't allow me to upload files with no extension.

## HOW-to-merge
1. https://github.com/krb5/krb5/pull/1378
2. https://github.com/krb5/krb5/pull/1379
3. https://github.com/krb5/krb5/pull/1366

### DIFF
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -146,7 +146,7 @@ k5_pac_locate_buffer(krb5_context context, const krb5_pac pac, uint32_t type,
     if (buffer == NULL)
         return ENOENT;
 
-    assert(buffer->offset < pac->data.length);
+    assert(buffer->offset <= pac->data.length);
     assert(buffer->size <= pac->data.length - buffer->offset);
 
     if (data_out != NULL)
--- a/src/lib/krb5/krb/pac.c
+++ b/src/lib/krb5/krb/pac.c
@@ -557,9 +557,11 @@ verify_pac_checksums(krb5_context context, const krb5_pac pac,
         ret = k5_pac_locate_buffer(context, pac, KRB5_PAC_SERVER_CHECKSUM,
                                    &server_checksum);
         if (ret)
-            return ret;
-        if (server_checksum.length < PAC_SIGNATURE_DATA_LENGTH)
-            return KRB5_BAD_MSIZE;
+            goto cleanup;
+        if (server_checksum.length < PAC_SIGNATURE_DATA_LENGTH) {
+            ret = KRB5_BAD_MSIZE;
+            goto cleanup;
+        }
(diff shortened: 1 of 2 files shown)


greghudson:
I'm a little unclear on the value of disabling include and includedir for fuzzing builds.  Wouldn't it yield just as much coverage to remove those statements from the corpora?  It seems unlikely that a fuzzing engine would generate include/includedir statements from nothing.

pkillarjun:
> I'm a little unclear on the value of disabling include and includedir for fuzzing builds. Wouldn't it yield just as much coverage to remove those statements from the corpora? It seems unlikely that a fuzzing engine would generate include/includedir statements from nothing.

This is highly unlikely; also, this way is very unreliable. You never know when someone will make a PR with corpus files that have `include/includedir` statements (Yo bro, new coverage hit, let's add that). To be fair, they don't have the context of this.

Also, more importantly, `oss-fuzz` has the cache of `fuzz_profile` corpora with `include/includedir` statements. Your hacky way does make sense, but it would be useless on the `oss-fuzz` infra.

PR #1380, opened 2024-10-16
Revert "Remove pre-CMS PKINIT compatibility code"

theoparis:
This reverts commit 7280d55ea3a1ffe5aa1bee82075e62177d79e1fe and fixes building with AWS-LC/BoringSSL. Without this reverted commit it fails with `openssl/cms.h` not being found, since it does not exist.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1163,6 +1163,7 @@ fi
 if test "$k5_cv_openssl_version_okay" = yes && (test "$enable_pkinit" = yes || test "$enable_pkinit" = try); then
   K5_GEN_MAKEFILE(plugins/preauth/pkinit)
   PKINIT=yes
+  AC_CHECK_LIB(crypto, CMS_get0_content, [AC_DEFINE([HAVE_OPENSSL_CMS], 1, [Define if OpenSSL supports cms.])])
 elif test "$k5_cv_openssl_version_okay" = no && test "$enable_pkinit" = yes; then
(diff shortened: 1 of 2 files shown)


greghudson:
PKINIT originally used OpenSSL's PKCS7 support, because there wasn't CMS support in OpenSSL at the time.  But PKCS#7 is a subset of CMS, and we observed some interoperability problems as a result [1].  So in release 1.10 we switched to the CMS support from OpenSSL 1.0.  We included those compatibility macros because OpenSSL 0.9.x versions were still in circulation, and it wasn't reasonable to break the PKINIT build on platforms that used OpenSSL 0.9.x as the system libcrypto.  We knew that with the passage of time, platforms would adopt OpenSSL 1.0 and the interoperability problems would be alleviated.  Eventually enough time passed that we no longer supported OpenSSL 0.9.x for other reasons, so I got rid of the compatibility macros a couple of years ago.

This PR proposes to re-enable the macros for the sake of OpenSSL forks that omit the CMS code.  (For some reason this seems to be every major fork.  I don't think the CMS code within OpenSSL is unusually bad, but I guess it isn't widely used; PKINIT may have bet on the wrong standard to reference.)  We would do so knowing that the resulting build isn't a correct implementation of PKINIT, with no assurance that the passage of time will correct the problem--in fact, we would be removing any pressure for those forks to implement CMS support.  So I am tentatively inclined to say no.

[1] https://krbdev.mit.edu/rt/Ticket/Display.html?id=6851

PR #1386, opened 2024-10-29
Fix: two use-of-uninitialized-value bug

pkillarjun:
## 1-Fix https://github.com/krb5/krb5/pull/1386/commits/06fa148feecb6094042dec64b710fbc2dc3053a5

```bash
./tests/fuzzing/fuzz_kdc crash-fuzz_kdc.zip
```
### MSan log
```log
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1221491691
INFO: Loaded 1 modules   (14578 inline 8-bit counters): 14578 [0xaec580, 0xaefe72), 
INFO: Loaded 1 PC tables (14578 PCs): 14578 [0xa65188,0xa9e0a8), 
./tests/fuzzing/fuzz_kdc: Running 1 inputs 1 time(s) each.
Running: crash-fuzz_kdc.zip
==1722250==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x9fde48 in load_64_le /home/magic/krb5.review/krb5.chnages/build/util/support/../.././../src/include/k5-platform.h:718:5
    #1 0x9fe95a in k5_hashtab_create /home/magic/krb5.review/krb5.chnages/build/util/support/../.././../src/util/support/hashtab.c:136:18
    #2 0x505f1a in kdc_init_lookaside /home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/../.././../src/tests/fuzzing/../../kdc/replay.c:139:11
    #3 0x509716 in LLVMFuzzerTestOneInput /home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/../.././../src/tests/fuzzing/fuzz_kdc.c:56:11
    #4 0x4518a4 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_kdc+0x4518a4) (BuildId: 2317b0e6f782d90b77a32ca501d194fe93307392)
    #5 0x43a6fa in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_kdc+0x43a6fa) (BuildId: 2317b0e6f782d90b77a32ca501d194fe93307392)
    #6 0x440040 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_kdc+0x440040) (BuildId: 2317b0e6f782d90b77a32ca501d194fe93307392)
    #7 0x46bdb6 in main (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_kdc+0x46bdb6) (BuildId: 2317b0e6f782d90b77a32ca501d194fe93307392)
    #8 0x7fa00b639087 in __libc_start_call_main (/lib64/libc.so.6+0x2a087) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #9 0x7fa00b63914a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2a14a) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #10 0x434d14 in _start (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_kdc+0x434d14) (BuildId: 2317b0e6f782d90b77a32ca501d194fe93307392)

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/magic/krb5.review/krb5.chnages/build/util/support/../.././../src/include/k5-platform.h:718:5 in load_64_le
Exiting
```

## 2-Fix https://github.com/krb5/krb5/pull/1386/commits/770332ff2d9d5c5324c6aef5923a7311ad87576a

```bash
./tests/fuzzing/fuzz_gss crash-fuzz_gss.zip
```

### MSan log
```log
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1839722398
INFO: Loaded 1 modules   (24216 inline 8-bit counters): 24216 [0xdde2a0, 0xde4138), 
INFO: Loaded 1 PC tables (24216 PCs): 24216 [0xd21db8,0xd80738), 
./tests/fuzzing/fuzz_gss: Running 1 inputs 1 time(s) each.
Running: crash-fuzz_gss.zip
Uninitialized bytes in strchr at offset 0 inside [0x7ffc34512650, 5)
==1728296==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7df026 in krb5_kt_resolve /home/magic/krb5.review/krb5.chnages/build/lib/krb5/keytab/../../.././../src/lib/krb5/keytab/ktbase.c:163:10
    #1 0x7e0efa in krb5_kt_default /home/magic/krb5.review/krb5.chnages/build/lib/krb5/keytab/../../.././../src/lib/krb5/keytab/ktdefault.c:45:12
    #2 0x73d7d2 in acquire_accept_cred /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/acquire_cred.c:190:20
    #3 0x73b2e7 in acquire_cred_context /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/acquire_cred.c:845:15
    #4 0x732548 in acquire_cred /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/acquire_cred.c:948:11
    #5 0x731dc5 in krb5_gss_acquire_cred /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/acquire_cred.c:1102:12
    #6 0x715a66 in kg_accept_krb5 /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/accept_sec_context.c:734:24
    #7 0x7134f2 in krb5_gss_accept_sec_context_ext /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/accept_sec_context.c:1286:12
    #8 0x71fa4e in krb5_gss_accept_sec_context /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/krb5/../../.././../src/lib/gssapi/krb5/accept_sec_context.c:1308:12
    #9 0x508672 in gss_accept_sec_context /home/magic/krb5.review/krb5.chnages/build/lib/gssapi/mechglue/../../.././../src/lib/gssapi/mechglue/g_accept_sec_context.c:249:15
    #10 0x5060e5 in LLVMFuzzerTestOneInput /home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/../.././../src/tests/fuzzing/fuzz_gss.c:62:5
    #11 0x4518d4 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_gss+0x4518d4) (BuildId: 43732c7c14fc2a79b0ca258f30f0284a5e0996b8)
    #12 0x43a72a in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_gss+0x43a72a) (BuildId: 43732c7c14fc2a79b0ca258f30f0284a5e0996b8)
    #13 0x440070 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_gss+0x440070) (BuildId: 43732c7c14fc2a79b0ca258f30f0284a5e0996b8)
    #14 0x46bde6 in main (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_gss+0x46bde6) (BuildId: 43732c7c14fc2a79b0ca258f30f0284a5e0996b8)
    #15 0x7fa912639087 in __libc_start_call_main (/lib64/libc.so.6+0x2a087) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #16 0x7fa91263914a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2a14a) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #17 0x434d44 in _start (/home/magic/krb5.review/krb5.chnages/build/tests/fuzzing/fuzz_gss+0x434d44) (BuildId: 43732c7c14fc2a79b0ca258f30f0284a5e0996b8)

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/magic/krb5.review/krb5.chnages/build/lib/krb5/keytab/../../.././../src/lib/krb5/keytab/ktbase.c:163:10 in krb5_kt_resolve
Exiting
```

[crash-fuzz_kdc.zip](https://github.com/user-attachments/files/17554579/crash-fuzz_kdc.zip)
[crash-fuzz_gss.zip](https://github.com/user-attachments/files/17554681/crash-fuzz_gss.zip)

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -466,7 +466,7 @@ AC_CONFIG_HEADERS(include/autoconf.h, [echo timestamp > include/autoconf.stamp])
 AC_C_CONST
 AC_HEADER_DIRENT
 AC_FUNC_STRERROR_R
-AC_CHECK_FUNCS(strdup setvbuf seteuid setresuid setreuid setegid setresgid setregid setsid flock fchmod chmod strptime geteuid setenv unsetenv getenv gmtime_r localtime_r bswap16 bswap64 mkstemp getusershell access getcwd srand48 srand srandom stat strchr strerror timegm explicit_bzero explicit_memset getresuid getresgid)
+AC_CHECK_FUNCS(strdup setvbuf seteuid setresuid setreuid setegid setresgid setregid setsid flock fchmod chmod strptime geteuid setenv unsetenv getenv gmtime_r localtime_r bswap16 bswap64 mkstemp getusershell access getcwd srand48 srand srandom stat strchr strerror timegm explicit_bzero explicit_memset getresuid getresgid getentropy)
 
 AC_CHECK_FUNC(mkstemp,
 [MKSTEMP_ST_OBJ=
--- a/src/kdc/replay.c
+++ b/src/kdc/replay.c
@@ -130,7 +130,7 @@ krb5_error_code
 kdc_init_lookaside(krb5_context context)
 {
     krb5_error_code ret;
-    uint8_t seed[K5_HASH_SEED_LEN];
+    uint8_t seed[K5_HASH_SEED_LEN] = {0};
     krb5_data d = make_data(seed, sizeof(seed));
 
     ret = krb5_c_random_make_octets(context, &d);
--- a/src/lib/crypto/krb/prng.c
+++ b/src/lib/crypto/krb/prng.c
@@ -96,7 +96,27 @@ cleanup:
 static krb5_boolean
 get_os_entropy(unsigned char *buf, size_t len)
 {
-#if defined(__linux__) && defined(SYS_getrandom)
+#if defined(HAVE_GETENTROPY)
+    int r;
+    size_t seg;
+
+    /* getentropy() has a maximum length of 256. */
+    while (len > 0) {
+        seg = (len > 256) ? 256 : len;
+        r = getentropy(buf, len);
+        if (r != 0)
+            break;
+        len -= seg;
+        buf += seg;
+    }
+    if (len == 0)
+        return TRUE;
+#elif defined(__linux__) && defined(SYS_getrandom)
+    /*
+     * Linux added SYS_getrandom in 3.17 (2014), but glibc did not have a
+     * wrapper until 2.25 (2017).  This block can be deleted when that interval
+     * is far in the past.
+     */
     int r;
 
     while (len > 0) {
@@ -104,9 +124,6 @@ get_os_entropy(unsigned char *buf, size_t len)
          * Pull from the /dev/urandom pool, but require it to have been seeded.
          * This ensures strong randomness while only blocking during first
          * system boot.
-         *
-         * glibc does not currently provide a binding for getrandom:
-         * https://sourceware.org/bugzilla/show_bug.cgi?id=17252
          */
         errno = 0;
         r = syscall(SYS_getrandom, buf, len, 0);
--- a/src/lib/krb5/keytab/ktdefault.c
+++ b/src/lib/krb5/keytab/ktdefault.c
@@ -37,7 +37,7 @@
 krb5_error_code KRB5_CALLCONV
 krb5_kt_default(krb5_context context, krb5_keytab *id)
 {
-    char defname[BUFSIZ];
+    char defname[BUFSIZ] = {0};
     krb5_error_code retval;
 
     if ((retval = krb5_kt_default_name(context, defname, sizeof(defname))))


greghudson:
I can't reproduce either of these using the given inputs or find a bug in either case.  In kdc_init_lookaside(), a return of 0 from krb5_c_random_make_octets() implies that seed is initialized, and in krb5_kt_default(), a return of 0 from krb5_kt_default_name() implies that defname contains a C string.  In neither case do I see a code path through the called function that would return 0 without writing to the output buffer.

pkillarjun:
> I can't reproduce either of these using the given inputs or find a bug in either case. In kdc_init_lookaside(), a return of 0 from krb5_c_random_make_octets() implies that seed is initialized, and in krb5_kt_default(), a return of 0 from krb5_kt_default_name() implies that defname contains a C string. In neither case do I see a code path through the called function that would return 0 without writing to the output buffer.

Ah, my bad. Build with memory sanitizer.
```bash

export CC=clang
export CXX=clang++
export CFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=memory -fno-omit-frame-pointer -g -fsanitize=fuzzer-no-link"
export CXXFLAGS="-g -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=memory  -fno-omit-frame-pointer -g -fsanitize=fuzzer-no-link"
export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"

mkdir -p build
cd build

./../src/configure CFLAGS="-fcommon $CFLAGS" CXXFLAGS="-fcommon $CXXFLAGS" --enable-static --disable-shared --enable-ossfuzz
make -j$(nproc)
````

**Comment by greghudson** [comment]
I rebuilt with those instructions.  I still can't reproduce a failure with fuzz_gss.  With fuzz_kdc, I see a different failure later in the execution:
```
==2253100==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x5635f53837a1 in k5_hashtab_add /me/krb5/fmbuild/util/support/../.././../src/util/support/hashtab.c:216:5
    #1 0x5635f4e6ba2e in insert_entry /me/krb5/fmbuild/tests/fuzzing/../.././../src/tests/fuzzing/../../kdc/replay.c:94:11
    #2 0x5635f4e6c9aa in LLVMFuzzerTestOneInput /me/krb5/fmbuild/tests/fuzzing/../.././../src/tests/fuzzing/fuzz_kdc.c:63:9
    [...]
```
but I can't find a corresponding bug for that either.  hashtab.c:216 is `K5_SLIST_INSERT_HEAD(&ht->buckets[i], ent, next);` which expands to `ent->next.sle_next = ht->buckets[i].slh_first; ht->buckets[i].slh_first = ent;`.  ht->buckets was initialized with `ht->buckets = calloc(ht->nbuckets, sizeof(*ht->buckets));` and appears properly zero-filled.  Adding a loop to manually K5_SLIST_INIT() each bucket does not change the result, so I don't think we are somehow being dinged for assuming that zero-filled memory contains null pointers.  valgrind does not report anything amiss in the execution (using a separate build with fuzzing enabled but without asan/msan).

**Comment by greghudson** [comment]
fuzz_kdc: I realized that both failures probably result from msan not thinking that seed is initialized.  In my version of msan the failure isn't reported until `i` (the result of `siphash24(key, klen, ht->k0, ht->k1)` where `ht->k0` and `ht->k1` are derived from seed) is used as an array index, while in Arjun's version the failure is reported as soon as we load from seed.  I also realized that the likely reason msan doesn't think seed is initialized because in prng.c we do `syscall(SYS_getrandom, buf, len, 0)`, and I would guess that msan (reasonably) doesn't know what all of the syscall()s do.  glibc added a binding for getrandom() in 2.25; if we detect and use that in favor of syscall() this problem should go away.  (Or we could add an initializer for seed[] as this PR suggests.  But I don't like that fix; we don't want valgrind or msan to think that everything is okay if the seed isn't set by krb5_c_random_make_octets().)

fuzz_gss: The difference between my and Arjun's experiences is probably related to the process environment, since the default keytab name does not depend on the gss_accept_sec_context() input token.  I still haven't found a path through  krb5_kt_default() that could return 0 without setting output.  The most complicated path involves k5_expand_path_tokens(), which checks out as far as I can tell, but I will keep looking.

**Comment by pkillarjun** [comment]
```bash
$ clang -v
clang version 18.1.8 (Fedora 18.1.8-1.fc40)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/x86_64-redhat-linux-gnu-clang.cfg
System configuration file directory: /etc/clang/
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/14
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/14
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64

$ git clone -b fuzzing-harness https://github.com/pkillarjun/krb5.git
```


### Ok, let's debug `fuzz_kdc`

File `src/lib/krb5/keytab/ktdefault.c`
```c
#ifndef LEAN_CLIENT
krb5_error_code KRB5_CALLCONV
krb5_kt_default(krb5_context context, krb5_keytab *id)
{
    char defname[BUFSIZ];
    krb5_error_code retval;

    if ((retval = krb5_kt_default_name(context, defname, sizeof(defname)))) //$(gdb) p defname
        return retval;
    return krb5_kt_resolve(context, defname, id); //$(gdb) p defname
}
```

#### gdb session

```gdb
$(gdb) # Do you see the junk? This is what MSAN will report on
$(gdb) p defname
$6 = "/\000\000\000\000\000\000\000\340\000\000\000 p\000\000@\002\000\000 p\000\000\000\000\000\000\000\000\000\001", '\000' <repeats 24 times>, "\250\331\335", '\000' <repeats 13 times>, "\001\000\000\000\000\000\000\001\254\331\335", '\000' <repeats 29 times>, "\320\345\260\000\000\000\000\000\240\331\335", '\000' <repeats 13 times>, "\260\331\335\000\000\000\000\000\340\000\000\000 p\000\000\b\252\377\377\377\177\000\000\000\000\000\000\000\000\000\000\300\247\377\377\377\177", '\000' <repeats 2522 times>...

$(gdb) n

$(gdb) # Now we have the valid data
$(gdb) # But with the junk (this junk is marked "region" by MSAN, meaning if this is ever read, MSAN will crash)
$(gdb) p defname
$7 = "FILE:/etc/krb5.keytab\000\000\000\000\000\000\000\000\000\000\001", '\000' <repeats 24 times>, "\250\331\335", '\000' <repeats 13 times>, "\001\000\000\000\000\000\000\001\254\331\335", '\000' <repeats 29 times>, "\320\345\260\000\000\000\000\000\240\331\335", '\000' <repeats 13 times>, "\260\331\335\000\000\000\000\000\340\000\000\000 p\000\000\b\252\377\377\377\177\000\000\000\000\000\000\000\000\000\000\300\247\377\377\377\177", '\000' <repeats 2522 times>...
```

```c
#include <ctype.h>
krb5_error_code KRB5_CALLCONV
krb5_kt_resolve (krb5_context context, const char *name, krb5_keytab *ktid)
{
    /* ... */

    *ktid = NULL;

    cp = strchr (name, ':'); // This is where MSAN crashes
    if (!cp)
        return (*krb5_kt_dfl_ops.resolve)(context, name, ktid);

    pfxlen = cp - name;
```

While this shouldn't be a problem, But it is for MSAN because of the `strchr`.

```bash
$ man strchr

strchr - locate character in string
```

### Final thoughts

Well, this is not a bug, not even a problem, because this is in controlled code (initialization code).

But I need this to be fixed, because this is going to be a roadblock for OSS-Fuzz infra, meaning the `fuzz_kdc` won't be fuzzed under MSAN until this "bug" is fixed.

pkillarjun:
```bash
$ clang -v
clang version 18.1.8 (Fedora 18.1.8-1.fc40)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/x86_64-redhat-linux-gnu-clang.cfg
System configuration file directory: /etc/clang/
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/14
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/14
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64

$ git clone -b fuzzing-harness https://github.com/pkillarjun/krb5.git
```


### Ok, let's debug `fuzz_gss`

File `src/lib/krb5/keytab/ktdefault.c`
```c
#ifndef LEAN_CLIENT
krb5_error_code KRB5_CALLCONV
krb5_kt_default(krb5_context context, krb5_keytab *id)
{
    char defname[BUFSIZ];
    krb5_error_code retval;

    if ((retval = krb5_kt_default_name(context, defname, sizeof(defname)))) //$(gdb) p defname
        return retval;
    return krb5_kt_resolve(context, defname, id); //$(gdb) p defname
}
```

#### gdb session

```gdb
$(gdb) # Do you see the junk? This is what MSAN will report on
$(gdb) p defname
$6 = "/\000\000\000\000\000\000\000\340\000\000\000 p\000\000@\002\000\000 p\000\000\000\000\000\000\000\000\000\001", '\000' <repeats 24 times>, "\250\331\335", '\000' <repeats 13 times>, "\001\000\000\000\000\000\000\001\254\331\335", '\000' <repeats 29 times>, "\320\345\260\000\000\000\000\000\240\331\335", '\000' <repeats 13 times>, "\260\331\335\000\000\000\000\000\340\000\000\000 p\000\000\b\252\377\377\377\177\000\000\000\000\000\000\000\000\000\000\300\247\377\377\377\177", '\000' <repeats 2522 times>...

$(gdb) n

$(gdb) # Now we have the valid data
$(gdb) # But with the junk (this junk is marked "region" by MSAN, meaning if this is ever read, MSAN will crash)
$(gdb) p defname
$7 = "FILE:/etc/krb5.keytab\000\000\000\000\000\000\000\000\000\000\001", '\000' <repeats 24 times>, "\250\331\335", '\000' <repeats 13 times>, "\001\000\000\000\000\000\000\001\254\331\335", '\000' <repeats 29 times>, "\320\345\260\000\000\000\000\000\240\331\335", '\000' <repeats 13 times>, "\260\331\335\000\000\000\000\000\340\000\000\000 p\000\000\b\252\377\377\377\177\000\000\000\000\000\000\000\000\000\000\300\247\377\377\377\177", '\000' <repeats 2522 times>...
```

```c
#include <ctype.h>
krb5_error_code KRB5_CALLCONV
krb5_kt_resolve (krb5_context context, const char *name, krb5_keytab *ktid)
{
    /* ... */

    *ktid = NULL;

    cp = strchr (name, ':'); // This is where MSAN crashes
    if (!cp)
        return (*krb5_kt_dfl_ops.resolve)(context, name, ktid);

    pfxlen = cp - name;
```

While this shouldn't be a problem, But it is for MSAN because of the `strchr`.

```bash
$ man strchr

strchr - locate character in string
```

### Final thoughts

Well, this is not a bug, not even a problem, because this is in controlled code (initialization code).

But I need this to be fixed, because this is going to be a roadblock for OSS-Fuzz infra, meaning the `fuzz_gss` won't be fuzzed under MSAN until this "bug" is fixed.

Note: in my last messages I misplaced the names; it's fuzz_gss, not fuzz_kdc.

greghudson:
Let me know if 482b8663e31be7518048bb1a03181add47e139fb fixes the fuzz_kdc problem without 4a9378abb718866450e5b9bc9399370a6dc2d1dc (which I left it in the PR branch).

I will look into reproducing the fuzz_gss problem.  It seems strange that a successful strlcpy() into a buffer followed by strchr() produces a failure.

pkillarjun:
> I will look into reproducing the fuzz_gss problem. It seems strange that a successful strlcpy() into a buffer followed by strchr() produces a failure.

```c
#ifndef LEAN_CLIENT
krb5_error_code KRB5_CALLCONV
krb5_kt_default(krb5_context context, krb5_keytab *id)
{
    char defname[BUFSIZ];
    krb5_error_code retval;

    if ((retval = krb5_kt_default_name(context, defname, sizeof(defname))))
        return retval;

    // The program won't crash here
    strncpy(defname, "AAAA", 4);
    for (int i = 0; i < 4; i++) {
        printf("%02x ", defname[i]);
    }
    printf("\n");

    // The program won't crash here
    memset(defname, 0x00, 8);
    for (int i = 0; i < 8; i++) {
        printf("%02x ", defname[i]);
    }
    printf("\n");

    // The program shouldn't crash, But it dose even on 10;
    strlcpy(defname, "AAAAAAAAAAAA", 12);
    for (int i = 0; i < 10; i++) {
        printf("%02x ", defname[i]);
    }
    printf("\n");

    return krb5_kt_resolve(context, defname, id);
}
```

greghudson:
strlcpy() was added to glibc in 2.38.  My dev system (which doesn't reproduce the failure) has glibc 2.35, so my builds are using the fallback implementation in util/support/strlcpy.c.  Your system probably has a newer glibc, so your builds are probably using the libc version.  Since the system libc isn't built with memory sanitization, I assume MemorySanitizer needs some per-function magic.  I did a quick search and found https://github.com/llvm/llvm-project/blob/6da5968f5ecc2a2e8b0697e335f4dec1b3bbfd01/compiler-rt/lib/msan/msan_interceptors.cpp which has interceptors for strncpy() and memset() but not strlcpy().  I found no search results mentioning strlcpy and MemorySanitizer, so we may be the first people to notice this gap.

So, I think we adequately understand the problem, but still have to work around it sooner than we could reasonably expect clang to fix it and the fixed version to make its way into the OSS-Fuzz infrastructure.  Off the top of my head:

* We could do what the PR currently does and add initializers to buffers that we strlcpy() into, whenever we run into this problem with a fuzzing target.  Adding unnecessary initializers carries the risk of failing to detect bugs (using the initial value when we don't expect to), so I'd like to think about alternatives.
* We could find a way to make the fuzzer build use the fallback strlcpy() implementation even if the system libc has one.  This might be a little fiddly, but there's an obvious place for this (the existing configure.ac code for `--enable-ossfuzz`) and we'd just need to reset the results of the earlier strlcpy() test with a comment.
* We could look for an improvement that happens to avoid using strlcpy().  In this case, we could save a copy in krb5_kt_default() by making kt_default_name() visible as k5_kt_default_name() (similar to k5_kt_client_default_name()).

pkillarjun:
> I did a quick search and found https://github.com/llvm/llvm-project/blob/6da5968f5ecc2a2e8b0697e335f4dec1b3bbfd01/compiler-rt/lib/msan/msan_interceptors.cpp which has interceptors for strncpy() and memset() but not strlcpy(). I found no search results mentioning strlcpy and MemorySanitizer, so we may be the first people to notice this gap.

I guess so, I did open a issue  https://github.com/google/sanitizers/issues/1810

> So, I think we adequately understand the problem, but still have to work around it sooner than we could reasonably expect clang to fix it and the fixed version to make its way into the OSS-Fuzz infrastructure.

```bash
root@65eb29ba8830:/# clang -v
clang version 18.1.8 (https://github.com/llvm/llvm-project.git 3b5b5c1ec4a3095ab096dd780e84d7ab81f3d7ff)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Candidate multilib: x32;@mx32
Selected multilib: .;@m64
```
They are with me; it would definitely take some time.

> We could do what the PR currently does and add initializers to buffers that we strlcpy() into, whenever we run into this problem with a fuzzing target. Adding unnecessary initializers carries the risk of failing to detect bugs (using the initial value when we don't expect to), so I'd like to think about alternatives.

> We could find a way to make the fuzzer build use the fallback strlcpy() implementation even if the system libc has one. This might be a little fiddly, but there's an obvious place for this (the existing configure.ac code for --enable-ossfuzz) and we'd just need to reset the results of the earlier strlcpy() test with a comment.

> We could look for an improvement that happens to avoid using strlcpy(). In this case, we could save a copy in krb5_kt_default() by making kt_default_name() visible as k5_kt_default_name() (similar to k5_kt_client_default_name()).


I don't want this PR to be merged until I see a bug after the MSan fix. Everything else is too much voodoo and unnecessary changes.

>> But I need this to be fixed, because this is going to be a roadblock for OSS-Fuzz infra, meaning the fuzz_gss won't be fuzzed under MSAN until this "bug" is fixed.

If this was a krb5 bug, I want this to be fixed; but now it's an MSan problem.

pkillarjun:
> I don't see interceptors for getrandom() or getentropy() so my proposed fix for fuzz_kdc may also not work, unfortunately.

Yeah, you are right, it won't work because it's a totally different bug.

> `uint8_t seed[K5_HASH_SEED_LEN];`

I have a second thought on this one. Wouldn't it be good to have random junk as seed for `k5_hashtab_create()`?

If so, then this patch might work, but it is very ugly.
```patch
diff --git a/src/kdc/replay.c b/src/kdc/replay.c
index 5125bc63c..fe4eb7ffe 100644
--- a/src/kdc/replay.c
+++ b/src/kdc/replay.c
@@ -126,6 +126,9 @@ discard_entry(krb5_context context, struct entry *entry)
 }
 
 /* Initialize the lookaside cache structures and randomize the hash seed. */
+#if __has_feature(memory_sanitizer)
+__attribute__((no_sanitize("memory")))
+#endif
 krb5_error_code
 kdc_init_lookaside(krb5_context context)
 {
```

pkillarjun:
> > I don't see interceptors for getrandom() or getentropy() so my proposed fix for fuzz_kdc may also not work, unfortunately.
> 
> Yeah, you are right, it won't work because it's a totally different bug.
> 
> > `uint8_t seed[K5_HASH_SEED_LEN];`
> 
> I have a second thought on this one. Wouldn't it be good to have random junk as seed for `k5_hashtab_create()`?
> 
> If so, then this patch might work, but it is very ugly.
> 
> ```diff
> diff --git a/src/kdc/replay.c b/src/kdc/replay.c
> index 5125bc63c..fe4eb7ffe 100644
> --- a/src/kdc/replay.c
> +++ b/src/kdc/replay.c
> @@ -126,6 +126,9 @@ discard_entry(krb5_context context, struct entry *entry)
>  }
>  
>  /* Initialize the lookaside cache structures and randomize the hash seed. */
> +#if __has_feature(memory_sanitizer)
> +__attribute__((no_sanitize("memory")))
> +#endif
>  krb5_error_code
>  kdc_init_lookaside(krb5_context context)
>  {
> ```

Ah, shit. My bad—I didn't read the comment on `kdc_init_lookaside`.

pkillarjun:
> I found no search results mentioning strlcpy and MemorySanitizer, so we may be the first people to notice this gap.

Quite funny, though: [glibc: Implement strlcpy and strlcat](https://sourceware.org/git/?p=glibc.git;a=commit;h=454a20c8756c9c1d55419153255fc7692b3d2199) is not the only one that is not supported; memory changes after `syscall` are also not supported.

> memory changes after `syscall` are also not supported.

This looks impossible to fix, but I think they can do it.


```c
/* src/kdc/replay.c */
static krb5_boolean
_get_os_entropy(unsigned char *buf, size_t len)
{
    int r;
    while (len > 0) {
        /*
         * Pull from the /dev/urandom pool, but require it to have been seeded.
         * This ensures strong randomness while only blocking during first
         * system boot.
         *
         * glibc does not currently provide a binding for getrandom:
         * https://sourceware.org/bugzilla/show_bug.cgi?id=17252
         */
        errno = 0;
        r = syscall(318, buf, len, 0);
        if (r <= 0) {
            if (errno == EINTR)
                continue;

            /* ENOSYS or other unrecoverable failure */
            break;
        }
        len -= r;
        buf += r;
    }
    if (len == 0)
        return TRUE;

    return FALSE;
}

/* Initialize the lookaside cache structures and randomize the hash seed. */
krb5_error_code
kdc_init_lookaside(krb5_context context)
{
    krb5_error_code ret;
    uint8_t seed[K5_HASH_SEED_LEN];
    krb5_data d = make_data(seed, sizeof(seed));

    _get_os_entropy(seed, sizeof(seed));
    for (int i = 0; i < 16; i++) {
        printf("%02x ", seed[i]);
    }
    printf("\n");

    ret = krb5_c_random_make_octets(context, &d);
    if (ret)
        return ret;
    ret = k5_hashtab_create(seed, 8192, &hash_table);
    if (ret)
        return ret;
    K5_TAILQ_INIT(&expiration_queue);
    return 0;
}
```

#### gdb

```c
Thread 1 "fuzz_kdc" hit Breakpoint 2, kdc_init_lookaside (context=0x714000000000) at ../.././../src/tests/fuzzing/../../kdc/replay.c:165
165         krb5_data d = make_data(seed, sizeof(seed));
(gdb) p seed
$3 = "K\000\000\000\000\000\000\000\230\320\377\377\377\177\000"
(gdb) n
167         _get_os_entropy(seed, sizeof(seed));
(gdb) n
168         for (int i = 0; i < 16; i++) {
(gdb) p seed
$4 = "\276\344\273\343_\351\361\033\271\033\343\375\375\224", <incomplete sequence \331>
(gdb) 
```

pkillarjun:
I will close this PR for now; I think msan needs some changes.

PR #1390, opened 2024-11-13
Remove missing SecurID header file

kenh:
It turns out the SecurID SDK no longer provides sdi_athd.h, so remove our include of it.

### DIFF
--- a/src/plugins/preauth/securid_sam2/securid2.c
+++ b/src/plugins/preauth/securid_sam2/securid2.c
@@ -45,7 +45,6 @@
 #include <syslog.h>
 #include <acexport.h>
 #include <sdi_defs.h>
-#include <sdi_athd.h>
 #include "extern.h"
 
 #define KRB5_SAM_SECURID_NEXT_CHALLENGE_MAGIC 0x5ec1d000


greghudson:
Just to confirm: this change is sufficient to make the code work with the current version of the SDK?

kenh:
I confirmed with the person that uses this code (our site doesn't use it, but I am involved with others that do) that yes, it works fine with this change; I know they use it in an operational capacity.

kenh:
Ahhh, hold on one second. I was actually just told there is ONE additional change. Will amend my pull request.

kenh:
Alright, so ... some more information.

The other change used by this site is a change that strips of the instance of a principal before it is sent to the ACE server (this is right after the code used to strip off the realm name). This is so if you have "user/admin" you can use the same SecurID token as your regular account.

This obviously isn't required to make the code compile or work. My first thought is that change belongs in separate pull request, and the behavior should be configurable. Thoughts?

greghudson:
Yeah, if it's not related to the SDK version then it's a separate matter.  I will go ahead and merge this soon.

kenh:
Thanks! Some further communication with the site that uses this code made us realize that we thought that maybe this was configurable as an alias on the ACE server, but that turned out not to work. I will see if we can cook up something more suitable for a pull request later.

PR #1391, opened 2024-11-22
Add a SBOM template in CycloneDX format

hughsie:
Hi,

My name is Richard Hughes and I'm a developer at Red Hat. I'm the maintainer of fwupd and LVFS, and am trying to improve software supply chain security by encouraging OEMs, ODMs and IBVs to ship Software Bill of Materials with each firmware binary blob (SBOMs).

I'm working alongside lots of other companies proactively trying to do the right thing. The reason I've opened this pull request is because krb5 is either used in the *build process* of a firmware we care about (e.g. EDK II, or coreboot) or is built *into the firmware binary* itself. Although my personal focus is on firmware, the SBOM file is in CycloneDX format (one of the most popular industry standards) which makes it also useful when building containers or OS images too.

I would like to contribute this template SBOM file into your project that gets included into source control with substituted values that get populated automatically. I'm not super familiar with the history of Kerberos, and so I've done my best populating the project values -- but please point out any that are incorrect (especially being careful with the license list) and I'll fix them up. I've also put the `sbom.cdx.json` file in what I feel is the right place, but please say if you want me to put it somewhere different or name it a different thing; the directory and `sbom` prefix are unimportant. I also wasn’t 100% sure whether to mark the component as a *library* or *application*, so advice is welcome.

The various firmware build tools will take these incomplete SBOM files and then build them into a complete composite SBOM to represent the firmware. Having an upstream reference to what the PURL and CPE values should be means we have something we can trust; I could quite easily spin up a web-service that we say "what CPE do we use for X" -> "cpe:2.3:a:Y:Z::::::::` but we don't actually know if that's still true, up to date, or what the maintainer actually wants them to be. Putting the template upstream means we can trust the values we find in the checked out code during the build process.

Also, if you’re not happy with MIT being being labelled the *supplier* (which seems more appropriate from a SBOM point of view, but makes some open source projects uncomfortable) we can remove that bit.

I've written a bit more about this proposal here https://blogs.gnome.org/hughsie/2024/11/14/firmware-sboms-for-open-source-projects/ and there's also lot more information about firmware SBOMs here: https://lvfs.readthedocs.io/en/latest/sbom.html – many thanks for your time and all the work that you do.

### DIFF
--- a/doc/sbom.cdx.json
+++ b/doc/sbom.cdx.json
@@ -1,0 +1,127 @@
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.6",
+  "version": 1,
+  "metadata": {
+    "authors": [
+      {
+        "name": "@VCS_SBOM_AUTHORS@"
+      }
+    ]
+  },
+  "components": [
+    {
+      "type": "library",
+      "bom-ref": "pkg:github/krb5/krb5@@VCS_TAG@",
+      "cpe": "cpe:2.3:a:mit:kerberos_5:@VCS_TAG@:*:*:*:*:*:*:*",
+      "name": "krb5",
+      "version": "@VCS_VERSION@",
+      "description": "Reference implementation of Kerberos, a network authentication protocol",
+      "supplier": {
+        "name": "Massachusetts Institute of Technology"
+      },
+      "authors": [
+        {
+          "name": "@VCS_AUTHORS@"
+        }
+      ],
+      "licenses": [
+        {
+          "license": {
+            "id": "BSD-2-Clause"
+          }
+        },
+        {
+          "license": {
+            "id": "BSD-3-Clause"
+          }
+        },
+        {
+          "license": {
+            "id": "BSD-4-Clause"
+          }
+        },
+        {
+          "license": {
+            "id": "Brian-Gladman-2-Clause"
+          }
+        },
+        {
+          "license": {
+            "id": "FSFULLRWD"
+          }
+        },
+        {
+          "license": {
+            "id": "HPND"
+          }
+        },
+        {
+          "license": {
+            "id": "HPND-export-US"
+          }
+        },
+        {
+          "license": {
+            "id": "HPND-export2-US"
+          }
+        },
+        {
+          "license": {
+            "id": "HPND-export-US-acknowledgement"
+          }
+        },
+        {
+          "license": {
+            "id": "HPND-export-US-modify"
+          }
+        },
+        {
+          "license": {
+            "id": "ISC"
+          }
+        },
+        {
+          "license": {
+            "id": "CMU-Mach-nodoc"
+          }
+        },
+        {
+          "license": {
+            "id": "GPL-2.0-or-later"
+          }
+        },
+        {
+          "license": {
+            "id": "MIT"
+          }
+        },
+        {
+          "license": {
+            "id": "MIT-CMU"
+          }
+        },
+        {
+          "license": {
+            "id": "OLDAP-2.8"
+          }
+        },
+        {
+          "license": {
+            "id": "OpenVision"
+          }
+        }
+      ],
+      "externalReferences": [
+        {
+          "type": "website",
+          "url": "https://kerberos.org/"
+        },
+        {
+          "type": "vcs",
+          "url": "https://github.com/krb5/krb5"
+        }
+      ]
+    }
+  ]
+}


abbra:
@hughsie thank you for the submission. The file looks OK to me, I just have one question that is not really covered in the proposal: in a situation when a firmware was built using a distribution-provided packages or release tarballs, how VCS_* symbols will be populated? Shouldn't then release scripts be updated to generate this information at the release time?

greghudson:
Apologies for the long delay.

My general reaction is that this seems burdensome to maintain.  Unlike many open source projects, we have numerous different licenses--all permissive open source, but with plenty of differences in detail.  We already collect these licenses into doc/notice.rst, and there is some risk of even that one file becoming out of sync with the source code.  Moreover, CycloneDX license descriptions are references to the SPDX license list, and while that list is quite long, I'm not sure it maps 1:1 with all of our licenses, or will do so in the future.  For instance, there is an ancient FundsXpress INC license on much of the lib/krb5/crypto code.  It's very similar to the MIT license of a certain era except for the asserted holder of copyright, but the body of the license contains "the name of FundsXpress. not be used in advertising or publicity" which is of course not going to be in the text of any other license.  I don't really want to risk making inaccurate legal statements about the licenses of copyright holders other than MIT.

I have some specific reactions, which are only relevant if I can be argued into accepting the general idea:

* The source tree contains a number of libraries and applications.  Does CycloneDX or SBOM anticipate including a component for each one?  (If so, the license lists explode in length.)  Simply asserting that there is one library component named "krb5" seems inaccurate.

* I found a statement elsewhere (cyclonedx-maven-plugin issue 67) that "CPE is deprecated by the NVD and by CycloneDX", implying that maybe swid is better.  (cpe isn't actually marked as deprecated in the v1.6 CycloneDX documentation, however.)
 
* I didn't exhaustively check the license list, but while the GPL is referenced a number of times in the tree, it is the intent that everything in MIT krb5 be distributable under a permissive license.  For instance, src/lib/krb5/deltat.c has the GPL license text, but it is actually GPL-2.0-with-bison-exception, with the exception being the important part.  Except it isn't exactly that--it references GPL 3, and the SPDX list doesn't seem to have a GPL-3-with-bison-exception entry.

* The top-level authors field is supposed to be "The person(s) who created the BOM."  Is @VCS_SBOM_AUTHORS@ going to be just the person who made the last (possibly trivial) modification?

* The component authors field is supposed to be "The person(s) who created the component."  Does that refer to the actual code making up the component (in which case the authors list would be large and burdensome to maintain), or the BOM component description?

hughsie:
> Apologies for the long delay.

No worries.

> My general reaction is that this seems burdensome to maintain. 

If it helps, krb5 is one of the more complicated SBOM files I've submitted; it's probably the oldest in terms of "first commit date" so maybe that's not a huge surprise. :)

> Unlike many open source projects, we have numerous different licenses--all permissive open source, but with plenty of differences in detail. We already collect these licenses into doc/notice.rst, and there is some risk of even that one file becoming out of sync with the source code. 

Right; agreed. If it helps, I'm a huge proponent in automation for things like this -- e.g. scanning each file for `SPDX-License-Identifier` and then building the list for the tarball.

> Moreover, CycloneDX license descriptions are references to the SPDX license list, and while that list is quite long, I'm not sure it maps 1:1 with all of our licenses, or will do so in the future. 

There's certainly an impedance mismatch, but I do think it's the future. Debian and Red Hat are 100% committed to expressing licenses in SPDX format, so if there's not an exact variant of SPDX that's correct, there soon should be.

> For instance, there is an ancient FundsXpress INC license on much of the lib/krb5/crypto code. It's very similar to the MIT license of a certain era except for the asserted holder of copyright, but the body of the license contains "the name of FundsXpress. not be used in advertising or publicity" which is of course not going to be in the text of any other license. I don't really want to risk making inaccurate legal statements about the licenses of copyright holders other than MIT.

Right. I didn't know this, and I suspect quite a few people shipping krb5 in random firmware blobs don't either! I strongly believe "institutional knowledge" like this needs to be documented somewhere machine readable.

> I have some specific reactions, which are only relevant if I can be argued into accepting the general idea:

Ohh -- I certainly don't want to argue; if you're not happy with the concept it doesn't go upstream and the downstreams shipping krb5 can make up something good enough for their compliance department. At least with a Red Hat on I'm comfortable using the SPDX licence list from kerberos.spec when building a container -- it just feels like the kind of information that upstream should _control_.

>     The source tree contains a number of libraries and applications. Does CycloneDX or SBOM anticipate including a component for each one? (If so, the license lists explode in length.) Simply asserting that there is one library component named "krb5" seems inaccurate.

I think it would be better to split them out as separate components -- I didn't do this as I wasn't sure if you'd accept this kind of thing upstream at all. If this is the direction you'd like me to explore just tell me the _granularity_ you'd like, e.g. directories you'd like to treat as separate things. This is also helpful in other projects like openssl as if a subcomponent doesn't get built then it shouldn't be included in the composite SBOM.

>     I found a statement elsewhere (cyclonedx-maven-plugin issue 67) that "CPE is deprecated by the NVD and by CycloneDX", implying that maybe swid is better. (cpe isn't actually marked as deprecated in the v1.6 CycloneDX documentation, however.)

I think quite a lot of software depends on CPE; in most cases I've used the PURL as the identifier for a project and include the CPE as an additional ID so we can "connect" up existing CVEs that do use the CPE as the identifier. If you'd rather not use CPE at all that's fine, but it's probably quite useful to have even if deprecated -- although I was surprised to see hear that it was, so I need to do some more research myself.

>     I didn't exhaustively check the license list, but while the GPL is referenced a number of times in the tree, it is the intent that everything in MIT krb5 be distributable under a permissive license. For instance, src/lib/krb5/deltat.c contains the GPL license text, but it is actually GPL-2.0-with-bison-exception, with the exception being the important part. Except it isn't exactly that--it references GPL 3, and the SPDX list doesn't seem to have a GPL-3-with-bison-exception entry.

Ahh, for exceptions the syntax is a bit different from SPDX 2.0 -- e.g. in this case it should be `GPL-3.0-only WITH Bison-exception-2.2` I think.

>     The top-level authors field is supposed to be "The person(s) who created the BOM." Is @VCS_SBOM_AUTHORS@ going to be just the person who made the last (possibly trivial) modification?

Nah, if you use uswid then it'll be everyone (as a list) that has contributed more than 10% of the file contents, e.g. https://github.com/hughsie/python-uswid/blob/main/uswid/vcs.py -- but if that makes you uncomfortable, you can either hardcode a list of names or leave it out entirely -- if you do that then tools like uswid use git to work out the list and fixup the sbom file before it's merged to the composite artifact.

>     The component authors field is supposed to be "The person(s) who created the component." Does that refer to the actual code making up the component (in which case the authors list would be large and burdensome to maintain), or the BOM component description?

The component authors is everyone who wrote the source code, and in the case of krb5 it's a long list indeed. If we split the one file to multiple subdirectories the list might be more sane.

PR #1393, opened 2024-11-25
db2 and klmdb: support principal aliases

abbra:
Allow specifying existing Kerberos principal as an alias when creating a principal via principal DB argument 'alias=existing.principal@REALM'.

DB arguments get stored as TL data with a tag KRB5_TL_DB_ARGS. As LMDB stores DB entries by using principal name as the key, in get_principal() we can check if retrieved entry has TL data of KRB5_TL_DB_ARGS with 'alias=' prefix and retrieve this new entry instead.

TODO: take flags into account when resolving aliases

### DIFF
--- a/src/plugins/kdb/ldap/libkdb_ldap/ldap_principal.c
+++ b/src/plugins/kdb/ldap/libkdb_ldap/ldap_principal.c
@@ -128,6 +128,76 @@ krb5_dbe_free_contents(krb5_context context, krb5_db_entry *entry)
     return;
 }
 
+static krb5_error_code
+iterate_entry(krb5_context context, krb5_ldap_context *ldap_context,
+              LDAP *ld, LDAPMessage *ent,
+              krb5_error_code (*func)(krb5_pointer, krb5_db_entry *),
+              krb5_pointer func_arg)
+{
+    krb5_error_code ret = 0;
+    krb5_principal cprinc = NULL, nprinc = NULL;
+    krb5_db_entry entry = { 0 }, *entptr;
+    char **canon = NULL, **names = NULL, **list;
+    size_t i;
+
+    canon = ldap_get_values(ld, ent, "krbCanonicalName");
+    names = ldap_get_values(ld, ent, "krbPrincipalName");
+    if (canon == NULL && names == NULL)
+        return 0;
+
+    /* Output an entry for the canonical name if one is given.  Otherwise
+     * output an entry for the first name within the realm. */
+    list = (canon != NULL) ? canon : names;
+    for (i = 0; list[i] != NULL; i++) {
+        krb5_free_principal(context, nprinc);
+        nprinc = NULL;
+        ret = krb5_ldap_parse_name(context, list[i], &nprinc);
+        if (ret)
+            goto cleanup;
+
+        if (is_principal_in_realm(ldap_context, nprinc)) {
+            ret = populate_krb5_db_entry(context, ldap_context, ld, ent,
+                                         nprinc, &entry);
+            if (ret)
+                goto cleanup;
+            ret = (*func)(func_arg, &entry);
+            krb5_dbe_free_contents(context, &entry);
+            if (ret)
+                goto cleanup;
+            break;
+        }
+    }
+
+    /* Output alias entries for each non-canonical name. */
+    if (canon != NULL && names != NULL) {
+        ret = krb5_ldap_parse_name(context, canon[0], &cprinc);
+        if (ret)
+            goto cleanup;
+        for (i = 0; names[i] != NULL; i++) {
+            if (strcmp(names[i], canon[0]) == 0)
+                continue;
+            krb5_free_principal(context, nprinc);
+            nprinc = NULL;
+            ret = krb5_ldap_parse_name(context, names[i], &nprinc);
+            if (ret)
+                goto cleanup;
+            ret = krb5_dbe_make_alias_entry(context, nprinc, cprinc, &entptr);
+            if (ret)
+                goto cleanup;
+            ret = (*func)(func_arg, entptr);
+            krb5_db_free_principal(context, entptr);
+            if (ret)
+                goto cleanup;
+        }
+    }
+
+cleanup:
+    krb5_free_principal(context, cprinc);
+    krb5_free_principal(context, nprinc);
+    ldap_value_free(canon);
+    ldap_value_free(names);
+    return ret;
+}
 
 krb5_error_code
 krb5_ldap_iterate(krb5_context context, char *match_expr,
@@ -135,9 +205,8 @@ krb5_ldap_iterate(krb5_context context, char *match_expr,
                   krb5_pointer func_arg, krb5_flags iterflags)
 {
     krb5_db_entry            entry;
-    krb5_principal           principal;
-    char                     **subtree=NULL, *princ_name=NULL, *realm=NULL, **values=NULL, *filter=NULL;
-    size_t                   tree=0, ntree=1, i=0;
+    char                     **subtree=NULL, *realm=NULL, *filter=NULL;
+    size_t                   tree=0, ntree=1;
     krb5_error_code          st=0, tempst=0;
     LDAP                     *ld=NULL;
     LDAPMessage              *result=NULL, *ent=NULL;
@@ -181,33 +250,10 @@ krb5_ldap_iterate(krb5_context context, char *match_expr,
 
         LDAP_SEARCH(subtree[tree], ldap_context->lrparams->search_scope, filter, principal_attributes);
         for (ent=ldap_first_entry(ld, result); ent != NULL; ent=ldap_next_entry(ld, ent)) {
-            values=ldap_get_values(ld, ent, "krbcanonicalname");
-            if (values == NULL)
-                values=ldap_get_values(ld, ent, "krbprincipalname");
-            if (values != NULL) {
-                for (i=0; values[i] != NULL; ++i) {
-                    if (krb5_ldap_parse_principal_name(values[i], &princ_name) != 0)
-                        continue;
-                    st = krb5_parse_name(context, princ_name, &principal);
-                    free(princ_name);
-                    if (st)
-                        continue;
-
-                    if (is_principal_in_realm(ldap_context, principal)) {
-                        st = populate_krb5_db_entry(context, ldap_context, ld,
-                                                    ent, principal, &entry);
-                        krb5_free_principal(context, principal);
-                        if (st)
-                            goto cleanup;
-                        (*func)(func_arg, &entry);
-                        krb5_dbe_free_contents(context, &entry);
-                        break;
-                    }
-                    (void) krb5_free_principal(context, principal);
-                }
-                ldap_value_free(values);
-            }
-        } /* end of for (ent= ... */
+            st = iterate_entry(context, ldap_context, ld, ent, func, func_arg);
+            if (st)
+                goto cleanup;
+        }
         ldap_msgfree(result);
         result = NULL;
     } /* end of for (tree= ... */
@@ -268,15 +314,17 @@ krb5_ldap_delete_principal(krb5_context context,
 
     GET_HANDLE();
 
-    if (ptype == KDB_STANDALONE_PRINCIPAL_OBJECT) {
+    if (ptype == KDB_STANDALONE_PRINCIPAL_OBJECT &&
+        (pcount == 1 ||
+         krb5_principal_compare(context, searchfor, entry->princ))) {
         st = ldap_delete_ext_s(ld, DN, NULL, NULL);
         if (st != LDAP_SUCCESS) {
             st = set_ldap_error (context, st, OP_DEL);
             goto cleanup;
         }
     } else {
-        if (((st=krb5_unparse_name(context, searchfor, &user)) != 0)
-            || ((st=krb5_ldap_unparse_principal_name(user)) != 0))
+        st = krb5_ldap_unparse_name(context, searchfor, &user);
+        if (st)
             goto cleanup;
 
         memset(strval, 0, sizeof(strval));
@@ -360,35 +408,6 @@ is_standalone_principal(krb5_context kcontext, krb5_db_entry *entry, int *res)
     if (!code)
         *res = (*res == KDB_STANDALONE_PRINCIPAL_OBJECT) ? 1 : 0;
     return code;
-}
-
-/*
- * Unparse princ in the format used for LDAP attributes, and set *user to the
- * result.
- */
-static krb5_error_code
-unparse_principal_name(krb5_context context, krb5_const_principal princ,
-                       char **user_out)
-{
-    krb5_error_code st;
-    char *luser = NULL;
-
-    *user_out = NULL;
-
-    st = krb5_unparse_name(context, princ, &luser);
-    if (st)
-        goto cleanup;
-
-    st = krb5_ldap_unparse_principal_name(luser);
-    if (st)
-        goto cleanup;
-
-    *user_out = luser;
-    luser = NULL;
-
-cleanup:
-    free(luser);
-    return st;
 }
 
 /*
@@ -478,10 +497,10 @@ krb5_ldap_rename_principal(krb5_context context, krb5_const_principal source,
         goto cleanup;
     }
 
-    st = unparse_principal_name(context, source, &suser);
+    st = krb5_ldap_unparse_name(context, source, &suser);
     if (st)
         goto cleanup;
-    st = unparse_principal_name(context, target, &tuser);
+    st = krb5_ldap_unparse_name(context, target, &tuser);
     if (st)
         goto cleanup;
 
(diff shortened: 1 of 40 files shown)


abbra:
@greghudson @cryptomilk This PR implements aliases for both DB2 and LMDB KDB backends. It is not ready for use yet but a minimal test works.

We still need to adjust behavior of `get_principal()` by taking flags into account as not every alias should be resolvable to the original name.

src/plugins/kdb/lmdb/kdb_lmdb.c
```diff
@@ -752,20 +790,37 @@ klmdb_put_principal(krb5_context context, krb5_db_entry *entry, char **db_args)
...
-    }
-
     if (dbc == NULL)
         return KRB5_KDB_DBNOTINITED;
 
+    if (db_args != NULL) {
+        ret = EINVAL;
+        for (i = 0; db_args[i] != NULL; i++) {
+            if (strncmp(db_args[i], ALIAS_PREFIX, ALIAS_PREFIX_LEN) == 0) {
+                /* find alias target */
```
cryptomilk, line 804: Shouldn't you check with krb5_parse_name() if the alias is a valid principal?

src/plugins/kdb/db2/kdb_db2.c
```diff
@@ -829,6 +858,44 @@ krb5_db2_put_principal(krb5_context context, krb5_db_entry *entry,
...
+                    k5_setmsg(context, KRB5_KDB_NOENTRY,
+                              _("Alias \"%s\" not found in the database"),
+                              db_args[i] + ALIAS_PREFIX_LEN);
+                    goto cleanup;
+                }
+                break;
+            }
+        }
+        if (retval != 0) {
+            /* DB2 does not support db_args DB arguments for principal */
```
cryptomilk, line 892: This comment can be removed.

cryptomilk:
Now smbclient works establishing a connection from my machine to a VM using the mdns name.

```
$ bin/smbclient //sambakdc.local/testshare -Uasn@SAMBAKDC.LOCALKDC.SITE -d5
[...]
Starting GENSEC submechanism gse_iakerb                                                                                                                                             Password for [asn@SAMBAKDC.LOCALKDC.SITE]:
gensec_gse_iakerb_client_start: Acquire credentials for asn@SAMBAKDC.LOCALKDC.SITE to access cifs/sambakdc.local
[...]
Try "help" to get a list of possible commands.
smb: \>
```

abbra:
Added kadmin man page update that documents `addprinc -x alias=...` for DB2 and LMDB modules.

greghudson:
Two initial reactions:

* The LDAP module already supports aliases, so we could presumably add support for the put_principal DB argument there.

* I'd like to explore representing aliases in a more direct way than a dummy entry with a tl-data entry.  A separate key representation isn't ideal because it would require an extra lookup (though that could be done second, so the cost is only paid when the entry doesn't exist).  It doesn't look like we have any kind of outer tag or version on marshalled principal entries in either DB2 or LDBM, so we'd be picking a representation that's easily distinguishable from principal entries.  I will look into this.

abbra:
Thanks. I'll look into adding aliases to ldap module in put_principal.

greghudson:
I have a concern that it might be possible to bypass access controls using aliases.  If you have create/read access to some part of the principal namespace but no access to the rest, you could create an alias to a principal outside of your namespace and getprinc that.  Fixing this would require teaching kadmind about KRB5_TL_DB_ARGS, and I'd prefer not to expand that mechanism (which I don't particularly like) to more components.

If we instead created a new kadmin command/RPC, kadmind could properly check access controls when creating aliases.  This approach would also create a better admin user experience.  The downside is that both the kadmin client and server would have to be upgraded to enable this feature, instead of just the client.

abbra:
I'm ok with a new kadmin command because also a change to ldap driver doesn't really fit well with addprinc out modprinc. An explicit alias command is better there because if it's assumed to operate on already existing principal.

src/plugins/kdb/lmdb/kdb_lmdb.c
```diff
@@ -737,13 +745,87 @@ klmdb_get_principal(krb5_context context, krb5_const_principal searchfor,
...
+                    *entry_out = NULL;
+                    k5_setmsg(context, KRB5_KDB_NOENTRY,
+                            _("Alias \"%s\" not found in the database"),
+                            db_args[i] + ALIAS_PREFIX_LEN);
+                    ret = KRB5_KDB_NOENTRY;
+                    goto cleanup;
+                }
+                /* replace entry->princ by the aliasing entry princ */
+                krb5_free_principal(context, temp_entry->princ);
+                krb5_copy_principal(context, (*entry_out)->princ, &temp_entry->princ);
```
greghudson, line 772: Why this?  The behavior documented in kdb.h is that the module should return an entry with the canonical name, not the requested name.

greghudson:
Two edge cases I want to think about:

* Alias chains and cycles.  With the current PR code chains are allowed, and the consequence of a cycle is that a fetch operation will blow the stack.  We could allow alias chains up to a reasonable limit (like 10) using a depth parameter, or we could disallow alias chains entirely and use a non-recursive implementation.  It's probably less total code to allow chains up to a limit (because otherwise I would feel like we should check for an alias chain at alias creation time).  But I'm not sure there are any good applications for alias chains, so perhaps it's cleaner to disallow them.

* modprinc/chpw/etc. applied to an alias name.  With the current PR code I believe the operation will succeed but have no effect, as it will write to the dummy alias entry.  The most admin-friendly behavior is probably to apply the modification to the alias target, but it is probably easier and safer to make the operation fail.  (Safer because otherwise kadmind has to be careful to do the ACL check against the canonical name and not the requested name.)

src/plugins/kdb/lmdb/kdb_lmdb.c
abbra: My mistake, I'll fix it.

abbra:
Alias chains: a very common case is making SPNs for AD compatibility. `host/...` is the primary principal, `cifs/...` is an alias, as well as any other SPN principal which can be derived from either `host/..` or `cifs/..`. I'd rather allow a limited chain depth than to disable them completely.

Applying `modprinc`/`chpw`/etc. to an alias target would mean we need to apply ACL check deeper. Do we have means to do that? As far as I can see, we don't really call `kadm5_get_principal()` in all code paths prior to applying an ACL...

greghudson:
I have no opposition to allowing alias chains if they're convenient, but I don't totally understand the SPN example.  Where would an alias to an alias come in there?

We don't currently have the means to apply ACL checks to alias targets, so we'd have to add that if we wanted to support modprinc/etc. on aliases.

Last month I said I wanted to explore representing aliases directly in the DB rather than via typed data inside a stub principal entry.  I started in on an experiment along those lines, and found value representations that wouldn't realistically conflict with principal entries (start with anything other than 0x2600 for DB2, start with 0xFFFFFFFF for LMDB).  I pretty quickly realized we would also have to extend the iprop protocol and dump format for aliases if we go this route.  If we don't extend those explicitly, we will be extending them implicitly via these stub principal entries.  This has some implications for LDAP; we support dumping a DB2 database and loading it into LDAP, which already has a different way of representing aliases.  I'll have to think on this more.

greghudson:
After more thought, I can see two different design possibilities:

1. Represent aliases with stub principal entries (probably using a new tl-data type rather than TL_DB_ARGS).  LDAP can generate stub entries in its iterate() method and interpret them in its add_principal() method to be dump/iprop-compatible with DB2/LMDB.  For this direction, I would probably add libkdb5 utility functions to create and interpret alias principal entries.  I might also consider interpreting them in krb5_db_get_principal() so that the key-value modules wouldn't need any changes at all.

2. Explicitly extend the KDB model (as seen by administrative tooling and the key-value modules) to include alias entries: add a KDB method to add an alias, extend the iterate() method to have a second callback for aliases, add a line type for aliases to the dump format, and extend iprop to support aliases.

While I think there are simplicity arguments for the end state of (2), it's looking like a lot of work compared to (1) especially in the iprop department.  It would also cause more cross-version compatibility headaches.  I will work on revising the PR in the direction of (1) for now.

greghudson:
Similar to a point--the logic is in libkdb5 and there's a new kadmin command.  I've pushed a rough cut.

abbra:
Thanks for the changes. They look fine for lmdb/db2 backends' usage.

For ldap backend we would need to modify `populate_krb5_db_entry()` to return TL records for aliases in the attributes of the entry. For `put_principal()` path we also need to handle the new TL records accordingly but also need to identify that an alias entry is being written and replace the write to write to the original entry's `krbprincipalname` attribute instead. This means potential `put_principal()` recursion?

greghudson:
> For ldap backend we would need to modify populate_krb5_db_entry() to return TL records for aliases in the attributes of the entry.

I don't think this is right.  KRB5_TL_ALIAS_TARGET conveys the target name of an alias stub, not the alias names of a canonical principal record.  Since the LDAP module doesn't need recursive lookups to resolve aliases, we shouldn't need to make any changes to get_principal(); that method will continue to just return the principal entry with the principal field set to the canonical name.

However, to be dump-compatible with the key-value modules, we do need to change put_principal() as you surmised, and we also need to change iterate() to generate alias stub entries with any alias names in addition to populated entries with canonical names.

greghudson:
> modprinc/chpw/etc. applied to an alias name.

I started work on this under the assumption that we would deny these operations, but I ran into an apparent conflict: we have a special 'e' privilege for extracting keys without randomizing them (xst -norandkey).  We will have a privilege escalation through aliases exploit unless we:

1. Prohibit this operation on aliases, which I think this would be a significant hindrance to using aliases for services.  At a minimum, we'd want to add an xst option to change the principal under which the extracted keys are recorded in the keytab.
2. Require the 'e' privilege on the target to create an aliases.  Also a significant limitation on using aliases.
3. Check the ACL on the canonical name.

So there's a fairly strong argument for implementing canonicalization for ACL checks for this one operation, at which point it probably makes sense to do it for all of them.

abbra:
I agree. Operations on aliases should be controlled against ACL on the canonical name. After all, any of the aliases effectively gives you the same key as the original principal.

greghudson:
This is a bit far in the weeds, but I also had some thoughts about the semantics of aliases that don't resolve (point to a nonexistent principal name, or hit the depth limit).  Currently we return KRB5_KDB_NOENTRY for these, which means they have a kind of dubious existence: they show up in dumps and in kadmin listprincs, but appear as nonexistent to inquire operations and don't block kadmin addprinc or kadmin renprinc.  And of course they spring back into full alias existence if their target is created.

Several options here:

1. Leave it alone.  The emergent semantics are a little weird but not harmful, and these should be rare.
2. Try to prevent non-resolving aliases.  We can't robustly do this efficiently in the current design, but we could prevent the most  obvious route by requiring the target to exist when creating an alias.
3. Return an error different from KRB5_KDB_NOENTRY when we encounter a non-resolving alias.

There is some potential for information leakage (determining whether a principal exists in a part of the DB you don't have inquire privilege on) which might vary depending on what we choose, but it's probably not worth worrying about.

greghudson:
I pushed kadmind changes to canonicalize modprinc/etc. target principals.  Still to address:

* rename_principal on an alias (for klmdb/db2) doesn't behave appropriately.  The default implementation starts with get_principal(), which resolves the alias, so the destination receives a copy of the canonical principal entry instead of an alias to the same target.  After some thought, I think we might want to just prohibit renaming aliases, as it's easy enough to add a new alias and delete the old one.

* LDAP: delete_principal needs to recognize aliases and remove the alias instead of deleting the whole entry, or we have a privilege escalation issue.  put_principal should recognize alias principals and add an alias to the canonical entry.  iterate should produce alias entries when a principal has multiple names and a canonical name.  rename_principal should deny renames of aliases (assuming that's the decision we make for the first bullet point) if it isn't denied at the libkdb5 layer.

* Consider whether to change the ACL permissions needed to make an alias in light of the decision to canonicalize target principals.  Right now it's 'a' (addition) on source and 'i' (inquire) on destination.  I don't think any permission check on the target is necessary to avoid attacks now that kadmind is canonicalizing; however, I can see an argument for strengthening the required target permissions to 'c' (change) because a restricted admin doesn't really have any business making aliases to a principal they don't control.

* Consider whether to add a kdb5_util tabdump table for aliases.

greghudson:
I think that's everything.  I changed krb5_db_rename_principal() to require that the source name be canonical, changed the permission requirements on alias targets to 'm', implemented LDAP support and added appropriate comments to kdb.h, and added a tabdump table for aliases.  I expect to make only small changes before merging at this point.

abbra:
Thank you. Sorry for not reacting earlier, I was traveling and now sick.

abbra (commented):
One comment regarding ldap driver rewrite. Otherwise it looks good, thank you!

src/plugins/kdb/ldap/libkdb_ldap/ldap_principal.c
```diff
@@ -128,16 +128,85 @@ krb5_dbe_free_contents(krb5_context context, krb5_db_entry *entry)
...
+            ret = populate_krb5_db_entry(context, ldap_context, ld, ent,
+                                         nprinc, &entry);
+            if (ret)
+                goto cleanup;
+            ret = (*func)(func_arg, &entry);
+            krb5_dbe_free_contents(context, &entry);
+            if (ret)
+                goto cleanup;
+            break;
+        }
```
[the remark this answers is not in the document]
greghudson: The LDAP module has the is_principal_in_realm() check for get_principal, put_principal, and iterate.  It accepts krbtgt/REALM@* as well as */REALM.  There is no corresponding check in DB2 or LMDB.

I don't have any documentation Novell's original motivation for this aspect of the design.  I would speculate that they wanted to support scenarios where multiple Kerberos realms could be layered on top of the same hierarchy of pre-existing LDAP user and service objects using krbPrincipalAux.  That may be what they had in mind in allowing multiple krbPrincipalName attributes to be associated with an object with no indication of which one is canonical (krbCanonicalName was added later by me).  There's a comment in krb5_ldap_delete_principal() suggesting that they at one point planned for multiple key sets in an object:

    NOTE: krbsecretkey attribute has per principal entries. There can be chances that the                                                           
    other principals' keys are existing/left-over. So delete all the values.                                                                        

but they didn't get there; "krbsecretkey" isn't the name of an attribute, and the krbPrincipalKey attribute doesn't have anything to distinguish between keys for different principals.  A multiple-realm setup within a single hierarchy could still currently work if the principals don't collide (i.e. each LDAP object corresponds to a Kerberos principal in at most one realm), though I doubt anyone is doing this.

I couldn't find anything in the links you provided about the A$@B entry (I'm probably just missing it).  Does it ever get looked up within realm A by that name?  I know it's important in Samba to replicate Microsoft behavior with high fidelity, but with the stock LDAP KDB module it may not be as important.  If it is important to allow creating and dumping principals by that name, I think it would be out of scope for this PR.

> On the other hand, the alias handling below does allow to see aliases outside of the realm, so interdomain trust account for outbound entry would be visible. At least, it is inconsistent.

This brings up the broader question of whether to allow cross-realm aliases (as created by kadmin; we can't prevent them from being set up manually in LDAP) and what they mean if created.  They could be used to indicate realm referrals, but (a) we'd have to do additional work in each KDB module to handle that, (b) the name part of the alias wouldn't be meaningful, only the realm, and (c) it wouldn't work with LDAP, or at least you'd have to create a out-of-realm principal object (manually at the moment, since put_principal denies out-of-realm principals) to hang the alias on.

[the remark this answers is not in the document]
abbra: >  couldn't find anything in the links you provided about the A$@B entry (I'm probably just missing it). Does it ever get looked up within realm A by that name? I know it's important in Samba to replicate Microsoft behavior with high fidelity, but with the stock LDAP KDB module it may not be as important. If it is important to allow creating and dumping principals by that name, I think it would be out of scope for this PR.

We store it this way in FreeIPA. Here is an example from IPA-IPA trust work-in-progress demo (https://github.com/abbra/freeipa-local-tests/tree/main/ipalab-config/ipa-trust). The same structure is used for IPA-AD forest trust already. Once trust is established between two IPA deployments (`ipa1demo.test` and `ipa2demo.test`), we'd have following LDAP entries in each deployment's LDAP trees (this is for bidirectional trust):

```
# krbtgt/IPA1DEMO.TEST@IPA2DEMO.TEST, ipa2demo.test, ad, trusts, ipa1demo.test
dn: krbPrincipalName=krbtgt/IPA1DEMO.TEST@IPA2DEMO.TEST,cn=ipa2demo.test,cn=ad,cn=trusts,dc=ipa1demo,dc=test
objectClass: krbPrincipal
objectClass: krbPrincipalAux
objectClass: krbTicketPolicyAux
objectClass: top
krbCanonicalName: krbtgt/IPA1DEMO.TEST@IPA2DEMO.TEST
krbPrincipalName: krbtgt/IPA1DEMO.TEST@IPA2DEMO.TEST
krbPrincipalKey:: some value
krbLastPwdChange: 20250219070326Z
krbExtraData:: some value

# krbtgt/M1@IPA2DEMO.TEST, ipa2demo.test, ad, trusts, ipa1demo.test
dn: krbPrincipalName=krbtgt/M1@IPA2DEMO.TEST,cn=ipa2demo.test,cn=ad,cn=trusts,dc=ipa1demo,dc=test
objectClass: krbPrincipal
objectClass: krbPrincipalAux
objectClass: krbTicketPolicyAux
objectClass: ipaAllowedOperations
objectClass: top
krbCanonicalName: krbtgt/M1@IPA2DEMO.TEST
krbPrincipalName: krbtgt/M1@IPA2DEMO.TEST
krbPrincipalName: M1$@IPA2DEMO.TEST
krbTicketFlags: 64
ipaAllowedToPerform;read_keys: cn=adtrust agents,cn=sysaccounts,cn=etc,dc=ipa1demo,dc=test
ipaAllowedToPerform;read_keys: cn=trust admins,cn=groups,cn=accounts,dc=ipa1demo,dc=test
krbPrincipalKey:: some value
krbLastPwdChange: 20250219070326Z
krbExtraData:: some value

# krbtgt/IPA2DEMO.TEST@IPA1DEMO.TEST, ipa2demo.test, ad, trusts, ipa1demo.test
dn: krbPrincipalName=krbtgt/IPA2DEMO.TEST@IPA1DEMO.TEST,cn=ipa2demo.test,cn=ad,cn=trusts,dc=ipa1demo,dc=test
objectClass: krbPrincipal
objectClass: krbPrincipalAux
objectClass: krbTicketPolicyAux
objectClass: top
krbCanonicalName: krbtgt/IPA2DEMO.TEST@IPA1DEMO.TEST
krbPrincipalName: krbtgt/IPA2DEMO.TEST@IPA1DEMO.TEST
krbPrincipalKey:: some value
krbLastPwdChange: 20250219070326Z
krbExtraData:: some value

# krbtgt/M2@IPA1DEMO.TEST, ipa2demo.test, ad, trusts, ipa1demo.test
dn: krbPrincipalName=krbtgt/M2@IPA1DEMO.TEST,cn=ipa2demo.test,cn=ad,cn=trusts,dc=ipa1demo,dc=test
objectClass: krbPrincipal
objectClass: krbPrincipalAux
objectClass: krbTicketPolicyAux
objectClass: top
krbCanonicalName: krbtgt/M2@IPA1DEMO.TEST
krbPrincipalName: krbtgt/M2@IPA1DEMO.TEST
krbPrincipalName: M2$@IPA1DEMO.TEST
krbPrincipalKey:: some value
krbLastPwdChange: 20250219070326Z
krbExtraData:: some value

```

The entries with a NetBIOS name alias (`M1$@IPA2DEMO.TEST` and `M2$@IPA1DEMO.TEST`) are these `A$@B` aliases. They are using salt of `krbtgt/A@B`, not `krbtgt/FQDN_A@B`. In IPA case we don't use them within KDC directly but rather obtain the keytab for `A$@B` and use in SSSD to authenticate to domain controllers of `B`.

I agree that changes to LDAP KDB driver do not affect this as FreeIPA is using different KDB driver. I just wanted to raise this point because eventually I'd like to make us to make KDB drivers composable and reuse common components as much as possible across multiple consumers.

Let's skip this point for the LDAP driver for now.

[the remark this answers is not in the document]
abbra: > I couldn't find anything in the links you provided about the A$@B entry 

No, this is not exactly written down in MSFT specs because cross-realm `krbtgt/..` keys are synthesized out of TDO content at runtime. So internal database structure is less important for them. But we (FreeIPA) try to keep closer to traditional krb5 schema (one principal - one db entry, more or less) with varying success. We aren't tied to AD LDAP schema and AD LDAP DIT, Samba AD does. So most of points I have are a mixture of two projects' needs...

In AD it is expected that a remote trusted domain's domain controller will authenticate to a trusting domain's DC using their interdomain trust account (which is a "normal" machine account, kind-of, in the trusting domain): covered by [MS-ADTS 6.1.6.8 Essential Attributes of Interdomain Trust Accounts](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/ac527b5b-0e88-48a1-8c73-497d40388d04). As I said above, in ([MS-KILE 3.3.5.6.2 Referrals](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/856a8663-6c36-46f0-9f6f-1a28a700d0b5) and [3.3.5.7.5 Cross-Domain Trust and Referrals](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/bac4dc69-352d-416c-a9f4-730b81ababb3)) there is an implication that:
- trusting domain's DC can derive krbtgt/.. principal's keys using TDO information
- trusted domain's DC can derive interdomain trust account's keys using their own TDO information and constructing an account in the trusting domain's namespace for authentication

Mapping of that to 'single LDAP entry for a principal' brought us to those '4 entries per bi-directional trust' situation I outlined above, where one entry (`M1$@IPA2DEMO.TEST` in `IPA1DEMO.TEST` realm) is technical, purely to allow picking up the keytab for authenticating against `IPA2DEMO.TEST` and reusing tools we have already for handling Kerberos keys. The other entry (`M2#@IPA1DEMO.TEST` in `IPA1DEMO.TEST`) is our analogue of interdomain trust account in AD. This is how the other domain will authenticate in all cross-domain situations in AD.

So, yes, aliasing is important for those entries to be treated properly from the KDC point of view because it influences the salt (MS-KILE 3.3.5.6.2). But the need to handle a foreign realm principal `M1$@IPA2DEMO.TEST` is our internal FreeIPA business here, driven by the aim to normalize access to all kerberos principals using the same means. If we were to use LDAP KDB driver as it is now, it wouldn't be able to find this principal at all. Samba AD wouldn't need to handle that itself as they don't have these foreign realm principals in the database in the first place.

PR #1395, opened 2024-11-28
Fix the memory leak issue when Kerberos fails to connect to OpenLDAP.

gardonkoo:
The function "initialize_server" leads to memory leak when the function "authenticate" returns non-success.
We forget to free the memory of "server->ldap_handle".
We should free "server->ldap_handle" before we free "server".

### DIFF
--- a/src/plugins/kdb/ldap/libkdb_ldap/kdb_ldap_conn.c
+++ b/src/plugins/kdb/ldap/libkdb_ldap/kdb_ldap_conn.c
@@ -189,6 +189,7 @@ initialize_server(krb5_ldap_context *ldap_context, krb5_ldap_server_info *info)
     if (ret) {
         info->server_status = OFF;
         time(&info->downtime);
+        ldap_unbind_ext_s(server->ldap_handle, NULL, NULL);
         free(server);
         return ret;
     }


greghudson:
Have you tested the fix?  Also, have you encountered the leak in practice, or was it discovered through source code analysis?

The reason I ask is that the original LDAP module we received from Novell had a commented-out call to release the handle in this block:
```
       server_info->server_status = OFF;                                       
       time(&server_info->downtime);                                           
       /* ldap_unbind_s(ldap_server_handle->ldap_handle); */                   
       free(ldap_server_handle);                                               
```
There's never been a way to inquire why (we didn't get their version control logs, if they had any), but I've declined to fix it at least once out of concern that I was missing something.

gardonkoo:
> Have you tested the fix? Also, have you encountered the leak in practice, or was it discovered through source code analysis?
> Yes, I encountered the leak in practice, and I have tested the fix.  To construct the leak, we could  kill slapd server straightly , and then do some "kinit"  things.

the function may be called as follows:
LDAP_SEARCH->LDAP_SEARCH_1->krb5_ldap_rebind->krb5_ldap_request_next_handle_from_pool->krb5_retry_get_ldap_handle->krb5_ldap_db_single_init->initialize_server->authenticate

initialize_server calls the "ldap_initialize" of OpenLDAP, let's see the implementation of ldap_initialize:
image
LINE 301, ldap_create actually calls malloc to apply memory for ld
but the slapd server has been shut down, then authenticate will return non-zero.

greghudson:
Thanks for the clarification.

PR #1396, opened 2024-11-29
Allow kdb plugin reentrancy

cryptomilk:
This allows to write a plugin which will load another plugin to extend its functionality.

### DIFF
--- a/src/include/kdb.h
+++ b/src/include/kdb.h
@@ -353,7 +353,17 @@ extern char *krb5_mkey_pwd_prompt2;
 #define KRB5_DB_LOCKMODE_EXCLUSIVE    0x0002
 #define KRB5_DB_LOCKMODE_PERMANENT    0x0008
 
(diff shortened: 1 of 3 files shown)


cryptomilk:
Here is the module which just wraps klmdb.so for now:

https://gitlab.com/cryptomilk/localkdc/-/blob/main/src/kdb/kdb_userdb.c?ref_type=heads

This seems to work just fine.

greghudson:
I pushed some changes, most notably changing the API name to krb5_db_load_module().

In the wrapper module you linked, I think you can just pass "klmdb" (without ".so") for improved portability; the libkrb5support plugin functions will add a platform-appropriate suffix.  Also, I am curious if using krb5_copy_context() works instead of creating a new context from the profile object.

cryptomilk:
I've did the changes to the module, `krb5_copy_context()` works fine too.

PR #1398, opened 2024-12-02
Default kdc_tcp_listen to kdc_listen

greghudson:
If kdc_tcp_listen is not specified in the realm or in [kdcdefaults], use the same listeners as were given for UDP instead of separately defaulting to port 88.  This change makes the kdc_listen and kpasswd_listen more consistent, while still allowing UDP and TCP listening to be separately configured when required for the KDC.

[As discussed in #1359]


cryptomilk:
lgtm

PR #1400, opened 2024-12-04
Add support for systemd socket activation

cryptomilk:
This adds support for systemd socket activation.

See https://0pointer.de/blog/projects/socket-activation.html

This requires PR #1359 to be merged first.

### DIFF
--- a/src/include/socket-utils.h
+++ b/src/include/socket-utils.h
@@ -43,6 +43,8 @@
 #ifndef SOCKET_UTILS_H
 #define SOCKET_UTILS_H
 
+#include <stdbool.h>
+
 /* Some useful stuff cross-platform for manipulating socket addresses.
    We assume at least ipv4 sockaddr_in support.  The sockaddr_storage
    stuff comes from the ipv6 socket api enhancements; socklen_t is
@@ -157,6 +159,39 @@ sa_socklen(const struct sockaddr *sa)
 #endif
     else
         abort();
+}
+
+/* Return true if a and b are the same address (and port if applicable). */
+static inline bool
+sa_equal(const struct sockaddr *a, const struct sockaddr *b)
+{
+    if (a == NULL || b == NULL || a->sa_family != b->sa_family)
+        return false;
+
+    if (a->sa_family == AF_INET) {
+        const struct sockaddr_in *x = sa2sin(a);
+        const struct sockaddr_in *y = sa2sin(b);
+
+        if (x->sin_port != y->sin_port)
+            return false;
+        return memcmp(&x->sin_addr, &y->sin_addr, sizeof(x->sin_addr)) == 0;
+    } else if (a->sa_family == AF_INET6) {
+        const struct sockaddr_in6 *x = sa2sin6(a);
+        const struct sockaddr_in6 *y = sa2sin6(b);
+
+        if (x->sin6_port != y->sin6_port)
+            return false;
+        return memcmp(&x->sin6_addr, &y->sin6_addr, sizeof(x->sin6_addr)) == 0;
+#ifndef _WIN32
+    } else if (a->sa_family == AF_UNIX) {
+        const struct sockaddr_un *x = sa2sun(a);
+        const struct sockaddr_un *y = sa2sun(b);
+
+        return strcmp(x->sun_path, y->sun_path) == 0;
+#endif
+    }
+
+    return false;
 }
 
 #endif /* SOCKET_UTILS_H */
--- a/.gitignore
+++ b/.gitignore
@@ -258,6 +258,7 @@ local.properties
 /src/kdc/rtest
 /src/kdc/t_ndr
 /src/kdc/t_replay
+/src/kdc/t_sockact
 
 /src/lib/k5sprt32.def
 
--- a/doc/admin/admin_commands/kadmind.rst
+++ b/doc/admin/admin_commands/kadmind.rst
@@ -121,6 +121,14 @@ ENVIRONMENT
 See :ref:`kerberos(7)` for a description of Kerberos environment
 variables.
 
+As of release 1.22, kadmind supports systemd socket activation via the
+LISTEN_PID and LISTEN_FDS environment variables.  Sockets provided by
+the caller must correspond to configured listener addresses (via the
+**kadmind_listen** or **kpasswd_listen** variables or equivalents) or
+they will be ignored.  Any configured listener addresses that do not
+correspond to caller-provided sockets will be ignored if socket
+activation is used.
+
 
 SEE ALSO
 --------
--- a/doc/admin/admin_commands/krb5kdc.rst
+++ b/doc/admin/admin_commands/krb5kdc.rst
@@ -106,6 +106,13 @@ ENVIRONMENT
 See :ref:`kerberos(7)` for a description of Kerberos environment
 variables.
 
+As of release 1.22, krb5kdc supports systemd socket activation via the
+LISTEN_PID and LISTEN_FDS environment variables.  Sockets provided by
+the caller must correspond to configured listener addresses (via the
+**kdc_listen** variable or equivalent) or they will be ignored.  Any
+configured listener addresses that do not correspond to
+caller-provided sockets will be ignored if socket activation is used.
+
 
 SEE ALSO
 --------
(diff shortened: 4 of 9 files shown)


cryptomilk:
Rebased on master and improved documentation.

greghudson:
I will want to restructure this code, but I need to confirm my understanding of the -ENOTCONN logic first.  The idea is that if we received one or more socket activation fds in the 3... range, we should ignore any configured listeners which don't match those sockets?

For the KDC, it seems like maybe it should be unnecessary to match the kdc.conf listener configuration to the systemd configuration--if there are socket activations, we can just listen for packets/connections on those sockets.

For kadmind it's not as simple, as a TCP or UNIX domain socket could be for either the kadmin RPC protocol or the kpasswd protocol.  I could see some ways around this such as a command-line flag to indicate that all of the activated fds are for one protocol or the other, but I'm not familiar enough with systemd to know if this would make system configuration simpler or more complicated.

cryptomilk:
> I will want to restructure this code, but I need to confirm my understanding of the -ENOTCONN logic first. The idea is that if we received one or more socket activation fds in the 3... range, we should ignore any configured listeners which don't match those sockets?

Yes, exactly.
 
> For the KDC, it seems like maybe it should be unnecessary to match the kdc.conf listener configuration to the systemd configuration--if there are socket activations, we can just listen for packets/connections on those sockets.

I more or less followed https://0pointer.de/blog/projects/socket-activation2.html

greghudson:
> I more or less followed https://0pointer.de/blog/projects/socket-activation2.html

That says:
> If we got more sockets passed than configured in CUPS we implicitly add configuration for them. If the CUPS configuration includes definitions for more listening sockets those will be bound natively in CUPS. That way we'll very robustly listen on all ports that are listed in either systemd or CUPS configuration.
Is there a reason we're doing the intersection instead of the union?

cryptomilk:
No, there is no reason, we also could do a union.

greghudson:
The steps skipped for socket activation fds are:

1. Create the socket
2. Verify the fd is in range for select()
3. Set SO_REUSEADDR
4. Disable dual-stack on IPv6 sockets
5. Bind the socket
6. listen()
7. Set non-blocking I/O (for non-RPC)
8. Set SO_LINGER with zero timeout
9. Turn on pktinfo for UDP sockets

Steps 1 and 5 should definitely be skipped.  I assume step 6 is taken care of by systemd, and steps 3 and 4 are probably unnecessary if systemd is binding the server sockets.  I'm not concerned with step 2 (it's perhaps obsolete as any modern system will use poll(), and it would be very unexpected for there to be socket activations with fds too high for select()).  However, steps 7, 8, and 9 seem like they could still be important.

greghudson:
I pushed some significant changes.  I don't have a test setup for socket activation, so it's possible that I broke something.  These changes reintroduce steps 7-9 for socket activation fds.

cryptomilk:
Looks like everything is still working:

```
Jan 03 14:06:39 krikkit krb5kdc[494891](info): Loaded
Jan 03 14:06:39 krikkit krb5kdc[494891](info): setting up network...
Jan 03 14:06:39 krikkit krb5kdc[494891](info): set up 1 sockets
Jan 03 14:06:39 krikkit krb5kdc[494892](info): creating 1 worker processes
Jan 03 14:06:39 krikkit krb5kdc[494892](info): closing down fd 3
Jan 03 14:06:39 krikkit krb5kdc[494894](info): commencing operation
Jan 03 14:06:39 krikkit krb5kdc[494894](info): AS_REQ (6 etypes {aes256-cts-hmac-sha384-192(20), aes128-cts-hmac-sha256-128(19), aes256-cts-hmac-sha1-96(18), aes128-cts-hmac-sha1-96(17), camellia256-cts-cmac(26), camellia128-cts-cmac(25)}) /run/localkdc/kdc.sock: NEEDED_PREAUTH: asn@KRIKKIT.LOCALKDC.SITE for krbtgt/KRIKKIT.LOCALKDC.SITE@KRIKKIT.LOCALKDC.SITE, Additional pre-authentication required
Jan 03 14:06:39 krikkit krb5kdc[494894](info): closing down fd 13
Jan 03 14:06:43 krikkit krb5kdc[494894](info): AS_REQ (6 etypes {aes256-cts-hmac-sha384-192(20), aes128-cts-hmac-sha256-128(19), aes256-cts-hmac-sha1-96(18), aes128-cts-hmac-sha1-96(17), camellia256-cts-cmac(26), camellia128-cts-cmac(25)}) /run/localkdc/kdc.sock: ISSUE: authtime 1735909603, etypes {rep=aes256-cts-hmac-sha384-192(20), tkt=aes256-cts-hmac-sha384-192(20), ses=aes256-cts-hmac-sha384-192(20)}, asn@KRIKKIT.LOCALKDC.SITE for krbtgt/KRIKKIT.LOCALKDC.SITE@KRIKKIT.LOCALKDC.SITE
Jan 03 14:06:43 krikkit krb5kdc[494894](info): closing down fd 13
```

src/include/socket-utils.h
```diff
@@ -159,4 +161,46 @@ sa_socklen(const struct sockaddr *sa)
...
 }
 
+/**
+ * @brief Compare if two IP addresses are equal.
+ *
+ * @param sa1 The first IP address to compare.
+ *
+ * @param sa2 The second IP address to compare.
+ *
+ * @return Return 0 if they are equal and non-zero if they aren't equal.
```
cryptomilk, line 171: I think this needs to be adapted to say: `Returns true when equal, false otherwise.`. The next sentence can be removed.

greghudson:
I felt like we can and should have a test case for this PR even if we can only simulate what a systemd unit file would do.  While doing research I found systemd-socket-activate(1), which is almost tailor-made for this purpose and should be installed on practically every Debian-based system (it's part of the systemd package).  There are unfortunately a couple of impedance points which may lead me to write my own socket activation simulator, even though that would further remove the test case from the systemd source base:

* It doesn't pass through environment variables unless you tell it to one variable at a time.  Right now I'm telling it everything in realm.env, which could lead to a very long argument list if the "make check" environment contains a lot of variables.  I think Linux supports big argument lists, but I'm not certain how big.

* It waits for a connection before executing the named daemon, just like systemd would.  For the purposes of exercising the net-server.c code, we don't need that, and it gets in the way of using it in k5test.py.  k5test.py wants a sentinel string so it knows when the backgrounded daemon process is ready to receive connections, and then doesn't read any more input from the daemon afterwards.  systemd-socket-activate prints:
```
Listening on <address1> as 3.
Listening on <address2> as 4.
...
Communication attempt on fd <n>.
Execing <kdcpath> (<kdcpath> -n)
```
The smaller problem is that if we want to listen on multiple addresses, there is no sentinel line after all of the listeners are set up, so we'd have to look for `Listening on <addressN>` where `<addressN>` is the systemd stringified form of the last address we want to pass in.  The bigger problem is that if `<kdcpath>` is long, the "Execing" line will plug the pipe buffer and the process will hang before it starts the KDC.

greghudson:
I'm still trying to figure out how socket activation should interact with kdc.conf listener configuration.  After my changes, we listen on all kdc.conf-configured listener addresses (using a socket activation fd if one matches, creating our own socket otherwise) and ignore any remaining socket activation fds.  I couldn't find a prescriptive systemd document about how this should work, so I'm resorting to reasoning about consequences and bits of prior art.

As I previously mentioned, https://0pointer.de/blog/projects/socket-activation2.html talks about making cupsd listen to the union of socket activation fds and configured interfaces.  To do that we would need a way to turn a socket address into a listener configuration.  For kadmind that is tricky as a TCP socket could be for either the kadmin or kpasswd protocol.  There is a defined way to associate names with sockets (LISTEN_FDNAMES), so we could make it work if we want to bear the complexity.

cupsd has a default listener configuration, but it's machine-local, whereas our default is network-wide.  So there is some risk that an admin might configure local socket activations and not realize that their KDC or kadmind is listening on the network as well.  We could consider suppressing default listeners that don't correspond to socket activation fds, while still honoring (creating our own sockets for) explicitly configured listeners.  That feels kind of awkward, so I'm not sure.

greghudson:
I pushed changes to:

* Ignore configured listeners that don't match caller-provided sockets, as suggested.
* Match on socket type as well as address.  Otherwise we can try to use a TCP socket for UDP listening or vice versa, which does not end well.
* Include a better-fitting test harness and use it in the tests.
* Add documentation.

cryptomilk:
review+

PR #1401, opened 2024-12-08
Add support for PKINIT on Windows

kenh:
These changes implement support for the PKINIT plugin on Windows platforms

This has been split into two parts: the aforementioned std::regex glue code for Windows platforms and the actual building of the PKINIT plugin.

The ```std::regex``` changes were as discussed previously, except that the question I posed on ```krbdev``` regarding specifying the ```-entry:DllMain``` option was never answered. My solution here is to remove this for building the krb5 DLLs; since the system default DLL entry point automatically calls ```DllMain``` if one is provided, it seems like this isn't necessary and things seem to work fine (you get unknown symbols when building the DLL otherwise). The use of ```-entry``` for building the CCAPI DLL and KFWLogin DLL has not been changed, but probably should be looked at by someone smarter than I.

Building the PKINIT plugin is mostly straightforward; there are some generic changes that need to be made, like converting the certificate loading routines that scan directories to the ```k5_dir_filenames``` interface. I also had to add a ```DllMain``` function to the pkinit plugin so it would call the plugin DLL initializer function; I decided not to use the routines that did that in ```win_glue.c``` as that seemed to be overkill for this. I also had to change a ```free()``` call to ```OPENSSL_free()``` as an exception was generated otherwise.

My current solution for the OpenSSL dependency is to require the setting of a variable ```OPENSSL_DIR``` when building Kerberos; that has to point to the root of a OpenSSL install tree; it is expected that include files are under ```%OPENSSL_DIR%\include``` and the import ```LIB``` files are under ```%OPENSSL_DIR%\lib```. If that variable isn't set then the plugin will not be built. This was tested with OpenSSL 3.4.0.

I haven't yet updated the build instructions or done anything about the installer; I wanted to get some feedback first on the direction of these changes. I tested this in our realm with a PKCS#11 library that interfaces with my work-issued smartcard; works fine.

I expect Leash will need some changes as it requires a password; if you give it a dummy password it will do PKINIT fine. That is also TBD.

Feedback is welcome.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,6 +70,8 @@ jobs:
         runs-on: windows-latest
         env:
             KRB_INSTALL_DIR: C:\kfw
+            OPENSSL_DIR: C:\Program Files\OpenSSL
+            OPENSSL_VERSION: 1_1
         steps:
             - name: Checkout repository
               uses: actions/checkout@v1
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -32,7 +32,8 @@ SUBDIRS=util include lib \
 	plugins/tls/k5tls \
 	kdc kadmin kprop clients appl tests \
 	config-files build-tools man doc @po@
-WINSUBDIRS=include util lib ccapi windows clients appl plugins\preauth\spake
+WINSUBDIRS=include util lib ccapi windows clients appl plugins\preauth\spake \
+	$(PKINIT_SUBDIR)
 BUILDTOP=$(REL).
 
 SRCS =  
@@ -106,6 +107,17 @@ config-windows: Makefile-windows
 #	cd ..
 
 #
+# Build the pkinit plugin if OpenSSL was configured
+#
+##DOS##!ifdef OPENSSL_DIR
+##DOS##PKINIT_SUBDIR=plugins\preauth\pkinit
+##DOS##PKINIT_MAKEFILE=$(PKINIT_SUBDIR)\Makefile
+##DOS##!else
+##DOS##PKINIT_SUBDIR=
+##DOS##PKINIT_MAKEFILE=
+##DOS##!endif
+
+#
 # We need outpre-dir explicitly in here because we may
 # try to build wconfig on a config-windows.
 #
@@ -154,7 +166,7 @@ WINMAKEFILES=Makefile \
 	windows\Makefile windows\lib\Makefile windows\ms2mit\Makefile \
 	windows\kfwlogon\Makefile windows\leashdll\Makefile \
 	windows\leash\Makefile windows\leash\htmlhelp\Makefile \
-	plugins\preauth\spake\Makefile
+	plugins\preauth\spake\Makefile $(PKINIT_MAKEFILE)
 
 ##DOS##Makefile-windows: $(MKFDEP) $(WINMAKEFILES)
 
@@ -277,6 +289,8 @@ WINMAKEFILES=Makefile \
 ##DOS##windows\leash\htmlhelp\Makefile: windows\leash\htmlhelp\Makefile.in $(MKFDEP)
 ##DOS##	$(WCONFIG) config < $@.in > $@
 ##DOS##plugins\preauth\spake\Makefile: plugins\preauth\spake\Makefile.in $(MKFDEP)
+##DOS##	$(WCONFIG) config < $@.in > $@
+##DOS##plugins\preauth\pkinit\Makefile: plugins\preauth\pkinit\Makefile.in $(MKFDEP)
 ##DOS##	$(WCONFIG) config < $@.in > $@
 
 clean-windows:: Makefile-windows
@@ -485,6 +499,10 @@ install-windows:
 	$(INSTALLDBGSYMS) clients\kswitch\$(OUTPRE)kswitch.pdb "$(KRB_INSTALL_DIR)\bin\."
 	copy plugins\preauth\spake\$(OUTPRE)$(SPAKELIB).dll "$(KRB_INSTALL_DIR)\bin\plugins\preauth\."
 	$(INSTALLDBGSYMS) plugins\preauth\spake\$(OUTPRE)$(SPAKELIB).pdb "$(KRB_INSTALL_DIR)\bin\plugins\preauth\."
+##DOS##!ifdef OPENSSL_DIR
+	copy plugins\preauth\pkinit\$(OUTPRE)$(PKINITLIB).dll "$(KRB_INSTALL_DIR)\bin\plugins\preauth\."
+	$(INSTALLDBGSYMS) plugins\preauth\pkinit\$(OUTPRE)$(PKINITLIB).pdb "$(KRB_INSTALL_DIR)\bin\plugins\preauth\."
+##DOS##!endif
 
 check-prerecurse: runenv.py
 	$(RM) $(SKIPTESTS)
--- a/src/config/win-pre.in
+++ b/src/config/win-pre.in
@@ -117,7 +117,7 @@ KFWFLAGS=-DUSE_LEASH=1 -DDEBUG -D_CRTDBG_MAP_ALLOC
 CC=cl
 
 PDB_OPTS=-Fd$(OUTPRE)\ -FD
-CPPFLAGS=-I$(top_srcdir)\include -I$(top_srcdir)\include\krb5 $(DNSFLAGS) -DWIN32_LEAN_AND_MEAN -DKRB5_DEPRECATED=1 -DKRB5_PRIVATE -D_CRT_SECURE_NO_DEPRECATE $(KFWFLAGS) $(TIME_T_FLAGS)
+CPPFLAGS=-I$(top_srcdir)\include -I$(top_srcdir)\include\krb5 $(DNSFLAGS) -DWIN32_LEAN_AND_MEAN -DKRB5_DEPRECATED=1 -DKRB5_PRIVATE -D_CRT_SECURE_NO_DEPRECATE $(KFWFLAGS) $(TIME_T_FLAGS) $(OSSLINCLUDE)
 # Treat the following warnings as errors:
 #  4020: too many actual parameters
 #  4024: different types for formal and actual parameter
@@ -188,6 +188,16 @@ SLIB=$(BUILDTOP)\lib\$(OUTPRE)k5sprt$(BITS).lib
 GLIB=$(BUILDTOP)\lib\$(OUTPRE)gssapi$(BITS).lib
 CCLIB=krbcc$(BITS)
 SPAKELIB=spake$(BITS)
+
+!ifdef OPENSSL_DIR
+OSSLLIB="$(OPENSSL_DIR)\lib\libcrypto.lib"
+OSSLINC="-I$(OPENSSL_DIR)\include"
+PKINITLIB=pkinit$(BITS)
+!else
+OSSLLIB=
+OSSLINC=
+PKINITLIB=
+!endif
 
 KRB4_INCLUDES=-I$(BUILDTOP)/include/kerberosIV
 
--- a/src/include/win-mac.h
+++ b/src/include/win-mac.h
@@ -236,4 +236,8 @@ HINSTANCE get_lib_instance(void);
 #define KRB5_CALLCONV_C
 #endif
 
+#ifndef PKCS11_MODNAME
+#define PKCS11_MODNAME "C:\\Program Files\\OpenSC Project\\OpenSC\\pkcs11\\opensc-pkcs11.dll"
+#endif
+
 #endif /* _KRB5_WIN_MAC_H */
--- a/src/lib/krb5_32.def
+++ b/src/lib/krb5_32.def
@@ -510,3 +510,10 @@ EXPORTS
 	k5_sname_compare				@474 ; PRIVATE GSSAPI
 	krb5_kdc_sign_ticket                            @475 ;
 	krb5_kdc_verify_ticket                          @476 ;
+
+; new in 1.22
+; private symbols used by PKINIT module
+	encode_krb5_sp80056a_other_info			@477 ; PRIVATE
+	encode_krb5_pkinit_supp_pub_info		@478 ; PRIVATE
+	krb5int_copy_data_contents			@479 ; PRIVATE
+	krb5_free_pa_data				@480 ; PRIVATE
--- a/src/plugins/preauth/pkinit/Makefile.in
+++ b/src/plugins/preauth/pkinit/Makefile.in
@@ -12,6 +12,9 @@ SHLIB_EXPDEPS = \
 	$(TOPLIBD)/libkrb5$(SHLIBEXT)
 SHLIB_EXPLIBS= -lkrb5 $(COM_ERR_LIB) -lk5crypto -lcrypto $(DL_LIB) $(SUPPORT_LIB) $(LIBS)
 
+WINLIBS = $(KLIB) $(SLIB) $(PLIB) $(CLIB) $(OSSLLIB)
+OSSLINCLUDE = $(OSSLINC)
+
 STLIBOBJS= \
 	pkinit_accessor.o \
 	pkinit_srv.o \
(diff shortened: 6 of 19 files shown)


greghudson:
I don't know of any reason why we specifically wanted `-entry:DllMain` in the linker options.  The option first appeared in 1997 (commit 441763d0b69e15def56d6141345e0eac7fe580de) with the message "win_glue.c: win32 dll entry/exit code added".  The "We should probably use DllMainCRTStartup" comment was added in 1999 by a different developer, suggesting to me that the option was wrongly added but that the person who observed that didn't feel empowered to fix it.

In the mail you said that removing the `-entry` option led to a kinit failure loading the 
service_locator plugin.  I don't have any insight into that; did you make progress on it?  I didn't see anything related in these commits.

For the first commit:

* I like the idea of storing the error into a fixed-length buffer in regex_t more than I like the idea of having our own errcode-to-string translation in k5_regerror().  Alternatively, we could add a free function to be used after regcomp() failure in our tree, which would be a no-op under POSIX (since regfree() after a regcomp() failure appears to be undefined) and would free the allocated error string on Windows.
* Remember to get rid of the XXX comment in win-pre.in.

I didn't see any red flags in either commit.  My biggest questions are what we will recommend in windows/README, what we will do in the Github Actions build script, and what we will do in the installer.  From what I can find, the most popular source of premade OpenSSL Windows builds is Shining Light Productions, a one-person software house.  Chocolatey and winget both pull from there, as does the Github Actions runner.  From the Chocolatey package comments, we might expect some periods of skew between package manager metadata and the versions available for download.  The Github Actions runner software list says the Windows Server 2022 image includes OpenSSL 1.1.1w, though Shining Light doesn't appear to offer 1.1.1w.  We could recommend people do their own OpenSSL build from source; I am okay with this option for the build instructions, but I don't like it for the Github Actions script as it would blow up the time it takes to run CI on pull requests.

kenh:
> I don't know of any reason why we specifically wanted `-entry:DllMain` in the linker options. The option first appeared in 1997 (commit [441763d](https://github.com/krb5/krb5/commit/441763d0b69e15def56d6141345e0eac7fe580de)) with the message "win_glue.c: win32 dll entry/exit code added". The "We should probably use DllMainCRTStartup" comment was added in 1999 by a different developer, suggesting to me that the option was wrongly added but that the person who observed that didn't feel empowered to fix it.

Fair enough; that suggests to me that -entry should be removed everywhere. If you want me to do that I can, otherwise we can defer that.

MOSTLY unrelated to that ... we have received information that when Windows is configured with all relevant DOD security settings our Kerberos kit stops working with a credential cache error. I realize this is super vague and not helpful; for various complicated reasons the information flow between users and developers in our program has broken down, but new personnel are being hired specifically to do Windows development so hopefully this will be resolved. The reason I mention that is according to the documentation I read part of the things that the system default DLL entry point does is set up permissions, and maybe the use of custom DLL entry points is preventing some needed permissions from being set up properly? That is just a wild guess and I know we need a developer to dig into that issue; I just wanted to mention that as possible motivation for cleaning that up.

> In the mail you said that removing the `-entry` option led to a kinit failure loading the service_locator plugin. I don't have any insight into that; did you make progress on it? I didn't see anything related in these commits.

Oh, yes, that was resolved by #1399 (it turned out that none of my changes caused that at all).

>     * I like the idea of storing the error into a fixed-length buffer in regex_t more than I like the idea of having our own errcode-to-string translation in k5_regerror().  Alternatively, we could add a free function to be used after regcomp() failure in our tree, which would be a no-op under POSIX (since regfree() after a regcomp() failure appears to be undefined) and would free the allocated error string on Windows.

A fixed-length buffer is fine with me and it strikes me as the most robust option that doesn't require any code changes to every caller of regex functions; I'll make that change. BTW, I forgot to mention that my Windows PKINIT test did have ```pkinit_cert_matching``` lines so the regex glue code was actually tested on Windows.

> 
>     * Remember to get rid of the XXX comment in win-pre.in.

Got it.

> 
> I didn't see any red flags in either commit. My biggest questions are what we will recommend in windows/README, what we will do in the Github Actions build script, and what we will do in the installer.

FWIW, building OpenSSL on Windows is actually very easy. The only issue I ran into is that the Perl included with Git for Windows does not work with the OpenSSL ```Configure``` script; Strawberry Perl works fine.

If you want to do 32 and 64 bit versions of Kerberos with PKINIT support you need to build both 32 and 64 bit versions of OpenSSL and set ```OPENSSL_DIR``` appropriately during each build run; I admit I only tested 64 bit. I also admit it is not clear to me how useful 32 bit builds are in 2024.

I can add some stuff to ```windows/README``` if you want on the next commit.

For the installer I guess it depends if you want to make OpenSSL a hard or soft requirement.  I believe the OpenSSL DLL has a version string in the name so it might be a bit complicated to include the filename automatically in the installer build process (I did not do any installer work yet). It does occur to me that having a KfW installer that include PKINIT support sure would be useful to a lot of people.

>From what I can find, the most popular source of premade OpenSSL Windows builds is Shining Light Productions, a one-person software house.

Yeah, I also got the same vibe from them as well. I know that for us, we're going to build OpenSSL locally.  If the GitHub Actions runner says it includes OpenSSL, I think it makes the most sense to use that.

greghudson:
> MOSTLY unrelated to that ... we have received information that when Windows is configured with all relevant DOD security settings our Kerberos kit stops working with a credential cache error.

My impression is that if you turn on Credential Guard, user-space processes can't use the LSA any more.  So KfW would continue working with CCAPI or FILE ccaches, but not the native ccache.  Maybe this is it, or maybe it's something else.
If I'm right, the only good options we have are to improve error messages and documentation.  We could conceivably make our GSSAPI krb5 mech act as a stub around SSPI when the LSA is used, but none of our usual libkrb5 behaviors would be operative, and I'm not sure why someone would want to use our implementation at that point.

kenh:
> My impression is that if you turn on Credential Guard, user-space processes can't use the LSA any more. So KfW would continue working with CCAPI or FILE ccaches, but not the native ccache. Maybe this is it, or maybe it's something else.

I'm. not getting wonderful bug reports (right now it's all second or third hand reports), but we exclusively use CCAPI in our deployment so I don't think that's the issue. Hopefully we can get some more detailed information later.

kenh:
Just FYI, I pushed new commits that had the recommended changes to the regex layer and I made some sample changes to ```windows/README``` about building with OpenSSL.

kenh:
I rebased ```pkinit-window``` to ```master``` and did a test to make sure that ```pkinit``` still works (it does).

greghudson:
I pushed changes to the first commit.  I hope I didn't break anything, but I haven't set up a Windows VM to test this yet.  We will need copyright/license statements in k5-regex.h and k5-regex.cpp.

kenh:
I added appropriate copyrights.  However, build fails with this error:
```
k5sprt64.lib(k5-regex.obj) : error LNK2019: unresolved external symbol "unsigned __int64 __cdecl krb5int_strlcpy(char *,char const *,unsigned __int64)" (?krb5int_strlcpy@@YA_KPEADPEBD_K@Z) referenced in function k5_regerror
  Hint on symbols that are defined and could potentially match:
    krb5int_strlcpy
```
I'm a little puzzled by that ... are we missing an ```extern "C"``` somewhere?

kenh:
Ah, yes, that was it. I put an ```extern "C" { ... }``` around the include of ```k5-platform.h``` and that resolved it. I don't know if your preference is to put that in ```k5-platform.h``` or leave it in ```k5-regex.cpp```.

greghudson:
That's an interesting question.  Our public headers are of course responsible for being usable from C++.  For our private headers like k5-platform.h it's less clear.  A few of our existing private headers do, but most don't.  Most of our C++ files are inside ccapi/ and windows/; a couple of them do include headers inside extern "C" blocks, but most don't have a need to.

So it's kind of a coin flip, but I will vote for making k5-platform.h usable from C++, especially since there are some __cplusplus conditionals within the header.  A basic whole-file guard should suffice, like we have in include/iprop.h.

kenh:
Done. Builds now (and ```pkinit``` works)

greghudson:
I was going to merge the first commit to get it out of the way, but I wound up making a few more changes than I had planned (chiefly, renaming k5-regex.cpp to regex.cpp for consistency with other libkrb5support files) and also the clang CI jobs failed, probably due to a runner change and not anything in this PR.  So I modified the PR branch for now, and will have another go at merging that commit tomorrow.

I have separately been working on a PR to remove 32-bit builds from the Windows CI, installer, and README.

kenh:
Sounds good to me, but ... did you have to remove the vi mode lines? Ouch. I mean, I'll live; I just was using those to more easily conform to the MIT indent style.

I'm obviously fine with removing the 32-bit builds; I'm not even sure there is x86 hardware that can run a modern Windows release. It looks like we haven't been distributing 32 bit DLLs or executables for a while.

greghudson:
I'm happy to have those in during the iteration on a PR (so I only took them out of the files in the first commit, shortly prior to merging it), but they aren't part of the standard boilerplate.

greghudson:
I've merged the regex commit, and I made some small changes to the PKINIT commit.

I'm not sure what I think about setting PKCS11_MODNAME to the OpenSC typical installation path.  On other platforms the default PKCS11_MODNAME relies on the system library search path.  Windows has a search process for DLL filenames[1], and it includes PATH as well as the system and Windows folders; I don't know if a system with OpenSC installed is likely to find the DLL in one of those places.

[1] https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-unpackaged-apps

greghudson:
> I will confess that I do not know what API sets, SxS manifest redirection, the loaded-module list, or known DLLs are

API sets seem unlikely to apply.  SxS manifests aren't helpful either (it's metadata we could supply giving the location of named DLLs, but if we knew that we could just pass the correct path).  The loaded-module list would only apply if the process already had the OpenSC DLL loaded.

Known DLLs are a possibility.  They are configured in the registry (HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\KnownDLLs) and it's possible that OpenSC adds itself there.

> I think that it's reasonable for the defaults to work out of the box with OpenSC as I understand that's the intention on other platforms

That is the intent.  It would just be unfortunate if we made the default work only when OpenSC is installed in its typical location, if using only the name would work regardless of its install path.

> (although in my experience that works very rarely because the PKCS11 library is not in the default search path on other platforms).

Debian/Ubuntu's opensc-pkcs11 package installs /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so ; I'd be surprised if that wasn't found via the default search path.

kenh:
> Debian/Ubuntu's opensc-pkcs11 package installs /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so ; I'd be surprised if that wasn't found via the default search path.

Oh RHEL-based systems (which is the most common in our environment) opensc installs the PKCS11 module in ```${libdir}/pkcs11/opensc-pkcs11.so``` and on MacOS X it seems like those libraries are also in a non-default search path location.

kenh:
> That is the intent. It would just be unfortunate if we made the default work only when OpenSC is installed in its typical location, if using only the name would work regardless of its install path.

I installed OpenSC on my system with the test build on it using the default installer.  First off, it seems if you don't specify ANY kind of ```X509_user_identity``` at all either on the command line or in the config file, PKINIT gives up very early. It looks like you only get the default ```PKCS11_MODNAME``` if you just specify ```PKCS11:``` without a residual; I am not sure how many people actually do that! I understand the complex balance between wanting PKINIT to work with minimal configuration versus avoiding PKINIT failures when it isn't configured; I don't have a good answer to what should be done here. I just want to mention that I don't think in practice ```PKCS11_MODNAME``` is used all that often.

However:

```
pkinit -X X509_user_identity=PKCS11: username
```

works (with the current default explicit pathname for the ```opensc-pkcs11.dll```)

```
pkinit -X X509_user_identity=PKCS11:opensc-pkcs11.dll username
```

Does NOT work (cannot find the DLL).

greghudson:
Thanks for the testing.  Given that, we can leave the default as it is in the PR.

Red Hat asked for the default module name to be configurable at build time, so that they could change it to p11-kit in their packages.  I agree that this is of limited utility if you have to tell PKINIT to use PKCS11 in the first place.

There's a reasonable case for using PKCS11 with zero client configuration if (1) a PKCS11 token is available with no PIN, (2) it contains a proper PKINIT anchor cert (not an SSL server cert that we'd be willing to use with EKU checking turned off), and (3) it contains a single proper PKINIT client cert.  But I don't think a lot of environments would consider it very secure to regularly authenticate with PIN-less PKINIT tokens.  If a token does have a PIN, I don't think we can typically check whether (2) and (3) are true before prompting the user for the PIN, and that would be a bad user experience if the PKCS11 token isn't there for Kerberos.

kenh:
> There's a reasonable case for using PKCS11 with zero client configuration if (1) a PKCS11 token is available with no PIN, (2) it contains a proper PKINIT anchor cert (not an SSL server cert that we'd be willing to use with EKU checking turned off), and (3) it contains a single proper PKINIT client cert. But I don't think a lot of environments would consider it very secure to regularly authenticate with PIN-less PKINIT tokens. If a token does have a PIN, I don't think we can typically check whether (2) and (3) are true before prompting the user for the PIN, and that would be a bad user experience if the PKCS11 token isn't there for Kerberos.

It's not clear to me that there is even a way to indicate via the PKCS11 API consistently that a no-PIN key is available; I am aware of the ```CKF_LOGIN_REQUIRED``` flag but even that doesn't quite cover all of the corner cases. In the case of PIV tokens (the ones I am familiar with) one of the slots on the card is specifically designated as 'no PIN required' (the idea is that it is used for things like building access). There doesn't seem to be a way in PKCS#11 to say that a PIN is required for some slots but not others and typically all keys on a card are presented as multiple objects in a single PKCS#11 slot. I've also never seen a PIV card with anything other than just the client certificates on them (no anchor certificates). I am certainly willing to believe such tokens exist, I just have no idea how common they are. I could see cases where maybe you'd want to use a PIN-less token for some kind of PKINIT access, but like you said I think the zero-configuration use cases are extremely rare to nonexistent.

greghudson:
I felt like we can and should have a test case for this PR even if we can only simulate what a systemd unit file would do.  While doing research I found systemd-socket-activate(1), which is almost tailor-made for this purpose and should be installed on practically every Debian-based system (it's part of the systemd package).  There are unfortunately a couple of impedance points which may lead me to write my own socket activation simulator, even though that would further remove the test case from the systemd source base:

* It doesn't pass through environment variables unless you tell it to one variable at a time.  Right now I'm telling it everything in realm.env, which could lead to a very long argument list if the "make check" environment contains a lot of variables.  I think Linux supports big argument lists, but I'm not certain how big.

* It waits for a connection before executing the named daemon, just like systemd would.  For the purposes of exercising the net-server.c code, we don't need that, and it gets in the way of using it in k5test.py.  k5test.py wants a sentinel string so it knows when the backgrounded daemon process is ready to receive connections, and then doesn't read any more input from the daemon afterwards.  systemd-socket-activate prints:
```
Listening on <address1> as 3.
Listening on <address2> as 4.
Communication attempt on fd <n>.
Execing <kdcpath> (<kdcpath> -n)
...
The smaller problem is that if we want to listen on multiple addresses, there is no sentinel line after all of the listeners are set up, so we'd have to look for "Listening on <addressN>" where <addressN> is the systemd stringified form of the last address we want to pass in.  The bigger problem is that if <path> is long, the "Execing" line will overflow the pipe buffer and the process will hang before it starts the KDC.

kenh:
Was this comment intended for #1400? I'm fine with doing a test, but I am not sure there are any extant tests on Windows.

greghudson:
I haven't tested the resulting installer (I didn't have an easy time grabbing suitable OpenSSL binaries for my dev VM), but these changes at least pass the CI.

kenh:
I'll try testing the installer, either today or tomorrow. FWIW, building OpenSSL from source was very straightforward.

greghudson:
I was able to do a build in a VM after building OpenSSL from source, and test the installer.  I verified that the PKINIT and libcrypto DLLs are installed (and fixed the name of the libcrypto DLL just now), and made some README changes to account for the installer integration.  I did not actually test PKINIT in the installed binaries, but I think I'm ready to merge this if you can do that.

kenh:
Okay! I built the installer and installed it in a clean VM snapshot and once I added the necessary configuration entries to krb5.ini (and the certificates) I was able to PKINIT successfully, using both the command line ```kinit``` and the GUI tool. One note: I had to install the "C++ 2022 Redistributable MSMs" to build the Installer, which I guess is not the default?

I used the latest Wix3 to build the installer, I guess that's Wix 3.14.1. I don't know if the latest Wix (version 5) would work or not.

greghudson:
I may have conflated the merge module component with the component for the MFC libraries themselves.  I will go through a Visual Studio installation and update the instructions.

kenh:
> (I could also give Powershell instructions, a la https://github.com/krb5/krbdev-services/blob/master/kfw/kfwsetup.ps1 . Not sure if that would be easier or harder for people to use.)

I am not sure we would use it since we have kind of weird requirements, but I can certainly see how it would be easier for some people. I say include it if it's not too much effort.

PR #1403, opened 2024-12-12
Remove explicit use of -entry flag for DLLs on Windows

kenh:
Remove the explicit use of the -entry flag to specify a DLL entry point on Windows; this is no longer recommended.  The default Windows entry point will call a DllMain function if one is provided; make sure we provide when necessary.

### DIFF
--- a/src/ccapi/lib/win/Makefile.in
+++ b/src/ccapi/lib/win/Makefile.in
@@ -95,7 +95,7 @@ $(CCLIBRES): $(VERSIONRC)
 	$(RC) $(RCFLAGS) -DCCAPI_LIB -fo $@ -r $**
 
 $(OUTPRE)$(CCLIB).dll: $(OBJS) $(CCLIB).def $(CCLIBRES)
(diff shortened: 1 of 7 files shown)


kenh:
Note that my presumption is pkinit-windows is going to be merged eventually; this just removes the OTHER uses of ```-entry```.

greghudson:
My preference is to add in the win-pre.in change and merge this first.

kenh:
Do you mean the changes in ```win-pre.in``` from #1401? If so, I can do that and then rebase #1401 after this is merged.

greghudson:
Yes.  (Not all of the changes to win-pre.in in that PR, of course, just the one removing the -entry flag.)

kenh:
Done

Arvin9433:
Okey

kenh:
Did you mean to force-push those changes? It seems like those are some of the ones from the Unix domain socket work, unless I am misunderstanding what git is telling me (which is always possible).

greghudson:
That was just a rebase onto head, with no changes.

PR #1404, opened 2024-12-21
WIP: Fix CI errors on Ubuntu 24.04

greghudson:
Github is transitioning the ubuntu-latest image from 22.04 to 24.04, which means our CI jobs sometimes run on one and sometimes on the other.

Most of the changes here are simple fixes to problems that previous escaped the various nets we've cast around them.  I am currently a bit stymied by an asan problem.  When asan has something to report, it tries to use llvm-symbolizer and fails with an error like:

```
/usr/bin/llvm-symbolizer-18: symbol lookup error: /home/runner/work/krb5/krb5/src/./lib/libkrb5support.so.0: undefined symbol: __asan_option_detect_stack_use_after_return
5131
==29597==WARNING: Can't read from symbolizer at fd 3
5132
```

In some cases this appears to cause the program to exit with status -13 (SIGPIPE).  Usually this is just a minor hindrance to analyzing a problem we want to fix anyway (and can debug locally with valgrind).  But in some cases, we intentionally cause a test program to exit early with an error.  In this cases, asan reports memory leaks even though the "leaked" memory is still accessible from the call stack leading up to the exit() call (valgrind does as well).  Currently that doesn't cause a `make check` failure because the test program's exit status is 1 with or without an asan leak report.  But if the exit status changes to -13 due to the symbolizer problem, the Python script doesn't see the status it expects and `make check` fails.

Possibly all of those tests should be restructured so that no leaks are reported (and perhaps so that any leaks that are reported cause `make check` failures).  But the symbolizer issue is annoying regardless.  We could run the asan build on ubuntu-22.04 instead of ubuntu-latest, which would kick the can down the road about two years, by which time the symbolizer issue might have been fixed by other people.  But I didn't find any useful search results for the error message, suggesting that the problem might be specific to our use of asan, in which case it's less likely to get fixed if we don't track it down ourselves.

The changes to shlib.conf are an unsuccessful attempt to fix the symbolizer issue (by passing `-fsanitize=address` when building shared libraries) and will likely not be retained.  Likewise the change to ignore SIGPIPE in t_credstore.c.

The changes to build.yml are just for the PR branch, and won't be merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,8 +13,13 @@ jobs:
             matrix:
                 name: [linux-clang, linux-clang-openssl, linux-gcc]
                 include:
+                    # llvm-symbolizer fails on Ubuntu 24.04 until we
+                    # can avoid using LD_LIBRARY_PATH in the test
+                    # suite.  This causes test suite failures due to
+                    # SIGPIPE exits from test programs.  See
+                    # https://github.com/llvm/llvm-project/issues/120915
                     - name: linux-clang
-                      os: ubuntu-latest
+                      os: ubuntu-22.04
                       compiler: clang
                       makevars: CPPFLAGS=-Werror
                       configureopts: --enable-asan
--- a/src/lib/krb5/krb/x-deltat.y
+++ b/src/lib/krb5/krb/x-deltat.y
@@ -44,6 +44,7 @@
 #ifdef __GNUC__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wuninitialized"
+#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
 #endif
 
 #include "k5-int.h"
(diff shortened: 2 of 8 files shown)


greghudson:
In 24.04, llvm-symbolizer has a dependency on libcurl, as part of an LLVM feature that allows debug symbols to be downloaded at runtime.  libcurl depends on libgssapi_krb5.  So in the test environment for an asan build, when llvm-symbolizer gets run, the dynamic linker sees a bunch of asan-compiled krb5 libraries in LD_LIBRARY_PATH, tries to link them in, and fails because the binary (a non-asan-compiled system program) doesn't have the asan symbols.

This seems tough to work around, and it isn't likely to get spontaneously fixed since it only affects the developers of a couple of dozen libraries.

greghudson:
I filed https://github.com/llvm/llvm-project/issues/120915 but it doesn't seem like an easy upstream fix.  They might be able to fix the SIGPIPE problem without too much difficulty, which would help a bit.

For this PR I will pin the clang asan job to Ubuntu 22.04.  For the longer term I think we are looking at avoiding LD_LIBRARY_PATH in the test suite.  This will require building a second version of each shared library and binary with its runpath pointed at the build tree.

PR #1406, opened 2024-12-23
Work around llvm-symbolizer LD_LIBRARY_PATH issue

greghudson:
On some platforms llvm-symbolizer is linked against libcurl, which depends on the krb5 libraries.  When the test suite runs programs with LD_LIBRARY_PATH pointed at asan-compiled krb5 libraries, llvm-symbolizer will encounter a dynamic linker error at startup, interfering with the display of stack traces.

In k5test.py, when the build enables asan, wrap the platform's symbolizer in a script that unsets LD_LIBRARY_PATH (or similar) and then runs the real llvm-symbolizer.

[I thought of this workaround for the symbolizer problem described in PR #1404, and it appears to work without requiring major changes to the build system.  Unfortunately, fixing it revealed another detected leak in t_kdb.py.  I'm not sure this leak is ours, since the allocations occur inside a call to ldap_sasl_interactive_bind_s() and it does not appear that the entire handle is leaked.  But we'll have to address it in some fashion.]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,13 +13,8 @@ jobs:
             matrix:
                 name: [linux-clang, linux-clang-openssl, linux-gcc]
                 include:
-                    # llvm-symbolizer fails on Ubuntu 24.04 until we
-                    # can avoid using LD_LIBRARY_PATH in the test
-                    # suite.  This causes test suite failures due to
-                    # SIGPIPE exits from test programs.  See
-                    # https://github.com/llvm/llvm-project/issues/120915
                     - name: linux-clang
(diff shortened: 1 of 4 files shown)


greghudson:
I tracked down the memory leak to a Debian patch in the cyrus-sasl2 package.  I filed https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1091694 and added a commit to work around the problem.  I will see if I can think of any way to more narrowly scope the workaround, as it could prevent detection of memory leaks in parts of our own code, and as there isn't an obvious trigger for undoing the workaround in the future.

PR #1411, opened 2025-02-25
Add PKINIT paChecksum2 from MS-PKCA v20230920

jrisc:
In 2023, Microsoft [updated MS-PKCA](https://winprotocoldoc.z19.web.core.windows.net/MS-PKCA/%5bMS-PKCA%5d-230920-diff.pdf) to add the optional `paChecksum2` element in the `PKAuthenticator` sequence. This checksum accepts SHA-1, SHA-256, SHA-384, and SHA-512 digests.

In Windows Server 2025, this checksum becomes mandatory when using PKINIT with FFDH (but strangely not with ECDH if SHA-1 is configured as allowed).

When failing due to the absence of `paChecksum2`, the `KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED` error code is returned by the AD KDC.

This topic was discussed with Microsoft's DocHelp in the Samba mailing list:
* [`paChecksum2` ASN.1 definition](https://lists.samba.org/archive/cifs-protocol/2025-January/004415.html)
* [`paChecksum2` position in PKAuthenticator](https://lists.samba.org/archive/cifs-protocol/2025-January/004417.html)
* [ECDH exception](https://lists.samba.org/archive/cifs-protocol/2025-January/004423.html)
* [List of supported OIDs](https://lists.samba.org/archive/cifs-protocol/2025-February/004432.html)

For the record, I submitted a [PR to support `paChecksum2` decoding in Wireshark](https://gitlab.com/wireshark/wireshark/-/merge_requests/18834) (merged).

I tested the present PR against Windows Server 2025.

There is still a decision to be made about retro-compatibility:
1. Should `paChecksum2` be generated by default (which would require updating all previous versions to, at least, recognize this element)?
2. If not, should the client send a first request without `paChecksum2`, and retry with `paChecksum2` if `KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED` is returned (at the cost of generating more traffic for AD realms and have the least safe option as default)?

A compromise could be to try `paChecksum2` by default, but falling back to `paChecksum` only if it fails, but I am not sure the returned error code will be the same for all Kerberos implementations in this case.


greghudson:
> Should paChecksum2 be generated by default (which would require updating all previous versions to, at least, recognize this element)?

PKAuthenticator is an extensible ASN.1 sequence, meaning that optional fields can be added to it without breaking decoding.  Extensibility is indicated by the "..." at the end of the sequence specification.  (We treat all ASN.1 sequences as extensible, but this one is actually standardized as such, unlike the sequences in RFC 4120.)  The freshness token was added in this manner.  Change control of this sequence is supposed to be in the hands of the IETF, not Microsoft, but I guess that ship has sailed.

I noticed a pkinit_pachecksum2_copy() that was added and didn't seem to be used; is that a relic of a previous iteration?

jrisc:
> PKAuthenticator is an extensible ASN.1 sequence, meaning that optional fields can be added to it without breaking decoding.

Great, this will avoid the cross-version compatibility problem.

> I noticed a pkinit_pachecksum2_copy() that was added and didn't seem to be used; is that a relic of a previous iteration?

Yes, my bad. I was initially using it to copy the checksum in `auth_pack` in `pkinit_as_req_create()` before I realize this is a local variable only used for encoding. So just casting the `paCheksum2` pointer argument as mutable structure was enough.

I will remove `pkinit_pachecksum2_copy()` and rebase the PR.

greghudson:
I did a big refactor; this will need to be tested against Windows again.

jrisc:
I confirm the PR is still working against Windows Server 2025 after refactoring.

I see that AD returns `KRB_AP_ERR_MODIFIED` in case the verification of paChecksum or paChecksum2 fails. This is likely because, as you mentioned in one of your comments, RFC4556 does not mention which error code should be returned in this case. Should we stick to the behaviour of Windows Server 2025 in this case (at least for paChecksum2)?

I think that returning `KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED` in case paChecksum2 is missing, and `KRB_AP_ERR_MODIFIED` in case its value is invalid is a better approach. In the future, if the MIT KDC starts enforcing the presence of paChecksum2, old clients might be lacking support for this element. Thus the `KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED` error would be a clear indicator of the problem (since I suppose all implementations of the PKINIT client already implement SHA-1 paChecksum). This would simplify troubleshooting, because `KRB_AP_ERR_MODIFIED` is a more generic error in comparison.

I can ask Microsoft's DocHelp to add the error code to be returned in this case in the next MS-PKCA version to make sure this remains a coherent behaviour across the different implementations.

greghudson:
Returning KRB_AP_ERR_MODIFIED is fine.  I will change it.

Some background on KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED: RFC 4556 makes paChecksum optional in the ASN.1 pkAuthenticator sequence, anticipating that a future protocol revision might replace it.  Implementations are supposed to return KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED if paChecksum is missing, so that these hypothetical future clients can know to retry with paChecksum instead of the hypothetical future protocol element.  Our implementation does _not_ make paChecksum optional; if a client omits it, our KDC will see a decoding error from k5int_decode_krb5_auth_pack() and  will probably send back something unhelpful (GENERIC or PREAUTH_FAILED or PREAUTH_REQUIRED, not sure which).

jrisc:
I think you've replaced `KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED` by `KRB_AP_ERR_MODIFIED` in case paChecksum(2) is missing or invalid. But what I meant is we should differentiate both cases:
* If paChecksum2 is missing: `KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED`
* If paChecksum2 is invalid: `KRB_AP_ERR_MODIFIED`

greghudson:
If paChecksum2 is missing, the current PR code just doesn't verify it (no error).

If paChecksum is missing, the current code will fail to decode the PKAuthenticator, and the PR doesn't change that.

I only changed KDC_ERR_PA_CHECKSUM_MUST_BE_INCLUDED to KRB_AP_ERR_MODIFIED in cases where one of the checksums is present but invalid.

jrisc:
My bad, I missed the `if (cksum2 == NULL)` test in `crypto_verify_checksums()` somehow.

I'm okay with the PR as it is now.

jrisc:
I asked MS DocHelp to include error codes in MS-PKCA:
https://lists.samba.org/archive/cifs-protocol/2025-March/004438.html

greghudson:
The commit message says "this checksum becomes mandatory when using PKINIT with FFDH".  Are there are any other conditionals on that, like a SHA-256 checksum on the CMS messages?

jrisc:
My understanding is that the AD `Computer Configuration\Policies\Administrative Templates\System\KDC\Configure hash algorithms for certificate logon` group policy controls hash function requirement for `paChecksum2`, the CMS data digest and signature (or rather MAC I suppose, since a shared key obtained by DH is always used since we removed the "RSA" method), and the signature of the user certificate used for authentication.

We are coherent regarding `paChecksum2` and CMS data digest+MAC, because we use SHA-256 for both. But afaik there is no setting for requiring a minimum hash function for `paChecksum2`, CMS data digest+MAC, or user certificate signature.

If compiled with OpenSSL, access to algorithms using the EVP interface is controlled by OpenSSL system configuration though. As an example, on CentOS Stream/RHEL 9 by default it is not possible to use user certificates signed using SHA-1 or process CMS data digested using SHA-1.

PR #1412, opened 2025-03-02
Use k5_path_join() in krb5int_open_plugin_dirs()

kenh:
Use k5_path_join() to create pathnames in the function krb5int_open_plugin_dirs() for Windows compatibility.

### DIFF
--- a/src/util/support/plugins.c
+++ b/src/util/support/plugins.c
@@ -403,12 +403,8 @@ krb5int_open_plugin_dirs (const char * const *dirnames,
                 struct plugin_file_handle *handle = NULL;
                 char *filepath = NULL;
 
-                if (!err) {
-                    if (asprintf(&filepath, "%s/%s", dirnames[i], filenames[j]) < 0) {
-                        filepath = NULL;
-                        err = ENOMEM;
-                    }
-                }
+                if (!err)
+                    err = k5_path_join(dirnames[i], filenames[j], &filepath);
 
                 if (!err && krb5int_open_plugin(filepath, &handle, ep) == 0) {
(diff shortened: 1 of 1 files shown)


greghudson:
The change seems fine in concept (though the code can be further simplified), but I'd like to understand the motivation.  It looks like krb5int_open_plugin_dirs() is used by kdb5.c, authdata.c, and locate_kdc.c.  Are we trying to get authdata or locate modules working on Windows?

kenh:
I was thinking of future proofing and some local code cleanup.

I was going through a set of local changes (these weren't done by me) and this was included. But as you note, ```krb5int_open_plugin_dirs()``` isn't actually used by anything currently on Windows (although I could maybe see some use for a locate plugin in the future, maybe?) I am not actually sure WHY this change was made; digging back through our history shows that was done as a commit titled "Enable HPCMP client pkinit plugin to build on Win", but clearly that wasn't really needed. So part of my motivation was to eliminate one more change that we had that MIT Kerberos did not.

At the end of the day I would argue that the rest of the plugin loading code is very careful to use ```k5_path_join()``` where appropriate and I can't think of a reason why this code shouldn't also use it.

greghudson:
Thanks for the context.  If the changes were made long enough ago, the client preauth interface might have been using krb5int_open_plugin_dirs() instead of k5_plugin_load().

kenh:
I was curious and I looked; this change dates from 2015 and was for MIT Kerberos 1.14.2, but even then the client preauth interface didn't use ```krb5int_open_plugin_dirs()```, but it might have been a change from earlier that was brought forward from an earlier release.

PR #1413, opened 2025-03-05
Add NetBIOS name support when displaying ticket addresses

toreonify:
Active Directory domain controllers such as Microsoft AD and Samba are including a NetBIOS name in list of client addresses.

Currently when using `klist` NetBIOS names are displayed as `unknown addrtype`.


greghudson:
I pushed some changes; can you double-check that it still works?  I don't have a convenient test case.

toreonify:
Tested on two Linux VMs, works fine. Short names printed without trailing spaces and long names are printed okay as well.

Didn't mention but I wrote only a check for space/NUL because I don't think that `klist` should check correctness of a name itself (allowed characters, etc.), because it doesn't affect it's functionality.

PR #1414, opened 2025-03-07
Fix gss_iakerb client realm discovery

cryptomilk:
Hi Greg,

Alexander and I tried to implement client realm discovery as described in the RFC.

[3.1 Enterprise principal names](https://datatracker.ietf.org/doc/html/draft-ietf-kitten-iakerb-03#section-3.1)

However the default realm is always is always added in the gss_krb5_import_name(). The question is how do we detect that we are doing iakerb and and skip adding the default realm.

The top commit in my asn-iakerb samba branch shows the changes we did on the client:
https://git.samba.org/?p=asn/samba.git;a=shortlog;h=refs/heads/asn-iakerb

Thanks for your help!

### DIFF
--- a/src/lib/gssapi/krb5/import_name.c
+++ b/src/lib/gssapi/krb5/import_name.c
@@ -215,6 +215,9 @@ krb5_gss_import_name(OM_uint32 *minor_status, gss_buffer_t input_name_buffer,
         } else if (g_OID_equal(input_name_type, GSS_KRB5_NT_ENTERPRISE_NAME)) {
             stringrep = tmp;
             flags |= KRB5_PRINCIPAL_PARSE_ENTERPRISE;
+            // if (how to detect if this is iakerb???)
+            //    /* realm discovery */
+            //    flags |= KRB5_PRINCIPAL_PARSE_NO_DEF_REALM
 #ifndef NO_PASSWORD
         } else if (g_OID_equal(input_name_type, gss_nt_machine_uid_name)) {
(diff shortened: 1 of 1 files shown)


abbra:
Right now iakerb mechanism reuses `krb5_gss_import_name()` as it is. I guess we need a wrapper over it to enforce additional flags for the enterprise name.

greghudson:
I would call this adding client-side realm discovery support rather than fixing it; PR #1357 only added acceptor-side support.  In that PR, I wrote:

> If we want to add initiator support, we need to figure out what circumstances (API calls and environment) should trigger it. Section 3.1 of the draft says "A client using IAKERB may not have any realm information, even for the realm of the client's host, or may know that the client host's realm is not appropriate for a given enterprise principal name." I'm not aware of any current API signalling we could use for knowing that the client host's realm is inappropriate, but it seems reasonable to use realm discovery if the application specifies an enterprise name and the client has no default_realm. Unfortunately, I think currently we won't get as far as IAKERB in this case, as the krb5 gss_import_name() call will fail, so we'd have to start thinking about more wide-ranging changes to the krb5 mechanism.

abbra:
I created https://github.com/krb5/krb5/pull/1415 which implements a basic discovery mechanism for the case when enterprise principal is imported and it contains no realm. It works for me with a modified gss-sample application. The modification is basically to use enterprise principal oid when passing a principal without realm.

PR #1415, opened 2025-03-10
iakerb: discover default realm for enteprise principals

abbra:
When GSSAPI application imports an enterprise principal that has no realm set, support discovery of the target realm.

gss_import_name() for IAKerb mechanism should allow to import names without known realm. On the client side this should trigger realm discovery over IAKerb protocol.


abbra:
Next step would be to identify conditions upon which we could handle the target service realm in the same way. Perhaps we should store the fact that initial client realm was actually discovered? We can store the discovered realm in the iakerb context and identify that a target name's realm is the same as the local realm (as it would be if we didn't know the realm), then if the client principal's realm was discovered do replace the target name's realm.

greghudson:
Just so I understand the use case:

* The user does not have existing tickets (or we would use the ccache realm and get a referral), and expects to get initial tickets during context establishment.
* The client host configuration may not have a default_realm.  (Do we want discovery to happen even if it does?)
* The client application is written for this purpose (it has to select IAKERB as we don't select it by default).
* The client application imports a GSS_KRB5_NT_ENTERPRISE_NAME name for the initiator name.  (We could also conceivably make this for for GSS_C_NT_USER_NAME.)
* The client application imports a GSS_C_HOSTBASED_SERVICE name for the target name.
* We want to use IAKERB realm discovery to discover the service's realm name, to find out where to send the AS request for initial tickets.  The service realm might not be the client realm, but it should be close enough that it can issue WRONG_REALM referrals for an enterprise principal.
* After getting initial tickets, we want to use the discovered realm when making the TGS request for service tickets.

Obstacles I'm aware of are:

* The enterprise name import will fail without a default_realm.  (This is covered in the PR.)
* The host-based name import will fail without a default_realm.  Commit 66587fcd6380eac2c53674df4f64a827d337aee5 made libgssapi_krb5 save the parts of the host-based name, but punted on deferring the principal composition.

I have notes on the current PR code:

* The krb5_gss_import_name() change reacts to a GSS_KRB5_NT_ENTERPRISE_NAME import without an at-sign.  This seems wrong; the domain after the at-sign in an enterprise name is not a Kerberos realm.
* We only use `goto` for `goto cleanup`-like constructions, in new code.  iakerb_initiator_step() will have to be refactored to avoid the `send_discovery_request` label.
* We don't use variable names with leading underscores in C code.

abbra:
Yes, those are mostly correct assumptions about the use case. Main use case we aim to solve is SMB3 authentication for a standalone client to a standalone server. In this case a client initiates an SMB connection to a server and provides user credentials for authentication. IAKerb is used to allow Kerberos be a replacement to NTLM usage. In practice, for users there is no reason to specify actual realm (they might not even know about it). macOS implementation of local KDC used a realm name derived from the SHA-1 hash of the PKINIT certificate registered in Apple's service, to which redirect was issued from a well-known realm name (https://github.com/appleopen/KerberosHelper/blob/master/Source/KerberosHelper.c details that logic).

From my recent discussion with Steve Syfuhs from Windows Authentication team, Microsoft does not want to go this way. Instead, they want to use heuristics and a very basic discovery mechanism. It means for a standalone use case we'd see unqualified user names to be used (e.g. 'some-user' vs 'some-user@REALM', or 'some-user' vs 'WORKSTATION\some-user'). For the target service we would know a host name at best and no realm.

Fragment of that discussion:
image

We still have no access to Windows version that actually enables IAKErb, I am hoping to get something next month at SMB3 IOLab after SambaXP conference.

abbra:
I'll address requested changes to the code tomorrow, hopefully.

abbra:
> The krb5_gss_import_name() change reacts to a GSS_KRB5_NT_ENTERPRISE_NAME import without an at-sign. This seems wrong; the domain after the at-sign in an enterprise name is not a Kerberos realm.

We definitely don't want to use a client default realm in the case an enterprise name was specified for IAKerb case. IAKerb most likely means we want to operate on the terms of the realm accessible to the target service. The problem is that `krb5_parse_name_flags()` will add that client default realm in case a realm is missing. This will lead to a reverse issue: target service's Kerberos library will potentially not know about the client's realm at all and will not be able to perform proxying on its behalf. At application level we also almost always will be expecting a local (target service's) account, not the client system's account.

This is why I added a special check for the '@' in the name. Perhaps, we can replace this by yet another `krb5_parse_name_flags()` usage to detect that a realm is missing and then set the additional flag but it would be a repetition...

greghudson:
> This is why I added a special check for the '@' in the name. Perhaps, we can replace this by yet another krb5_parse_name_flags() usage to detect that a realm is missing and then set the additional flag but it would be a repetition...

An enterprise name with a single '@' does not specify a Kerberos realm; it parses to "a@b" in the first data component.  Given what you've said about IAKERB wanting to use the service realm, why would we want to add the default realm in that case?

abbra:
I tried to pass both flags and it seems to work fine. I still left the check in `gss-sample` to allow switching to enteprise name.

abbra:
Successful localkdc + iakerb demo using this PR: https://vda.li/images/localkdc-smb3-iakerb-demo.webm

Network trace for IAKerb discovery and authentication over SMB3 is here: https://abbra.fedorapeople.org/.paste/iakerb-in-samba-with-realm-discovery.tcp, keytab: https://abbra.fedorapeople.org/.paste/localkdc.keytab

src/lib/gssapi/krb5/iakerb.c
```diff
@@ -992,6 +996,7 @@ iakerb_gss_init_sec_context(OM_uint32 *minor_status,
     }
 
     kname = (krb5_gss_name_t)target_name;
+    atarget_name = target_name;
```
cryptomilk, line 999: You already set this in line 974.

src/lib/gssapi/krb5/iakerb.c
```diff
@@ -1041,11 +1046,33 @@ iakerb_gss_init_sec_context(OM_uint32 *minor_status,
         if (ctx->gssc == GSS_C_NO_CONTEXT)
             input_token = GSS_C_NO_BUFFER;
 
+        atarget_name = (gss_name_t) target_name;
```
cryptomilk, line 1049: You already set this in line 974 and 999.

src/lib/gssapi/krb5/iakerb.c
abbra: fixed

src/lib/gssapi/krb5/iakerb.c
abbra: fixed

cryptomilk:
We implemented support for ccache in Samba client. Once we have a service ticket and try to use it we run into a segfault:

```
#0  0x00007fe364d4a3a7 in __GI___wait4 (pid=178545, stat_loc=stat_loc@entry=0x7ffc5ad05b68, options=options@entry=0, usage=usage@entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
        sc_ret = -512
        sc_ret = <optimized out>
#1  0x00007fe364d4a4eb in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc@entry=0x7ffc5ad05b68, options=options@entry=0) at waitpid.c:38
No locals.
#2  0x00007fe364c9bc1b in do_system (line=line@entry=0x55dd9e0684d0 "/bin/sleep 999999999") at ../sysdeps/posix/system.c:172
        __result = <optimized out>
        _cleanup_start_doit = true
        _buffer = {__routine = 0x7fe364c9bca0 <cancel_handler>, __arg = 0x7ffc5ad05b70, __canceltype = 0, __prev = 0x0}
        _cleanup_routine = 0x7fe364c9bca0 <cancel_handler>
        cancel_args = {quit = 0x7fe364e58560 <quit>, intr = 0x7fe364e58600 <intr>, pid = 178545}
        status = -1
        ret = 0
        pid = 178545
        sa = {__sigaction_handler = {sa_handler = 0x1, sa_sigaction = 0x1}, sa_mask = {__val = {65536, 140721832090904, 140721832090948, 140721832091104, 0, 140720308486149, 140614585277840, 140614627710656, 65, 140614627711104, 2, 140721832091088, 140614627695508, 140614627710328, 122, 140614627711104}}, sa_flags = 0, sa_restorer = 0x7ffc5ad05dc0}
        omask = {__val = {7296, 140721832090656, 140721832091328, 0, 3458905235652632128, 6156726823249762048, 3458905235652631776, 6156726823249762048, 0, 140721832090720, 0, 140614630633680, 2, 6156726823249762048, 140721832090912, 27}}
        reset = {__val = {6, 65, 140614627711104, 2, 140721832090976, 140614626050899, 206158430232, 140721832090992, 140721832090800, 6156726823249762048, 140721832091024, 140721832090864, 140721832091184, 140614627711104, 0, 0}}
        spawn_attr = {__flags = 12, __pgrp = 0, __sd = {__val = {6, 65, 140614627711104, 2, 140721832090976, 140614626050899, 206158430232, 140721832090992, 140721832090800, 6156726823249762048, 140721832091024, 140721832090864, 140721832091184, 140614627711104, 0, 0}}, __ss = {__val = {7296, 140721832090656, 140721832091328, 0, 3458905235652632128, 6156726823249762048, 3458905235652631776, 6156726823249762048, 0, 140721832090720, 0, 140614630633680, 2, 6156726823249762048, 140721832090912, 27}}, __sp = {sched_priority = 0}, __policy = 0, __cgroup = 0, __pad = {0 <repeats 15 times>}}
#3  0x00007fe364c9bdbe in __libc_system (line=line@entry=0x55dd9e0684d0 "/bin/sleep 999999999") at ../sysdeps/posix/system.c:208
No locals.
#4  0x00007fe365204cd2 in call_panic_action (why=<optimized out>, as_root=as_root@entry=false) at ../../source3/lib/util.c:713
        lp_sub = <optimized out>
        cmd = 0x55dd9e0684d0 "/bin/sleep 999999999"
        result = <optimized out>
        __func__ = "call_panic_action"
#5  0x00007fe365205d44 in smb_panic_s3 (why=<optimized out>) at ../../source3/lib/util.c:729
No locals.
#6  0x00007fe364ede38b in smb_panic (why=why@entry=0x7ffc5ad05f60 "Signal 11: Segmentation fault") at ../../lib/util/fault.c:209
No locals.
#7  0x00007fe364ede400 in fault_report (sig=11) at ../../lib/util/fault.c:83
        signal_string = "Signal 11: Segmentation fault\000\000\000\000\000\000\000\001\000\000\000\n\000\000\000\374\177\000\000 d\320Z\374\177\000\000\000\000\000\000\343\177", '\000' <repeats 18 times>, "P\352\356d\000\000\000\000x5\312d\343\177\000\000\000\000\000\000\000\000\000\000\312\330\300d\343\177\000\000\000f\320Z\374\177\000\0000e\320Z\374\177\000"
        counter = 1
#8  sig_fault (sig=11) at ../../lib/util/fault.c:94
No locals.
#9  <signal handler called>
No locals.
#10 0x00007fe363e995cc in iakerb_gss_inquire_sec_context_by_oid (minor_status=0x7ffc5ad06d6c, context_handle=0x0, desired_object=0x7fe364415d20 <gse_sesskey_inq_oid>, data_set=0x7ffc5ad06d60) at krb5/iakerb.c:1297
        ctx = <optimized out>
#11 0x00007fe363e94d42 in gss_inquire_sec_context_by_oid (minor_status=0x7ffc5ad06d6c, context_handle=0x55dd9e01c8c0, desired_object=0x7fe364415d20 <gse_sesskey_inq_oid>, data_set=0x7ffc5ad06d60) at mechglue/g_inq_context_oid.c:61
        status = <optimized out>
        ctx = 0x55dd9e01c8c0
        mech = 0x55dd9e0115d0
        status = <optimized out>
        ctx = <optimized out>
        mech = <optimized out>
#12 gss_inquire_sec_context_by_oid (minor_status=minor_status@entry=0x7ffc5ad06d6c, context_handle=0x55dd9e01c8c0, desired_object=desired_object@entry=0x7fe364415d20 <gse_sesskey_inq_oid>, data_set=data_set@entry=0x7ffc5ad06d60) at mechglue/g_inq_context_oid.c:30
        status = <optimized out>
        ctx = <optimized out>
        mech = <optimized out>
#13 0x00007fe364411447 in gssapi_get_session_key (mem_ctx=mem_ctx@entry=0x55dd9e037a80, gssapi_context=<optimized out>, session_key=session_key@entry=0x55dd9e037aa8, keytype=keytype@entry=0x0) at ../../auth/kerberos/gssapi_pac.c:204
        gss_min = 0
        gss_maj = <optimized out>
        set = 0x0
        __FUNCTION__ = "gssapi_get_session_key"
#14 0x00007fe3644489eb in gensec_gse_session_key (gensec_security=gensec_security@entry=0x55dd9e016e60, mem_ctx=0x55dd9e037a80, session_key=0x55dd9e037aa8) at ../../source3/librpc/crypto/gse.c:1898
        gse_ctx = <optimized out>
#15 0x00007fe364448cdb in gensec_gse_session_info (gensec_security=0x55dd9e016e60, mem_ctx=0x55dd9dfee4d0, _session_info=0x55dd9dfee518) at ../../source3/librpc/crypto/gse.c:1966
        gse_ctx = <optimized out>
        nt_status = {v = 0}
        tmp_ctx = 0x55dd9dfee5a0
        session_info = 0x55dd9e037a80
        maj_stat = <optimized out>
        min_stat = 0
        pac_blob = {data = 0x55dd9dffae20 "\005", length = 232}
        pac_blob_ptr = <optimized out>
        name_token = {length = 0, value = 0x0}
        principal_string = 0x55dd9e015880 "asn@KRIKKIT.LOCALKDC.SITE"
        __FUNCTION__ = "gensec_gse_session_info"
#16 0x00007fe364493254 in gensec_session_info (gensec_security=0x55dd9e016e60, mem_ctx=<optimized out>, session_info=0x55dd9dfee518) at ../../auth/gensec/gensec.c:271
        status = <optimized out>
#17 0x00007fe364494feb in gensec_child_session_info (gensec_security=<optimized out>, mem_ctx=<optimized out>, session_info=<optimized out>) at ../../auth/gensec/gensec_util.c:322
No locals.
#18 0x00007fe364493254 in gensec_session_info (gensec_security=0x55dd9e013d40, mem_ctx=mem_ctx@entry=0x55dd9dfee4d0, session_info=session_info@entry=0x55dd9dfee518) at ../../auth/gensec/gensec.c:271
        status = <optimized out>
#19 0x00007fe3653985a4 in smbd_smb2_session_setup_gensec_done (subreq=<optimized out>) at ../../source3/smbd/smb2_sesssetup.c:943
        req = 0x55dd9dfee2f0
        state = 0x55dd9dfee4d0
        status = {v = 0}
#20 0x00007fe36504f092 in _tevent_req_notify_callback (req=req@entry=0x55dd9e0073f0, location=location@entry=0x7fe36447d0b0 "../../auth/gensec/gensec.c:570") at ../../lib/tevent/tevent_req.c:177
        new_depth = 1
#21 0x00007fe36504f152 in tevent_req_finish (req=req@entry=0x55dd9e0073f0, state=state@entry=TEVENT_REQ_DONE, location=location@entry=0x7fe36447d0b0 "../../auth/gensec/gensec.c:570") at ../../lib/tevent/tevent_req.c:234
        p = 0x0
#22 0x00007fe36504f16e in _tevent_req_done (req=req@entry=0x55dd9e0073f0, location=location@entry=0x7fe36447d0b0 "../../auth/gensec/gensec.c:570") at ../../lib/tevent/tevent_req.c:240
No locals.
#23 0x00007fe36449308a in gensec_update_done (subreq=<optimized out>) at ../../auth/gensec/gensec.c:570
        req = 0x55dd9e0073f0
        state = <optimized out>
        status = <optimized out>
        debug_subreq = 0x55dd9dfd33a0 "tevent_req[0x55dd9e013e60/../../auth/gensec/spnego.c:1614]: state[2] error[0 (0x0)]  state[struct gensec_spnego_update_state (0x55dd9e014040)] timer[(nil)] finish[../../auth/gensec/spnego.c:2109]"
        __func__ = "gensec_update_done"
#24 0x00007fe36504f092 in _tevent_req_notify_callback (req=req@entry=0x55dd9e013e60, location=location@entry=0x7fe36447e3d8 "../../auth/gensec/spnego.c:2109") at ../../lib/tevent/tevent_req.c:177
        new_depth = 2
#25 0x00007fe36504f152 in tevent_req_finish (req=req@entry=0x55dd9e013e60, state=state@entry=TEVENT_REQ_DONE, location=location@entry=0x7fe36447e3d8 "../../auth/gensec/spnego.c:2109") at ../../lib/tevent/tevent_req.c:234
        p = 0x0
#26 0x00007fe36504f16e in _tevent_req_done (req=req@entry=0x55dd9e013e60, location=location@entry=0x7fe36447e3d8 "../../auth/gensec/spnego.c:2109") at ../../lib/tevent/tevent_req.c:240
No locals.
#27 0x00007fe36449a711 in gensec_spnego_update_post (req=req@entry=0x55dd9e013e60) at ../../auth/gensec/spnego.c:2109
        state = 0x55dd9e014040
        spnego_state = 0x55dd9e00a9c0
        ops = <optimized out>
        status = <optimized out>
#28 0x00007fe36449b2b4 in gensec_spnego_update_done (subreq=<optimized out>) at ../../auth/gensec/spnego.c:1990
        req = 0x55dd9e013e60
        state = 0x55dd9e014040
        spnego_state = 0x55dd9e00a9c0
#29 0x00007fe36504f092 in _tevent_req_notify_callback (req=req@entry=0x55dd9e01b060, location=location@entry=0x7fe36447d0b0 "../../auth/gensec/gensec.c:570") at ../../lib/tevent/tevent_req.c:177
        new_depth = 3
#30 0x00007fe36504f152 in tevent_req_finish (req=req@entry=0x55dd9e01b060, state=state@entry=TEVENT_REQ_DONE, location=location@entry=0x7fe36447d0b0 "../../auth/gensec/gensec.c:570") at ../../lib/tevent/tevent_req.c:234
        p = 0x0
#31 0x00007fe36504f16e in _tevent_req_done (req=req@entry=0x55dd9e01b060, location=location@entry=0x7fe36447d0b0 "../../auth/gensec/gensec.c:570") at ../../lib/tevent/tevent_req.c:240
No locals.
#32 0x00007fe36449308a in gensec_update_done (subreq=<optimized out>) at ../../auth/gensec/gensec.c:570
        req = 0x55dd9e01b060
        state = <optimized out>
        status = <optimized out>
        debug_subreq = 0x55dd9e00b380 "netlogon"
        __func__ = "gensec_update_done"
#33 0x00007fe36504f092 in _tevent_req_notify_callback (req=req@entry=0x55dd9e008ca0, location=location@entry=0x7fe364435278 "../../source3/librpc/crypto/gse.c:1555") at ../../lib/tevent/tevent_req.c:177
        new_depth = 4
#34 0x00007fe36504f152 in tevent_req_finish (req=0x55dd9e008ca0, state=TEVENT_REQ_DONE, location=0x7fe364435278 "../../source3/librpc/crypto/gse.c:1555") at ../../lib/tevent/tevent_req.c:234
        p = 0x0
#35 0x00007fe36504f1fd in tevent_req_trigger (ev=<optimized out>, im=<optimized out>, private_data=<optimized out>) at ../../lib/tevent/tevent_req.c:291
        req = <optimized out>
#36 0x00007fe36504e67b in tevent_common_invoke_immediate_handler (im=0x55dd9e008db0, removed=removed@entry=0x0) at ../../lib/tevent/tevent_immediate.c:190
        handler_ev = 0x55dd9dfe6370
        ev = 0x55dd9dfe6370
        cur = {prev = <optimized out>, next = <optimized out>, event_ctx = 0x55dd9dfe6370, wrapper = 0x0, busy = <optimized out>, destroyed = <optimized out>, detach_ev_ctx = <optimized out>, handler = <optimized out>, private_data = <optimized out>, handler_name = <optimized out>, create_location = <optimized out>, schedule_location = <optimized out>, cancel_fn = <optimized out>, additional_data = <optimized out>, tag = <optimized out>}
#37 0x00007fe36504e69c in tevent_common_loop_immediate (ev=ev@entry=0x55dd9dfe6370) at ../../lib/tevent/tevent_immediate.c:236
        im = <optimized out>
        ret = <optimized out>
#38 0x00007fe3650549b4 in epoll_event_loop_once (ev=0x55dd9dfe6370, location=<optimized out>) at ../../lib/tevent/tevent_epoll.c:905
        epoll_ev = 0x55dd9dffdd30
        tval = {tv_sec = 133864215878712000, tv_usec = 94410326827888}
        panic_triggered = false
#39 0x00007fe365052957 in std_event_loop_once (ev=0x55dd9dfe6370, location=0x7fe3652f7ca8 "../../source3/smbd/smb2_process.c:2163") at ../../lib/tevent/tevent_standard.c:110
        glue_ptr = <optimized out>
        glue = 0x55dd9dfe4850
        ret = <optimized out>
#40 0x00007fe36504d50b in _tevent_loop_once (ev=ev@entry=0x55dd9dfe6370, location=location@entry=0x7fe3652f7ca8 "../../source3/smbd/smb2_process.c:2163") at ../../lib/tevent/tevent.c:820
        ret = <optimized out>
        nesting_stack_ptr = 0x0
        __func__ = "_tevent_loop_once"
#41 0x00007fe36504d71a in tevent_common_loop_wait (ev=0x55dd9dfe6370, location=0x7fe3652f7ca8 "../../source3/smbd/smb2_process.c:2163") at ../../lib/tevent/tevent.c:949
        ret = <optimized out>
#42 0x00007fe365052905 in std_event_loop_wait (ev=0x55dd9dfe6370, location=0x7fe3652f7ca8 "../../source3/smbd/smb2_process.c:2163") at ../../lib/tevent/tevent_standard.c:141
        glue_ptr = <optimized out>
        glue = 0x55dd9dfe4850
        ret = <optimized out>
#43 0x00007fe36504d776 in _tevent_loop_wait (ev=ev@entry=0x55dd9dfe6370, location=location@entry=0x7fe3652f7ca8 "../../source3/smbd/smb2_process.c:2163") at ../../lib/tevent/tevent.c:968
No locals.
#44 0x00007fe365380027 in smbd_process (ev_ctx=ev_ctx@entry=0x55dd9dfe6370, msg_ctx=msg_ctx@entry=0x55dd9dfcedd0, sock_fd=sock_fd@entry=33, interactive=interactive@entry=false) at ../../source3/smbd/smb2_process.c:2163
        lp_sub = 0x7fe365219c30 <s3_global_substitution>
        client = 0x55dd9dffbe60
        sconn = 0x55dd9dffc1a0
        xconn = 0x55dd9e003220
        locaddr = 0x55dd9e0018d0 "@\304\377\235\335U"
        remaddr = <optimized out>
        ret = <optimized out>
        status = <optimized out>
        tv = {tv_sec = 1741947987, tv_usec = 871200}
        trace_state = {ev = 0x55dd9dfe6370, frame = 0x55dd9e066510, profile_idle = {start = 0, stats = 0x0}, before_wait_tv = {tv_sec = 1741947987, tv_usec = 886908}, after_wait_tv = {tv_sec = 1741947987, tv_usec = 892426}}
        debug = true
        now = 133864215878712000
        chroot_dir = <optimized out>
        rc = <optimized out>
        __func__ = "smbd_process"
        __FUNCTION__ = "smbd_process"
#45 0x000055dd648c0c69 in smbd_accept_connection (ev=0x55dd9dfe6370, fde=<optimized out>, flags=<optimized out>, private_data=<optimized out>) at ../../source3/smbd/server.c:1039
        addrstr = "fe80::a538:f849:a6e3:4b02%wlp0s20f3\000\335U\000\000;l\te\343\177"
        status = {v = 0}
        s = 0x0
        msg_ctx = 0x55dd9dfcedd0
        addr = {ss_family = 10, __ss_padding = "\277,\000\000\000\000\376\200\000\000\000\000\000\000\2458\370I\246\343K\002\002\000\000\000PZ\213\310Є\376\235\335U\000\000p\204\376\235\335U\000\000\200v\320Z\374\177\000\000P\377\377\377\377\377\377\377\000\000\000\000\000\000\000\000`\204\376\235\335U\000\000\320t\320Z\374\177\000\000pc\376\235\335U\000\000\270\027\377\235\335U\000\000\275\001\000\000\000\000\000\000\220\027\377\235\335U\000", __ss_align = 140614629142006}
        in_addrlen = 28
        fd = 33
        pid = 0
        __FUNCTION__ = "smbd_accept_connection"
#46 0x00007fe36504e137 in tevent_common_invoke_fd_handler (fde=0x55dd9dfe8890, flags=<optimized out>, removed=removed@entry=0x0) at ../../lib/tevent/tevent_fd.c:174
        handler_ev = 0x55dd9dfe6370
#47 0x00007fe365054d8e in epoll_event_loop (epoll_ev=0x55dd9dfe6510, tvalp=0x7ffc5ad07530) at ../../lib/tevent/tevent_epoll.c:696
        fde = <optimized out>
        effective_flags = <optimized out>
        flags = <optimized out>
        got_error = <optimized out>
        selected = <optimized out>
        ret = 1
        i = 0
        timeout = <optimized out>
        wait_errno = 0
        events = {{events = 1, data = {ptr = 0x55dd9dfe8890, fd = -1644263280, u32 = 2650704016, u64 = 94410326837392}}}
        ret = <optimized out>
        i = <optimized out>
        events = <optimized out>
        timeout = <optimized out>
        wait_errno = <optimized out>
        fde = <optimized out>
        selected = <optimized out>
        effective_flags = <optimized out>
        flags = <optimized out>
        got_error = <optimized out>
        add_flags = <optimized out>
#48 epoll_event_loop_once (ev=<optimized out>, location=<optimized out>) at ../../lib/tevent/tevent_epoll.c:926
        epoll_ev = 0x55dd9dfe6510
        tval = {tv_sec = 806, tv_usec = 633871}
        panic_triggered = false
#49 0x00007fe365052957 in std_event_loop_once (ev=0x55dd9dfe6370, location=0x55dd648ba780 "../../source3/smbd/server.c:1388") at ../../lib/tevent/tevent_standard.c:110
        glue_ptr = <optimized out>
        glue = 0x55dd9dfe4850
        ret = <optimized out>
#50 0x00007fe36504d50b in _tevent_loop_once (ev=ev@entry=0x55dd9dfe6370, location=location@entry=0x55dd648ba780 "../../source3/smbd/server.c:1388") at ../../lib/tevent/tevent.c:820
        ret = <optimized out>
        nesting_stack_ptr = 0x0
        __func__ = "_tevent_loop_once"
#51 0x00007fe36504d71a in tevent_common_loop_wait (ev=0x55dd9dfe6370, location=0x55dd648ba780 "../../source3/smbd/server.c:1388") at ../../lib/tevent/tevent.c:949
        ret = <optimized out>
#52 0x00007fe365052905 in std_event_loop_wait (ev=0x55dd9dfe6370, location=0x55dd648ba780 "../../source3/smbd/server.c:1388") at ../../lib/tevent/tevent_standard.c:141
        glue_ptr = <optimized out>
        glue = 0x55dd9dfe4850
        ret = <optimized out>
#53 0x00007fe36504d776 in _tevent_loop_wait (ev=ev@entry=0x55dd9dfe6370, location=location@entry=0x55dd648ba780 "../../source3/smbd/server.c:1388") at ../../lib/tevent/tevent.c:968
No locals.
#54 0x000055dd648c32f1 in smbd_parent_loop (ev_ctx=0x55dd9dfe6370, parent=0x55dd9dfb7ff0) at ../../source3/smbd/server.c:1388
        trace_state = {frame = 0x55dd9dfe84d0}
        ret = 0
        trace_state = <optimized out>
        ret = <optimized out>
        __FUNCTION__ = "smbd_parent_loop"
#55 main (argc=<optimized out>, argv=<optimized out>) at ../../source3/smbd/server.c:2360
        cmdline_daemon_cfg = <optimized out>
        log_stdout = <optimized out>
        ports = 0x0
        profile_level = 0x0
        opt = <optimized out>
        pc = <optimized out>
        main_server_id = {pid = 178149, task_id = 0, vnn = 4294967295, unique_id = 15886692101031918144}
        long_options = {{longName = 0x0, shortName = 0 '\000', argInfo = 4, arg = 0x7fe364e6f160 <poptHelpOptions>, val = 0, descrip = 0x55dd648b93c8 "Help options:", argDescrip = 0x0}, {longName = 0x55dd648b961b "build-options", shortName = 98 'b', argInfo = 0, arg = 0x0, val = 98, descrip = 0x55dd648b907e "Print build options", argDescrip = 0x0}, {longName = 0x55dd648b9409 "port", shortName = 112 'p', argInfo = 1, arg = 0x7ffc5ad078d8, val = 0, descrip = 0x55dd648b90aa "Listen on the specified ports", argDescrip = 0x0}, {longName = 0x55dd648b954a "profiling-level", shortName = 80 'P', argInfo = 1, arg = 0x7ffc5ad078d0, val = 0, descrip = 0x55dd648b93df "Set profiling level", argDescrip = 0x55dd648b9629 "PROFILE_LEVEL"}, {longName = 0x0, shortName = 0 '\000', argInfo = 4, arg = 0x7fe365443360 <popt_common_samba>, val = 0, descrip = 0x55dd648b93f3 "Common Samba options:", argDescrip = 0x0}, {longName = 0x0, shortName = 0 '\000', argInfo = 4, arg = 0x7fe365442ba0 <popt_common_daemon>, val = 0, descrip = 0x55dd648b9637 "Daemon options:", argDescrip = 0x0}, {longName = 0x0, shortName = 0 '\000', argInfo = 4, arg = 0x7fe365442cc0 <popt_common_version>, val = 0, descrip = 0x55dd648b921f "Version options:", argDescrip = 0x0}, {longName = 0x0, shortName = 0 '\000', argInfo = 0, arg = 0x0, val = 0, descrip = 0x0, argDescrip = 0x0}}
        parent = 0x55dd9dfb7ff0
        frame = <optimized out>
        status = <optimized out>
        ev_ctx = 0x55dd9dfe6370
        msg_ctx = <optimized out>
        server_id = {pid = 178151, task_id = 0, vnn = 4294967295, unique_id = 3902111396800065021}
        se = <optimized out>
        profiling_level = <optimized out>
        np_dir = <optimized out>
        lp_sub = <optimized out>
        smbd_shim_fns = {change_to_root_user = 0x7fe365353de5 <smbd_change_to_root_user>, become_authenticated_pipe_user = 0x7fe365353e75 <smbd_become_authenticated_pipe_user>, unbecome_authenticated_pipe_user = 0x7fe365353f3b <smbd_unbecome_authenticated_pipe_user>, contend_level2_oplocks_begin = 0x7fe3653c94e9 <smbd_contend_level2_oplocks_begin>, contend_level2_oplocks_end = 0x7fe3653c9865 <smbd_contend_level2_oplocks_end>, become_root = 0x7fe365353f49 <smbd_become_root>, unbecome_root = 0x7fe365353f71 <smbd_unbecome_root>, exit_server = 0x7fe3653be3a7 <smbd_exit_server>, exit_server_cleanly = 0x7fe3653be3ba <smbd_exit_server_cleanly>}
        ok = <optimized out>
        __func__ = "main"
        __FUNCTION__ = "main"
Detaching from program: /home/asn/workspace/prj/oss/samba/asn-iakerb/bin/default/source3/smbd/smbd, process 178544
[Inferior 1 (process 178544) detached]

```

abbra:
It may be a half-initialized part of the acceptor implementation I did for AP-REQ.

src/lib/gssapi/krb5/iakerb.c
```diff
@@ -569,6 +640,23 @@ iakerb_initiator_step(iakerb_ctx_id_t ctx,
 
     switch (ctx->state) {
     case IAKERB_AS_REQ:
+        if (cred->name->princ->realm.length == 0 &&
+            (cred->name->princ->realm.data == NULL ||
+             cred->name->princ->realm.data[0] == '\0')) {
```
greghudson, line 645: If `realm.length` is 0 then it is potentially undefined behavior to fetch `realm.data[0]`.  The length test alone should be sufficient.

abbra: Fixed, thanks.

greghudson:
Added tests, though I might revise them to more explicitly check that we're doing realm discovery.

I was wrong about the service name import failing with no default_realm, since krb5_sname_to_principal() doesn't look up the default realm until the principal is used.

I noticed a pre-existing bug while debugging the tests: if iakerb_acceptor_realm() gets an error from krb5_get_default_realm(), it tries to send back an error, but krb5_mk_error() fails because error.server isn't set and it's a mandatory field.  Not in scope for this PR, but I will try to remember to fix it separately.

greghudson:
I'm trying to understand the necessity for modifying the target principal in iakerb_tkt_creds_ctx() and iakerb_gss_init_sec_context().

If the caller imports a host-based service, krb5_sname_to_principal() will omit the realm in the constructed principal name (for all values of dns_canonicalize_hostname) unless there is a [domain_realm] mapping for the service.  krb5_get_credentials() will then begin at the client principal realm and be receptive to referrals.

If there is a domain_realm mapping for the service, or if the caller imports a principal name with realm, then it seems incorrect to override the explicit realm selection with IAKERB realm discovery.

abbra:
I'll re-try without that fixup. In my previous tests it did somehow forced the client's realm before the discovery but it might be issue with our iakerb flow in experimental samba code.

abbra:
Yes, here is what happens:
- we receive a TGS-REP with `cifs/asn.localkdc.test@ASN.LOCALKDC.SITE` but the original target service had been `cifs/asn.localkdc.test@AB.LOCALKDC.SITE` (the client side realm)
- it is stored in the memory ccache
- krb5 will attempt to retrieve the ticket to `cifs/asn.localkdc.test@AB.LOCALKDC.SITE` and will not find it
- `gss_init_sec_context()` fails due to it, from application's point of view.

```
[695] 1742372916.129395: TGS reply is for testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@ASN.LOCALKDC.SITE with session key aes256-sha2/648C
[695] 1742372916.129396: TGS request result: 0/Success
[695] 1742372916.129397: Received creds for desired service cifs/asn.localkdc.test@ASN.LOCALKDC.SITE
[695] 1742372916.129398: Storing testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@ASN.LOCALKDC.SITE in MEMORY:jTSYh43
[695] 1742372916.129399: Getting credentials testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@AB.LOCALKDC.SITE using ccache MEMORY:jTSYh43
[695] 1742372916.129400: Retrieving testuser@ASN.LOCALKDC.SITE -> krb5_ccache_conf_data/start_realm@X-CACHECONF: from MEMORY:jTSYh43 with result: -1765328243/Matching credential not found
[695] 1742372916.129401: Retrieving testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@AB.LOCALKDC.SITE from MEMORY:jTSYh43 with result: -1765328243/Matching credential not found
gse_get_client_auth_token: gss_init_sec_context failed with [Unspecified GSS failure.  Minor code may provide more information: Matching credential not found](2529639053)
```

Full trace:
```
# KRB5_TRACE=/dev/stderr smbclient -d10 -U testuser%Secret123 --use-krb5-ccache=KCM:0 -L asn.localkdc.test
INFO: Current debug levels:
  all: 10
  tdb: 10
  printdrivers: 10
  lanman: 10
  smb: 10
  rpc_parse: 10
  rpc_srv: 10
  rpc_cli: 10
  passdb: 10
  sam: 10
  auth: 10
  winbind: 10
  vfs: 10
  idmap: 10
  quota: 10
  acls: 10
  locking: 10
  msdfs: 10
  dmapi: 10
  registry: 10
  scavenger: 10
  dns: 10
  ldb: 10
  tevent: 10
  auth_audit: 10
  auth_json_audit: 10
  kerberos: 10
  drs_repl: 10
  smb2: 10
  smb2_credits: 10
  dsdb_audit: 10
  dsdb_json_audit: 10
  dsdb_password_audit: 10
  dsdb_password_json_audit: 10
  dsdb_transaction_audit: 10
  dsdb_transaction_json_audit: 10
  dsdb_group_audit: 10
  dsdb_group_json_audit: 10
  ldapsrv: 10
lp_load_ex: refreshing parameters
Initialising global parameters
INFO: Current debug levels:
  all: 10
  tdb: 10
  printdrivers: 10
  lanman: 10
  smb: 10
  rpc_parse: 10
  rpc_srv: 10
  rpc_cli: 10
  passdb: 10
  sam: 10
  auth: 10
  winbind: 10
  vfs: 10
  idmap: 10
  quota: 10
  acls: 10
  locking: 10
  msdfs: 10
  dmapi: 10
  registry: 10
  scavenger: 10
  dns: 10
  ldb: 10
  tevent: 10
  auth_audit: 10
  auth_json_audit: 10
  kerberos: 10
  drs_repl: 10
  smb2: 10
  smb2_credits: 10
  dsdb_audit: 10
  dsdb_json_audit: 10
  dsdb_password_audit: 10
  dsdb_password_json_audit: 10
  dsdb_transaction_audit: 10
  dsdb_transaction_json_audit: 10
  dsdb_group_audit: 10
  dsdb_group_json_audit: 10
  ldapsrv: 10
Processing section "[global]"
doing parameter max smbd processes = 1000
doing parameter dedicated keytab file = FILE:/etc/samba/localkdc.keytab
doing parameter kerberos method = dedicated keytab
doing parameter log file = /var/log/samba/log.%m
doing parameter log level = 10 auth_audit:5@/var/log/samba/audit.log
doing parameter server role = standalone
doing parameter obey pam restrictions = yes
doing parameter realm = AB.LOCALKDC.SITE
doing parameter netbios name = AB
doing parameter workgroup = AB
doing parameter idmap config * : range = 0 - 0
doing parameter idmap config * : backend = tdb
doing parameter idmap config AB : range = 1000 - 10000
doing parameter idmap config AB : backend = sss
pm_process() returned Yes
lp_servicenumber: couldn't find homes
added interface eth0 ip=192.168.221.2 bcast=192.168.221.255 netmask=255.255.255.0
Client started (version 4.22.0).
Opening cache file at /var/lib/samba/lock/gencache.tdb
sitename_fetch: No stored sitename for realm 'AB.LOCALKDC.SITE'
internal_resolve_name: looking up asn.localkdc.test#20 (sitename (null))
namecache_fetch: name asn.localkdc.test#20 found.
remove_duplicate_addrs2: looking for duplicate address/port pairs
Connecting to 192.168.221.3 at port 445
socket options: SO_KEEPALIVE=0, SO_REUSEADDR=0, SO_BROADCAST=0, TCP_NODELAY=1, TCP_KEEPCNT=9, TCP_KEEPIDLE=7200, TCP_KEEPINTVL=75, IPTOS_LOWDELAY=0, IPTOS_THROUGHPUT=0, SO_REUSEPORT=0, SO_SNDBUF=87040, SO_RCVBUF=131072, SO_SNDLOWAT=1, SO_RCVLOWAT=1, SO_SNDTIMEO=0, SO_RCVTIMEO=0, TCP_QUICKACK=1, TCP_DEFER_ACCEPT=0, TCP_USER_TIMEOUT=0
 session request ok
 negotiated dialect[SMB3_11] against server[asn.localkdc.test]
cli_session_setup_spnego_send: Connect to asn.localkdc.test as testuser@AB.LOCALKDC.SITE using SPNEGO
GENSEC backend 'gssapi_spnego' registered
GENSEC backend 'gssapi_krb5' registered
GENSEC backend 'gssapi_krb5_sasl' registered
GENSEC backend 'spnego' registered
GENSEC backend 'schannel' registered
GENSEC backend 'ncalrpc_as_system' registered
GENSEC backend 'sasl-EXTERNAL' registered
GENSEC backend 'ntlmssp' registered
GENSEC backend 'ntlmssp_resume_ccache' registered
GENSEC backend 'http_basic' registered
GENSEC backend 'http_ntlm' registered
GENSEC backend 'http_negotiate' registered
Starting GENSEC mechanism spnego
Starting GENSEC submechanism gse_iakerb
gensec_gse_iakerb_client_start: Kerberos required username[AB\testuser]
gensec_gse_iakerb_client_start: Acquire credentials for testuser to access cifs/asn.localkdc.test
[695] 1742372916.129344: Resolving unique ccache of type MEMORY
gse_get_client_auth_token: gss_init_sec_context(cifs/asn.localkdc.test@AB.LOCALKDC.SITE)gensec_update_send: gse_iakerb[0x5613af47a1f0]: subreq: 0x5613af460300
gensec_update_send: spnego[0x5613af478dd0]: subreq: 0x5613af4795b0
gensec_update_done: gse_iakerb[0x5613af47a1f0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af460300/../../source3/librpc/crypto/gse.c:1488]: state[2] error[0 (0x0)]  state[struct gensec_gse_update_state (0x5613af4604e0)] timer[(nil)] finish[../../source3/librpc/crypto/gse.c:1499]
gensec_update_done: spnego[0x5613af478dd0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af4795b0/../../auth/gensec/spnego.c:1614]: state[2] error[0 (0x0)]  state[struct gensec_spnego_update_state (0x5613af479790)] timer[(nil)] finish[../../auth/gensec/spnego.c:2109]
[695] 1742372916.129345: Getting initial credentials for testuser@ASN.LOCALKDC.SITE
[695] 1742372916.129347: Sending unauthenticated request
gensec_update_send: gse_iakerb[0x5613af47a1f0]: subreq: 0x5613af474d20
gensec_update_send: spnego[0x5613af478dd0]: subreq: 0x5613af48cca0
gensec_update_done: gse_iakerb[0x5613af47a1f0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af474d20/../../source3/librpc/crypto/gse.c:1488]: state[2] error[0 (0x0)]  state[struct gensec_gse_update_state (0x5613af474f00)] timer[(nil)] finish[../../source3/librpc/crypto/gse.c:1499]
gensec_update_done: spnego[0x5613af478dd0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af48cca0/../../auth/gensec/spnego.c:1614]: state[2] error[0 (0x0)]  state[struct gensec_spnego_update_state (0x5613af48ce80)] timer[(nil)] finish[../../auth/gensec/spnego.c:2109]
[695] 1742372916.129348: Received error from KDC: -1765328359/Additional pre-authentication required
[695] 1742372916.129351: Preauthenticating using KDC method data
[695] 1742372916.129352: Processing preauth types: PA-PK-AS-REQ (16), PA-FX-FAST (136), PA-ETYPE-INFO2 (19), PA-PKINIT-KX (147), PA-SPAKE (151), PA-ENC-TIMESTAMP (2), PA_AS_FRESHNESS (150), PA-FX-COOKIE (133)
[695] 1742372916.129353: Selected etype info: etype aes256-sha2, salt "ASN.LOCALKDC.SITEtestuser", params ""
[695] 1742372916.129354: Received cookie: MIT1\x00\x00\x00\x01\x8c\xbb\x1dp\xfe[\xc5\x9e\x07\xa5q\xd9\xf1\x86)\x91\xe0\x8a\xda(\x0dc\xea\xe0\xa6\x8b\x83\xdc\xc9\xd0\xb1\x94\xde@\xd0_\xefS\xfbJ\xb1%\xef\xd62\x17\xc7=Nt\x1c\xad\xb6hE\xeca\xc4h\x9dE\xc4\xef\xba\x1e\x965\xb06v\xa1o\x8fTh\xc8\xf8\x13\x81\x11\xe5Q\xd2\xca\x8f6Wl\xbd\x87\xcep\xa5\x85\x85f\x82`{1\x14\xc1\x04\xcf\x178\x93\xb0\xe4\x9b\xd0)\x8b\xd6\x00\xdf\x08\xe5\x86s\xca9\xef@;\x93x\xe4C\x98\xf6H\x98p\xd1q\xd3\x046k\xca\x1d
[695] 1742372916.129355: PKINIT client has no configured identity; giving up
[695] 1742372916.129356: Preauth module pkinit (147) (info) returned: 0/Success
[695] 1742372916.129357: PKINIT client received freshness token from KDC
[695] 1742372916.129358: Preauth module pkinit (150) (info) returned: 0/Success
[695] 1742372916.129359: PKINIT client has no configured identity; giving up
[695] 1742372916.129360: Preauth module pkinit (16) (real) returned: 22/Invalid argument
[695] 1742372916.129361: SPAKE challenge received with group 1, pubkey E7250568973E0C57C2CB2674C5D7F5EB082D6AE43629502645CF9DC5101E592F
[695] 1742372916.129362: SPAKE key generated with pubkey 5FD835AA3D275799759E137B2E2E652B15EE6E1B014D6C2DCAA5BBFC0E90D427
[695] 1742372916.129363: SPAKE algorithm result: 5FE43D48F1063D8C9DB8843DF372394F3A3A105715084D2964A12C11C546C2E0
[695] 1742372916.129364: SPAKE final transcript hash: 70D711761AF3403B60814B2F6E3E6384E0DB33271AF5EBDD13A6C6B84F9924E8
[695] 1742372916.129365: Sending SPAKE response
[695] 1742372916.129366: Preauth module spake (151) (real) returned: 0/Success
[695] 1742372916.129367: Produced preauth for next request: PA-FX-COOKIE (133), PA-SPAKE (151)
gensec_update_send: gse_iakerb[0x5613af47a1f0]: subreq: 0x5613af460300
gensec_update_send: spnego[0x5613af478dd0]: subreq: 0x5613af48c630
gensec_update_done: gse_iakerb[0x5613af47a1f0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af460300/../../source3/librpc/crypto/gse.c:1488]: state[2] error[0 (0x0)]  state[struct gensec_gse_update_state (0x5613af4604e0)] timer[(nil)] finish[../../source3/librpc/crypto/gse.c:1499]
gensec_update_done: spnego[0x5613af478dd0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af48c630/../../auth/gensec/spnego.c:1614]: state[2] error[0 (0x0)]  state[struct gensec_spnego_update_state (0x5613af48c810)] timer[(nil)] finish[../../auth/gensec/spnego.c:2109]
[695] 1742372916.129368: Processing preauth types: PA-ETYPE-INFO2 (19)
[695] 1742372916.129369: Selected etype info: etype aes256-sha2, salt "ASN.LOCALKDC.SITEtestuser", params ""
[695] 1742372916.129370: Produced preauth for next request: (empty)
[695] 1742372916.129371: AS key determined by preauth: aes256-sha2/29CD
[695] 1742372916.129372: Decrypted AS reply; session key is: aes256-sha2/61CC
[695] 1742372916.129373: FAST negotiation: available
[695] 1742372916.129374: Resolving unique ccache of type MEMORY
[695] 1742372916.129375: Initializing MEMORY:yvlvEyv with default princ testuser@ASN.LOCALKDC.SITE
[695] 1742372916.129376: Storing config in MEMORY:yvlvEyv for krbtgt/ASN.LOCALKDC.SITE@ASN.LOCALKDC.SITE: fast_avail: yes
[695] 1742372916.129377: Storing testuser@ASN.LOCALKDC.SITE -> krb5_ccache_conf_data/fast_avail/krbtgt\/ASN.LOCALKDC.SITE\@ASN.LOCALKDC.SITE@X-CACHECONF: in MEMORY:yvlvEyv
[695] 1742372916.129378: Storing config in MEMORY:yvlvEyv for krbtgt/ASN.LOCALKDC.SITE@ASN.LOCALKDC.SITE: pa_type: 151
[695] 1742372916.129379: Storing testuser@ASN.LOCALKDC.SITE -> krb5_ccache_conf_data/pa_type/krbtgt\/ASN.LOCALKDC.SITE\@ASN.LOCALKDC.SITE@X-CACHECONF: in MEMORY:yvlvEyv
[695] 1742372916.129380: Storing testuser@ASN.LOCALKDC.SITE -> krbtgt/ASN.LOCALKDC.SITE@ASN.LOCALKDC.SITE in MEMORY:yvlvEyv
[695] 1742372916.129381: Moving ccache MEMORY:yvlvEyv to MEMORY:jTSYh43
[695] 1742372916.129382: Destroying ccache MEMORY:yvlvEyv
[695] 1742372916.129383: Getting credentials testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@ASN.LOCALKDC.SITE using ccache MEMORY:jTSYh43
[695] 1742372916.129384: Retrieving testuser@ASN.LOCALKDC.SITE -> krb5_ccache_conf_data/start_realm@X-CACHECONF: from MEMORY:jTSYh43 with result: -1765328243/Matching credential not found
[695] 1742372916.129385: Retrieving testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@ASN.LOCALKDC.SITE from MEMORY:jTSYh43 with result: -1765328243/Matching credential not found
[695] 1742372916.129386: Retrieving testuser@ASN.LOCALKDC.SITE -> krbtgt/ASN.LOCALKDC.SITE@ASN.LOCALKDC.SITE from MEMORY:jTSYh43 with result: 0/Success
[695] 1742372916.129387: Starting with TGT for client realm: testuser@ASN.LOCALKDC.SITE -> krbtgt/ASN.LOCALKDC.SITE@ASN.LOCALKDC.SITE
[695] 1742372916.129388: Requesting tickets for cifs/asn.localkdc.test@ASN.LOCALKDC.SITE, referrals on
[695] 1742372916.129389: Generated subkey for TGS request: aes256-sha2/5AB0
[695] 1742372916.129390: etypes requested in TGS request: aes256-sha2, aes128-sha2, aes256-cts, aes128-cts, camellia256-cts, camellia128-cts
[695] 1742372916.129392: Encoding request body and padata into FAST request
gensec_update_send: gse_iakerb[0x5613af47a1f0]: subreq: 0x5613af474d20
gensec_update_send: spnego[0x5613af478dd0]: subreq: 0x5613af4a9c70
gensec_update_done: gse_iakerb[0x5613af47a1f0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af474d20/../../source3/librpc/crypto/gse.c:1488]: state[2] error[0 (0x0)]  state[struct gensec_gse_update_state (0x5613af474f00)] timer[(nil)] finish[../../source3/librpc/crypto/gse.c:1499]
gensec_update_done: spnego[0x5613af478dd0]: NT_STATUS_MORE_PROCESSING_REQUIRED tevent_req[0x5613af4a9c70/../../auth/gensec/spnego.c:1614]: state[2] error[0 (0x0)]  state[struct gensec_spnego_update_state (0x5613af4a9e50)] timer[(nil)] finish[../../auth/gensec/spnego.c:2109]
[695] 1742372916.129393: Decoding FAST response
[695] 1742372916.129394: FAST reply key: aes256-sha2/88C9
[695] 1742372916.129395: TGS reply is for testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@ASN.LOCALKDC.SITE with session key aes256-sha2/648C
[695] 1742372916.129396: TGS request result: 0/Success
[695] 1742372916.129397: Received creds for desired service cifs/asn.localkdc.test@ASN.LOCALKDC.SITE
[695] 1742372916.129398: Storing testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@ASN.LOCALKDC.SITE in MEMORY:jTSYh43
[695] 1742372916.129399: Getting credentials testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@AB.LOCALKDC.SITE using ccache MEMORY:jTSYh43
[695] 1742372916.129400: Retrieving testuser@ASN.LOCALKDC.SITE -> krb5_ccache_conf_data/start_realm@X-CACHECONF: from MEMORY:jTSYh43 with result: -1765328243/Matching credential not found
[695] 1742372916.129401: Retrieving testuser@ASN.LOCALKDC.SITE -> cifs/asn.localkdc.test@AB.LOCALKDC.SITE from MEMORY:jTSYh43 with result: -1765328243/Matching credential not found
gse_get_client_auth_token: gss_init_sec_context failed with [Unspecified GSS failure.  Minor code may provide more information: Matching credential not found](2529639053)
gensec_update_send: gse_iakerb[0x5613af47a1f0]: subreq: 0x5613af460300
gensec_update_send: spnego[0x5613af478dd0]: subreq: 0x5613af4a9c70
gensec_update_done: gse_iakerb[0x5613af47a1f0]: NT_STATUS_LOGON_FAILURE tevent_req[0x5613af460300/../../source3/librpc/crypto/gse.c:1488]: state[3] error[-7963671676338569107 (0x917B5ACDC000006D)]  state[struct gensec_gse_update_state (0x5613af4604e0)] timer[(nil)] finish[../../source3/librpc/crypto/gse.c:1502]
gensec_spnego_client_negTokenTarg_step: SPNEGO(gse_iakerb) login failed: NT_STATUS_LOGON_FAILURE
gensec_update_done: spnego[0x5613af478dd0]: NT_STATUS_LOGON_FAILURE tevent_req[0x5613af4a9c70/../../auth/gensec/spnego.c:1614]: state[3] error[-7963671676338569107 (0x917B5ACDC000006D)]  state[struct gensec_spnego_update_state (0x5613af4a9e50)] timer[(nil)] finish[../../auth/gensec/spnego.c:2032]
[695] 1742372916.129402: Destroying ccache MEMORY:jTSYh43
SPNEGO login failed: The attempted logon is invalid. This is either due to a bad username or authentication information.
session setup failed: NT_STATUS_LOGON_FAILURE
Freeing parametrics:
```

abbra:
More debugging: it looks like Samba hard-codes the realm when importing the target name:
```
...
#5  0x00007fb7a6de5fe3 in gss_init_sec_context (minor_status=minor_status@entry=0x7ffe07e00ea8, claimant_cred_handle=<optimized out>, 
    context_handle=context_handle@entry=0x559f731a3650, target_name=0x559f731b1b70, req_mech_type=req_mech_type@entry=0x559f731a36a8, req_flags=46, time_req=0, 
    input_chan_bindings=0x0, input_token=0x7ffe07e00ec0, actual_mech_type=0x0, output_token=0x7ffe07e00ed0, ret_flags=0x559f731a366c, time_rec=0x7ffe07e00eac)
    at mechglue/g_init_sec_context.c:192
        status = <optimized out>
        temp_minor_status = 1600415092
        union_name = 0x559f731b1b70
        union_cred = <optimized out>
        internal_name = 0x559f731bdf50
        union_ctx_id = 0x559f73178cd0
        selected_mech = 0x559f731bda20
        mech = 0x559f731bda20
        input_cred_handle = <optimized out>
#6  0x00007fb7a70a310b in gse_get_client_auth_token (mem_ctx=0x559f731a34e0, gensec_security=<optimized out>, token_in=<synthetic pointer>, token_out=0x559f731a34e8)
    at ../../source3/librpc/crypto/gse.c:559
...
(gdb) print (char*)(*(*(gss_name_t)0x559f731b1b70)->external_name)->value
$9 = 0x559f731ba720 "cifs/asn.localkdc.test@AB.LOCALKDC.SITE"
```

I'll see if fixing Samba client side to not use explicit realm in target name in case of iakerb will help.

abbra:
Update. Fixing Samba helped: https://git.samba.org/?p=ab/samba.git/.git;a=commitdiff;h=ceecabce946dc93527e563c851e4d9fa8f95119e

greghudson:
One of the other changes I'd like to make is to apply KRB5_PRINCIPAL_PARSE_NO_DEF_REALM to all IAKERB name imports when we use krb5_parse_name(), in which case you could avoid that leading "@".  (And possibly import the same name for IAKERB and regular krb5, but only if the realm being hardcoded is the default realm for the krb5_context.)

greghudson:
I made that change and some other small ones.

I am concerned about the manipulation of cred->name->princ->realm in iakerb_set_discovered_realm(), as it is a side-effect on the claimant credential passed by the caller.  As an example of how this could cause a problem, imagine that a client imports an enterprise name, acquires a cred with it, tries to use that as a claimant cred for IAKERB to service 1, fails, and tries again with service 2.  Since the cred now has an explicit realm, the second IAKERB attempt will not use realm discovery.

abbra:
We probably need to reset discovery state if we get back to AS-REQ processing.

greghudson:
On further investigation, krb5 gss_init_sec_context()  already has side effects on the claimant cred handle before this PR, in the form of kg_cred_resolve().  The same category of failures is possible; if a claimant cred is used for two different gss_init_sec_context() operations with different targets, cache selection for the first operation could cause the wrong cache to be used for the second.  I would guess that GSSAPI consumers generally don't care; if they acquire claimant creds at all, they acquire them for one gss_init_sec_context() operation and then release them.

Fixing this is outside the scope of the current PR, if it's necessary at all.  My current understanding for the (hypothetical) fix would be to add krb5_gss_ctx_id_rec fields for the ccache, client name, and impersonator principal, set those fields with kg_cred_resolve() instead of modifying the claimant cred, and use those fields in get_credentials() etc..  IAKERB realm discovery could then follow suit.  But for this PR I think we can let stand the behavior of iakerb_set_discovered_realm().

greghudson:
I pushed some more changes.  Most notably, I removed discovered_realm from the IAKERB context as we don't need it any more; with that gone, iakerb_set_discovery_realm() wasn't doing much and could be folded into the one caller.  I also added a discovery state to the machine to avoid adding to the amount of conditional logic in the AS_REQ state.

We could conceivably check for creds in the cache after discovering the server realm, and potentially skip the AS or TGS requests.  But I don't think we expect that to be a valuable optimization given our assumptions of how IAKERB will be used.

abbra:
> I pushed some more changes. Most notably, I removed discovered_realm from the IAKERB context as we don't need it any more; with that gone, iakerb_set_discovery_realm() wasn't doing much and could be folded into the one caller. I also added a discovery state to the machine to avoid adding to the amount of conditional logic in the AS_REQ state.
> 
> We could conceivably check for creds in the cache after discovering the server realm, and potentially skip the AS or TGS requests. But I don't think we expect that to be a valuable optimization given our assumptions of how IAKERB will be used.

The latter might actually be useful. One of common use cases for Local KDC we plan is to allow machine A to have a cross-realm trust to machine B. In that case local KDC on machine A can issue the TGT to the local user principal with whatever pre-authentication mechanism they support  but not possible to initiate over GSSAPI (e.g. PKINIT, idp, passkey, or otp) and that ticket can later be used to authenticate to a remote system over IAKerb.

This wouldn't work against Windows systems as they have no support for all those pre-authentication methods aside PKINIT (and they wouldn't support PKINIT over IAKerb yet) but it will help a lot for Linux systems where we absolutely want to explore and make it useful.

abbra:
Code-wise, I think changes are fine. One thing I miss in the tests is a combination of SPNEGO + IAKerb over an application protocol. This would simulate what we do in Samba and the crash described by @cryptomilk above happens in that flow once we reuse existing service ticket in a ccache, thus skipping AS-REQ + TGS-REQ over IAKerb.

greghudson:
I'm not sure how you would do IAKERB inside SPNEGO in our implementation; as far as I recall, our SPNEGO only works with mechanisms that are available by default, which IAKERB is not.

I had glossed over the crash report.  The backtrace is somewhat obscured by the use of -O, though I can see that it has to do with iakerb_gss_inquire_sec_context_by_oid().

abbra:
The `iakerb_gss_inquire_sec_context_by_oid()` is too late -- it merely dereferences NULL inner context handle which was passed to it. That handle was supposed to be iakerb context but it was never set up during AP-REQ processing, it seems.

abbra:
@greghudson I added a small commit that should fix the problem with the crash. In case of directly proceeding to AP-REQ we end up skipping storing the internal context handle and that leads to the crash.

greghudson:
I hadn't actually pushed the changes I said I had.  That's done now.

I will move your iakerb_accept_sec_context() fix to a separate PR.  The bug it fixes is a regression in commit e04f0283516e80d2f93366e0d479d13c9b5c8c2a .

greghudson:
>> We could conceivably check for creds in the cache after discovering the server realm, and potentially skip the AS or TGS requests. But I don't think we expect that to be a valuable optimization given our assumptions of how IAKERB will be used.
> The latter might actually be useful.

I took a stab at changing the IAKERB state machine to work this way, but I think bigger changes are necessary to make it work.  If we gss_import_name() a realmless principal and do a basic gss_acquire_cred() (as opposed to gss_acquire_cred_with_password() or gss_acquire_cred_from() with a password), the library will fail out because it can't find a cache in the collection matching the realmless principal.

abbra:
> I took a stab at changing the IAKERB state machine to work this way, but I think bigger changes are necessary to make it work. If we gss_import_name() a realmless principal and do a basic gss_acquire_cred() (as opposed to gss_acquire_cred_with_password() or gss_acquire_cred_from() with a password), the library will fail out because it can't find a cache in the collection matching the realmless principal.

yes, I saw another issue last week: if both the client and the server run local KDCs and those have cross-realm agreement, then a client using ccache with TGT from its local KDC cannot acquire a service ticket to the server over IAKERB. This is due to the fact that IAKERB on the server side will attempt to find the KDC for the client's realm and since that one is only accessible to the client itself, the server side's of IAKERB will fail.

In the network trace below .221.2 is the client, .221.3 is the server. The server attempts to lookup client's realm through a locator plugin over DNS, fails and sends a failure:
```
$ tshark -r ~/Downloads/iakerb-two-realms.tcp 
    1   0.000000 192.168.221.2 → 192.168.221.3 TCP 74 51336 → 445 [SYN] Seq=0 Win=64240 Len=0 MSS=1460 SACK_PERM TSval=3001061984 TSecr=0 WS=128
    2   0.000018 192.168.221.3 → 192.168.221.2 TCP 74 445 → 51336 [SYN, ACK] Seq=0 Ack=1 Win=65160 Len=0 MSS=1460 SACK_PERM TSval=3521582579 TSecr=3001061984 WS=128
    3   0.000026 192.168.221.2 → 192.168.221.3 TCP 66 51336 → 445 [ACK] Seq=1 Ack=1 Win=64256 Len=0 TSval=3001061984 TSecr=3521582579
    4   0.000144 192.168.221.2 → 192.168.221.3 SMB2 342 Negotiate Protocol Request
    5   0.000150 192.168.221.3 → 192.168.221.2 TCP 66 445 → 51336 [ACK] Seq=1 Ack=277 Win=64896 Len=0 TSval=3521582579 TSecr=3001061984
    6   0.004212 192.168.221.3 → 192.168.221.2 SMB2 378 Negotiate Protocol Response
    7   0.004233 192.168.221.2 → 192.168.221.3 TCP 66 51336 → 445 [ACK] Seq=277 Ack=313 Win=64128 Len=0 TSval=3001061988 TSecr=3521582583
    8   0.010218 192.168.221.2 → 192.168.221.3 SMB2 1423 Session Setup Request
    9   0.012756 192.168.221.3 → 192.168.221.1 DNS 86 Standard query 0x2068 URI _kerberos.AB.LOCALKDC.SITE
   10   0.014026 192.168.221.1 → 192.168.221.3 DNS 151 Standard query response 0x2068 No such name URI _kerberos.AB.LOCALKDC.SITE SOA ns0.centralnic.net
   11   0.014163 192.168.221.3 → 192.168.221.1 DNS 91 Standard query 0xa805 SRV _kerberos._udp.AB.LOCALKDC.SITE
   12   0.014548 192.168.221.1 → 192.168.221.3 DNS 156 Standard query response 0xa805 No such name SRV _kerberos._udp.AB.LOCALKDC.SITE SOA ns0.centralnic.net
   13   0.014599 192.168.221.3 → 192.168.221.1 DNS 91 Standard query 0x3f7a SRV _kerberos._tcp.AB.LOCALKDC.SITE
   14   0.014883 192.168.221.1 → 192.168.221.3 DNS 156 Standard query response 0x3f7a No such name SRV _kerberos._tcp.AB.LOCALKDC.SITE SOA ns0.centralnic.net
   15   0.015138 192.168.221.3 → 192.168.221.2 SMB2 143 Session Setup Response, Error: STATUS_LOGON_FAILURE

```

PR #1416, opened 2025-03-21
Fix IAKERB accept_sec_context null pointer crash

greghudson:
A regression in commit e04f0283516e80d2f93366e0d479d13c9b5c8c2a leads to iakerb_gss_init_sec_context() returning successfully with *context_handle still set to null, in the case where no IAKERB steps are necessary.  Subsequent GSS calls using this context will dereference a null pointer and crash.

Fix this bug by moving the *context_handle assignment to the cleanup handler, alongside the code to clean up new contexts on error.

[ghudson@mit.edu: moved fix to cleanup handler to avoid code duplication; rewrote commit message]

### DIFF
--- a/src/appl/gss-sample/t_gss_sample.py
+++ b/src/appl/gss-sample/t_gss_sample.py
@@ -116,6 +116,13 @@ for realm in multipass_realms():
     # test default (i.e., krb5) mechanism with GSS_C_DCE_STYLE
     tgs_test(realm, ['-dce'])
 
+    mark('AP')
+    ccache_save(realm)
+    tgs_test(realm, ['-krb5'])
+    tgs_test(realm, ['-spnego'])
+    tgs_test(realm, ['-iakerb'], ['-iakerb'])
+    tgs_test(realm, ['-dce'])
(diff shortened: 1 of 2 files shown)


greghudson:
I want to add a test case to this.  That will be most easily done after PR #1415 is merged (I plan to just add a call to the rewritten t_iakerb.c), so I'll wait for that.

[Tagging @abbra]

greghudson:
Looking more closely, I think this wasn't really a regression in commit e04f0283516e80d2f93366e0d479d13c9b5c8c2a, in that it was half-broken in the same way before the commit.  Prior to that commit, iakerb_gss_accept_sec_context() would assign to *context_handle if krb5_gss_accept_sec_context_ext() returned GSS_S_COMPLETE (whether or not it was processing the first token), but if accept_sec_context returns GSS_S_CONTINUE_NEEDED as it would for mutual auth, it would exhibit the same bug.  I will adjust the commit message.

abbra:
LGTM, thank you for the tests and separating this out.

PR #1417, opened 2025-03-26
Fix IAKERB error handling

greghudson:
KRB-ERROR has non-optional realm and sname fields, so we must set a server field in the krb5_error object before calling krb5_mk_error().

Add KRB-ERROR processing to the initiator state machine.


abbra:
LGTM

PR #1418, opened 2025-03-27
Fix argument type in getsockname

RullDeef:


greghudson:
Was this bug discovered through investigation of a crash, or from reading/analysis of the source code?  If we haven't seen this crop up at runtime, I am inclined to take this bug as evidence that local_addr_fallback_kludge() is never reached, and remove the function altogether.

RullDeef:
that was discovered from running static analysis. I think that fact that there were no crashes yet is because there were no such cases when `gethostname` or `gethostbyname` would fail.
looking at `localaddr.c`, `local_addr_fallback_kludge` is reachable and cannot be removed easily.

greghudson:
My reasoning was that if local_addr_fallback_kludge() has done nothing but crash since it was introduced twenty years ago, and nobody has complained about it, then clearly no one is calling it, and we can just make the Windows version of krb5_os_localaddr() error out if gethostname() or gethostbyname() fails.

But after more reflection I've decided to go with the original fix, just to avoid uncertainty.

PR #1420, opened 2025-04-04
possible typo bug

pseudometric:
It looks like this `if` should be an `else if` (?).


greghudson:
Merged in commit 6f8292ca20bddf1de23b9f525e54a3a674c46b8b.  (I didn't have permissions to force-push to the PR branch, so github will show this PR as not merged.)

pseudometric:
I see, thanks; next time I'll see if I can adjust the PR permissions to allow that.

PR #1421, opened 2025-04-04
gssapi: Only handle IAKerb errors in initiator step

cryptomilk:
This is similar to the acceptor step which handles this correctly already. Else we return an error for NEED_PREAUTH ...

### DIFF
--- a/src/lib/gssapi/krb5/iakerb.c
+++ b/src/lib/gssapi/krb5/iakerb.c
@@ -349,8 +349,8 @@ iakerb_acceptor_realm(iakerb_ctx_id_t ctx, gss_cred_id_t verifier_cred,
         ret = krb5_get_default_realm(ctx->k5c, &defrealm);
         if (ret) {
             /* Generate an error reply if there is no default realm. */
-            ret = iakerb_mk_error(ctx->k5c, verifier_cred, KRB_ERR_GENERIC,
-                                  &reply);
+            ret = iakerb_mk_error(ctx->k5c, verifier_cred,
+                                  KRB_AP_ERR_IAKERB_KDC_NOT_FOUND, &reply);
             if (ret)
                 goto cleanup;
         } else {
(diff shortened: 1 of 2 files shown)


cryptomilk:
This fixes smbclient against our localkdc.

abbra:
I think the test needs a fix.

abbra:
@greghudson the refactoring for error handling did not take into account that an error returned by the proxied KDC is something that should be handled by the client side. For example, if pre-authentication is required, this will be returned as an error. As a result, iakerb mechanism will fail to work -- initiator side will instead exit. It should only exit when IAKerb-specific errors were returned, as described in the IAKerb specification.

abbra:
Perhaps we should first try the realm of the acceptor (verifier) cred if acceptor is not empty and only then fallback to the default realm?

Samba always uses a concrete principal loaded from its keytab. So at least in the SMB IAKerb case we would always have acceptor defined and with a specific realm.

This does not remove the need to indicate a failure of the discovery process. On the other hand, using KRB_AP_ERR_IAKERB_KDC_NOT_FOUND seems close: the initiator mech knows it was asking for the realm discovery already, it is unlikely that an IAKerb-enabled service would not know its realm at all, so this error can be differentiated from the server-to-KDC connectivity.

greghudson:
> Perhaps we should first try the realm of the acceptor (verifier) cred if acceptor is not empty and only then fallback to the default realm?

iakerb_acceptor_realm() already does this.

cryptomilk:
Hmm, this didn't work with our localkdc, we run into an error because the default realm was not set in the configuration.

greghudson:
I'd be interested in knowing where the error comes from.  The tests include a case where the acceptor has no default realm but still satisfies realm discovery.

I reminded myself of how to set an extended error message in the GSS krb5 mech and set one for the KRB_AP_ERR_IAKERB_KDC_NOT_FOUND realm discovery case, which should help disambiguate.

PR #1430, opened 2025-05-11
libkrb5: fix client preauth handling of fallback_disabled

pseudometric:
I've run across an unexpected behavior with SPAKE in MIT Kerberos 1.21.3, which may be a bug. It occurs when some KDCs have SPAKE enabled (by setting `spake_preauth_groups`), while it is off in other KDCs for the same realm. Consider:

KDC A has:

```
[libdefaults]
spake_preauth_groups = edwards25519

[kdcdefaults]
spake_preauth_kdc_challenge = edwards25519
```

while KDC B has both unset. Suppose that the only other supported preauthentication mechanism is encrypted-timestamp. The following can occur with client C:

```
C -> A : AS-REQ
A -> C : KRB5KDC_ERR_PREAUTH_REQUIRED + SPAKE challenge
C -> B : AS-REQ with SPAKE response
B -> C : KRB5KDC_ERR_PREAUTH_REQUIRED (because B does not support SPAKE)

```
At this point the client should stop: it has replaced the reply key as part of sending the SPAKE challenge, and called `disable_fallback()` to ensure it will not continue to another mechanism and use the wrong reply key (`spake_client.c`). However, it *does* in fact continue, generating an AS-REQ with encrypted-timestamp using the abandoned SPAKE reply key, which then fails no matter which KDC it picks for that next attempt. This happens because the code to respect `disable_fallback()` is this (`get_in_tkt.c`):

```
    /* Don't continue if fallback is disabled. */
    if (code && ctx->fallback_disabled)
        goto cleanup;
```

but here, `code == 0,` so this case is not triggered.

My best guess at a fix is this:

```
    /* Don't continue if fallback is disabled. */
    if (ctx->fallback_disabled) {
        if (code == 0)
            code = KRB5_PREAUTH_FAILED;
        goto cleanup;
    }
```

which seems fine in that the client stops, but `kinit(1)` gives an odd error message:

`kinit: No user identity options specified while getting initial credentials`

so I'm not sure this is quite right.

The same thing can happen without the optimistic challenge; it just requires that the client talk to a KDC with SPAKE enabled the first two times, then to one without it next (so a bit less likely). And of course sometimes the failure doesn't happen at all if the problematic sequence of KDC choices doesn't happen (random via SRV records in my case), so it occurs sporadically with frequency depending on how many SPAKE vs non-SPAKE KDCs there are.

Obviously, ordinarily all the KDCs for a realm will agree on their SPAKE configuration and so this would not happen; I noticed it when I enabled SPAKE in just one KDC out of several as a test. Nonetheless, this behavior seems like a bug.

Separately, I’m also wondering why fallback must be disabled in the first place in this situation. It certainly seems better, ideally, if the client can continue to try another shared-key preauthentication method, just resetting the reply key to its initial one when starting over. I imagine it has something to do with the general preauthentication state model laid out in RFC-6113?

Thanks,

-- Richard

### DIFF
--- a/src/lib/krb5/krb/get_in_tkt.c
+++ b/src/lib/krb5/krb/get_in_tkt.c
@@ -1331,9 +1331,6 @@ init_creds_step_request(krb5_context context,
     /* Don't continue after a keyboard interrupt. */
     if (code == KRB5_LIBOS_PWDINTR)
         goto cleanup;
-    /* Don't continue if fallback is disabled. */
-    if (code && ctx->fallback_disabled)
-        goto cleanup;
     if (code) {
         /* See if we can try a different preauth mech before giving up. */
         k5_save_ctx_error(context, code, &save);
--- a/src/lib/krb5/krb/init_creds_ctx.h
+++ b/src/lib/krb5/krb/init_creds_ctx.h
@@ -63,9 +63,9 @@ struct _krb5_init_creds_context {
     krb5_enctype etype;
     krb5_boolean info_pa_permitted;
     krb5_boolean restarted;
-    krb5_boolean fallback_disabled;
     krb5_boolean encts_disabled;
     struct krb5_responder_context_st rctx;
+    krb5_preauthtype current_preauth_type;
     krb5_preauthtype selected_preauth_type;
     krb5_preauthtype allowed_preauth_type;
     k5_json_object cc_config_in;
--- a/src/lib/krb5/krb/preauth2.c
+++ b/src/lib/krb5/krb/preauth2.c
@@ -552,7 +552,9 @@ set_cc_config(krb5_context context, krb5_clpreauth_rock rock,
 static void
 disable_fallback(krb5_context context, krb5_clpreauth_rock rock)
 {
-    ((krb5_init_creds_context)rock)->fallback_disabled = TRUE;
+    krb5_init_creds_context ctx = (krb5_init_creds_context)rock;
+
+    ctx->allowed_preauth_type = ctx->current_preauth_type;
 }
 
 static struct krb5_clpreauth_callbacks_st callbacks = {
@@ -676,6 +678,7 @@ process_pa_data(krb5_context context, krb5_init_creds_context ctx,
             if (real && previously_failed(ctx, pa->pa_type))
                 continue;
             mod_pa = NULL;
+            ctx->current_preauth_type = pa->pa_type;
             ret = clpreauth_process(context, h, modreq, ctx->opt, &callbacks,
                                     (krb5_clpreauth_rock)ctx, ctx->request,
                                     ctx->inner_request_body,
@@ -908,6 +911,7 @@ k5_preauth_tryagain(krb5_context context, krb5_init_creds_context ctx,
     if (h == NULL)
         return KRB5KRB_ERR_GENERIC;
     mod_pa = NULL;
+    ctx->current_preauth_type = pa_type;
     ret = clpreauth_tryagain(context, h, modreq, ctx->opt, &callbacks,
                              (krb5_clpreauth_rock)ctx, ctx->request,
                              ctx->inner_request_body,
@@ -954,6 +958,7 @@ fill_response_items(krb5_context context, krb5_init_creds_context ctx,
         h = find_module(context, ctx, pa->pa_type, &modreq);
         if (h == NULL)
             continue;
+        ctx->current_preauth_type = pa->pa_type;
         ret = clpreauth_prep_questions(context, h, modreq, ctx->opt,
                                        &callbacks, (krb5_clpreauth_rock)ctx,
                                        ctx->request, ctx->inner_request_body,


pseudometric:
And now I see that this is definitely not right, since it causes PKINIT tests to fail. I assume in some cases the reply key will have been replaced and `code == 0` because everything has worked, but it still reaches that point in the code, and so it fails incorrectly.

Before digging into this further though, I'd like to hear what MIT folks think about this situation.

greghudson:
More context on the disable_fallback variable can be found here: https://krbdev.mit.edu/rt/Ticket/Display.html?id=8654

The initial motivation is that if SPAKE password authentication fails due to a mistyped password, we don't want to transmit a dictionary-attackable ciphertext in the course of failing, as that could allow a passive eavesdropper to recover the mistyped password.

In the scenario you describe, the client behavior does appear to be a poor reaction to an unexpected KDC response.  The client doesn't really expect the KDC to respond with PREAUTH_REQUIRED to a preauthenticated request.  When it does (because KDC B doesn't see the request as preauthenticated), that slips the fallback_disabled net.

I think a fix would have to specifically look for a PREAUTH_REQUIRED response with fallback_disabled set.  I believe the current PR fix breaks PKINIT when the KDC returns an error like KDC_ERR_DH_KEY_PARAMETERS_NOT_ACCEPTED after the client makes a preauthenticated request.  In that case we go through the k5_preauth_tryagain() clause to generate a new preauthenticated request, and when that succeeds we have code == 0 and ctx->fallback_disabled set.  SPAKE 2FA (if it were implemented) could also run afoul of this, as we would pass through the "ctx->more_padata != NULL" clause after setting fallback_disabled.

pseudometric:
> I think a fix would have to specifically look for a PREAUTH_REQUIRED response with fallback_disabled set.

Thanks. That's what I was thinking too; it just seemed a little ad-hoc, so I wondered if there were a more general way to address this that I'm not seeing. I'll make a stab at fixing it this way, then.

greghudson:
I think there might be a more robust way to handle this, by moving the logic into preauth2.c.  We have support for setting a single allowed "real" preauth type, but currently only set that field by loading it from the ccache.  We could start setting that field when a mech calls disable_fallback(), which would prevent other mechs from being invoked later in the exchange.  Then we wouldn't need the fallback_disabled field or its check in get_in_tkt.c.

pseudometric:
> I think there might be a more robust way to handle this…

That does sound better, and I’m happy to try it instead if you want. But, I’m a little unsure about the whole notion of disabling fallback to begin with. I understand this:

> The primary motivation is that if SPAKE password authentication fails due to a mistyped password, we don't want to transmit a dictionary-attackable ciphertext in the course of failing, as that could allow a passive eavesdropper to recover the mistyped password.

… but I think there’s more to it: Does the mere fact that the client tried to use SPAKE really indicate that it *requires* the higher level of security SPAKE provides? One could counter that the fact that the client and KDC both have encrypted-timestamp enabled means that they are both willing to use it, and the client would just *prefer* to use SPAKE if it’s available and works. After all, if the client doesn’t want to use encrypted-timestamp, it can just turn it off.

This actually has practical implications: I would like to deploy SPAKE at work. Since it’s new, I was assuming I’d roll it out as the preferred preauth mechanism, knowing that clients can fall back to the existing encrypted-{timestamp,challenge} if the SPAKE fails for some reason. We log the preauth mechanism used, so we can track this. But if I understand correctly, SPAKE is actually right now hard-coded to *not* allow that fallback. And in fact, it would not fall back to encrypted-challenge either, correct? — even though that doesn’t have the same cryptographic weakness. That’s what the code suggests anyway, I haven’t tried it yet. Am I missing some elements of all this?

I’ll also note that in our case, we’re not actually much worried about dictionary attacks; almost all our clients in the realms at issue are software identities using randomly generated keys AES keys. So while we do want to move away from encrypted-timestamp just on general grounds, this motivation for prohibiting fallback to encrypted-timestamp doesn’t carry much weight for us.

I realize this may all seem a bit academic, since (as the RFC says) normally all the shared-key preauthentication methods will either all work or all fail. But things do not always work normally :), and the lack of a fallback when rolling out a new preauth mechanism will really give me pause. In fact, now that I’ve thought about it I’m pretty sure I will add it as an option myself to our MIT Kerberos build, if we don’t end up adding it here.

greghudson:
> Does the mere fact that the client tried to use SPAKE really indicate that it requires the higher level of security SPAKE provides?

If the client and KDC negotiated a SPAKE group, then the client has strong evidence that it is not necessary to expose the long-term key to dictionary attacks in order to conduct an AS exchange.

To draw an analogy, a client and KDC might both be willing to tolerate rc4-hmac for the sake of interoperating with old Microsoft endpoints.  But it would be pretty bad if a client, after failing an AS-REQ with aes256-cts encrypted timestamp, tried again with rc4-hmac (which is unsalted) before giving up.  It would be doing pointless extra work, and it would be making dictionary attacks by passive eavesdroppers easier by multiple orders of magnitude.  The same reasoning applies to falling back from SPAKE to encrypted timestamp.

The more general reasoning behind commit 7a24a088c16d326127dd2b29084d4ca085c70d10 is that once a client has negotiated support for a mechanism, acquired credentials, and constructively tried to authenticate, trying a different mechanism is unlikely to succeed and is somewhat likely to cause a problem of some sort.  The problem might be UX-related rather than security-related; for example, when we fall back from PKINIT to encrypted timestamp because the principal entry has long-term keys (typically randomized and in nobody's possession but the KDC's), we ask for a password rather than reporting a potentially useful error message about the PKINIT failure.

I consider these concerns more important than concerns that SPAKE preauth might fail after negotiating a group.  The scenario of mixed KDC support within a realm is interesting, but that's a configuration that could lead to hard-to-diagnose failures for any multi-step preauth mechanism.

> And in fact, it would not fall back to encrypted-challenge either, correct? — even though that doesn’t have the same cryptographic weakness. That’s what the code suggests anyway, I haven’t tried it yet.

That is correct, and is a consequence of the more general reasoning about fallback.  Assuming that the security of the FAST channel holds, falling back to encrypted challenge would be pointless additional work but not a weakening of security.

> I’ll also note that in our case, we’re not actually much worried about dictionary attacks; almost all our clients in the realms at issue are software identities using randomly generated AES keys.

I can see how that would affect your risk calculus, but I also don't think there is a lot of upside to deploying SPAKE in this environment.

pseudometric:
Thanks for explaining. It’s true, we certainly don’t have to use SPAKE just because it’s there. I want to add FAST to these AS-REQs separately to improve the basic security of the exchange; once that happens it will use encrypted-challenge, and we could call that good enough. Part of the issue is that we do have *some* human users in these realms, for admin and other role principals. I certainly want to get away from encrypted-timestamp for them. We can use wrappers around human tools to arrange that like so by setting `KRB5_CONFIG` with e.g.:

```
[libdefaults]
preferred_preauth_types = ...

[plugins]
clpreauth = {
  disable = encrypted_timestamp
}
```

but it’s a bit awkward, which pushes toward the idea of using SPAKE globally if we’re going to use it at all. We could of course just not use passwords at all for humans by removing the keys from those principals, but it’s not always feasible to provide OTP or PKINIT for them since that requires more setup. I’ve thought about adding a principal string attribute to control which preauth methods the KDC will offer for a given principal. We already get that effect in part with OTP, which is only offered if the principal has an `otp` attribute, and for shared-key preauth in that those methods are only offered if the principal has long-term keys attached. This would generalize it.

Anyway, back to the topic of this PR: this is no longer really an issue for us, now that I understand what’s going on. I was concerned when I turned on SPAKE and saw these sporadic encrypted-timestamp preauth failures, but now we know that it’s just a side effect of having a mixed configuration, and it won’t happen when we apply it to all the KDCs. Do you want to improve this, or just document it as a known effect and leave it as is? I can try a patch to use the preauth method constraint you mentioned instead to replace the existing fallback flag, if you want. Or, maybe you want to put on your to-do list for your team to address at some point. Just let me know.

greghudson:
I pushed a commit implementing what I had in mind.

pseudometric:
Cool; I'll try it out and report back.

PR #1432, opened 2025-05-17
Implement fault tolerance for empty TXT record strings

evyzacq:
When the TXT-DATA value returned by DNS is empty, it causes the program to crash. 
This fix adds fault tolerance to prevent the program from crashing in such cases.


greghudson:
Thanks for submitting this.  I have pushed a modified commit which moves the check and adds an additional check, as well as changing the commit message.  Let me know if it still seems okay.

evyzacq:
Thanks for the update! It's totally fine to move the checks ahead, but just a note — the new placement already covers the check at line 477, so that line can be removed to avoid redundancy.

evyzacq:
Looks good to me, thanks for the updates! Let’s get it merged.

PR #1434, opened 2025-05-20
kdb/ldap: fix error handling in krb5_ldap_read_policy()

naya451:
In case policy = NULL, program goes to cleanup and NULL-dereference happens. Return without goto to avoid it.


greghudson:
I reviewed libkdb_ldap to find similarly problematic functions, and made additional changes based on what I found.

I am considering removing the defensive null checks in libkdb_ldap instead.  They are very spotty--mostly only in policy-related functions--and the risk would be low as libkdb_ldap isn't a general-use library.

PR #1435, opened 2025-05-22
Refactor dnsglue functions using k5input

greghudson:
Use k5input to make the dnsglue.c and dnssrv.c functions simpler and more robust.  Rename the dnsglue functions to use the now-preferred k5_ prefix instead of krb5int_.  Use stdint type names instead of unsigned char and unsigned short.  Remove some out-of-date comments.

### DIFF
--- a/src/lib/krb5/os/dnsglue.c
+++ b/src/lib/krb5/os/dnsglue.c
@@ -37,34 +37,24 @@
 #endif
 
 /*
- * Only use res_ninit() if there's also a res_ndestroy(), to avoid
- * memory leaks (Linux & Solaris) and outright corruption (AIX 4.x,
- * 5.x).  While we're at it, make sure res_nsearch() is there too.
- *
- * In any case, it is probable that platforms having broken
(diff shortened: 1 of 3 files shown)


greghudson:
The query loop in krb5int_dns_init() is more problematic than I initially gave it credit for:

* The comparison "(size_t) len > maxincr" is equivalent to "len < 0" since len is of type int and maxincr is INT_MAX.  We have another check for "len < 0" a few lines later, which is redundant.

* If len is larger than the current buffer size, we search for a power of two that is at least len and try again with that buffer size.  But we're never going to use this buffer again.  Why not allocate at least len bytes?

* This code expects SEARCH() to return the necessary buffer length if the answer is larger than anslen bytes.  In reality, res_nsearch() returns anslen.  This behavior might have changed in 2002 due to https://www.kb.cert.org/vuls/id/738331 . 
(I'm not sure how macOS's dns_search() works in this regard; there is no documentation and I couldn't find source code for it.)  So I think in practice this code only works because answers always fit within a 4K buffer.

PR #1436, opened 2025-06-02
pkinit: fix error handling in pkinit_server_verify_padata()

naya451:
In case pkinit_init_kdc_req_context() fails to allocate reqctx, NULL-dereference in cleanup is possible.
In this point there is no need to free any resources so it is possible to respond and return immediatly.


greghudson:
In this case I think it is more robust to fix the cleanup handler than to avoid it.  cleanup labels in this code base are supposed to execute correctly if any subset of the allocated locals are null.  I pushed a change to take this approach instead.

PR #1439, opened 2025-06-23
util/profile: fix NULL-dereference of iter->file

naya451:
iter->file may be NULL. Add a check before dereference.


greghudson:
I don't think iter->file can be null when iter->node is non-null.  Can you identify a case where it can be?

My reasoning: iter->node is only assigned a non-null value at lines 608 and 644.  On line 608, iter->file being null is ruled out by line 551.  On line 644, iter->node already had a non-null value.  Also, iter->file can only go from non-null to null at lines 601 and 636, and in both of those cases iter->null is assigned null.

PR #1440, opened 2025-06-24
Avoid undefined memcpy in asn1_encode.c

greghudson:
The C standard specifies that passing null pointers to most standard library functions results in undefined behavior (C99 7.1.4).  This applies to memcpy() even when the length is 0.  insert_bytes() in asn1_encode.c may be called with a null pointer from an empty krb5_data or other counted value in a structure to be encoded.  Do not call memcpy() in this case.


greghudson:
This was actually merged in commit 4b4a720cacec8827c9b3f65b4920ac1b0075f38e, but I forgot to push to the PR branch first.

PR #1445, opened 2025-07-26
keyring ccache tests can fail if session keyring revoked

greghudson:
On Linux, the tests in lib/krb5/ccache rely on the session keyring being valid.  In unusual process environments this won't be true, such as when a buildbot worker is started from a login session which is then closed.  In this case the test suite will fail with an error containing "Key has been revoked".  The failure can be simulated with `keyctl revoke @s` within a disposable login session.  Both t_cccol.py and t_cc.c are affected.

We could ignore this problem as it is rare, or we could address it in one of two ways:

1. Make the tests work in this condition.  This could be done using the "keyctl session" command-line process wrapper, or the keyctl_join_session_keyring() API within a process.

2. Skip the keyring tests in this condition, by probing the session keyring (`keyctl list @s`).

Both of these options are awkward given the current structure of the tests.  We could do (1) via Makefile.in using an autoconf test, defining a make variable containing "keyctl session -" when keyctl is available or nothing when it isn't, and using that variable as a command prefix when invoking t_cccol.py and t_cc.  Or we could do it within t_cccol.py and t_cc.c, at the cost of adding a test suite dependency for the Python keyutils bindings.

For (2), in t_cccol.py we want to run `keyctl list @s` with output ignored, and skip the keyring tests if the exit status isn't 0.  In t_cc.c we would need to add a conditional call to something like keyctl_describe().

The PR branch currently contains an implementation of (2) for t_cccol.py but not t_cc.c.  (1) has the tidiness benefit of better isolating the test suite from the containing login session, so I may switch to that approach.


greghudson:
I switched to an implementation of (1) using autoconf to define a command wrapper for Makefile.in.

t_cc could still fail if libkeyutils is available, keyctl is not, and the session keyring is revoked.  This could be fixed by doing the session join inside t_cc, but I will consider that scenario too rare to be worth worrying about for now.

RUNPYTEST does not play nicely with command wrappers as it is a combination of an environment variable setting and a command.  My workaround is to use "env" in between the command wrapper and `$(RUNPYTEST)`.

greghudson:
There are also keyring tests in tests/t_ccache.py, which the current PR commit misses.

PR #1446, opened 2025-08-01
Reunite xdrproc_t with its comment

neuschaefer:


greghudson:
I tracked down where the comment got separated (it was in my revisions to a kenh PR from 2023) and amended the commit message.

The comment also seems pretty inaccurate, so I edited it down to something more correct.  The same comment appears in the Solaris and libtirpc versions of xdr.h, but (1) the second argument is a direct pointer to the data object, not "a pointer to an opaque pointer", and (2) the per-type functions do not check whether the second argument is (or points to) null to decide whether to allocate memory; they allocate when the XDR handle's mode is decoding and the data type is a pointer type (xdr_pointer()/xdr_array()).

neuschaefer:
@greghudson thanks for looking into this!

PR #1447, opened 2025-08-03
FAST: obtain anonymous PKINIT as an armor automatically

abbra:
For multi-factor preauthentication methods one has to use an armor ccache to create a FAST channel. This can be done by calling kinit twice:

 kinit -c ./fast_armor @REALM
 kinit -T ./fast_armor principal@REALM

Automate these steps by using a memory ccache.

### DIFF
--- a/src/man/krb5.conf.man
+++ b/src/man/krb5.conf.man
@@ -665,6 +665,19 @@ servers yet.  New in release 1.19.
 Specifies the name of the host\(aqs site for the purpose of DNS\-based
 KDC discovery for this realm.  New in release 1.22.
 .TP
+\fBauto_fast_armor\fP
+If this flag is true, then initial ticket request will use Anonymous
+PKINIT to protect the communication as a FAST channel in case an application
+did not provide its own FAST channel. This is useful for deployments where
+pre-authentication methods require use of the FAST channel, such as
+passwordless methods provided by FreeIPA. Microsoft Active Directory
+implementation of PKINIT does not support Anonymous PKINIT feature.
+As a result, \fIauto_fast_armor\fP defaults to false.
+.sp
+Use of \fIauto_fast_armor = true\fP requires properly configured PKINIT and
+WELLKNOWN/ANONYMOUS principal defined on the KDC side. Consult KDC documentation
+for details.
+.TP
 \fBv4_instance_convert\fP
 This subsection allows the administrator to configure exceptions
 to the \fBdefault_domain\fP mapping rule.  It contains V4 instances
--- a/src/appl/gss-sample/t_gss_sample.py
+++ b/src/appl/gss-sample/t_gss_sample.py
@@ -142,4 +142,43 @@ for realm in multipass_realms():
     kt_test(realm, ['-iakerb'], ['-iakerb'])
     kt_test(realm, ['-dce'])
 
+# Test IAKerb with auto_fast_armor: exercise the IAKERB_ARMOR_AS_REQ state,
+# where the initiator transparently acquires an Anonymous PKINIT armor ticket
+# through the IAKerb proxy before sending the main password-based AS-REQ.
+if pkinit_enabled:
+    mark('IAKerb auto_fast_armor')
+
+    armor_krb5_conf = {'realms': {'$realm': {'auto_fast_armor': 'true'}}}
+    armor_kdc_conf = {'realms': {'$realm': {
+        'pkinit_eku_checking': 'none'}}}
+
+    armor_realm = K5Realm(krb5_conf=armor_krb5_conf, kdc_conf=armor_kdc_conf,
+                          pkinit=True)
+    armor_realm.addprinc('WELLKNOWN/ANONYMOUS')
+
+    # IAKerb realm discovery followed by armor acquisition and password AS-REQ,
+    # all routed through the IAKerb proxy (no direct KDC contact).
+    pw_test(armor_realm, ['-iakerb'], ['-iakerb'])
+
+    armor_realm.stop()
+
+    mark('IAKerb auto_fast_armor with preauth required')
+
+    preauth_kdc_conf = {'realms': {'$realm': {
+        'pkinit_eku_checking': 'none',
+        'default_principal_flags': '+preauth'}}}
+
+    preauth_realm = K5Realm(krb5_conf=armor_krb5_conf,
+                            kdc_conf=preauth_kdc_conf, pkinit=True)
+    preauth_realm.addprinc('WELLKNOWN/ANONYMOUS')
+
+    # Same as above but with pre-authentication required for all principals,
+    # as in a typical FreeIPA deployment.  The armor exchange requires an
+    # extra round trip (KDC_ERR_PREAUTH_REQUIRED for WELLKNOWN/ANONYMOUS),
+    # which previously triggered a use-after-free on the armor init-creds
+    # options and caused "No credentials cache found".
+    pw_test(preauth_realm, ['-iakerb'], ['-iakerb'])
+
+    preauth_realm.stop()
+
 success('GSS sample application')
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -2432,4 +2432,9 @@ k5_print_addr(const struct sockaddr *sa, char *buf, size_t len);
 void
 k5_print_addr_port(const struct sockaddr *sa, char *buf, size_t len);
 
+/* Check if Automated PKINIT is allowed for the realm */
+krb5_boolean
+k5_fast_is_pkinit_allowed(krb5_context context, krb5_data *realm);
+
+
 #endif /* _KRB5_INT_H */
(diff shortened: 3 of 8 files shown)


abbra:
I tried to add a test to t_otp.py but haven't been satisfied with a result. Maybe we can have a separate test. 

Otherwise, tested with my FreeIPA deployment which has anonymous pkinit enabled by default. 

A missing part is an error recovery in case pkinit anchors aren't configured because we would get asked to enter a password for the well-known principal. Maybe we should check for the anchors being configured?

abbra:
One more thought: perhaps it makes sense to automatically trigger this path in case pkinit anchors are configured and no armor ccache was specified. This would be preferred for environments like FreeIPA but we can also add a `libdefaults` setting to enable it.

greghudson:
References to a 2014 discussion on this topic (don't take anything therein as gospel, even if I said it):

https://mailman.mit.edu/pipermail/krbdev/2014-May/011958.html
https://mailman.mit.edu/pipermail/krbdev/2014-June/011969.html

> A missing part is an error recovery in case pkinit anchors aren't configured because we would get asked to enter a password for the well-known principal. Maybe we should check for the anchors being configured?

Ideally WELLKNOWN/ANONYMOUS won't have keys so encrypted timestamp won't be offered.  However, I think `addprinc -nokey` didn't exist when anonymous support was first added, so our documentation recommends `kadmin -q 'addprinc -randkey WELLKNOWN/ANONYMOUS'` instead of `kadmin addprinc -nokey WELLKNOWN/ANONYMOUS`.  So we probably want the client code to know that a fully anonymous ticket request should only use PKINIT.

> perhaps it makes sense to automatically trigger this path in case pkinit anchors are configured and no armor ccache was specified. This would be preferred for environments like FreeIPA but we can also add a libdefaults setting to enable it.

I do like the idea of handling this inside krb5_get_init_creds_*() rather than kinit.  But I think the vast majority of realms don't support anonymous PKINIT, so trying it without good evidence that it's desired is more likely to degrade performance than make an authentication work.  Because PKINIT has a client configuration requirement, I don't think it's too burdensome to require an additional profile setting (probably a realm variable, not libdefaults) to automate using anonymous armor.

The 2014 thread does contain some ideas for getting rid of the PKINIT anchor configuration requirement, in order to achieve a zero-client-configuration path to automatic FAST.  I don't think any of the presented ideas are fully baked, and we are more likely to get something done if we don't go down that path.  But if we did magically get rid of the anchor configuration requirement for PKINIT, it might make sense to look into another one of that thread's ideas: interpreting an initial PREAUTH_REQUIRED error containing PA-FX-FAST and no real preauth mechanisms as a hint to try anonymous PKINIT for FAST armor.  (Note that a current MIT KDC will typically offer PKINIT if anonymous PKINIT is turned on, because it has no idea which client principals are expected to be able to use PKINIT.  So some kind of KDC change would be required to offer a FAST-only response.)


A final note: in the linked thread I talk about caching the wellknown/anonymous ticket.  I no longer think it's important to worry about caching that ticket from an efficiency perspective.  The "default FAST ccache" idea still has merit as a way for pam_krb5/pam_sssd to set up a keytab-based FAST armor ticket for the user, but I don't think we need to pursue that here.  (Handing the user unrestricted access to a keytab-based armor ticket relies heavily on AD-fx-fast-armor for security.  By my understanding, Heimdal implemented AS FAST support in 2010 but didn't implement AD-fx-fast-armor until 2021, so that line of armor ticket deployment is questionable.)

abbra:
Thank you for these references. I forgot about that discussion completely.

Getting to `krb5_get_init_creds_*()` proposal, I think it could be a good compromise as it covers all apps, pretty much, not just kinit. The only worry I have is that of GSSAPI krb5 behavior with collection-based ccaches: if we would use a memory-based ccache to store a FAST ccache, could it be treated within krb5 GSSAPI mechanism to pick up a wrong credential in unrelated operations?

greghudson:
> The only worry I have is that of GSSAPI krb5 behavior with collection-based ccaches: if we would use a memory-based ccache to store a FAST ccache, could it be treated within krb5 GSSAPI mechanism to pick up a wrong credential in unrelated operations?

Memory caches aren't added to the global collection (since commit 49bb627fed70c5258c151c5135ac3d95ed1ee55d).

We don't necessarily need to cache the armor ticket at all, although the existing get_in_tkt.c internals may make it easier if we do.

abbra:
@greghudson I implemented support for automatically acquiring Anonymous PKINIT if `pkinit_anchors` is configured for the realm (or globally). It works for me with FreeIPA and allows to automatically handle passwordless authentication methods.

```bash
# using Fedora's krb5 build: 
[admin@dc ~]$ kinit admin
Password for admin@MINIMAL.TEST: 
kinit: Password read interrupted while getting initial credentials

# using custom krb5 build with this patch:
[admin@dc ~]$ cd krb5/src
[admin@dc src]$ LD_LIBRARY_PATH=lib clients/kinit/kinit 
Enter OTP Token Value: 
[admin@dc src]$ klist -C
Ticket cache: KCM:61000
Default principal: admin@MINIMAL.TEST

config: pa_type(krbtgt/MINIMAL.TEST@MINIMAL.TEST) = 141
config: fast_avail(krbtgt/MINIMAL.TEST@MINIMAL.TEST) = yes
09/21/25 08:28:05  09/22/25 07:59:41  krbtgt/MINIMAL.TEST@MINIMAL.TEST
```

abbra:
I haven't looked into failing `t_pkinit.py` yet.

abbra:
Ok, the issues fixed (avoided leaking internal request's retval, only store the internal ccache in the state rather than gic opts).

I'm not sure how to address Active Directory use case without adding an explicit configuration option because AD DCs do not support Anonymous PKINIT and we have no way to detect that library-wise. PKINIT can be enabled in AD (and in most federal government deployments is enabled because they rely on CAC) but Anonymous PKINIT will be missing, so we cannot default to use of Anonymous PKINIT for all environments...

greghudson:
Per the preceding conversation, I thought we were planning to add a new realm variable.  I suggest `auto_fast_armor`.

abbra:
Ok, should I still check for the presence of the `pkinit_anchors` or just ignore a failure like I do now? Assume whoever set the `auto_fast_armor = true` has done the other setup as well?

abbra:
Modified. Now `auto_fast_armor` defaults to `false` and documentation has been updated.

greghudson:
pkinit_libdefault_strings() was a mistake in my opinion.  I do not want to see that kind of logic propagating its way into libkrb5.  Just check the realm section.

abbra:
Changed.

greghudson:
krb5int_fast_as_armor() is called from restart_init_creds_loop(), which is part of the async AS state machine.  These changes call the synchronous krb5_get_init_creds_password() to get the armor ticket.  So, for example, an IAKERB client would attempt to do anonymous PKINIT directly instead of via the GSS exchange.

(I think there is already a bug of this nature, but I can't find the writeup and I think it's pretty obscure.  I think it's when we need a cross-realm TGT for TGS FAST armor, or something.)

Modifying the async state machine to handle this is of course harder than writing synchronous code.  I can work on it, but it might be a while because of some internal MIT work.

abbra:
Thanks for the analysis. I will read through the async AS state machine to see how we can add the condition for this subrequest.

src/man/krb5.conf.man
```diff
@@ -665,6 +665,19 @@ servers yet.  New in release 1.19.
 Specifies the name of the host\(aqs site for the purpose of DNS\-based
 KDC discovery for this realm.  New in release 1.22.
 .TP
+\fBauto_fast_armor\fP
+If this flag is true, then initial ticket request will use Anonymous
+PKINIT to protect the communication as a FAST channel in case an application
+did not provide its own FAST channel.This is useful for deployments where
```
Carbenium, line 671: Just a small typo.
```suggestion
did not provide its own FAST channel. This is useful for deployments where
```

abbra:
One bug in OTP pre-auth found by Fedora rawhide testers: when GSSAPI is in use, we cannot acquire cred that expired and requires any prompting. The bug was hidden because in order to even acquire OTP, one needs to use armor and GSSAPI doesn't provide armor by itself. With automated FAST armor we proceed to handle OTP preauth and crash.

greghudson:
My preferred fix for that bug is to test for `prompter == NULL` in doprompt().

Based on other uses of the prompter in the code base, the appropriate error code for not having a prompter is either EIO ("Input/output error", used by gic_pwd.c and PKINIT) or KRB5_LIBOS_CANTREADPWD ("Cannot read password", used by PKINIT PKCS11 and SAM-2 preauth).  Technically we're reading a PIN rather than a password, so I guess EIO is slightly more correct here.

greghudson:
I merged the OTP fix.

abbra:
@greghudson as https://github.com/krb5/krb5/pull/1534 was merged, I think we can close this PR. Do you agree?

greghudson:
Yes, closing.

PR #1451, opened 2025-09-03
fix mechListMIC tag in get_negTokenResp

dovsiannikov:
The SPNEGO `NegTokenTarg` ASN.1 definition as per rfc2478 specifies the optional `mechListMIC` field as:

    mechListMIC    [3] OCTET STRING OPTIONAL

During token parsing we look for a tag `0xA4` while in practice the generated tokens contain tag `0xA3.`  As a result the MIC field is not extracted, leading to various issues like incorrect sequence numbering and `GSS_S_GAP_TOKEN` later.


greghudson:
I added a test case (or rather, restored an existing one in t_spnego.c that stopped working some time ago) that would have caught this bug.

PR #1452, opened 2025-09-10
Fix TSan report in Kerberos

rschu1ze:
We ([ClickHouse](https://github.com/ClickHouse/ClickHouse/)), carried this small patch in our krb5 fork around since two years and we like to upstream it. It addresses an issue found by [tsan](https://clang.llvm.org/docs/ThreadSanitizer.html). I unfortunately did not find an associated original error report (e.g. crashdump), so I can't tell you more information. My kind ask would be that you double-check and suggest improvements to this fix, thanks.

### DIFF
--- a/src/util/profile/prof_file.c
+++ b/src/util/profile/prof_file.c
@@ -220,9 +220,11 @@ errcode_t profile_open_file(const_profile_filespec_t filespec,
     }
     if (data) {
         data->refcount++;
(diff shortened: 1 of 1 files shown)


greghudson:
I adjusted the fix so that we only lock the data object once, and fleshed out the commit message.

rschu1ze:
Thanks, and feel free to merge (I just see a message "2 workflows awaiting approval" and can't merge by myself).

PR #1453, opened 2025-09-10
Fix MSan report in prng

rschu1ze:
We ([ClickHouse](https://github.com/ClickHouse/ClickHouse/)), carried this small patch in our krb5 fork around since four years and we like to upstream it. It addresses an issue found by [msan](https://clang.llvm.org/docs/MemorySanitizer.html). I unfortunately did not find an associated original error report (e.g. crashdump), so I can't tell you more information. My kind ask would be that you double-check and suggest improvements to this fix, thanks.

### DIFF
--- a/src/lib/crypto/krb/prng.c
+++ b/src/lib/crypto/krb/prng.c
@@ -26,6 +26,12 @@
 
 #include "crypto_int.h"
 
+#if defined(__has_feature)
+#if __has_feature(memory_sanitizer)
+#include <sanitizer/msan_interface.h>
+#endif
+#endif
+
 krb5_error_code KRB5_CALLCONV
 krb5_c_random_seed(krb5_context context, krb5_data *data)
 {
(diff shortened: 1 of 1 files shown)


greghudson:
This is an issue in msan rather than an issue found by msan.  It has been discussed previously in PR #1386, although it may be a bit confusing to follow that discussion as get_os_entropy() was using syscall() instead of getentropy() at the time.  There is no "interceptor" in msan for getentropy() in https://github.com/llvm/llvm-project/commits/main/compiler-rt/lib/msan/msan_interceptors.cpp so msan can't see that the result is no longer uninitialized.  The same applies to strlcpy() and strlcat(), which we use.

I am reluctant to add workarounds for misbehaving sanitization tools to the MIT krb5 code base, as there is potentially no limit to the amount of goop that would need to be added to make every tool happy.

rschu1ze:
Thanks for the nice explanation. I agree that we shouldn't have workarounds for msan in upstream krb5.

PR #1457, opened 2025-09-22
Add k5test --trace option

greghudson:
Make it easier to collect trace logs for commands invoked by test scripts.  Trace output will be placed in testtrace.<cmdnum> for daemons, or collected in the log for regular command invocations.

### DIFF
--- a/src/util/k5test.py
+++ b/src/util/k5test.py
@@ -595,7 +595,7 @@ def _find_symbolizer():
 # be used by the test script.
 def _parse_args():
     global args, verbose, testpass, _debug, _debugger_command
-    global _stop_before, _stop_after, _shell_before, _shell_after
+    global _stop_before, _stop_after, _shell_before, _shell_after, _trace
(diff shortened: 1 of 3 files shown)


src/util/k5test.py
```diff
@@ -621,6 +623,7 @@ def _parse_args():
     _stop_after = _parse_cmdnum('--stop-after', options.stopa)
     _shell_before = _parse_cmdnum('--shell-before', options.shellb)
     _shell_after = _parse_cmdnum('--shell-after', options.shella)
+    _trace = _parse_cmdnum('--shell-after', options.trace)
```
abbra, line 626: `_parse_cmdnum('--trace', ..)` instead of `--shell-after`

abbra (changes requested):

src/util/k5test.py
greghudson: Fixed, thanks.

abbra:
LGTM

PR #1471, opened 2025-12-10
Fix use of strchr with new GCC

abbra:
According to C specification, 7.28.5.1, strchr() is a generic function and its string argument's type is promoted to the result. So if it is const char*, the result will be const char* as well.

gcc in Rawhide (15.2.1 20251111) implements this conformance, resulting in a compilation fail with -Werror=discarded-qualifiers.

We either have to change target variable definition to be 'const char *' or cast the argument to 'char *'.

### DIFF
--- a/src/lib/krb5/ccache/ccbase.c
+++ b/src/lib/krb5/ccache/ccbase.c
@@ -201,8 +201,8 @@ krb5_cc_register(krb5_context context, const krb5_cc_ops *ops,
 krb5_error_code KRB5_CALLCONV
 krb5_cc_resolve (krb5_context context, const char *name, krb5_ccache *cache)
 {
-    char *pfx, *cp;
-    const char *resid;
+    char *pfx;
+    const char *cp, *resid;
     unsigned int pfxlen;
     krb5_error_code err;
     const krb5_cc_ops *ops;
--- a/src/lib/krb5/os/expand_path.c
+++ b/src/lib/krb5/os/expand_path.c
@@ -454,7 +454,8 @@ k5_expand_path_tokens_extra(krb5_context context, const char *path_in,
 {
     krb5_error_code ret;
     struct k5buf buf;
-    char *tok_begin, *tok_end, *tok_val, **extra_tokens = NULL, *path;
+    const char *tok_begin, *tok_end;
+    char *tok_val, **extra_tokens = NULL, *path;
     const char *path_left;
     size_t nargs = 0, i;
     va_list ap;
--- a/src/lib/krb5/os/locate_kdc.c
+++ b/src/lib/krb5/os/locate_kdc.c
@@ -214,8 +214,8 @@ oom:
 }
 
 static void
-parse_uri_if_https(const char *host_or_uri, k5_transport *transport,
-                   const char **host, const char **uri_path)
+parse_uri_if_https(char *host_or_uri, k5_transport *transport,
+                   char **host, const char **uri_path)
 {
     char *cp;
 
@@ -257,8 +257,7 @@ locate_srv_conf_1(krb5_context context, const krb5_data *realm,
                   k5_transport transport, int udpport)
 {
     const char *realm_srv_names[4];
-    char **hostlist = NULL, *realmstr = NULL, *host = NULL;
-    const char *hostspec;
+    char **hostlist = NULL, *realmstr = NULL, *host = NULL, *hostspec;
     krb5_error_code code;
     size_t i;
     int default_port;
@@ -587,8 +586,8 @@ prof_locate_server(krb5_context context, const krb5_data *realm,
  * Return a NULL *host_out if there are any problems parsing the URI.
  */
 static void
-parse_uri_fields(const char *uri, k5_transport *transport_out,
-                 const char **host_out, int *primary_out)
+parse_uri_fields(char *uri, k5_transport *transport_out,
+                 char **host_out, int *primary_out)
 
 {
     k5_transport transport;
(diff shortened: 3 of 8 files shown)


abbra:
Sample of a failed build (scroll to the end): https://kojipkgs.fedoraproject.org//work/tasks/9032/139609032/build.log

```
making all in lib/krb5/ccache...
make[3]: Entering directory '/builddir/build/BUILD/krb5-1.21.3-build/krb5-1.21.3/src/lib/krb5/ccache'
gcc -fPIC -DSHARED -DHAVE_CONFIG_H  -I../../../include -I../../../include -I./ccapi -I. -I. -DKRB5_DEPRECATED=1 -DKRB5_PRIVATE -I/usr/include/et -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Wno-complain-wrong-lang -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -march=z14 -mtune=z15 -fasynchronous-unwind-tables -fstack-clash-protection -I/usr/include/et -fPIC -fno-strict-aliasing -fstack-protector-all -Wall -Wcast-align -Wshadow -Wmissing-prototypes -Wno-format-zero-length -Woverflow -Wstrict-overflow -Wmissing-format-attribute -Wmissing-prototypes -Wreturn-type -Wmissing-braces -Wparentheses -Wswitch -Wunused-function -Wunused-label -Wunused-variable -Wunused-value -Wunknown-pragmas -Wsign-compare -Werror=uninitialized -Wno-maybe-uninitialized -Werror=pointer-arith -Werror=int-conversion -Werror=incompatible-pointer-types -Werror=discarded-qualifiers -Werror=implicit-int -Werror=strict-prototypes -Werror=declaration-after-statement -Werror-implicit-function-declaration -pthread  -c ccbase.c -o ccbase.so.o && mv -f ccbase.so.o ccbase.so
ccbase.c: In function ‘krb5_cc_resolve’:
ccbase.c:213:8: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  213 |     cp = strchr (name, ':');
      |        ^
cc1: some warnings being treated as errors
make[3]: *** [Makefile:787: ccbase.so] Error 1
make[3]: Leaving directory '/builddir/build/BUILD/krb5-1.21.3-build/krb5-1.21.3/src/lib/krb5/ccache'
make[2]: *** [Makefile:1113: all-recurse] Error 1
make[2]: Leaving directory '/builddir/build/BUILD/krb5-1.21.3-build/krb5-1.21.3/src/lib/krb5'
make[1]: *** [Makefile:974: all-recurse] Error 1
make[1]: Leaving directory '/builddir/build/BUILD/krb5-1.21.3-build/krb5-1.21.3/src/lib'
make: *** [Makefile:1520: all-recurse] Error 1
RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.QM3hsw (%build)
    Bad exit status from /var/tmp/rpm-tmp.QM3hsw (%build)
Child return code was: 1
```

abbra:
C specification draft that can be used to view section 7.28.5.1: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3685.pdf

jrisc:
I think there should be no cast on the input string parameter of `strchr()` to discard the `const` qualifier. This will likely result in the original string to be modified via the returned pointer. This is the case in the `parse_uri_if_https()` function where `cp` is used to write a null terminator in the string originally provided by the `host_or_uri` argument which is supposed to point to an immutable string. In such case, it means that the signature of the function is wrong, and the problem should be addressed at this level.

I reworked this commit to create a test patch for Fedora Rawhide with additional changes in order not to use casts (it builds successfully):
https://src.fedoraproject.org/rpms/krb5/pull-request/69

abbra:
Thanks, I've updated the patchset.

abbra:
@greghudson there is a typo in the rewritten commit message, 7.26.5.1 instead of 7.28.5.1. Section 7.26 only has three subsections.

greghudson:
What document are you looking at?  I found https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf and it has strchr in 7.26.5.3; section 7.28 there is <threads.h>.

(Looking back I see you gave a link already; I will check that out and compare.)

greghudson:
Okay, I assume n3685 is newer than n3096 just based on the number.  I had found <https://www.reddit.com/r/C_Programming/comments/w5hl80/c23_now_finalized/> which I guess wasn't reliable.  I will change the reference in the commit message back to 7.28.

abbra:
Yes, I used n3685, as commented with the link earlier in this thread

jrisc:
There is another case in `tests/responder.c`:
```
responder.c: In function ‘responder’:
responder.c:285:17: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  285 |             key = strchr(data->otp_answer, '=');
      |                 ^
```

It can be fixed this way:
```diff
index 82f870ea5d..4221a20283 100644
--- a/src/tests/responder.c
+++ b/src/tests/responder.c
@@ -282,8 +282,7 @@ responder(krb5_context ctx, void *rawdata, krb5_responder_context rctx)
     /* Provide a particular response for an OTP challenge. */
     if (data->otp_answer != NULL) {
         if (krb5_responder_otp_get_challenge(ctx, rctx, &ochl) == 0) {
-            key = strchr(data->otp_answer, '=');
-            if (key != NULL) {
+            if (strchr(data->otp_answer, '=') != NULL) {
                 /* Make a copy of the answer that we can chop up. */
                 key = strdup(data->otp_answer);
                 if (key == NULL)
```

avivanoff:
Since this change is not in 1.22.2, it fails to compile on Ubuntu 26, GCC 15.2 (probably other platforms as well). Are you planning a release with this fix in?

PR #1482, opened 2026-02-12
Fix iprop termination race in kpropd

zborbely:
There's a race condition in iprop code handling the SIGTERM signal. The signal terminates the process unconditionally, even in critical regions updating the database. The error manifests during the next kpropd startup as a SIGSEGV (when interrupted in the db code) or SIGABRT (if there's a replication gap).

The probability of this race is low, but can be increased if the database is locked by "kdb5_util dump" during the shutdown, forcing ulog_replay() to wait. This patch intercepts the signal and notifies the main loop to terminate instead of delivering at once. This solution is similar that is implemented in kadmind.

As an extra safety measure, the second patch triggers full resync upon detecting replication gaps before an assertation failure activates in plugins/kdb/db2/kdb_xdr.c:
```
164     /* Put extended fields here */
165     if (entry->len != KRB5_KDB_V1_BASE_LENGTH)
166         abort();
```

**Legal**
THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE MIT LICENSE. YOU MAY OBTAIN A COPY OF THE LICENSE AT https://opensource.org/licenses/MIT.
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.

### DIFF
--- a/src/kprop/kpropd.c
+++ b/src/kprop/kpropd.c
@@ -142,6 +142,8 @@ static const char *port = KPROP_SERVICE;
 static char **db_args = NULL;
 static size_t db_args_size = 0;
 
+static volatile sig_atomic_t signal_requests_exit = 0;
+
 static void parse_args(int argc, char **argv);
 static void do_standalone(void);
 static void doit(int fd);
@@ -233,6 +235,12 @@ kill_do_standalone(int sig)
                     (int)fullprop_child);
         }
         kill(fullprop_child, sig);
+        if (sig == SIGINT || sig == SIGTERM) {
+            /* Defer termination to avoid exiting in the middle of a critical
+             * DB or ulog section. */
+            signal_requests_exit = 1;
+            return;
+        }
     }
     /* Make sure our exit status code reflects our having been signaled */
     signal_wrapper(sig, SIG_DFL);
@@ -757,6 +765,9 @@ reinit:
         incr_ret = NULL;
         full_ret = NULL;
 
+        if (signal_requests_exit)
+            goto done;
+
         /*
          * Get the most recent ulog entry sno + ts, which
          * we package in the request to the primary KDC
@@ -999,7 +1010,7 @@ done:
     ulog_fini(kpropd_context);
     krb5_free_context(kpropd_context);
 
-    return (runonce == 1) ? 0 : 1;
+    return (runonce || signal_requests_exit) ? 0 : 1;
 }
 
 
--- a/src/lib/kdb/kdb_convert.c
+++ b/src/lib/kdb/kdb_convert.c
@@ -745,6 +745,18 @@ ulog_conv_2dbentry(krb5_context context, krb5_db_entry **entry,
 #undef u
     }
 
+    if (ent->len == 0) {
+        /*
+         * The len field should have been set in the existing principal or (for
+         * a creation update) set in the update entry.  If was not set, we just
+         * applied a non-creation update to a nonexistent principal, implying
+         * that our database is out of sync with our ulog metadata.  Return an
+         * error to force a full resync.
+         */
+        ret = KRB5_KDB_NOENTRY;
+        goto cleanup;
+    }
+
     /*
      * process mod_princ_data request
      */
--- a/src/plugins/kdb/db2/kdb_xdr.c
+++ b/src/plugins/kdb/db2/kdb_xdr.c
@@ -161,9 +161,11 @@ krb5_encode_princ_entry(krb5_context context, krb5_data *content,
     krb5_kdb_encode_int16(entry->n_key_data, nextloc);
     nextloc += 2;
 
-    /* Put extended fields here */
-    if (entry->len != KRB5_KDB_V1_BASE_LENGTH)
-        abort();
+    if (entry->len != KRB5_KDB_V1_BASE_LENGTH) {
+        retval = KRB5_KDB_TRUNCATED_RECORD;
+        k5_setmsg(context, retval, _("KDB entry length must be at least %d"),
+                  KRB5_KDB_V1_BASE_LENGTH);
+    }
 
     /* Any extra data that this version doesn't understand. */
     if (entry->e_length) {


greghudson:
Apologies for the long delay.

Making kpropd terminate gracefully on signal delivery seems like a good idea, and I plan to merge that commit with some changes.

I do not think I agree with some of the analysis in the PR about "replication gaps".  From the description and comments, it sounds like the proximal symptom after ungraceful termination (in some cases) was the abort() at kdb_xdr.c:166 due to `entry->len != KRB5_KDB_V1_BASE_LENGTH`.  (This is a very old bit of code, and I do not know why it chooses to abort rather than return an error.)  Attempting to infer backwards:

* krb5_db2_put_principal() was called with an entry with a zero `len` field.
* ulog_replay() assembled a krb5_db_entry with a zero `len` field.
* ulog_conv_2dbentry() processed a new principal entry with no AT_LEN attribute.  (An existing entry would have `len` set by krb5_db_get_principal().)
* kadmind sent a kdb_incr_result_t entry for a new principal entry with no AT_LEN attribute.
* Within kadmind, ulog_get_entries() produced a new principal update entry with no AT_LEN.
* ulog_get_entries() successfully parsed a new principal update entry with no AT_LEN from the ulog file using xdr_kdb_incr_update_t().

I infer that kadmind in this scenario is running on an intermediate secondary KDC (using multi-level iprop) rather than the actual admin server, and the offending state was caused by ungraceful kpropd termination on the intermediate.

From analysis of kdb_log.c, I cannot discern a moment where sudden termination would leave behind a ulog containing a well-formed but partial update entry.  It's easier to conceive of there being a KDB entry with zero `len`, although I'm not coming up with a scenario where such an entry is then marshalled into an offending update entry.

The PR description also asserts that the likelihood of termination in a critical code section increases if the database is locked by a dump because ulog_replay() has to wait.  I cannot follow this claim; if ulog_replay() blocks in a call to krb5int_put_principal_no_log(), it's waiting in a non-critical section of code.

zborbely:
Feel free to modify my PR as you wish.

About the "replication gap":
Yes, we're using hierarchical iprop replication, and it happened on an intermediate host. We haven't seen similar on the primary KDC. We haven't seen similar on replicas either until an automated task calling kdb5_util dump started to run in the reboot window. Reboot only affected the crashed instance, not its upstream.

I tried to reproduce this situation on my devel box, but I couldn't. I have only partial results and facts, but all of them point to the same direction.

Some more facts I realized during my analysis:
- the affected principal had been created just that time
- right after the keys of this new principal had been changed by a new operation, producing an update entry
- the crash happened updating this entry (having only the AT_PRINC and AT_KEYDATA attributes, no len or others)
- the upd entry in ulog_replay() seems valid and sane
- in krb5_encode_princ_entry() all fields (including entry->len) are zero except what got in upd

My only theory left is that the abnormal termination of kpropd caused some db corruption when the original entry creation got lost (e.g. ulog synced before the db or so), and caused crash during updating the non-existing principal after reboot.

I managed to partially verify this idea with the following test setup:
- primary/replica pair with working and synchronized iprop
- stop kpropd on the replica
- save the ulog on the replica
- delete a principal on the replica
- restore the saved ulog (with these steps I simulated a "lost principal create" situation)
- restart kpropd on the replica
- try to do anything with this entry on the primary (e.g. cpw -randkey princ)

It will produce exactly the same crash as I experienced before.

This race window can be pretty small, and I'm definitely sure "keeping the database locked and releasing in the worst time" is a key part of this scenario. My theory is that during reboot the systemd terminates "kdb5_util dump" first which releases the lock, and ulog_reply() can continue. kpropd will be terminated right after, now in a critical section somewhere in DB or ULOG update.

We experienced not just this "replication gap" aborts, but kpropd segfaults upon next boot in the db2 code.

greghudson:
Thanks for the additional information.  It does look like ulog_conv_2dbentry() does not currently have any safety against a non-creation principal update being applied to a nonexistent DB entry, and checking for `ent->len == 0` may  be the most straightforward way to detect the mismatch.

zborbely:
Exactly. Fortunately returning an error code has a hidden bonus, it will trigger a full resync so the db inconsistency will be auto-resolved :-)

src/kprop/kpropd.c
```diff
@@ -142,6 +142,8 @@ static const char *port = KPROP_SERVICE;
 static char **db_args = NULL;
 static size_t db_args_size = 0;
 
+static int signal_received = 0;
```
nicowilliams: This should in principle be of type `sig_atomic_t`.

src/kprop/kpropd.c
```diff
@@ -743,6 +751,8 @@ do_iprop(void)
     if (debug)
         fprintf(stderr, _("kadm5 initialization succeeded\n"));
 
+    signal_wrapper(SIGTERM, iprop_signal_handler);
```
nicowilliams: Why not also catch `SIGINT` and `SIGQUIT`?

nicowilliams:
This is a good patch, but the system needs to be able to recover from power failures at inopportune times too.

src/kprop/kpropd.c
zborbely: I was not sure about `SIGQUIT`, so I did only a minimal change. You're right, `SIGINT` and `SIGQUIT` have been added also, just like in `loop_setup_signals()`.

nicowilliams: I agree, yeah, `SIGQUIT` is not necessary.  If I type `^\` in my terminal I probably want a core dump _now_.

zborbely: Following your suggestion I added `SIGINT` and `SIGQUIT` also, now the implementation is similar to kadmind (unification is always good). I did nothing with `SIGHUP`...

src/lib/kdb/kdb_convert.c
```diff
@@ -745,6 +745,18 @@ ulog_conv_2dbentry(krb5_context context, krb5_db_entry **entry,
...
 
+    if (ent->len == 0) {
+        /*
+         * The len field should have been set in the existing principal or (for
+         * a creation update) set in the update entry.  If was not set, we just
+         * applied a non-creation update to a nonexistent principal, implying
+         * that our database is out of sync with our ulog metadata.  Return an
+         * error to force a full resync.
+         */
+        ret = KRB5_KDB_NOENTRY;
```
nicowilliams, line 756: Idk what iprop in MIT looks like nowadays, but if you have a whole entry from upstream then I'd rather accept it than reject it in this case.

greghudson:
I added a third commit to change the abort() in kdb_xdr.c to an error return.  The ulog code won't trigger the condition any more with the `ent->len == 0` check, but it's still tangentially related to the PR.

I noted that kpropd already had iprop-specific signal handlers for the termination signals, which kill the fullprop child process.  I changed the final commit to integrate with the existing signal handler rather than set a different one.

src/kprop/kpropd.c
```diff
@@ -233,6 +235,12 @@ kill_do_standalone(int sig)
                     (int)fullprop_child);
         }
         kill(fullprop_child, sig);
+        if (sig == SIGINT || sig == SIGQUIT || sig == SIGTERM) {
```
nicowilliams: For `SIGQUIT` you should set a global `sig_atomic_t` boolean that says to `abort()`.  The point is that `SIGQUIT` is generated by the terminal driver when you type `<ctrl>-\`, and that it causes a core dump by default (so you can then inspect it with a debugger).  So if one catches `SIGQUIT` then one should arrange to dump core (i.e., call `abort()`) rather than `_exit()`.

src/lib/kdb/kdb_convert.c
greghudson: I'm not sure what you mean.  As the comment states, this condition detects that we had a partial entry from upstream, not a whole one, but we did not have an existing principal entry to apply it to.

This area of the iprop code has changed very little from its initial incarnation.

src/kprop/kpropd.c
greghudson: I included SIGQUIT because kadmind and krb5kdc also exit normally on a SIGQUIT (forever for kadmind, since 1.10 for krb5kdc).  I think this exiting behavior is mostly harmless because daemons don't run in a terminal (so no ctrl-\) and sending SIGQUIT to a daemon process generally isn't the best way to debug it.  But by the same token, it's mostly useless--there's no reason to have more letters of code to more gracefully handle an event we don't expect to see in a production environment.  So I'll take it out and just let SIGQUIT terminate the process immediately.

src/lib/kdb/kdb_convert.c
nicowilliams: Oh, in Heimdal updates include the whole principal record, which is what I was referring to by "but if you have a whole entry from upstream", but MIT doesn't have it, so never mind.

src/kprop/kpropd.c
zborbely: Can I add anything to move this forward?

PR #1492, opened 2026-03-08
Fix S4U2Self ignoring time_req when acquiring impersonated credentials

abbra:
gss_acquire_cred() for the service locates the existing TGT in the ccache and records its KDC-granted endtime as cred->expire; it never requests a shorter-lived TGT.  So when gss_acquire_cred_impersonate_name() calls kg_impersonate_name() with a time_req, the impersonator TGT already carries a KDC-defined lifetime, not the requested one.

kg_impersonate_name() compounds this by zero-initialising in_creds (leaving in_creds.times.endtime = 0) and never writing time_req into it. When that reaches send_tgs.c:
```
    req.till = desired->times.endtime ? desired->times.endtime
                                      : tgt->times.endtime;
```
the zero endtime causes the fallback to tgt->times.endtime — the impersonator TGT's full KDC-granted lifetime.  The KDC therefore issues the S4U2Self ticket with KDC-policy lifetime, not the requested time_req.

The subsequent S4U2Proxy request via gss_init_sec_context() gets the correct lifetime because that is an entirely separate call with its own time_req: kg_new_connection() converts it to an absolute timestamp (ctx->krb_times.endtime = ts_incr(now, time_req)) and get_credentials() places it into in_creds.times.endtime before calling krb5_get_credentials(). send_tgs.c then takes the first branch and sends the requested deadline. This has nothing to do with what gss_acquire_cred_impersonate_name() did.

Fix kg_impersonate_name() to apply the same time_req-to-endtime conversion that kg_new_connection() already performs, so the S4U2Self TGS-REQ carries the caller's requested deadline rather than silently falling back to the impersonator TGT's full lifetime.

### DIFF
--- a/src/lib/gssapi/krb5/s4u_gss_glue.c
+++ b/src/lib/gssapi/krb5/s4u_gss_glue.c
@@ -53,6 +53,17 @@ kg_impersonate_name(OM_uint32 *minor_status,
     *output_cred = NULL;
     memset(&in_creds, 0, sizeof(in_creds));
 
+    if (time_req != 0 && time_req != GSS_C_INDEFINITE) {
+        krb5_timestamp now;
+
+        code = krb5_timeofday(context, &now);
+        if (code != 0) {
+            *minor_status = code;
+            return GSS_S_FAILURE;
+        }
+        in_creds.times.endtime = ts_incr(now, time_req);
+    }
+
     if (user->is_cert)
         subject_cert = user->princ->data;
     else
--- a/src/lib/gssapi/krb5/s4u_gss_glue.c
+++ b/src/lib/gssapi/krb5/s4u_gss_glue.c
@@ -192,14 +192,35 @@ krb5_gss_acquire_cred_impersonate_name(OM_uint32 *minor_status,
  */
 static krb5_error_code
 make_proxy_cred(krb5_context context, krb5_gss_cred_id_t cred,
-                krb5_gss_cred_id_t impersonator_cred)
+                krb5_gss_cred_id_t impersonator_cred,
+                krb5_timestamp max_endtime)
 {
     krb5_error_code code;
     krb5_data data;
+    krb5_cc_cursor cur = 0;
+    krb5_creds cur_creds;
     char *str;
 
-    code = krb5_cc_copy_creds(context, impersonator_cred->ccache,
-                              cred->ccache);
+    /* Copy credentials from the impersonator ccache, bounding endtime when
+     * requested so copied TGTs don't outlive the S4U2Self evidence ticket. */
+    code = krb5_cc_start_seq_get(context, impersonator_cred->ccache, &cur);
+    if (code)
+        return code;
+
+    while (!(code = krb5_cc_next_cred(context, impersonator_cred->ccache,
+                                      &cur, &cur_creds))) {
+        if (max_endtime != 0 && cur_creds.times.endtime > max_endtime)
+            cur_creds.times.endtime = max_endtime;
+        code = krb5_cc_store_cred(context, cred->ccache, &cur_creds);
+        krb5_free_cred_contents(context, &cur_creds);
+        if (code)
+            break;
+    }
+
+    if (cur)
+        krb5_cc_end_seq_get(context, impersonator_cred->ccache, &cur);
+    if (code == KRB5_CC_END)
+        code = 0;
     if (code)
         return code;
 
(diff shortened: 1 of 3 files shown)


abbra:
Without this patch I get from an application that performs protocol transition and then S4U2Proxy regardless what limited time_req I added:
```
admin@m1:~$ klist
Ticket cache: KCM:864800000:6179
Default principal: admin@IPA.TEST

03/09/2026 00:12:16  03/09/2026 23:18:36  host/m1.ipa.test@IPA.TEST
03/09/2026 00:12:16  03/09/2026 23:18:36  krbtgt/IPA.TEST@IPA.TEST
	for client host/m1.ipa.test@IPA.TEST
03/09/2026 00:12:16  03/09/2026 00:22:16  HTTP/m1.ipa.test@IPA.TEST
```
E.g. both krbtgt and host/ tickets have KDC-dictated ~24hr lifetime (FreeIPA does jitter of the last hour to spread renewals). S4U2Proxy's HTTP/.. ticket is limited to 10 minutes how application was instructed.

With this patch I get
```
admin@m1:~$ klist
Ticket cache: KCM:864800000:88798
Default principal: admin@IPA.TEST

03/09/2026 00:25:41  03/09/2026 00:35:41  host/m1.ipa.test@IPA.TEST
03/09/2026 00:25:41  03/10/2026 00:08:34  krbtgt/IPA.TEST@IPA.TEST
	for client host/m1.ipa.test@IPA.TEST
03/09/2026 00:25:42  03/09/2026 00:35:42  HTTP/m1.ipa.test@IPA.TEST
```
There is still problem with krbtgt/ ticket but host/ ticket got restricted as well.

abbra:
Now TGT is also constrained to time_req:
```
$ klist
Ticket cache: KCM:864800000:99564
Default principal: admin@IPA.TEST

03/09/2026 09:22:26  03/09/2026 09:32:26  host/m1.ipa.test@IPA.TEST
03/09/2026 09:22:26  03/09/2026 09:32:26  krbtgt/IPA.TEST@IPA.TEST
	for client host/m1.ipa.test@IPA.TEST
03/09/2026 09:22:26  03/09/2026 09:32:26  HTTP/m1.ipa.test@IPA.TEST
```

greghudson:
> Now TGT is also constrained to time_req:

This statement confuses me.  Is the TGT being obtained automatically via a client keytab in this scenario, or is it already present?

To elaborate on the question: my understanding of an S4U2Self/S4U2Proxy flow is that you start with a TGT for intermediate-service, you make an S4U2Self request to get client->intermediate, and then you make an S4U2Proxy request with those two tickets to get client->target.  Neither of those steps should result in a new TGT unless there's another realm involved.

abbra:
The scenario I'm implementing is an app authenticating a user over a different protocol, then obtaining TGT, performing S4U2Self and optionally series of S4U2Proxy requests using its own keytab. Once this is done, a user gets access to a shell where credentials cache contains these tickets but otherwise the user has no access to the keytab.

My goal is to have one of three situations:
- user's credentials cache contains S4U2Self client->intermediate + S4U2Proxy tickets to selected services
- user's credentials cache contains only S4U2Self client->intermediate ticket
- user's credentials cache contains only S4U2Proxy tickets to selected services

Each ticket limited to the specified lifetime, optionally. That's why I start with TGT with lifetime limit, if enabled in the app's configuration.

abbra:
It looks like the TGT must be present in the ccache for S4U2Proxy tickets to be used, as you said (their usage fails otherwise, according to krb5 trace events). 

Direct use of the S4U2Self client->intermediate does not require TGT presence.

So I am able to have the following:
1. No TGT in ccache but S4U2Self client->intermediate ticket, optionally time-limited. Works for cases such as PAM authentication with `pam_sss_gss.so` module.
2. TGT + S4U2Self client->intermediate ticket, optionally time-limited. Works for further S4U2Proxy as well.
3. TGT + S4U2Self client->intermediate ticket + S4U2Proxy tickets, all optionally time-limited.

The latter case is an optimization to pre-fetch proxy tickets. It is useful for situations like a network file system mounting on logon.

greghudson:
Apologies for getting so far behind.

The first commit looks fine, but I do not like the second commit.  It adds a lot of complexity, and it basically lies by creating a ccache entry with an endtime that doesn't match the actual ticket endtime.

If the TGT lifetime being longer than time_req creates a practical problem, we should resolve it by figuring out what is looking at the TGT end time and making it look at the evidence ticket end time instead.

greghudson:
> If the TGT lifetime being longer than time_req creates a practical problem, we should resolve it by figuring out what is looking at the TGT end time and making it look at the evidence ticket end time instead.

Specifically, it looks like scan_ccache() in acquire_cred.c will currently prefer the impersonator TGT expiration time over the subject ticket expiration time, because (1) it prefers a TGT for the subject's realm, which the impersonator TGT usually will be, and (2) if it fails to find a TGT it prefers the first entry it finds, which the impersonator TGT definitely will be.

My suggested fix is to filter out ccache entries whose clients aren't the same as the subject, so adding at line 420 (just after the `krb5_is_config_principal()` block):

        if (!krb5_principal_compare(context, creds.client, cred->name->princ))
            continue;

I haven't tested this approach, however.

abbra:
Thank you, Greg. I'm at IETF and Hackathon this weekend and next week. Once I'm back, I'll look into this.

PR #1500, opened 2026-03-30
Asynchronous S4U interfaces

abbra:
This branch combines work from several other PRs to improve S4U operations over GSSAPI and IAKERB. The code refactoring required touching too many related places so I combined these PRs into a single one.

- Asynchronous S4U2Self and S4U2Proxy with _step interface
- Asynchronous S4U over IAKERB
- Asynchronous Anonymous PKINIT over IAKERB
- Automated FAST armor when Anonymous PKINIT is available
- Apply lifetime limitations when requesting S4U2Self
- Allow use of a cached S4U cred when the target service is the same already


src/lib/krb5/krb/get_creds.c
```diff
@@ -1163,6 +1251,70 @@ krb5_tkt_creds_init(krb5_context context, krb5_ccache ccache,
     return code;
 }
 
+/*
+ * Configure ctx for S4U2Self mode.  Must be called before the first
+ * krb5_tkt_creds_step() call.  Subsequent steps will include PA-S4U-X509-USER
+ * padata in TGS-REQs and verify the reply padata.
+ */
+krb5_error_code
+k5_tkt_creds_set_s4u2self(krb5_context context, krb5_tkt_creds_context ctx,
```
voronond, line 1375: This seems to be sufficient for the same realm scenarios (service realm is the same as client realm). It unfortunately doesn't work for the cross-realm (service in realm A and client from realm B). If we keep the regular tkt_step flow, then we receive a 'KDC policy reject' as we are trying to obtain a service ticket with our initial TGT.

We need to extend the current state-machine of `krb5_tkt_creds_step` to check whether the s4u_user was set and whether its realm is a cross-realm. If so, we need to negotiate cross-realm TGT to the user's realm and send the S4U2Self request there. Then follow it back to a service realm.

abbra: Thank you for testing. There is considerably more work there, now implemented.

voronond: I have it implemented already if that helps: https://github.com/voronond/krb5/pull/1/changes/46316daae81485fed632957459ceccc4c07d4644

I have also extended t_authdata.py, which already exercised the existing S4U with a new step-based async approach (so we have coverage for that as well). The changes are quite small and located in the get_creds.c file. I have already executed several tests in AD multi-domain infrastructure, and it works (even in cross-forest realms).

abbra: Thanks, I implemented that already, meant to say that it took some time to do so. The following commits add the cross-realm support:
- [get_creds: skip S4U padata in TGT-acquisition TGS-REQs](https://github.com/krb5/krb5/pull/1500/commits/cc1736388be4909d33926e51b4d9983d8361d5c2)
- [get_creds: navigate to user realm for cross-realm S4U2Self](https://github.com/krb5/krb5/pull/1500/commits/cfe3f0cf4dcc1a5206a39a7b5d158b8c61616e58)
- [get_creds: use enterprise principal for foreign-realm S4U2Self TGS-REQs](https://github.com/krb5/krb5/pull/1500/commits/5c2a413f4cd37085b01f158ad29ff182e0141849)
- [tests/gssapi: add cross-realm step-based S4U2Self test](https://github.com/krb5/krb5/pull/1500/commits/91c1627ff27a64ae89021714f3e12f1980d3c9f0)

src/include/k5-int.h
```diff
@@ -2448,4 +2448,20 @@ k5_print_addr(const struct sockaddr *sa, char *buf, size_t len);
...
 k5_print_addr_port(const struct sockaddr *sa, char *buf, size_t len);
 
+/* Check if Automated PKINIT is allowed for the realm */
+krb5_boolean
+k5_fast_is_pkinit_allowed(krb5_context context, krb5_data *realm);
+
+/* Configure a krb5_tkt_creds_context for S4U2Self mode.  Must be called after
+ * krb5_tkt_creds_init() and before the first krb5_tkt_creds_step() call. */
+krb5_error_code
+k5_tkt_creds_set_s4u2self(krb5_context context, krb5_tkt_creds_context ctx,
```
voronond, line 2458: Should it be added in `src/include/krb5/krb5.hin`? Otherwise, we won't be able to use it outside the library during the krb5_tkt_creds_init/krb5_tkt_creds_step flow, as it would not be exported in krb5.h.

PR #1501, opened 2026-03-30
klist: add -I keytab option to show authentication indicators

abbra:
Add a new -I keytab option to klist that decrypts service tickets using the provided keytab and displays any authentication indicators embedded in the ticket's CAMMAC authorization data.

Authentication indicators (RFC 8129) are stored in the ticket's encrypted body as IF-RELEVANT(CAMMAC(AUTH-INDICATOR)) authdata. The existing -d option only shows authdata types from cached credentials, where authdata is never populated. The new -I option works by:

1. Looking up the service's long-term key in the keytab
2. Decrypting the ticket's enc-part with krb5_decrypt_tkt_part()
3. Walking authorization_data for IF-RELEVANT containers
4. Verifying and unwrapping any CAMMAC using the service's long-term key (same key used to encrypt the ticket) via k5_unwrap_cammac_svc()
5. Decoding AUTH-INDICATOR elements with k5_authind_decode()

### DIFF
--- a/doc/user/user_commands/klist.rst
+++ b/doc/user/user_commands/klist.rst
@@ -8,7 +8,8 @@ SYNOPSIS
 
 **klist**
 [**-e**]
-[[**-c**] [**-l**] [**-A**] [**-f**] [**-s**] [**-a** [**-n**]]]
+[[**-c**] [**-l**] [**-A**] [**-f**] [**-s**] [**-a** [**-n**]]
+[**-I** *keytab*]]
 [**-C**]
 [**-k** [**-i**] [**-t**] [**-K**]]
 [**-V**]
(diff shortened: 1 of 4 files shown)


greghudson:
Apologies for the long delay.  Actionable feedback:

* There is a duplicate comment above show_auth_indicators().

* krb5_server_decrypt_ticket_keytab() exists, and collapses two steps of show_auth_indicators() into one.  It also tries harder to decrypt the ticket for enumerable keytabs, since server aliases mean that the ticket server doesn't always match what is in a keytab file.

* There's no test case.  It should be easy enough to add one to t_authdata.py.

* The user documentation does not need to mention CAMMACs or refer to RFC 9509.  Being within the encrypted part of the ticket is sufficient reason for needing the service key.

* The added symbols in krb5_32.def should be commented as being new in 1.23 and private symbols used by kinit.

Less actionable feedback that I just want a record of having thought about:

* The keytab is resolved for each ticket.  This is of little import (resolving a file keytab doesn't even open the file, just creates a memory object recording the name).

* There is no way to use the default keytab like there is for `klist -k`.  We could make "-" a synonym for using the default keytab.  We could make `-I` a flag and then take the keytab name from a second positional argument, but then we would need a way to use the default ccache with a non-default keytab, so that's not really an improvement.

* I checked for a conflicting meaning for `klist -I` in Heimdal and it doesn't currently use `-I`.

* Maybe there should be a public libkrb5 API for retrieving auth indicators from a decrypted ticket?  Currently we have the private APIs used in the PR, and the authdata APIs (also private) used by gss_get_name_attribute().

abbra:
Thank you for the feedback. I addressed all actionable items, including `klist -I -` which was an omission. 

Having public krb5 API would be nice, though most applications that would be interested in this information already use GSSAPI and can go by gss_get_name_attribute() already.

PR #1502, opened 2026-04-01
Add authentication indicators to audit plugin interface

abbra:
Fix audit plugin build and add authentication indicators to audit state so that they can be audited during AS-REQ

### DIFF
--- a/src/plugins/audit/kdc_j_encode.c
+++ b/src/plugins/audit/kdc_j_encode.c
@@ -63,6 +63,9 @@ rep_to_value(krb5_kdc_rep *rep, const krb5_boolean ev_success,
              k5_json_object obj);
 static krb5_error_code
 tkt_to_value(krb5_ticket *tkt, k5_json_object obj, const char *key);
+static krb5_error_code
+indicators_to_value(krb5_data **indicators, k5_json_object obj,
+                    const char *key);
 static char *map_patype(krb5_preauthtype pa_type);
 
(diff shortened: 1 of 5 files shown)


src/plugins/audit/kdc_j_encode.c
```diff
@@ -897,6 +904,46 @@ tkt_to_value(krb5_ticket *tkt, k5_json_object obj,
...
+
+    if (indicators == NULL || *indicators == NULL)
+        return 0;
+
+    ret = k5_json_array_create(&arr);
+    if (ret)
+        return ENOMEM;
+
+    for (ind = indicators; *ind != NULL; ind++) {
+        if ((*ind)->data == NULL || (*ind)->length == 0)
```
greghudson: Did you find this necessary or were you just being defensive?  Checking for a null data pointer seems especially defensive; if the length is non-zero and the data pointer is null then something went badly wrong in the KDC.

abbra: Defensive. As KDB driver may change indicators through the PAC processing, it is under influence of external code KDC itself cannot give guarantees about.

greghudson: A zero-length auth indicator would be a strange choice, but not strictly prohibited.  Filtering it out of the log and not the ticket would just makes the log less accurate.  A null/nonzero krb5_data value is a memory error, and would crash the KDC process when it encodes the ticket.  So I plan to take this check out.

PR #1506, opened 2026-04-13
Fix two NegoEx parsing vulnerabilities

greghudson:
I will push this immediately as it is a security fix.  The exposure is pretty minimal as the NegoEx parsing code isn't reachable unless a NegoEx GSS mechanism is registered in /etc/gss/mechs.  When the vulnerabilities are reachable, they only affect availability and don't lead to escalation of privilege.


twangboy:
Any plans to get a release out with this fix?

PR #1508, opened 2026-04-30
Add forward_lifetime to cap delegated credentials

mgkuhn:
Unconstrained credential delegation through GSSAPI hands the destination service a forwarded TGT whose lifetime matches the source TGT, typically 10 or 24 hours.  When the destination service is not configured for constrained delegation, this can expose more privilege than the session needs.

This PR adds a `[libdefaults]` `forward_lifetime` setting and a corresponding `KRB5_FORWARD_LIFETIME` environment variable. When either is set, `krb5_fwd_tgt_creds` caps the forwarded ticket's lifetime to the configured duration and clears its renewable flag.

Typical application:

```
$ KRB5_FORWARD_LIFETIME=5m ssh -K example.org quick-command
```

This came up while hardening `ssh -K` access to a `sec=krb5p` NFS-mounted home directory; constrained delegation isn't configured on the relevant servers and changing that is outside our remit.

Tested on Ubuntu 24.04


mgkuhn:
I, Markus Kuhn, hereby grant the MIT Kerberos consortium the right to use and distribute the changes I have submitted to krb5/krb5 under the terms of the existing project license.

PR #1510, opened 2026-05-18
Fix error propagation in get_pac_princ_with_realm

TristanInSec:
## Summary

The cleanup label in `get_pac_princ_with_realm()` unconditionally returns 0 instead of the error code in `ret`. When the PAC client name string has an unexpected number of `@` characters (0 or 3+), `ret` is set to `KRB5_PARSE_MALFORMED` but the function still returns success. The caller in `do_tgs_req.c` then proceeds with a NULL `stkt_pac_client` pointer (initialized to NULL at line 647), causing a NULL pointer dereference crash in the KDC during cross-realm S4U2Proxy (constrained delegation) processing.

## Fix

Change `return 0` to `return ret` in the cleanup label.

Note: the success path through `krb5_parse_name_flags()` (line 666) returns directly on success (line 668) or falls through to cleanup with the error in `ret`, so changing cleanup to return `ret` is correct for all paths:
- Malformed name (0 or 3+ `@`): `ret = KRB5_PARSE_MALFORMED`, cleanup returns error
- Parse failure: `ret` holds the parse error, cleanup returns it
- Success: returns at line 668, does not reach cleanup

ticket: new
tags: pullup
target_version: 1.22-next

### DIFF
--- a/src/kdc/kdc_util.c
+++ b/src/kdc/kdc_util.c
@@ -665,13 +665,13 @@ get_pac_princ_with_realm(krb5_context context, krb5_pac pac,
 
     ret = krb5_parse_name_flags(context, client_str, flags, princ_out);
     if (ret)
-        return ret;
+        goto cleanup;
 
     (*princ_out)->type = KRB5_NT_MS_PRINCIPAL;
 
 cleanup:
     free(client_str);
-    return 0;
+    return ret;
 }
 
 /* This probably wants to be updated if you support last_req stuff */


TristanInSec:
Hi @greghudson, the 3 fix PRs (#1509/#1510/#1511) are all merged -- would you like to assign CVEs for these, or shall I request them independently? Happy either way. Thanks.

greghudson:
I don't plan to request CVE assignments.  In general I think the CVE system is too heavyweight for availability-only bugs, particularly in the very recent era of LLM-assisted vulnerability research.

For the first two PRs I also think the nature of the attack surface makes the bugs impractical to exploit:

* For PR 1509 the attack surface is gss_import_sec_context().  The intended use of this function (described in RFC 2743 section 1.2.10) is transferring a GSS context between an application's own processes; there is no sane way to use it on untrusted input.

 * For PR 1510 the attack surface is data within a previously verified PAC, and the impact is a null pointer dereference.  At worst a KDC could crash a KDC in a realm it peers with.

I was going to make a similar argument for PR 1551, that krb5_rd_cred() is generally called within a disposable process like an sshd child process.  Unfortunately it appears that krb5_rd_cred() is reachable from the krb5 gss_accept_sec_context()  even if the application passes a null delegated_cred_handle.

We do have a security reporting email address, krbcore-security@mit.edu, which allows these conversations to be had before the bug is made public.  It's less important to use it for crash bugs than for privilege escalation bugs, of course.

TristanInSec:
> I don't plan to request CVE assignments. In general I think the CVE system is too heavyweight for availability-only bugs, particularly in the very recent era of LLM-assisted vulnerability research.
> 
> For the first two PRs I also think the nature of the attack surface makes the bugs impractical to exploit:
> 
>     * For PR 1509 the attack surface is gss_import_sec_context().  The intended use of this function (described in RFC 2743 section 1.2.10) is transferring a GSS context between an application's own processes; there is no sane way to use it on untrusted input.
> 
>     * For PR 1510 the attack surface is data within a previously verified PAC, and the impact is a null pointer dereference.  At worst a KDC could crash a KDC in a realm it peers with.
> 
> 
> I was going to make a similar argument for PR 1551, that krb5_rd_cred() is generally called within a disposable process like an sshd child process. Unfortunately it appears that krb5_rd_cred() is reachable from the krb5 gss_accept_sec_context() even if the application passes a null delegated_cred_handle.
> 
> We do have a security reporting email address, [krbcore-security@mit.edu](mailto:krbcore-security@mit.edu), which allows these conversations to be had before the bug is made public. It's less important to use it for crash bugs than for privilege escalation bugs, of course.

Thanks for the detailed response and for merging all 3 fixes. 
Good point on the limited attack surface for #1509 and #1510 -- but still, security bugs in a security-critical library
If you don't mind, not to add overhead, I will request the CVEs on my own and be conservative on CVSS scores.
Also - noted krbcore-security@mit.edu for future reports.

PR #1515, opened 2026-05-23
Always copy mech name_type for GSS union names

greghudson:
In gssint_convert_name_to_union_name(), make a copy of the name_type OID yielded by the mechanism's gss_display_name(), as it will later be released with gss_release_oid().  We usually get away with not copying the name type because gss_release_oid() usually ignores static mechanism OIDs, but this property is fragile (it relies on mechanisms implementing gss_internal_release_oid()).

Reported by Daniel Sands.

### DIFF
--- a/src/lib/gssapi/mechglue/g_glue.c
+++ b/src/lib/gssapi/mechglue/g_glue.c
@@ -372,6 +372,7 @@ gssint_convert_name_to_union_name(OM_uint32 *minor_status, gss_mechanism mech,
 {
     OM_uint32 major_status,tmp;
     gss_union_name_t union_name;
+    gss_OID name_type;
 
     union_name = (gss_union_name_t) malloc (sizeof(gss_union_name_desc));
     if (!union_name) {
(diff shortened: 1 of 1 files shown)


greghudson:
Before submitting this PR with the fix suggested by Daniel, I did an inventory of all the places where the mechglue stores and releases single OIDs (not OID sets).  The standardized GSSAPI traffics only in well-known, statically allocated OID objects and does not specify a gss_release_oid() function.  We have always provided a gss_release_oid(), and must continue to provide a working one for the sake of the gss_str_to_oid() extension if nothing else.

I found that the mechglue could _almost_ always treat stored OID objects as static and never release them, but there are two exceptions:

* importExportName() copies the mechanism OID decoded from the export token into the union name's mech_type field.  It first looks up the mech OID, so we could instead alias the mechanism's mech_type, assuming that name export tokens never contain interposer OIDs.

* gss_import_name() copies the caller's input_name_type into the union name's name_type field.  There is no lookup of input_name_type here and we can't assume that the caller passed a static OID constant (although they usually do), so this case would be hard to account for.

I didn't think it was wise to treat only some stored OIDs as static, so for now we will continue to treat all of them as allocated.

I didn't find any other cases where we directly store a mech-provided OID into a structure, so with this PR we should be correctly storing only allocated OIDs into mechglue objects.

The mechglue is not very organized about how it releases OIDs.  OID fields are released in various cleanup handlers as well as the relevant gss_release_ API for the containing structure, and each field is inconsistently freed with gss_release_oid() or generic_gss_release_oid().  I decided not to touch this as I would be more likely to create a bug than fix one.

PR #1526, opened 2026-06-16
OSS-Fuzz: Add new fuzzer targets keytab processing

arthurscchan:
This PR creates a new fuzzer targets keytab processing, together with disable of LeakSanitizer and corpora preparation.


greghudson:
Why do we need to disable leak detection for this fuzz target?

arthurscchan:
Hi sorry for the late reply. The major reason is that the new fuzzer actually discovered a memleak, and for some projects they don't care about leak and so I disable them to allow the fuzzer to go past the leak and continue fuzzing some deeper logic. I am sorry for the rush on this, I could always remove the disable of Leak detection and allow OSS-Fuzz to throw the leak as an issue. Which way do you prefer?

greghudson:
I'd prefer to fix the discovered leak and then add the fuzzer target without disabling leak detection.

arthurscchan:
OK. Give me some time to handle it.

PR #1527, opened 2026-06-18
Fix portability and static build issues on macOS

overcoil:
Three small independent fixes found while building krb5 statically on macOS
with Clang and vcpkg:

1. appl/simple: include arpa/inet.h for inet_ntoa
   sim_client.c calls inet_ntoa() without including <arpa/inet.h>.
   On strict POSIX systems this produces an implicit-declaration warning
   or error.

2. build: add MACOS_FRAMEWORK to static link flags
   When linking statically on macOS, the CoreFoundation and Security
   frameworks are not propagated to consumers. This causes linker
   failures in downstream builds that link against the static libraries
   without separately specifying the framework flags.
   - Added @MACOS_FRAMEWORK@ to Libs.private in mit-krb5.pc.in so
     pkg-config --static exposes it.
   - Added @MACOS_FRAMEWORK@ to KRB5_BASE_LIBS in config/pre.in for
     the Makefile build path.


greghudson:
I'm curious about the clang link failure with an uninitialized const variable with static storage.  It sounds like a bug.  But adding an initializer is cheap.

In the third commit, not only is `@MACOS_FRAMEWORK@` added to Libs.private, but `@LIBS@ @DL_LIB@ @PTHREAD_CFLAGS@`, without any accompanying justification in the commit message.  We use `$(LIBS)` in the libkrb5 SHLIB_EXPLIBS and` $(DL_LIB)` in the libkrb5support SHLIB_EXPLIBS, so those make sense even if they only tangentially related to the PR.  But we only use `$(PTHREAD_CFLAGS)` in the build system when compiling threaded test programs.  On most platforms libkrb5 tries to be agnostic about whether the binary is linked with thread support, using weak references to perform mutex operations only when libpthreads is present.

overcoil:
> I'm curious about the clang link failure with an uninitialized const variable with static storage. It sounds like a bug. But adding an initializer is cheap.

I decided to re-check this again and this is no longer needed with my latest CI environment. I'll remove this.

> In the third commit, not only is `@MACOS_FRAMEWORK@` added to Libs.private, but `@LIBS@ @DL_LIB@ @PTHREAD_CFLAGS@`, without any accompanying justification in the commit message. We use `$(LIBS)` in the libkrb5 SHLIB_EXPLIBS and` $(DL_LIB)` in the libkrb5support SHLIB_EXPLIBS, so those make sense even if they only tangentially related to the PR. But we only use `$(PTHREAD_CFLAGS)` in the build system when compiling threaded test programs. On most platforms libkrb5 tries to be agnostic about whether the binary is linked with thread support, using weak references to perform mutex operations only when libpthreads is present.

You're correct there... I've tested removing `@PTHREADS_CFLAGS@` from the CI runners and it is indeed superfluous. 

I'll resubmit the PR with these updates.

greghudson:
Merge as commit 9a27137bfa04f9ae79b19cce7cdb05680dfe844e .  I didn't appear to have permission to push to the PR branch, so the PR will show up as unmerged.

PR #1533, opened 2026-07-14
Build against OpenSSL-compatible libraries (AWS-LC)

WillChilds-Klein:
# Description

This PR proposes 3 small changes to make krb5 compatible with AWS-LC:

- **Gate builtin sha2 typedefs on `K5_BUILTIN_SHA2`**: guards the `SHA256_CTX`/`SHA512_CTX` struct/typedef block in `sha2.h` so it doesn't collide with AWS-LC's own typedefs in the OpenSSL build.
- **Disable Camellia enctypes when `OPENSSL_NO_CAMELLIA` is defined**: gates the `K5_OPENSSL_CAMELLIA` macro, Camellia enctype/checksum table entries, provider declarations, and test-suite references so krb5 builds and passes `make check` against libraries (like AWS-LC) that omit Camellia.
- **Guard the SPAKE `BN_FLG_CONSTTIME` hint for AWS-LC**: wraps `BN_set_flags(w, BN_FLG_CONSTTIME)` in `#if !defined(OPENSSL_IS_AWSLC)` so stock OpenSSL keeps the hint while AWS-LC (which dropped the flag and is constant-time by default) still compiles.

# Testing

AWS-LC [has added](https://github.com/aws/aws-lc/pull/3308) an integration test utilizing this patchset.


greghudson:
I don't plan to merge this PR because building against OpenSSL variants isn't a project goal.  I do have some comments on the commits:

1. This conditionalization is only necessary because aws-lc's `<openssl/opensslv.h>` drags too much stuff into the global namespace.  We don't include any other OpenSSL header when building the builtin/ files as stubs.
2. A simpler alternative is to use the built-in Camellia when building against aws-lc, by putting in crypto_int.h (where it currently defines K5_OPENSSL_CAMELLIA):
```
        #ifdef OPENSSL_IS_AWSLC
        #define K5_BUILTIN_CAMELLIA
        #else
        #define K5_OPENSSL_CAMELLIA
        #endif
```
3. aws-lc should provide BN_FLG_CONSTTIME as a no-op if it wants the benefit of compatibility with code written to OpenSSL.  Asking OpenSSL consumers to support the API difference is not the right answer.

WillChilds-Klein:
Thanks for your time and the feedback, I've adapted our patches to follow your guidance around `K5_BUILTIN_CAMELLIA`.

PR #1534, opened 2026-07-24
Add auto_fast_armor realm variable for FAST

jason34105533:
## Summary

Adds an `auto_fast_armor` realm variable. When set, initial credential
acquisition eagerly acquires an anonymous PKINIT ticket to use as FAST
armor before making the real request, unless an armor ccache is
already configured, then makes the real request FAST-armored from the
start. This lets FAST-dependent preauthentication mechanisms (such as
OTP) work without a pre-existing TGT or manual armor ccache setup. The
bootstrapped ticket is discarded once the real request completes.

The armor acquisition is driven through the same
`krb5_init_creds_step()` state machine as the outer request, so it
works transparently over any transport, including IAKERB, without any
transport-specific plumbing.

## Changes since the last revision

Per review feedback, this revision reconciles the design with #1447:

- Renamed the option to `auto_fast_armor` and moved it from
  `[libdefaults]` to `[realms]`.
- Changed the trigger from reactive (waiting to see PA-FX-FAST in the
  KDC's error padata) to eager: the armor ticket is now acquired
  before the first request is ever sent, saving a round trip. Since
  the variable is now realm-scoped opt-in, there's no need to wait for
  a signal that isn't informative anyway (most KDCs advertise FAST
  support regardless of whether anonymous PKINIT is configured).
- Split out the IAKERB OTP responder change into a separate,
  standalone change, to be submitted as its own PR once this one
  lands. (OTP preauthentication requires FAST armor to already be
  available before the KDC will offer it, and IAKERB has no other way
  to supply one, so that fix has no reachable regression test without
  this PR merged first — sequencing them avoids that dependency in
  CI/testing.)
- Replaced the OTP+RADIUS-mock IAKERB test with a simpler
  password-based IAKERB test that directly demonstrates the
  transport-agnostic bootstrap (no `pyrad` dependency needed).

## Test plan

- [x] `make check` passes (`--enable-maintainer-mode --with-crypto-impl=openssl`)
- [x] `src/util/cstyle.py` reports no style violations on this diff
- [x] `src/tests/t_pkinit.py`: covers the eager bootstrap via plain
      `kinit`, asserting the expected trace order (armor acquisition
      happens before any unauthenticated request for the real
      principal) and that it does not recurse for the anonymous
      principal
- [x] `src/tests/gssapi/t_gssapi.py`: covers the same bootstrap over
      IAKERB, confirming no IAKERB-specific code is needed

### DIFF
--- a/doc/admin/conf_files/krb5_conf.rst
+++ b/doc/admin/conf_files/krb5_conf.rst
@@ -487,6 +487,17 @@ following tags may be specified in the realm's subsection:
     names to local user names.  The tag is the mapping name, and the
     value is the corresponding local user name.
 
+**auto_fast_armor**
+    If this flag is true, initial credential acquisition will acquire
+    an anonymous PKINIT ticket to use as FAST armor before making the
+    real ticket request, unless an armor ccache is already configured
+    (as with kinit's **-T** option).  This flag allows
+    preauthentication mechanisms that require FAST, such as OTP, to
+    work without a pre-existing ticket.  This flag should only be set
+    for realms known to support anonymous PKINIT.  This flag has no
+    effect if the client principal is itself the anonymous principal.
+    The default value is false.  New in release 1.23.
+
 **default_domain**
     This tag specifies the domain used to expand hostnames when
     translating Kerberos 4 service principals to Kerberos 5 principals
--- a/src/include/k5-int.h
+++ b/src/include/k5-int.h
@@ -185,6 +185,7 @@ typedef unsigned char   u_char;
 #define KRB5_CONF_ALLOW_WEAK_CRYPTO            "allow_weak_crypto"
 #define KRB5_CONF_AUTH_TO_LOCAL                "auth_to_local"
 #define KRB5_CONF_AUTH_TO_LOCAL_NAMES          "auth_to_local_names"
+#define KRB5_CONF_AUTO_FAST_ARMOR              "auto_fast_armor"
 #define KRB5_CONF_CANONICALIZE                 "canonicalize"
 #define KRB5_CONF_CCACHE_TYPE                  "ccache_type"
 #define KRB5_CONF_CLOCKSKEW                    "clockskew"
--- a/src/include/k5-trace.h
+++ b/src/include/k5-trace.h
@@ -173,8 +173,8 @@ void krb5int_trace(krb5_context context, const char *fmt, ...);
 #define TRACE_DNS_URI_SEND(c, domain)                   \
     TRACE(c, "Sending DNS URI query for {str}", domain)
 
-#define TRACE_FAST_ARMOR_CCACHE(c, ccache_name)         \
-    TRACE(c, "FAST armor ccache: {str}", ccache_name)
+#define TRACE_FAST_ARMOR_CCACHE(c, ccache)              \
+    TRACE(c, "FAST armor ccache: {ccache}", ccache)
 #define TRACE_FAST_ARMOR_CCACHE_KEY(c, keyblock)                \
     TRACE(c, "Armor ccache session key: {keyblock}", keyblock)
 #define TRACE_FAST_ARMOR_KEY(c, keyblock)               \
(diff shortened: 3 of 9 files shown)


jason34105533:
This modification was prompted by testing IAKERB login against a localkdc environment. 
During testing, I noticed that the client was completely unable to handle the KDC's requirement for FAST armor, resulting in a confusing failure.

greghudson:
I do think the OTP responder should be separated out, ideally into a separate PR (although I know Github isn't the best at PR interdependencies).

The libkrb5 part of this PR overlaps significantly with PR #1447 .  In broad strokes this one is closer to mergeable, because it modifies the init_creds_step state machine rather than executing a synchronous fast_acquire_pkinit_armor() from within a krb5_init_creds_step() invocation.  But there are some differences I think should be rectified in favor of the older PR's behavior:

* In that PR we settled on the name `auto_fast_armor` for the config variable.  It should be a realm variable rather than a [libdefaults] variable; this is a natural fit because there will need to be realm configuration for PKINIT anyway.

* We will get better security properties (and one fewer round trip) if we start with getting an armor ticket right away, rather than waiting to see if the unprotected initial KDC reply indicates FAST support.  (That indication of support isn't really useful information.  Any KDC from the last decade plus will support FAST, but only a few will be configured to support anonymous PKINIT.  But that's what the realm config variable is for.)

jason34105533:
Thanks for the detailed feedback and for pointing me to #1447. I've updated the PR title/description and force-pushed a new revision that:

- Renames the config variable to `auto_fast_armor` and moves it under `[realms]`, matching #1447.
- Acquires the armor ticket eagerly instead of waiting for a PA-FX-FAST hint, so the first AS request is already armored.
- Drops the IAKERB OTP responder changes from this PR. I'll send those separately once `auto_fast_armor` lands, since there's currently no practical way to regression-test them without it (OTP requires FAST armor to be offered, and IAKERB has no other way to obtain one).
- Replaces the OTP + RADIUS mock IAKERB test with a password-based test that exercises the same transport-agnostic bootstrap without requiring `pyrad`.

I also verified locally that the eager acquisition removes the extra round trip you mentioned—the trace now shows the armor ticket being obtained immediately after `Getting initial credentials for ...`, without first sending an unauthenticated request for the target principal.

greghudson:
I pushed some changes to simplify the flow--we can set up the nested context at init time rather than waiting for the first _step request, and avoid some copy-paste by folding step_armor() into _step() and putting less responsibility into the (renamed) 

I noticed one issue that requires a bit of thought to resolve: the PR code modifies the caller's krb5_get_init_creds_opt object when it does:
```
        code = krb5_get_init_creds_opt_set_fast_ccache_name(context, ctx->opt,
                                                            ccname);
```
The options object is only supposed to be read from, not written to, by the get_init_creds functions.  For many callers this is harmless, but if the same options object is used for successive init_creds operations, the second operation will see a stale name of the FAST memory armor cache from the first operation and it won't behave correctly.

There doesn't appear to be a way to copy a GIC options object currently, and the current contract of krb5int_fast_as_armor() precludes managing the FAST ccache name in the get_in_tkt.c code.  I will revisit this before merging.

jason34105533:
Saw the XXX comment about mutating the caller's opt — figured I'd take a stab at one way to fix it, no worries at all if you'd rather go a different direction (or already have something in mind).

The idea: give `krb5int_fast_as_armor()` an extra `auto_armor_ccache` parameter so it can fall back to the ctx-owned armor ccache directly when the caller hasn't set one explicitly, instead of routing through `opt`. `finish_auto_armor()` ends up not touching `opt` at all.
(`get_in_tkt.c` is the only place this function is called in the entire codebase.)

To sanity-check that this was a real issue and not just theoretical, I put together a small throwaway C program that reuses one `krb5_get_init_creds_opt` across two `krb5_get_init_creds_password()` calls. On the current commit it reliably fails the second call ("No credentials cache found") once the first call's ephemeral ccache goes away. With this change both calls come back clean across repeated runs.

greghudson:
I like that idea, but decided to further refactor krb5int_fast_as_armor() so that we aren't extracting the cache name and re-resolving it in the auto-armor case.

jason34105533:
Confirmed your refactor works well — rebuilt from a clean tree and reran everything, all green. Thanks for tightening that up, this is cleaner than what I had.

abbra:
FYI, we replaced my (PR#1447) implementation by this one in Fedora builds and it immediately broke FreeIPA deployments. I think the current (PR#1536) patchset is incomplete and needs to be extended.

Details are in https://bodhi.fedoraproject.org/updates/FEDORA-2026-b887a90f4b#comment-4756801

PR #1535, opened 2026-07-27
Fix deserializing context with missing profile files

bugaevc:
The code path in `k5_internalize_context` tolerates some failures of `internalize_oscontext` and `profile_ser_internalize` (`EINVAL` and `ENOENT`), and attempts to proceed nevertheless. A specific non-fatal failure that can happen when loading the profile is missing configuration files (i.e. `/etc/krb5.conf`): `profile_init_flags` tries to read at least one file, returns `ENOENT` if they all turned out to be missing, and then `profile_ser_internalize` attempts to proceed despite that.

However, `profile_ser_internalize` was only updating the buffer pointer in case of complete success, which meant that, when `k5_internalize_context` tried to proceed after an error, it was actually re-reading profile data (starting with `PROF_MAGIC_PROFILE`), and not the trailer of the serialized context (`KV5M_CONTEXT`) that it expected. So it failed on that.

I observed this failure in the following scenario:
* An application using GSS API is running in the [Flatpak](https://flatpak.org/) environment;
* Kerberos is properly set up and configured on the host system;
* Flatpak provides the application with a different view of the file system compared to the host, relevantly for our case a separate `/etc`;
* [gssproxy](https://github.com/gssapi/gssproxy/) (cc @simo5) is used (transparently to the app, https://github.com/flatpak/flatpak/pull/4914) to proxy GSS API calls to the host system, where the Kerberos GSS implementation serves them;
* As a part of `gss_unwrap` implementation, gssproxy transfers the established context from the host system to the app inside Flatpak (`gssi_unwrap` -> `gpp_remote_to_local_ctx`);
* The serialized context representation encodes `/etc/krb5.conf`, but no such file exists in the Flatpak environment, therefore deserialization fails near the very end, in the manner described above.

It looks like this was intended to work, and missing conf files are not supposed to fail the entire deserialization process. To make it work, advance the buffer pointer as soon as `profile_ser_internalize` finishes actually reading the serialized profile data, whether `profile_init` succeeds or not.

With this change, `gss_unwrap` in Flatpak via gssproxy succeeds.

------------------

It looks like GitHub PRs is the preferred way of submitting patches now, but if it's not, I could post this to a mailing list instead.

### DIFF
--- a/src/util/profile/prof_init.c
+++ b/src/util/profile/prof_init.c
@@ -654,12 +654,12 @@ errcode_t profile_ser_internalize(profile_t *profilep,
         goto cleanup;
     }
 
-    if ((retval = profile_init((const_profile_filespec_t *) flist,
-                               profilep)))
-        goto cleanup;
-
+    /* Update the buffer pointer even if profile initialization fails, so the
+     * caller can keep deserializing from the correct position.  */
     *bufpp = bp;
     *remainp = remain;
+
+    retval = profile_init((const_profile_filespec_t *) flist, profilep);
 
 cleanup:
     if (flist) {


simo5:
LGTM,
@greghudson WDYT?

greghudson:
It's fine; I edited the commit message and comment for brevity and will merge it soon.

I don't really understand the intent of the existing code.  It's weird that it tolerates failures from internalize_oscontext(), which just reads in four numbers and stores them.  It might have been a backward-compatibility measure, but (a) backwards compatibility shouldn't really have been a factor for the serialization code, and (b) why tolerate ENOENT?  In a similar vein, I'm not sure why an EINVAL from profile_ser_internalize() is tolerated.  It's conceivable that both tolerances were sloppily written backwards-compatibility measures and that there was never an intention to tolerate missing config files.  But I think it's okay to tolerate them now, and I am fine with using a scalpel rather than a sledgehammer to make it work.

Just to make note of some edge cases: this tolerance creates a semantic difference between profile variables that are read from krb5_init_context() and variables that get read on the fly.  Configurations determined by the first class of variable will survive a serialize-deserialize operation even if the target has no access to the config files, while those determined by the second class will not.

Also, there unfortunately appears to be a set of configuration variables which is read from krb5_init_context() but not serialized, such as dns_canonicalize_hostname and allow_weak_crypto.  These won't survive a serialize-deserialize even if the target has the same config files as the source.  Fixing that is out of scope for this PR of course.

PR #1536, opened 2026-08-08
Add OTP responder for IAKERB initial credential acquisition

jason34105533:
IAKERB drives init_creds programmatically with no prompter available, so an OTP preauth challenge cannot be answered and the OTP client preauth module fails with EIO.
Register a responder that answers the OTP question using the caller-supplied password when one is available, leaving other preauth mechanisms unchanged.

Together with the auto_fast_armor variable from #1534, this allows OTP authentication over IAKERB without existing credentials or an armor ccache.

Add a test covering the full IAKERB + FAST + OTP path using the simulated RADIUS backend. The test verifies that the responder supplies the password to the OTP challenge and that the RADIUS server receives and accepts it.

### DIFF
--- a/doc/appdev/gssapi.rst
+++ b/doc/appdev/gssapi.rst
@@ -265,23 +265,34 @@ The following options are supported by the krb5 mechanism:
   be used for verification when initiator credentials are acquired
   using a password and verified.
 
+* **otp**: For acquiring initiator credentials, this option instructs
+  the mechanism to acquire fresh credentials into a unique memory
+  credential cache.  The value is used as a one-time password
+  credential if the KDC requests OTP preauthentication.  This option
+  may not be used with the **ccache** or **client_keytab** options,
+  and a *desired_name* must be specified.  (New in release 1.23.)
+
 * **password**: For acquiring initiator credentials, this option
   instructs the mechanism to acquire fresh credentials into a unique
-  memory credential cache.  This option may not be used with the
-  **ccache** or **client_keytab** options, and a *desired_name* must
-  be specified.  (New in release 1.19.)
+  memory credential cache.  The option value is used as the Kerberos
+  password.  This option may not be used with the **ccache** or
+  **client_keytab** options, and a *desired_name* must be specified.
+  This option may be used with the **otp** option, but only one of the
+  values will be used, depending on what preauthentication mechanisms
+  are made available by the KDC.  (New in release 1.19.)
 
 * **rcache**: For acquiring acceptor credentials, the name of the
   :ref:`replay cache <rcache_definition>` to be used when processing
   the initiator tokens.  (New in release 1.13.)
 
-* **verify**: For acquiring initiator credentials, this option
-  instructs the mechanism to verify the credentials by obtaining a
-  ticket to a service with a known key.  The service key is obtained
-  from the keytab specified with the **keytab** option or the default
-  keytab.  The value may be the name of a principal in the keytab, or
-  the empty string.  If the empty string is given, any ``host``
-  service principal in the keytab may be used.  (New in release 1.19.)
+* **verify**: When acquiring fresh initiator credentials with the
+  **password** or **otp** options, this option instructs the mechanism
+  to verify the credentials by obtaining a ticket to a service with a
+  known key.  The service key is obtained from the keytab specified
+  with the **keytab** option or the default keytab.  The value may be
+  the name of a principal in the keytab, or the empty string.  If the
+  empty string is given, any ``host`` service principal in the keytab
+  may be used.  (New in release 1.19.)
 
 In release 1.20 or later, if a collection name is specified for
 **cache** in a call to gss_store_cred_into(), an existing cache for
--- a/src/lib/gssapi/krb5/acquire_cred.c
+++ b/src/lib/gssapi/krb5/acquire_cred.c
@@ -325,7 +325,7 @@ can_get_initial_creds(krb5_context context, krb5_gss_cred_id_rec *cred)
 {
     krb5_error_code code;
 
-    if (cred->password != NULL)
+    if (cred->password != NULL || cred->otp != NULL)
         return TRUE;
 
     if (cred->client_keytab == NULL)
@@ -477,9 +477,8 @@ get_cache_for_name(krb5_context context, krb5_gss_cred_id_rec *cred)
     cctype = krb5_cc_get_type(context, defcc);
     have_collection = krb5_cc_support_switch(context, cctype);
 
-    /* We can use an empty default ccache if we're using a password or if
-     * there's no collection. */
-    if (cred->password != NULL || !have_collection) {
+    /* We can use an empty default ccache if there is no collection. */
+    if (!have_collection) {
         if (krb5_cc_get_principal(context, defcc, &princ) == KRB5_FCC_NOFILE) {
             cred->ccache = defcc;
             defcc = NULL;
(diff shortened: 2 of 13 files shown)


greghudson:
We have traditionally kept the Kerberos password input separate from other string inputs to preauth mechanisms, and  there are security ramifications to blurring the lines.  The Kerberos protocol was built around a specific set of security properties for the password.  They aren't the best properties (there is a lot of potential for dictionary attacks in the common and default configurations), but one of the important properties is phishing resistance.  With this PR a KDC can ask the client to hand over the password, so that property is lost.  This concern may be overblown, in that the KDC must be authenticated to the client and must be for a realm with PKINIT and fast_auto_armor configuration the client, suggesting an elevated level of trust.  But I am still uneasy.

I do get that without this PR, it is difficult to connect the dots between FAST OTP and IAKERB.  There is no GSS initial creds API with a prompter interface, and there is no libkrb5 API for IAKERB.  A good, mechanism-neutral prompter-enabled initial creds API would be a big undertaking to specify and implement.  (I believe I've seen a draft specification from Nico, but from my memory of it, implementing it in MIT krb5 would require turning the clpreauth pluggable interface inside out.)  We could do one of those ugly gss_krb5 specials like gss_krb5_import_cred(); there may be no better answer in the realm of reasonable effort.

jason34105533:
Thanks — the phishing-resistance point is fair, and the password-as-OTP shortcut was exactly where the line got blurred. I've reworked it locally to keep the two inputs separate:

- Add an "otp" credential store option (alongside the existing "password" option) for IAKERB initiator creds, with the same constraints as password: initiator-only, requires a desired_name, mutually exclusive with ccache/client_keytab, and rejected for the non-IAKERB krb5 mech.

- The IAKERB responder now only answers the OTP question with the value from the credential store. The Kerberos password is never used to answer any KDC question, so a KDC can no longer induce the client to hand it over.

- An OTP-only credential (no password at all) is sufficient to acquire initial credentials, since the OTP preauth module determines the reply key.

This stays within the cred-store extension mechanism, so there’s no new ABI surface. It’s arguably less ugly than introducing a new gss_krb5_* call, though it does mean the value is fixed at acquire time, which seems to match how OTP values are typically used anyway.

One point worth flagging: I serialized the otp field in the exported cred format, which grows the array from 14 to 15 elements. Import now accepts both lengths, so older tokens still import, but older libraries won’t import newer tokens. (creds without an otp value keep the existing 14-element format, so normal credentials interoperate with older libraries in both directions; only otp-bearing tokens add a 15th element, which older libraries cleanly reject. )

I think this idea is kind of tricky though, and it do possibly cause issues when communicating with old version libs.
So I would be happy to take the dedicated gss_krb5_* API route instead if you think that’s the cleaner long-term direction.

greghudson:
This approach should be workable; I didn't think about gss_acquire_cred_from() in my previous reply.  GSS import/export formats aren't required to be interoperable across versions.

Is there a reason this should only work for IAKERB?  Semantically it seems to make sense for normal gss_acquire_cred_from().

nicowilliams:
> This approach should be workable; I didn't think about gss_acquire_cred_from() in my previous reply. GSS import/export formats aren't required to be interoperable across versions.
> 
> Is there a reason this should only work for IAKERB? Semantically it seems to make sense for normal gss_acquire_cred_from().

Yes, using `gss_acquire_cred_from()` works for this, except you have to know a priori that you'll need an `otp` element.  Also, we don't have a way of programmatically indicating via GSS-API major/minor status combinations that the user needs to use an OTP, except that hopefully the user can get a user-meaningful error message from `gss_display_status()`.  If there are multiple options (smartcard _or_ OTP), then things will get messy.  Ditto if the user does not speak English (maybe the KDC knows the user's preferred language and localizes the `e-text` accordingly).  But yes, this approach will work.

~Except I don't see why we need to include the OTP in exported credentials -- it's going to expire shortly anyway, so what's the point?~  Well, It doesn't hurt, so I think it's fine to include the OTP in the exported credential token.

nicowilliams:
> One point worth flagging: I serialized the otp field in the exported cred format, which grows the array from 14 to 15 elements. Import now accepts both lengths, so older tokens still import, but older libraries won’t import newer tokens. (creds without an otp value keep the existing 14-element format, so normal credentials interoperate with older libraries in both directions; only otp-bearing tokens add a 15th element, which older libraries cleanly reject. )
> 
> I think this idea is kind of tricky though, and it do possibly cause issues when communicating with old version libs. So I would be happy to take the dedicated gss_krb5_* API route instead if you think that’s the cleaner long-term direction.

There is no point to including the OTP in the exported credential.  No one is going to `gss_acquire_cred_from()`, export the credential, import the credential, then use it with `gss_init_sec_context()`.  Sure, I suppose someone might just to see if it works, or maybe someone has a complex privsep scheme that forces it, but it seems very unlikely.  Still, it doesn't hurt to include the OTP in the exported credential token (the caller is expected to protect that token's confidentiality), plus the OTP is about to be used and expire, so I think it's fine to leave it in there.

jason34105533:
Thanks, both.

No deep reason for the IAKERB-only restriction — it was just scope-minimization for the first pass. Extending "otp" to the normal krb5 mechanism makes sense to me; I'll handle that in a later push.
On the exported cred format: OK, I'll keep it as is for now, since it doesn't hurt and I think my import already tolerates older tokens.

jason34105533:
Pushed the generalized version. The "otp" option is now accepted for the regular krb5 mechanism as well: the responder moved to a shared kg_otp_responder(), and the eager initial-credential acquisition path registers it when an otp value is present. An otp-only credential (no password) works there too, since the OTP preauth module determines the reply key; against a KDC that doesn't offer OTP, acquisition fails cleanly with the usual preauth failure.

For testing I added a -k option to t_credstore and a test that acquires with an otp value against the OTP KDC — with auto_fast_armor configured, the eager acquisition bootstraps anonymous PKINIT armor, answers the OTP challenge, and the simulated RADIUS backend verifies it.

greghudson:
I made some changes, the most substantive of which is that the verify option should apply to credentials gained using OTP.  (This should be largely unnecessary since auto_fast_armor will authenticate the KDC via its X.509 certificate, but it's easy to respect the option in this case so we should do so.  On the flip side I noted that IAKERB doesn't seem to implement the verify option; that's a little concerning but it's out of scope for this PR.)

I didn't see the need for the t_credstore.c -k option (it should use the krb5 mech by default) so I took it out.

greghudson:
> I didn't see the need for the t_credstore.c -k option (it should use the krb5 mech by default) so I took it out.

I put this back in with a comment in t_gssapi.py to explain why we use it.  Without it, we acquire krb5 and krb5-inside-SPNEGO credentials.  (Heimdal has an interesting solution for this which is to tag SPNEGO as a mechanism that can directly use union creds, but that's not important here.)  The second acquisition fires a second RADIUS request from the KDC, but our mock RADIUS daemon only handles one request, so we get a timeout.

In the process of debugging I noticed a small malfunction in the libkrb5 preauth framework: previously failed mechanisms are filtered out inf the process_pa_data() loop, but not in the fill_response_items() loop.  So after OTP fails the second time around, we get a pointless second call to otp_client_prep_questions().

I also noted a small memory leak when otp_client_prep_questions() is called multiple times during the same AS request; it overwrites `*modreq` with a new decode_krb5_pa_otp_challenge() result without freeing the previous value.  This is the visible reason for the CI failure on my previous round of changes; the SPNEGO-krb5 acquisition failing isn't enough to cause t_credstore to fail as one of the mechanisms succeeds, but the small memory leak resulting from the failure causes the asan job to fail.

I will fix these two problems outside of the PR, as they don't cause issues when the test behaves as intended.

